PR #1, opened 2017-03-24
Create Option type and base combinators

JMPergar:
Port of https://github.com/scala/scala/blob/v2.12.1/src/library/scala/Option.scala

### DIFF
--- a/library/src/main/java/com/finecinnamon/kats/Option.kt
+++ b/library/src/main/java/com/finecinnamon/kats/Option.kt
@@ -1,0 +1,159 @@
+/*
+ * Copyright (C) 2017
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.finecinnamon.kats
+
+import java.util.*
+
+/**
+ * Port of https://github.com/scala/scala/blob/v2.12.1/src/library/scala/Option.scala
+ *
+ * Represents optional values. Instances of `Option`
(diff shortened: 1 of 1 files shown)


JorgeCastilloPrz:
you are fast :)

library/src/main/java/com/finecinnamon/kats/Option.kt
```diff
@@ -0,0 +1,159 @@
...
+     * Returns true if the option is an instance of $some, false otherwise.
+     */
+    val isDefined: Boolean = !isEmpty
+
+    /**
+     * Returns the option's value.
+     * @note The option must be nonempty.
+     * @throws java.util.NoSuchElementException if the option is empty.
+     */
+    abstract fun get(): A
```
raulraja: :fire: Since this is not a standard lib these can be removed

library/src/main/java/com/finecinnamon/kats/Option.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Returns the option's value if the option is nonempty, otherwise
+     * return the result of evaluating `default`.
+     *
+     * @param default  the default expression.
+     */
+    inline fun getOrElse(default: () -> A): A = if (isEmpty) default() else get()
```
raulraja: can this be strict and not lazy and implemented in terms of `fold`?

library/src/main/java/com/finecinnamon/kats/Option.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Apply the given procedure $f to the option's value,
+     * if it is nonempty. Otherwise, do nothing.
+     *
+     * @param  f   the procedure to apply.
+     * @see map
+     * @see flatMap
+     */
+    inline fun <B> foreach(f: (A) -> B) { if (!isEmpty) f(get()) }
```
raulraja: :fire:

library/src/main/java/com/finecinnamon/kats/Option.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    class Some<A>(val value: A) : Option<A>() {
+        override val isEmpty = false
+        override fun get(): A = value
+    }
+
+    class None<A> : Option<A>() {
```
raulraja: Can this be an object instead of class so it's more effitient. `None` is a singleton that extends `Option<Nothing>`

PR #3, opened 2017-03-24
Create Option type with base combinators

JMPergar:
Issue #4 

Port of https://github.com/scala/scala/blob/v2.12.1/src/library/scala/Option.scala

### DIFF
--- a/core/src/main/java/kats/Option.kt
+++ b/core/src/main/java/kats/Option.kt
@@ -16,15 +16,13 @@
 
 package kats
 
-import java.util.*
-
 /**
  * Port of https://github.com/scala/scala/blob/v2.12.1/src/library/scala/Option.scala
  *
  * Represents optional values. Instances of `Option`
  * are either an instance of $some or the object $none.
  */
-sealed class Option<A> {
+sealed class Option<out A> {
 
     /**
      * Returns true if the option is $none, false otherwise.
@@ -37,19 +35,12 @@ sealed class Option<A> {
     val isDefined: Boolean = !isEmpty
 
     /**
-     * Returns the option's value.
-     * @note The option must be nonempty.
-     * @throws java.util.NoSuchElementException if the option is empty.
-     */
-    abstract fun get(): A
-
-    /**
      * Returns the option's value if the option is nonempty, otherwise
      * return the result of evaluating `default`.
      *
      * @param default  the default expression.
      */
-    inline fun getOrElse(default: () -> A): A = if (isEmpty) default() else get()
+    inline fun getOrElse(default: () -> A): A = fold({ default() }, { A -> A })
 
     /**
      * Returns a $some containing the result of applying $f to this $option's
@@ -62,7 +53,7 @@ sealed class Option<A> {
      *  @see flatMap
      *  @see foreach
      */
-    inline fun <B> map(f: (A) -> B): Option<B> = if (isEmpty) None() else Some(f(get()))
+    inline fun <B> map(f: (A) -> B): Option<B> = fold({ None }, { A -> Some(f(A)) })
 
     /**
      * Returns the result of applying $f to this $option's value if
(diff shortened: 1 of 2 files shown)


core/src/main/java/kats/Option.kt
```diff
@@ -0,0 +1,159 @@
...
+     * Returns true if the option is an instance of $some, false otherwise.
+     */
+    val isDefined: Boolean = !isEmpty
+
+    /**
+     * Returns the option's value.
+     * @note The option must be nonempty.
+     * @throws java.util.NoSuchElementException if the option is empty.
+     */
+    abstract fun get(): A
```
raulraja: :fire: no need for unpricipled APIs at this stage

core/src/main/java/kats/Option.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Returns the option's value if the option is nonempty, otherwise
+     * return the result of evaluating `default`.
+     *
+     * @param default  the default expression.
+     */
+    inline fun getOrElse(default: () -> A): A = if (isEmpty) default() else get()
```
raulraja: Can we implement all combinators in terms of `fold` once we remove `get`?

core/src/main/java/kats/Option.kt
```diff
@@ -0,0 +1,159 @@
...
+     * Returns the result of applying $f to this $option's
+     * value if the $option is nonempty.  Otherwise, evaluates
+     * expression `ifEmpty`.
+     *
+     * @note This is equivalent to `$option map f getOrElse ifEmpty`.
+     *
+     * @param  ifEmpty the expression to evaluate if empty.
+     * @param  f       the function to apply if nonempty.
+     */
+    inline fun <B> fold(ifEmpty: () -> B, f: (A) -> B): B = if (isEmpty) ifEmpty() else f(get())
```
raulraja:
```kotlin
when(this) {
 is None -> ...
 is Some -> ...
}
```

core/src/main/java/kats/Option.kt
```diff
@@ -0,0 +1,159 @@
...
+    val nonEmpty = isDefined
+
+    /**
+     * Tests whether the option contains a given value as an element.
+     *
+     * @param elem the element to test.
+     * @return `true` if the option has an element that is equal (as
+     * determined by `==`) to `elem`, `false` otherwise.
+     */
+    fun contains(elem: A): Boolean = !isEmpty && get() == elem
```
raulraja: :fire:

raulraja: We are not part of the Collection hierarchies yet so this not needed

core/src/main/java/kats/Option.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Apply the given procedure $f to the option's value,
+     * if it is nonempty. Otherwise, do nothing.
+     *
+     * @param  f   the procedure to apply.
+     * @see map
+     * @see flatMap
+     */
+    inline fun <B> foreach(f: (A) -> B) { if (!isEmpty) f(get()) }
```
raulraja: :fire:

core/src/main/java/kats/Option.kt
```diff
@@ -0,0 +1,159 @@
...
+
+    class Some<A>(val value: A) : Option<A>() {
+        override val isEmpty = false
+        override fun get(): A = value
+    }
+
+    class None<A> : Option<A>() {
```
raulraja: This may be a singleton. `object None extends Option<Nothing>`

core/src/main/java/kats/Option.kt
JMPergar: :+1:

core/src/main/java/kats/Option.kt
JMPergar: :+1:

core/src/main/java/kats/Option.kt
JMPergar: :+1:

PR #6, opened 2017-03-24
Create Try type and base combinators

JMPergar:
Issue #5 

Port of https://github.com/scala/scala/blob/v2.12.1/src/library/scala/util/Try.scala


core/src/main/java/kats/Try.kt
```diff
@@ -0,0 +1,144 @@
...
+     * Returns the value from this `Success` or the given `default` argument if this is a `Failure`.
+     *
+     * ''Note:'': This will throw an exception if it is not a success and default throws an exception.
+     */
+    abstract fun getOrElse(default: () -> A): A
+
+    /**
+     * Returns the value from this `Success` or throws the exception if this is a `Failure`.
+     */
+    abstract fun get(): A
```
raulraja: All we are learning in `Option` also applies here :fire:

core/src/main/java/kats/Try.kt
```diff
@@ -0,0 +1,144 @@
+/*
+ * Copyright (C) 2017
```
aballano: Missing "The Kats Authors"

PR #15, opened 2017-03-24
print verbose log of tests on console

JorgeCastilloPrz:
By adding the appended `gradle` configuration, you allow gradle to print on the log the result of each unit test running with the build. This is pretty handy to read CI logs.

The result can be checked on this build CI checks.

### DIFF
--- a/kats/build.gradle
+++ b/kats/build.gradle
@@ -16,6 +16,12 @@
 
 apply plugin: 'kotlin'
 
+test {
+    testLogging {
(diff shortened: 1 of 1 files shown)


jrgonzalezg:
Can this be put on the allprojects section of the main build.gradle for any future modules to also benefit from it?

JorgeCastilloPrz:
Would be cool, but I don't know how to do it, since the `allProjects` block does not accept a `test` block inside of it

PR #16, opened 2017-03-24
Add apply method to Option type

JMPergar:
Related with #1


jrgonzalezg:
`apply` is used in scala to be able to use the class as a function. It may be better to call it `invoke` in Kotlin, or to declare both with one being an alias of the other.

aballano:
Mmm it is possible that there's something with companion objects and invoke actually. https://discuss.kotlinlang.org/t/generic-object-creation/1663/5?u=shyish

JMPergar:
Good point @aballano, I'll work in this ;)

jrgonzalezg:
Also, if f() is costly it will be called twice, maybe it could be cached on a val.

kats/src/main/kotlin/kats/Option.kt
```diff
@@ -24,6 +24,11 @@ package kats
  */
 sealed class Option<out A> {
 
+    companion object {
+
+        inline operator fun <A> invoke(f: () -> A): Option<A> = f().let { if (it == null) None else Some(it) }
```
raulraja: This should not be a function but a strict non nullable argument. This factory method would violate the laws because if you pass in `null` you should get `Some(null)`. You are adding implicitly here a safeguard that may not be wanted by the user.

raulraja: Even though Scala does that. Additionally if you want this behaviour we should add a factory method `fromNullable` which does what you have implemented here

JMPergar: And what is the way if we want to compute with Option like the result? We should manage this every time?

raulraja: `fromNullable`

katz/src/main/kotlin/katz/Option.kt
```diff
@@ -24,6 +24,13 @@ package katz
  */
 sealed class Option<out A> {
 
+    companion object {
+
+        inline fun <A> fromNullable(f: () -> A): Option<A> = f().let { if (it == null) None else Some(it) }
+
+        operator fun <A> invoke(a: A): Option<A> = if (a == null) None else Some(a)
```
raulraja: this one should just be `= Some(a)` or not exist at all.

jrgonzalezg: `= Some(a)` seems good and should play well with Kotlin null safety

jrgonzalezg: You could also do:
`operator fun <A: Any> invoke(a: A): Option<A> = Some(a)`

raulraja:

PR #17, opened 2017-03-24
Created either type and base combinators

aballano:
Issue #7 

Port of https://github.com/scala/scala/blob/v2.12.1/src/library/scala/util/Either.scala


kats/src/main/kotlin/kats/Either.kt
```diff
@@ -0,0 +1,56 @@
...
+ * Represents a value of one of two possible types (a disjoint union.)
+ * An instance of Either is either an instance of [Left] or [Right].
+ */
+sealed class Either<out A, out B> {
+
+    inline fun <C> fold(fl: (A) -> C, fr: (B) -> C): C = when (this) {
+        is Right -> fr(b)
+        is Left -> fl(a)
+    }
+
```
jrgonzalezg: Name of functions should be `fa` and `fb`

kats/src/main/kotlin/kats/Either.kt
```diff
@@ -0,0 +1,56 @@
...
+    fun swap(): Either<B, A> =
+            fold({ Right<B, A>(it) }, { Left<B, A>(it) })
+
```
jrgonzalezg: This should be implemented using `when(it)`

kats/src/main/kotlin/kats/Either.kt
```diff
@@ -0,0 +1,56 @@
...
+            fold({ false }, { predicate(it) })
+
+    inline fun <AA, A> filterOrElse(predicate: (B) -> Boolean, default: AA): Either<AA, B> where A : AA =
+            fold({ this as Either<AA, B> }, { if (predicate(it)) this as Either<AA, B> else Left(default) })
+
+
+
+    class Left<out A, out B>(val a: A) : Either<A, B>()
+    class Right<out A, out B>(val b: B) : Either<A, B>()
+}
```
jrgonzalezg: Left and Right may better be data classes and both declare val named value as in Scala cats

kats/src/main/kotlin/kats/Either.kt
aballano: Why? Fold already does that for you, why not encapsulate the matching in there? 

Btw I'm removing the explicit types as they're not needed.

kats/src/main/kotlin/kats/Either.kt
```diff
@@ -0,0 +1,56 @@
...
+        is Right -> fr(b)
+        is Left -> fl(a)
+    }
+
+    fun swap(): Either<B, A> =
+            fold({ Right<B, A>(it) }, { Left<B, A>(it) })
+
+    inline fun <C> map(f: (B) -> C): Either<A, C> =
+            fold({ it as Either<A, C> }, { Right(f(it)) })
+
```
jrgonzalezg: Could be done as:
```
when(this) {
        is Right -> Right(f(value))
        is Left -> Left(value)
    }
```
which also does not seem to produce a warning like the casting version

aballano: I'll take a look at the warning indeed, is really annoying :D

jrgonzalezg: It seems the warning is also gone with:
```
fold({ Left(it) }, { Right(f(it)) })
```

aballano: You're right, i think i made couple of mistakes around, I'll work on creating a test case to be 100% sure :)

kats/src/main/kotlin/kats/Either.kt
```diff
@@ -0,0 +1,56 @@
...
+    fun <BB, B> contains(elem: BB): Boolean where B : BB =
+            fold({ false }, { it == elem })
+
+    inline fun exists(predicate: (B) -> Boolean): Boolean =
+            fold({ false }, { predicate(it) })
+
+    inline fun <AA, A> filterOrElse(predicate: (B) -> Boolean, default: AA): Either<AA, B> where A : AA =
```
raulraja: I did not know of this trick. @JMPergar this can replace the extension functions for getOrElse on Option.

kats/src/main/kotlin/kats/Either.kt
```diff
@@ -0,0 +1,56 @@
...
+            fold({ this as Either<AA, B> }, { if (predicate(it)) this as Either<AA, B> else Left(default) })
```
raulraja: The casting is fine but for correctness let's return the proper instance on a `Right` since we are not looking to optimize allocations here

kats/src/main/kotlin/kats/Either.kt
```diff
@@ -0,0 +1,56 @@
...
+
+
+    class Left<out A, out B>(val a: A) : Either<A, B>()
+    class Right<out A, out B>(val b: B) : Either<A, B>()
+}
+
+inline fun <AA, A, B, C> Either<A, B>.flatMap(f: (B) -> Either<AA, C>): Either<AA, C> where A : AA =
+        fold({ this as Either<AA, C> }, { f(it) })
+
+inline fun <A, B, BB> Either<A, B>.getOrElse(default: () -> BB): BB where B : BB = fold({ default() }, { b -> b })
```
raulraja: Why are these defined as extensions?

kats/src/main/kotlin/kats/Either.kt
JMPergar: I'm agree with @aballano, delegating to fold method is better way IMHO

kats/src/main/kotlin/kats/Either.kt
JMPergar: It's not valid because this B is not the that is declared in class.

raulraja:

katz/src/main/kotlin/katz/Either.kt
```diff
@@ -0,0 +1,194 @@
...
+}
+
+/**
+ * Binds the given function across [Either.Right].
+ *
+ * @param f The function to bind across [Either.Right].
+ */
+inline fun <AA, C, A, B> Either<A, B>.flatMap(f: (B) -> Either<AA, C>): Either<AA, C> where A : AA =
+        fold({ Either.Left(it) }, { f(it) })
+
```
jrgonzalezg: Use just <A, B> as type parameters, type bounds are obtained by covariance on Either itself

katz/src/main/kotlin/katz/Either.kt
```diff
@@ -0,0 +1,194 @@
...
+ *
+ * Example:
+ * ```
+ * Right(12).getOrElse(17) // Result: 12
+ * Left(12).getOrElse(17)  // Result: 17
+ * ```
+ */
+inline fun <BB, A, B> Either<A, B>.getOrElse(default: () -> BB): BB where B : BB =
+        fold({ default() }, { b -> b })
+
```
jrgonzalezg: Use just <A, B> as type parameters, type bounds are obtained by covariance on Either itself

katz/src/main/kotlin/katz/Either.kt
```diff
@@ -0,0 +1,194 @@
...
+ * ```
+ * Right(12).filterOrElse({ it > 10 }, { -1 }) // Result: Right(12)
+ * Right(7).filterOrElse({ it > 10 }, { -1 })  // Result: Left(-1)
+ *
+ * val left: Either<Int, Int> = Left(12)
+ * left.filterOrElse({ it > 10 }, { -1 })      // Result: Left(12)
+ * ```
+ */
+inline fun <AA, A, B> Either<A, B>.filterOrElse(predicate: (B) -> Boolean, default: () -> AA): Either<AA, B> where A : AA =
+        fold({ Either.Left(it) }, { if (predicate(it)) Either.Right(it) else Either.Left(default()) })
```
jrgonzalezg: Use just <A, B> as type parameters, type bounds are obtained by covariance on Either itself

katz/src/main/kotlin/katz/Either.kt
```diff
@@ -0,0 +1,194 @@
...
+ *
+ * Example:
+ * ```
+ * Right(12).getOrElse(17) // Result: 12
+ * Left(12).getOrElse(17)  // Result: 17
+ * ```
+ */
+inline fun <A : Any, B : Any> Either<A, B>.getOrElse(default: () -> B): B =
+        fold({ default() }, { b -> b })
+
```
jrgonzalezg: Maybe A can be removed here since it is not used. It may be possible to replace it by Nothing or by a star projection or something similar.

katz/src/main/kotlin/katz/Either.kt
```diff
@@ -0,0 +1,194 @@
...
+ * Returns the value from this [Either.Right] or the given argument if this is a [Either.Left].
+ *
+ * Example:
+ * ```
+ * Right(12).getOrElse(17) // Result: 12
+ * Left(12).getOrElse(17)  // Result: 17
+ * ```
+ */
+inline fun <B : Any> Either<*, B>.getOrElse(default: () -> B): B =
+        fold({ default() }, { b -> b })
```
JMPergar: I did it the same, but now I think that it's more clear in this way:

`+inline fun <B : Any> Either<*, B>.getOrElse(default: () -> B): B = fold({ default() }, { it })`

`it` better than `b -> b`

katz/src/main/kotlin/katz/Either.kt
```diff
@@ -0,0 +1,194 @@
...
+        override val isRight = false
+    }
+
+    /**
+     * The right side of the disjoint union, as opposed to the [Left] side.
+     */
+    data class Right<out B : Any>(val b: B) : Either<Nothing, B>() {
+        override val isLeft = true
+        override val isRight = false
+    }
```
jrgonzalezg: isLeft and isRight are reversed here

JMPergar: ouch!!

katz/src/main/kotlin/katz/Either.kt
```diff
@@ -0,0 +1,194 @@
...
+     * ```
+     * Right("something").contains { "something" } // Result: true
+     * Right("something").contains { "anything" }  // Result: false
+     * Left("something").contains { "something" }  // Result: false
+     *  ```
+     *
+     * @param elem    the element to test.
+     * @return `true` if the option has an element that is equal (as determined by `==`) to `elem`, `false` otherwise.
+     */
+    fun <BB, B> contains(elem: () -> BB): Boolean where B : BB =
```
JMPergar: I think I've been ambiguous here, in my last comment I refer to the param `elem`. `elem` should be a simple type not a lambda.

katz/src/main/kotlin/katz/Either.kt
aballano: image

katz/src/main/kotlin/katz/Either.kt
aballano: I wanted to make the value lazy in here, since if this Either is a Left, then `elem` doesn't have to be evaluated, in case is a call to a method or something similar. What do you think?

katz/src/main/kotlin/katz/Either.kt
```diff
@@ -0,0 +1,196 @@
...
+ * ```
+ * Right("something").contains { "something" } // Result: true
+ * Right("something").contains { "anything" }  // Result: false
+ * Left("something").contains { "something" }  // Result: false
+ *  ```
+ *
+ * @param elem    the element to test.
+ * @return `true` if the option has an element that is equal (as determined by `==`) to `elem`, `false` otherwise.
+ */
+fun <A : Any, B, BB : Any> Either<A, B>.contains(elem: () -> BB): Boolean where B : BB =
```
JMPergar: `elem` should be a simple type not a function, this is the way:

`fun <A : Any, B : Any> Either<A, B>.contains(elem: B): Boolean = fold({ false }, { it == elem })`

katz/src/main/kotlin/katz/Either.kt
```diff
@@ -0,0 +1,196 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package katz
+
+import katz.Either.Right
```
raulraja: Does this need to be imported?

raulraja:
@JMPergar merge is blocked awaiting you.

PR #18, opened 2017-03-24
Add reader type

### LINKED ISSUE #33: Add Reader Type

JorgeCastilloPrz:
Fixes #33 

**Tests are not in place since I still have the ones I had for the reader, which are probably not passing anymore, and should be adapted to use the alias**. I am waiting to create a Monad and polish everything a bit to create the tests. 

**Please avoid commenting about adding tests for now** 😝.

### Current status

`Kleisli` has been implemented and some typealiases using it to substitute `Reader`:
typealias ReaderT<F, D, A> = Kleisli<F, D, A>
typealias Reader<D, A> = ReaderT<Id, D, A]>

`Kleisli` is a higher level abstraction over the `Reader` behavior, so the `ReaderT` is going to be aliased to `Kleisli` and the `Reder` will be just an implementation of the `ReaderT` using the `Id` (identity) as it's monad value.

[Here are](http://typelevel.org/cats/datatypes/kleisli.html) the **typelevel/cats** docs for `Kleisli`, for more details.


kats/src/main/kotlin/kats/Reader.kt
```diff
@@ -0,0 +1,16 @@
+package kats
+
+/**
+ * Basic implementation of the Reader monad. Provides an "implicit" context (configuration) for
+ * function execution. Intended to provide Dependency Injection.
+ */
+class Reader<C, A>(val rd: (C) -> A) {
```
JMPergar: `A` should be covariant and the `C` contravariant and both types should be hierarchies from `Any`

kats/src/main/kotlin/kats/Reader.kt
```diff
@@ -0,0 +1,16 @@
...
+
+  inline fun <B> map(crossinline f: (A) -> B): Reader<C, B> = Reader { c -> f(rd(c)) }
+
+  inline fun <B> flatMap(crossinline f: (A) -> Reader<C, B>): Reader<C, B> = Reader { c ->
+    f(rd(c)).rd(c)
+  }
+
+  fun run(c: C) = rd(c)
```
JMPergar: I think it will be clearer if `rd` param name was `run` and `run` method was named `apply`.

kats/src/main/kotlin/kats/Reader.kt
JMPergar: Will be interesting add `zip` combinator

JMPergar: An extension function to transform function with one parameter in `Reader` will be a great point.

`fun <A, B> ((A) -> B).reader()`

kats/src/main/kotlin/kats/Reader.kt
raulraja: Reader is a transformer and its evaluation should be explicit. Current conventions are run or value, being run the most used one

kats/src/main/kotlin/kats/Reader.kt
raulraja: The most important missing combinator is 'local' which allows switching the environment to unify two different dependencies so you can compose readers with different dependencies.

raulraja:
We need the Id monad before this one can be properly finished. Once the hkt support is in this should become a ReaderT of ID but we need Id first or this reader won't be useful in apps that operate async over a Future or Task and not just over a sync context

kats/src/main/kotlin/kats/Reader.kt
JorgeCastilloPrz: * Fixing `A` to declare it as `out` as it's definitely a covariant type.
* Type `C` **cannot** be contravariant (`in`), as it's occurring also in an `out` position in the `flatMap()` method argument: `crossinline f: (A) -> Reader<C, B>`
* Fixing `A`, `C` and `B` types to make them extend from `Any`. (Would be cool to have an explanation of why exactly do you need this top type bound to be imposed here as I don't understand it completely.

JorgeCastilloPrz: Added `local` combinator @raulraja

JorgeCastilloPrz: I have added two pure methods inside of a `companion object Factory` right now to achieve two different ways of Lifting:
* One to lift a plain A value to a reader wrapping it inside of a function that takes no parameters at all.
* Another one to lift a read (`rd`) function to a Reader.

kats/src/main/kotlin/kats/Reader.kt
JorgeCastilloPrz: I will keep `run` for the moment as it's kind of the standard I have seen in other languages used for the run method of Readers.

JorgeCastilloPrz:
Can you both @JMPergar and @raulraja review the two pure lifting methods I have just added please ? 

Thanks guys

JorgeCastilloPrz:
Done @raulraja. I still have a doubt over it. I wanted to declare it as inline but the `rd` method usage is used from a not `inline` place so compiler forced me to tag the argument as `noinline`. I am not really convinced about this one tbh.

katz/src/main/kotlin/katz/Reader.kt
```diff
@@ -0,0 +1,40 @@
...
+   * you can compose readers with different type dependencies.
+   *
+   * D: type represents a bigger context than C.
+   * @param fd: function to convert from the bigger context D to a context of type C.
+   */
+  inline fun <D : Any> local(crossinline fd: (D) -> C): Reader<D, A> = Reader { bc ->
+    rd(fd(bc))
+  }
+
+  fun run(c: C) = rd(c)
```
raulraja: If you rename 'rd' to 'run' we can remove this

JorgeCastilloPrz: How would we run the reader then ? I mean, the `invoke` method is not running the reader by itself but returning a new reader from a given function passed in. Should than one run rd() directly or something ?

raulraja:
Since run its a function it can just be applied via reader.run(d), just
need to make run a public field.

JorgeCastilloPrz:
Oh, true story hehe. Renaming it. All the properties are public by default in kotlin so shouldn't need further changes.

katz/src/main/kotlin/katz/Reader.kt
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Basic implementation of the Reader monad. Provides an "implicit" context (configuration) for
+ * function execution. Intended to provide Dependency Injection.
+ */
+class Reader<C : Any, out A : Any>(val run: (C) -> A) {
+
+    inline fun <B : Any> map(crossinline fa: (A) -> B): Reader<C, B> = Reader { c -> fa(run(c)) }
+
+    inline fun <B : Any> flatMap(crossinline fa: (A) -> Reader<C, B>): Reader<C, B> = Reader { c ->
```
JMPergar: This is the way if you want set `C` as contravariant type:

```
inline fun <C : Any, A : Any, B : Any> Reader<C, A>.flatMap(crossinline fa: (A) -> Reader<C, B>): Reader<C, B> = 
        Reader { c -> fa(run(c)).run(c) }
```

Remove `flatMap` and implement it as an extension function.

katz/src/main/kotlin/katz/Reader.kt
```diff
@@ -0,0 +1,38 @@
+package kats
```
JMPergar: You forget the `Copyright (C) 2017 The Kats Authors`

katz/src/main/kotlin/katz/Reader.kt
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    /**
+     * local combinator allows switching the environment to unify two different dependency types, so
+     * you can compose readers with different type dependencies.
+     *
+     * D: type represents a bigger context than C.
+     * @param fd: function to convert from the bigger context D to a context of type C.
+     */
+    inline fun <D : Any> local(crossinline fd: (D) -> C): Reader<D, A> = Reader { d ->
```
JMPergar: And `zip` combinator?

katz/src/main/kotlin/katz/Reader.kt
```diff
@@ -0,0 +1,38 @@
...
+         * Lifts an A value to Reader wrapping it in a supplier function with a Nothing argument.
+         */
+        fun <C: Any, A : Any> pure(a: A): Reader<C, A> = Reader { _ -> a }
+
+        /**
+         * Lifts read function to Reader.
+         */
+        inline operator fun <C : Any, A : Any> invoke(noinline rd: (C) -> A) = Reader(rd)
+    }
+}
```
JMPergar: What do you think about the idea of this extension function?

```
fun <A : Any, B : Any> ((A) -> B).reader() : Reader<A, B> = Reader(this)
```

katz/src/main/kotlin/katz/Reader.kt
```diff
@@ -0,0 +1,38 @@
...
+
+        /**
+         * Lifts an A value to Reader wrapping it in a supplier function with a Nothing argument.
+         */
+        fun <C: Any, A : Any> pure(a: A): Reader<C, A> = Reader { _ -> a }
+
+        /**
+         * Lifts read function to Reader.
+         */
+        inline operator fun <C : Any, A : Any> invoke(noinline rd: (C) -> A) = Reader(rd)
```
JMPergar: I think this method have the same syntax that the main constructor, in fact, I think this overwrite the constructor, either way, this method doesn't bring any advantage.

raulraja:
@JorgeCastilloPrz Once this is master https://github.com/FineCinnamon/Katz/pull/48#pullrequestreview-29470511 you should be able to implement `Kleisli` which is `ReaderT`. and the father of all monads. then you can just set :
```
typealias ReaderT<F, D, A> = Kleisli<F, D, A>
typealias Reader<D, A> = ReaderT<Id, D, A]>
```

katz/src/main/kotlin/katz/Reader.kt
JorgeCastilloPrz: added

katz/src/main/kotlin/katz/Reader.kt
JorgeCastilloPrz: added, still working on tests

katz/src/main/kotlin/katz/Reader.kt
JorgeCastilloPrz: What's the benefit on declaring `C` as contravariant ? I have just added `zip` combinator, where `C` is used again as a covariant type (like in flatMap), so not really sure about whether it still makes sense to make it contravariant.

Can you share your opinion on this @raulraja ?

Oh, and your extension method doesn't seem to work properly with the usage I do of `flatMap` in tests. Your approach seem to lack something @JMPergar.

katz/src/main/kotlin/katz/Reader.kt
JorgeCastilloPrz: Makes sense, added it and a test for it.

JorgeCastilloPrz:
Sure

katz/src/main/kotlin/katz/Reader.kt
```diff
@@ -0,0 +1,56 @@
...
+ * limitations under the License.
+ */
+
+package katz
+
+/**
+ * Basic implementation of the Reader monad. Provides an "implicit" context (configuration) for
+ * function execution. Intended to provide Dependency Injection.
+ */
+class Reader<C : Any, out A : Any>(val run: (C) -> A) {
```
ffgiraldez: we finally take the decision to remove the `X: Any` from Generic

katz/src/main/kotlin/katz/Reader.kt
```diff
@@ -0,0 +1,56 @@
...
+    companion object Factory {
+
+        /**
+         * Lifts an A value to Reader wrapping it in a supplier function with a Nothing argument.
+         */
+        fun <C : Any, A : Any> pure(a: A): Reader<C, A> = Reader { _ -> a }
+    }
+}
+
+fun <A : Any, B : Any> ((A) -> B).reader() : Reader<A, B> = Reader(this)
```
ffgiraldez: ❤️  this 👍

katz/src/main/kotlin/katz/Reader.kt
```diff
@@ -0,0 +1,56 @@
...
+ */
+class Reader<C : Any, out A : Any>(val run: (C) -> A) {
+
+    inline fun <B : Any> map(crossinline fa: (A) -> B): Reader<C, B> = Reader { c -> fa(run(c)) }
+
+    inline fun <B : Any> flatMap(crossinline fa: (A) -> Reader<C, B>): Reader<C, B> = Reader { c ->
+        fa(run(c)).run(c)
+    }
+
+    fun <B : Any> zip(other: Reader<C, B>): Reader<C, Pair<A, B>> =
```
ffgiraldez: Should be Tuple2 once #51 merged

katz/src/main/kotlin/katz/Reader.kt
JorgeCastilloPrz: Cool. I will remove it but can I know the reasons please?

katz/src/main/kotlin/katz/Reader.kt
JorgeCastilloPrz: Yep. I am waiting for it!

katz/src/main/kotlin/katz/Reader.kt
JorgeCastilloPrz: JMP's idea :)

katz/src/main/kotlin/katz/Reader.kt
ffgiraldez: was discussed  in slack, i will try to find it send you the infor over there

JorgeCastilloPrz:
@raulraja can you take a look at the `Kleisli` early impl I have just pushed ? 

I have also added the following typealias: ``typealias ReaderT<F, D, A> = Kleisli<F, D, A>`

So it would be cool to know if the direction is correct to start thinking about removing the reader and start using for it the:
`typealias Reader<D, A> = ReaderT<Id, D, A]>`

katz/src/main/kotlin/katz/data/Kleisli.kt
```diff
@@ -0,0 +1,18 @@
+package katz.data
+
+import katz.Functor
+import katz.HK
+import katz.Monad
+
+class Kleisli<F, D, out A>(val run: (D) -> HK<F, A>) {
```
raulraja: Looks good! we should add `pure`, `ask`, `flatten` and any other combinators that may be useful. In addition we need a monad instance for this.

JorgeCastilloPrz: I have just removed `Reader`, added the `Reader alias to `ReaderT<Id, D, A>`, and added more combinators like ask, pure and flatten.

I am facing a problem with a simple extension function that we had for the `Reader` to lift it from any `(A) -> B` function. I need to lift a method like `d -> HKT<F,A>` from the function to return a proper reader instance using the alias, but I have not been able to do it yet. Some help on that one please ?

JorgeCastilloPrz:
I have removed Reader, added Reader alias to ReaderT<Id, D, A>, and more combinators and functions like ask, pure and flatten.

I am facing a problem with a simple extension function that we had for the Reader to lift it from any (A) -> B function. I want to port that behavior to the new way of declaring readers through the alias (and Kleisli).
I need to return a method like d -> HKT<F,A> from the function to return a proper reader instance using the alias, but I have not been able to do it yet. Some help on that one please ?

Everything is pushed.

JorgeCastilloPrz:
Ok. I have now removed the typealias for the Reader and recreated it as an extension of the `Kleisli` forwarding all it's combinators to it. Should be working.

Tests do not need to change as the `Reader` public API for the final user is equivalent to the one I already had.

There is some compilation error on `Coprodukt` class so I am not able to run tests 😓

JorgeCastilloPrz:
Thanks for your help 👍

PR #23, opened 2017-03-24
Initial NonEmptyList implementation

jrgonzalezg:
Taken from https://github.com/FineCinnamon/Bandhook-Kotlin/blob/rr-fp-refactor/core/src/main/java/com/github/finecinnamon/predef.kt

### DIFF
--- a/katz/src/main/kotlin/katz/NonEmptyList.kt
+++ b/katz/src/main/kotlin/katz/NonEmptyList.kt
@@ -38,11 +38,11 @@ class NonEmptyList<out A> private constructor(
 
     fun isEmpty(): Boolean = false
 
-    inline fun <reified B> map(f: (A) -> B): NonEmptyList<B> =
+    fun <B> map(f: (A) -> B): NonEmptyList<B> =
             NonEmptyList(f(head), tail.map(f))
 
-    inline fun <reified B> flatMap(f: (A) -> List<B>): NonEmptyList<B> =
-            unsafeFromList(all.flatMap(f))
+    fun <B> flatMap(f: (A) -> List<B>): NonEmptyList<B> =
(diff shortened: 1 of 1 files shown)


katz/src/main/kotlin/katz/NonEmptyList.kt
```diff
@@ -0,0 +1,77 @@
...
+
+package katz
+
+/**
+ * A List that can not be empty
+ */
+class NonEmptyList<out A> private constructor(
+        override val head: A,
+        override val tail: List<A>,
+        val all: List<A>) : NonEmptyCollection<A> {
```
raulraja: I think we should remove all references to being part of the  colections interface since we want to encourage proper APIs and not those broken java ones.

katz/src/main/kotlin/katz/NonEmptyList.kt
```diff
@@ -0,0 +1,77 @@
...
+        return all.hashCode()
+    }
+
+    override fun toString(): String {
+        return "NonEmptyList(all=$all)"
+    }
+
+    companion object Factory {
+        inline fun <reified A> of(head: A, vararg t: A): NonEmptyList<A> = NonEmptyList(head, t.asList())
+        fun <A> unsafeFromList(l: List<A>): NonEmptyList<A> = NonEmptyList(l)
```
raulraja: Do we need this at all?

jrgonzalezg: Removed the inheritance from the kotlin-stdlib collections on a new commit. The factory methods are used on some of the other methods and may be needed to create immutable lists so i am not sure if we should remove them.

jrgonzalezg:
Removed access to unsafe construction from outside too

raulraja:

PR #25, opened 2017-03-24
Create StateT type

### LINKED ISSUE #14: Create StateT type
This implementation will be a simplification as we don't have HKTs yet

arturogutierrez:
Fixes #14 

Super basic example of State Monad:

```kotlin
// State for a random 6-dice roll
val dice = State<Random, Int> { r -> Pair(r, r.nextInt(6) + 1) }
// Applying a +3 modifier to the roll
val modDice = dice.map { n -> (n + 3).coerceAtMost(6) } // return a new State<Random, Int> generating only 4-5-6 possible dice values
// Roll example
val d1 = modDice.runA(Random())
```


arturogutierrez:
PR updated! :)

katz/src/main/kotlin/katz/data/State.kt
```diff
@@ -0,0 +1,48 @@
...
+package katz
+
+/**
+ * State[S, A] is basically a function S => (S, A),
+ * where S is the type that represents your state
+ * and A is the result the function produces. In addition
+ * to returning the result of type A, the function
+ * returns a new S value, which is the updated state.
+ */
+class State<S, out A>(val runF: (S) -> Pair<S, A>) {
```
ffgiraldez: please use Tuple2 instead of Pair

ffgiraldez: make it implement `HK2<State.F,S,A>`

arturogutierrez: AFAIK, StateT is who has to support HKT, right? This State is using Identity as Monad Environment, isn't?

arturogutierrez: About Tuple2, I will change it :)

raulraja:
Now that we have `Id` instead of `State` this type should be `StateT`. StateT has three type args. 
`F`, `S` and `A` where F is some other monad. once StateT is implemented you can derive `State` by 
`type State[S, A] = StateT[Id, S, A]`.
StateT allows us to use state in other context beside a plain environment such as `Id` for example `Future`.

katz/src/main/kotlin/katz/data/State.kt
arturogutierrez: Raul's comment was very clear 😄 
I will try to implement `StateT` 💪

katz/src/main/kotlin/katz/data/State.kt
```diff
@@ -0,0 +1,48 @@
...
+    fun run(s: S): Tuple2<S, A> {
+        return runF(s)
+    }
+
+    fun runA(s: S): A = run(s).b
+
+    fun runS(s: S): S = run(s).a
+
+    fun <B> map(f: (A) -> B): State<S, B> {
+        return State { s1 ->
```
JorgeCastilloPrz: you can remove this return by using `fun <B> map(f: (A) -> B) = State<S, B> { ... }` syntax

raulraja:
@arturogutierrez take a look at how `Reader` integrates with `Kleisli` now in master, you are gonna have to do something similar for StateT and State.

ffgiraldez:
I will gonna update this PR soon

ffgiraldez:
i see that map2 and map2Eval was defined here in [cats](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/StateT.scala#L274)

Now that @raulraja are pushing laws i will not do more test on typeclasses and focus on help him to finish it.

PR #29, opened 2017-03-25
Remove support for null in our types

JMPergar:
Our types should accept only types non-nullable.

giphy

### DIFF
--- a/katz/src/main/kotlin/katz/NonEmptyList.kt
+++ b/katz/src/main/kotlin/katz/NonEmptyList.kt
@@ -19,7 +19,7 @@ package katz
 /**
  * A List that can not be empty
  */
-class NonEmptyList<out A> private constructor(
+class NonEmptyList<out A: Any> private constructor(
         val head: A,
         val tail: List<A>,
         val all: List<A>) {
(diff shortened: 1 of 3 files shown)


katz/src/main/kotlin/katz/NonEmptyList.kt
```diff
@@ -66,6 +66,6 @@ class NonEmptyList<out A> private constructor(
     }
 
     companion object Factory {
-        fun <A> of(head: A, vararg t: A): NonEmptyList<A> = NonEmptyList(head, t.asList())
+        fun <A: Any> of(head: A, vararg t: A): NonEmptyList<A> = NonEmptyList(head, t.asList())
     }
 }
```
jrgonzalezg: This change will forbid the creation of NonEmptyLists of nullable types linke NonEmptyLists<String?> which may cause interoperability problems.

katz/src/main/kotlin/katz/NonEmptyList.kt
```diff
@@ -66,6 +66,6 @@ class NonEmptyList<out A> private constructor(
     }
 
     companion object Factory {
-        fun <A> of(head: A, vararg t: A): NonEmptyList<A> = NonEmptyList(head, t.asList())
+        fun <A: Any> of(head: A, vararg t: A): NonEmptyList<A> = NonEmptyList(head, t.asList())
     }
```
jrgonzalezg: I am not sure about the type bound in of(). Since NonEmptyList is already parameterized with <A: Any> it is probably redundant on the companion object function. It may make the code more clear but it can also be seen as a duplicate code / specification. What do you think?

katz/src/main/kotlin/katz/NonEmptyList.kt
JMPergar: The idea is to name `null` as an unwelcome citizen. This system accepts nullable types but in the case, the type will be null the system crash in the moment that the null appears. Fail fast is better IMHO. For the interoperability, I think the way is use method like `fromNullable` on model with types like Option.

katz/src/main/kotlin/katz/NonEmptyList.kt
JMPergar: The `A` type in this function isn't the `A` type defined in the class so if you don't declare `<A: Any>` in the function the compiler will tell you "no no no"

jrgonzalezg: I thought the compiler could infer the restriction but then it is ok for me as is :)

PR #31, opened 2017-03-26
Initial encoding for lightweight HKs and a few typeclasses

raulraja:
This is ready for review and we can use it as base to implement some of the most important typeclasses. Let me know if you have any question so I can merge this and we can fix the Reader and State types.


raulraja:
Ready for review.

katz/src/test/kotlin/katz/OptionTest.kt
```diff
@@ -66,5 +66,32 @@ class OptionTest : UnitSpec() {
...
+        }
+
+        "Option.monad.binding" should "for comprehend over option" {
+            val result = Option.monad().binding {
+                val x = ! Option(1)
+                val y = ! Option(1)
+                yields (x + y)
+            }
+            result shouldBe Option(2)
+        }
```
aballano: This test is duplicated, could be causing the build to fail

raulraja: That's it then, out of a computer if you don't mind checking in a fix otherwise I'll change it tomorrow

katz/src/main/kotlin/katz/Id.kt
```diff
@@ -0,0 +1,31 @@
...
+
+    companion object : IdInstances
+}
+
+/**
+ * Companion interface for [Id] type
+ * outside due to https://youtrack.jetbrains.com/issue/KT-10532
+ */
+private interface IdInstances {
+    fun <A> HK<Id.F, A>.ev(): Id<A> = this as Id<A>
```
pakoito: Could this be inlined?

katz/src/main/kotlin/katz/predef.kt
```diff
@@ -0,0 +1,28 @@
...
+import kotlin.coroutines.experimental.startCoroutine
+
+typealias HK<F, A> = katz.typeclasses.HK<F, A>
+typealias Functor<F> = katz.typeclasses.Functor<F>
+typealias Applicative<F> = katz.typeclasses.Applicative<F>
+typealias ApplicativeError<F, E> = katz.typeclasses.ApplicativeError<F, E>
+typealias Monad<F> = katz.typeclasses.Monad<F>
+typealias MonadError<F, E> = katz.typeclasses.MonadError<F, E>
+
+fun <F, B> Monad<F>.binding(c: suspend MonadContinuation<F, *>.() -> HK<F, B>): HK<F, B> {
```
pakoito: It would be great if these functions were documented a bit.

katz/src/main/kotlin/katz/Id.kt
```diff
@@ -0,0 +1,31 @@
+package katz
+
+class Id<out A>(val value: A): HK<Id.F, A> {
+
+    class F
```
pakoito: I would like to know, does this F need to be a class, or could it be replaced with an object instead?

katz/src/main/kotlin/katz/typeclasses/Monad.kt
```diff
@@ -0,0 +1,50 @@
...
+        returnedMonad = value
+    }
+
+    override fun resumeWithException(exception: Throwable) {
+        throw exception
+    }
+
+    internal lateinit var returnedMonad: HK<F, A>
+
+    operator suspend fun <B> HK<F, B>.not(): B = bind { this }
```
pakoito: huehuehuehue

katz/src/main/kotlin/katz/Id.kt
ffgiraldez: could not be an object
compiler error with object instead of class

the only viable options is use `enum class` instead of `class`

katz/src/main/kotlin/katz/Id.kt
ffgiraldez: cannot inline virtual functions like this one declared inside an interface

katz/src/main/kotlin/katz/Id.kt
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Companion interface for [Id] type
+ * outside due to https://youtrack.jetbrains.com/issue/KT-10532
+ */
+private interface IdInstances {
+    fun <A> HK<Id.F, A>.ev(): Id<A> = this as Id<A>
+
+    fun monad() = object : Monad<Id.F> {
+        override fun <A, B> map(fa: HK<Id.F, A>, f: (A) -> B): HK<Id.F, B> =
```
JMPergar: Will `map` and `flatMap` methods always have the same form? In this case and Option the answers is yes, it's possible extract it to parent interface?

katz/src/main/kotlin/katz/Option.kt
```diff
@@ -134,3 +136,23 @@ sealed class Option<out A> {
...
+ * outside due to https://youtrack.jetbrains.com/issue/KT-10532
+ */
+private interface OptionInstances {
+    fun <A> HK<Option.F, A>.ev(): Option<A> = this as Option<A>
+
+    fun monad() = object : Monad<Option.F> {
+        override fun <A, B> map(fa: HK<Option.F, A>, f: (A) -> B): HK<Option.F, B> =
+                fa.ev().map(f)
+
+        override fun <A> pure(a: A): HK<Option.F, A> = Option.Some(a)
```
ffgiraldez: what do you think about use `fromNullable()` inside `pure()` instead of call directly Option.Some() ?? for that this PR need to be Rebased with master

katz/src/main/kotlin/katz/Option.kt
```diff
@@ -125,6 +125,8 @@ sealed class Option<out A> {
     object None : Option<Nothing>() {
         override val isEmpty = true
     }
+
+
```
JMPergar: Ejem! TwoBlankLinesException!! We need to add Detekt to the build system, it's official XD

katz/src/main/kotlin/katz/Option.kt
```diff
@@ -134,3 +136,23 @@ sealed class Option<out A> {
...
+/**
+ * Companion interface for [Option] type
+ * outside due to https://youtrack.jetbrains.com/issue/KT-10532
+ */
+private interface OptionInstances {
+    fun <A> HK<Option.F, A>.ev(): Option<A> = this as Option<A>
+
+    fun monad() = object : Monad<Option.F> {
+        override fun <A, B> map(fa: HK<Option.F, A>, f: (A) -> B): HK<Option.F, B> =
```
JMPergar: Question:

- `Applicative` implements `map` method and delegates its behaviour to  `ap` and `pure` methods.
- `Monad` implement `ap` method and delegates his behaviour to `flatMap` method.
- These kinds of interfaces implements `pure` and `flatMap` behaviours.

In this case, is override `map` method needed?

katz/src/main/kotlin/katz/typeclasses/Applicative.kt
```diff
@@ -0,0 +1,19 @@
+package katz.typeclasses
+
+import katz.typeclasses.Functor
```
JMPergar: These two imports are unused.

katz/src/main/kotlin/katz/Option.kt
jrgonzalezg: From nullable does conversions, we discussed this the other day, the pure implementation of pure(null) should be Some(null) like it is here

katz/src/main/kotlin/katz/predef.kt
```diff
@@ -0,0 +1,28 @@
...
+import katz.typeclasses.Monad
+import katz.typeclasses.MonadError
+import kotlin.coroutines.experimental.startCoroutine
+
+typealias HK<F, A> = katz.typeclasses.HK<F, A>
+typealias Functor<F> = katz.typeclasses.Functor<F>
+typealias Applicative<F> = katz.typeclasses.Applicative<F>
+typealias ApplicativeError<F, E> = katz.typeclasses.ApplicativeError<F, E>
+typealias Monad<F> = katz.typeclasses.Monad<F>
+typealias MonadError<F, E> = katz.typeclasses.MonadError<F, E>
```
jrgonzalezg: What is the purpose of this type aliases? Aren't they equivalent to the import?

katz/src/main/kotlin/katz/typeclasses/Applicative.kt
```diff
@@ -0,0 +1,19 @@
...
+    fun <A, B> ap(fa: HK<F, A>, ff: HK<F, (A) -> B>): HK<F, B>
+
+    fun <A, B> product(fa: HK<F, A>, fb: HK<F, B>): HK<F, Pair<A, B>> =
+            ap(fb, map(fa) { a: A -> { b: B -> Pair(a, b) } })
+
+    override fun <A, B> map(fa: HK<F, A>, f: (A) -> B): HK<F, B> = ap(fa, pure(f))
+
+    fun <A, B, Z> map2(fa: HK<F, A>, fb: HK<F, B>, f: (Pair<A, B>) -> Z): HK<F, Z> =
+            map(product(fa, fb), f)
+}
```
jrgonzalezg: Cool stuff :D

katz/src/main/kotlin/katz/typeclasses/ApplicativeError.kt
```diff
@@ -0,0 +1,20 @@
+package katz.typeclasses
+
+import katz.Either
+import katz.typeclasses.HK
```
JMPergar: Unused imports, please remove it

katz/src/main/kotlin/katz/typeclasses/Monad.kt
jrgonzalezg: I found ! potentially confusing too. It is a pity we don't have <-

katz/src/main/kotlin/katz/Option.kt
JMPergar: And if this is correct, the main `map` method could be delegated to the `monad map method`, no?

katz/src/main/kotlin/katz/Id.kt
```diff
@@ -0,0 +1,31 @@
...
+}
+
+/**
+ * Companion interface for [Id] type
+ * outside due to https://youtrack.jetbrains.com/issue/KT-10532
+ */
+private interface IdInstances {
+    fun <A> HK<Id.F, A>.ev(): Id<A> = this as Id<A>
+
+    fun monad() = object : Monad<Id.F> {
```
jrgonzalezg: How many objects does this create? One for the whole application or one per call? If it is one per call it will be probably better to extract it to a top level object.

katz/src/main/kotlin/katz/Id.kt
jrgonzalezg: It seems to compile with object too in my case. But what are the advantages of declaring it as an object? It will cause the existence of an instance and we probably do not need any.

katz/src/main/kotlin/katz/Id.kt
JMPergar: Yoy say in this way?

```
private interface OptionInstances : Monad<Option.F> {
    fun <A> HK<Option.F, A>.ev(): Option<A> = this as Option<A>
    
    override fun <A, B> map(fa: HK<Option.F, A>, f: (A) -> B): HK<Option.F, B> =
            fa.ev().map(f)

    override fun <A> pure(a: A): HK<Option.F, A> = Option.Some(a)

    override fun <A, B> flatMap(fa: HK<Option.F, A>, f: (A) -> HK<Option.F, B>): HK<Option.F, B> =
                fa.ev().flatMap { f(it).ev() }
}
```

I asked myself the same question.

katz/src/main/kotlin/katz/typeclasses/Monad.kt
ffgiraldez: i like the infix fun` await`
```kotlin
val x = await Option(1)
```

katz/src/main/kotlin/katz/typeclasses/Monad.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    infix suspend fun <B> bind(m: HK<F, B>): B = bind { m }
+
+    infix suspend fun <B> bind(m: () -> HK<F, B>): B = suspendCoroutineOrReturn { c ->
+            returnedMonad = M.flatMap(m(), { x ->
+                c.resume(x)
+                returnedMonad
+            })
+            COROUTINE_SUSPENDED
+        }
```
jrgonzalezg: Does this allow for asynchronous code? Isn't c.resume(x) called immediately/synchronously? I think we need at least a test of the do notation where one bind depends on the result value of another bind.

katz/src/main/kotlin/katz/predef.kt
```diff
@@ -0,0 +1,28 @@
+package katz
```
JMPergar: What is the convention for named this file in lowercase?

katz/src/main/kotlin/katz/typeclasses/Monad.kt
raulraja: Do you know where await is defined?

katz/src/main/kotlin/katz/Id.kt
raulraja: nope, they won't some types like `Ior` will require evidence of other typeclasses which change the method signature

katz/src/main/kotlin/katz/Option.kt
raulraja: overrides are necessary for performance in some cases and in others for behavior with non deterministic instances such as Applicative on Future to sopport paralellism, this is the standard impl from cats and others.

katz/src/main/kotlin/katz/typeclasses/Monad.kt
ffgiraldez: `await` it's not defined in Kotlin, just as @pakoito it's generally acceptable because C# define it as keyword for their corutines(async/await). but testing a bit around this i cannot properly infix it
```Kotlin
infix suspend fun <B> await(kind: HK<F, B>): B = bind { kind }

val x: Int = await Option(1) //Compiler cannot infer the type and expect varargs, 
```
But we can make an extension function that only be visible inside comprehension

```Kotlin
open class MonadContinuation<F, A>(val M: Monad<F>) : Serializable, Continuation<HK<F, A>> {
...
suspend fun <B>  HK<F, B>.await(): B = bind { this }
...
}
val result = Option.monad().binding {
  val x = Option(1).await()
  val y = Option(1).await()
  yields(x + y)
}

Option(3).await() // not compile await() not defined
```

WDYT guys?

katz/src/main/kotlin/katz/Option.kt
raulraja: Also sometimes you want to provide different instances for `Future<Option>` and `Monad<Option>`in some cases and you don't want to necessarily use inheritance until we have a solid strategy to receive instances in a lang without implicits.

katz/src/main/kotlin/katz/predef.kt
raulraja: by having these aliases here we can have the code organized in whichever package structure we want and provide users with an uber import `import katz.*` and also give them the option to import them a la carte.

katz/src/main/kotlin/katz/Id.kt
raulraja: as far as I know there can only be one `object` so this returns the same instance each time and if not we have to change it. @pakoito do you know?

katz/src/main/kotlin/katz/typeclasses/Monad.kt
```diff
@@ -0,0 +1,50 @@
...
+
+    fun <A, B> flatMap(fa: HK<F, A>, f: (A) -> HK<F, B>): HK<F, B>
+
+    override fun <A, B> ap(fa: HK<F, A>, ff: HK<F, (A) -> B>): HK<F, B> =
+            flatMap(ff, { f -> map(fa, f) })
+}
+
+@RestrictsSuspension
+open class MonadContinuation<F, A>(val M : Monad<F>) : Serializable, Continuation<HK<F, A>> {
+
```
raulraja: convention in FP instances of abstract types are usually capitalized. Also as for the top level vs inside companion we want a consistent API that is predictable when obtaining instances.  I think it's cleanr to say `Option.monad()` than `OptionMonad()` but I'm fine with whatever

katz/src/main/kotlin/katz/typeclasses/Monad.kt
raulraja: This is what it does since async monad such as Future perform their trampoline jump on `flatMap` such as in the case of the Bandhook demo app where this code was originally written. Unless I'm missing something from Coroutines due to my limited understanding this should work for all monads both sync and async.

katz/src/main/kotlin/katz/predef.kt
raulraja: package level objects in scala are usually lowercased and I carried that over but if this is not hipster approved in Kotlin I'll change it! in any case this file contains no instances or classes so a user will never interact with it directly. Now that you asked... What is the reason for camelcasing file names that contains no classes or instantiable objects and that are only pertinent to the compiler scope resolution? :trollface:

katz/src/main/kotlin/katz/typeclasses/Monad.kt
pakoito: I prefer it infix, even if it means a different name altogether.

katz/src/main/kotlin/katz/Id.kt
pakoito: If we want 0 instances then we should make that explicit.

katz/src/main/kotlin/katz/Id.kt
ffgiraldez: in every invocation you create a new instance because the scope it's a function not a class, but it's easy to test

katz/src/main/kotlin/katz/Id.kt
ffgiraldez: if compiler does not complain, should be an object, the worst case it's only one instance of F shared across all instances of the type.

katz/src/main/kotlin/katz/typeclasses/Monad.kt
ffgiraldez: i like capitalized name for HK types

katz/src/main/kotlin/katz/typeclasses/Monad.kt
ffgiraldez: the problem with the infix is that i was not able to compile it :(

katz/src/main/kotlin/katz/Id.kt
jrgonzalezg: Is it possible to make it so it has 0 instances? how? defining it as an abstract class? any other alternative?

jrgonzalezg: It seems Nothing is defined as:
```
public class Nothing private constructor()
```
so defining F as Nothing may be the way.

katz/src/main/kotlin/katz/typeclasses/Monad.kt
jrgonzalezg: I am not sure about this, and i was not sure it was correct on the Bandhook demo either. Maybe making some tests could help as verify it. I would start with a simple one like the one i said where one value depends on another previous value on the same do comprehension to check that sequential operations are working well. But i really don't know what suite of tests is needed to fully cover all uses cases here.

jrgonzalezg: Sorry, i have seen now that you already included more tests, it can be enough with that for now.

katz/src/main/kotlin/katz/Id.kt
jrgonzalezg: If that's the case then it needs to be created as top level object. Maybe we can preserve the monad() shortcut by making it return the top level object instead of an inlined one.

katz/src/main/kotlin/katz/typeclasses/Monad.kt
jrgonzalezg: It seems by @ffgiraldez comment above that the monad() methods are currently producing multiple object instances (see my other comment with respect to this)

raulraja: yeah, I'll change that

katz/src/main/kotlin/katz/Id.kt
pakoito: That's the approach to take, yes.

PR #47, opened 2017-03-27
Added codecov step to travis.

aballano:
Recommended extension: https://github.com/codecov/browser-extension

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,6 @@ jdk:
 
 script:
   ./gradlew build test
+
+after_success:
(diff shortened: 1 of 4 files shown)


ffgiraldez:
check https://github.com/codecov/example-gradle, i think we need to add more changes in order to add Codecov support

aballano:
Apparently the report seems to be fine on local but it fails to report properly once uploaded.

If anyone wants to check, the command to run it locally is this:

`./gradlew check && ./gradlew codeCoverageReport`

PR #48, opened 2017-03-28
Add HKTs for 2-5 types. Implement monad for Either and Ior

pakoito:
I'm not sure about the Ior implementation due to the shitty implicit semigroup I had to grokk for Ior. Also, I'm forced to use `*` for non-bias type over `Nothing` else flatMap couldn't be inlined due to the reified type C.

Other than some phantom UI errors by the Kotlin parser it seems to pass the test so...yeah. I'm happy to finally contribute to the project :D

### DIFF
--- a/katz/src/main/kotlin/katz/instances/IorMonad.kt
+++ b/katz/src/main/kotlin/katz/instances/IorMonad.kt
@@ -16,17 +16,12 @@
 
 package katz
 
-object IorMonad : Monad<HK<Ior.F, Nothing>> {
+class IorMonad<A>(val semigroup: Semigroup<A>) : Monad<HK<Ior.F, A>> {
+    override fun <B> pure(b: B): HK2<Ior.F, A, B> =
+            Ior.Right(b)
 
-    override fun <A> pure(a: A): HK2<Ior.F, Nothing, A> =
-            Ior.Right(a)
-
-    override fun <A, B> flatMap(fa: HK2<Ior.F, Nothing, A>, f: (A) -> HK2<Ior.F, Nothing, B>): HK2<Ior.F, Nothing, B> =
-            fa.ev().flatMap(combine()) { f(it).ev() }
-
-    fun <A> combine(): Semigroup<A> = object : Semigroup<A> {
-        override fun combine(a: A, b: A): A = combineAll(a, b)
-    }
+    override fun <B, C> flatMap(fa: HK<HK<Ior.F, A>, B>, f: (B) -> HK<HK<Ior.F, A>, C>): HK<HK<Ior.F, A>, C> =
+            fa.ev().flatMap(semigroup as Semigroup<Any?>) { f(it).ev() } as HK2<Ior.F, A, C>
 }
 
-fun <A> HK2<Ior.F, Nothing, A>.ev(): Ior<Nothing, A> = this as Ior<Nothing, A>
+fun <A> HK2<Ior.F, *, A>.ev(): Ior<*, A> = this as Ior<*, A>
--- a/katz/src/main/kotlin/katz/instances/IorMonad.kt
+++ b/katz/src/main/kotlin/katz/instances/IorMonad.kt
@@ -1,0 +1,33 @@
+/*
+ * Copyright (C) 2017 The Katz Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


katz/src/main/kotlin/katz/instances/IorMonad.kt
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package katz
+
+object IorMonad: Monad<HK<Ior.F, *>> {
+
+    override fun <A> pure(a: A): HK2<Ior.F, *, A> =
+            Ior.Right(a).ev()
+
```
jrgonzalezg: Remove .ev() here

katz/src/main/kotlin/katz/instances/IorMonad.kt
```diff
@@ -0,0 +1,32 @@
...
+    override fun <A, B> flatMap(fa: HK2<Ior.F, *, A>, f: (A) -> HK2<Ior.F, *, B>): HK2<Ior.F, *, B> =
+            fa.ev().flatMap(combine()) { f(it).ev() }
+
+    fun <A> combine(): Semigroup<A> = object: Semigroup<A> {
+        override fun combine(a: A, b: A): A = a
+    }
```
jrgonzalezg: Maybe a1 and a2 would be better as parameter names in combine

katz/src/main/kotlin/katz/typeclasses/HigherKinds.kt
```diff
@@ -18,3 +18,10 @@ package katz
 
 interface HK<out F, out A>
 
+typealias HK2<F, A, B> = HK<HK<F, A>, B>
+
+typealias HK3<F, A, B, C> = HK<HK2<F, A, B>, C>
+
+typealias HK4<F, A, B, C, D> = HK<HK3<F, A, B, C>, D>
+
+typealias HK5<F, A, B, C, D, E> = HK<HK4<F, A, B, C, D>, E>
```
jrgonzalezg: Cool! Higher Kinded Currying :D

katz/src/main/kotlin/katz/instances/IorMonad.kt
jrgonzalezg: It also seems that this may cause problems since the combine here is not combining at all. The problem is that the flatMap on Ior has a different signature than the one in Monad. I am not sure how we could overcome that on a proper way without having real implicits.

katz/src/test/kotlin/katz/data/IorTest.kt
```diff
@@ -132,5 +133,26 @@ class IorTest : UnitSpec() {
 
         }
 
+        "Ior.monad.flatMap should be consistent with Ior#flatMap" {
+            forAll { a: Int ->
+                val implicit: Semigroup<Number> = object: Semigroup<Number> {
+                    override fun combine(a: Number, b: Number): Number = a
+                }
```
jrgonzalezg: This implicit may be flaky or different than the one in IorMonad. You could use:
```
val implicit: Semigroup<Int> = IorMonad.combine<Int>()
```
to ensure both alternatives use the same combine. Despite of that, the approach to combining probably requires more changes (see my other comments)

katz/src/main/kotlin/katz/instances/IorMonad.kt
pakoito: This function is never visited, it only flatMaps on right, which is why any implementation is valid.

ffgiraldez: Also flatmap Ior.Both with an Ior.Left

pakoito: Wait, no.

pakoito: Yeah you're right, lemme fix.

katz/src/main/kotlin/katz/typeclasses/HigherKinds.kt
raulraja: :clap:

katz/src/test/kotlin/katz/data/IorTest.kt
raulraja: You can't create lawful monoids for `Double` and `Float` https://github.com/scalaz/scalaz/issues/334. Also to avoid confusion we should not name instances `implicits` since these are not real implicits and explicitly passed as args. I think for this anonymous instance a better name would be `intMonoidInstance` or just `IM`

raulraja:
@pakoito Excellent work :clap:

katz/src/test/kotlin/katz/data/IorTest.kt
pakoito: I got the name from the existing test above. I'll modify both now.

jrgonzalezg: I agree with the confusion of implicit. Please also rename the other implicit named things on IorTest like the ones i added (on master now).

PR #50, opened 2017-03-28
Improved / simplified flatMap on Ior and added tests for checking it …

jrgonzalezg:
…supports upper bounds on A

### DIFF
--- a/katz/src/main/kotlin/katz/data/Ior.kt
+++ b/katz/src/main/kotlin/katz/data/Ior.kt
@@ -266,15 +266,15 @@ sealed class Ior<out A, out B> {
  *
  * @param f The function to bind across [Ior.Right].
  */
-inline fun <A, B, reified C : A, D> Ior<A, B>.flatMap(aa: Semigroup<C>, crossinline f: (B) -> Ior<C, D>): Ior<C, D> = when (this) {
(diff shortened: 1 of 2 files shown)


katz/src/main/kotlin/katz/data/Ior.kt
```diff
@@ -266,17 +266,17 @@ sealed class Ior<out A, out B> {
  *
  * @param f The function to bind across [Ior.Right].
  */
-inline fun <A, B, reified C : A, D> Ior<A, B>.flatMap(aa: Semigroup<C>, crossinline f: (B) -> Ior<C, D>): Ior<C, D> = when (this) {
-    is Ior.Left -> Ior.Left(value as C)
+inline fun <A, B, D> Ior<A, B>.flatMap(AA: Semigroup<A>, crossinline f: (B) -> Ior<A, D>): Ior<A, D> = when (this) {
```
pakoito: This solves a problem on #48. I'll revisit it after this gets merged :D

Could you please use a name that's lowercase?

jrgonzalezg: If i understood raúl correctly it is a convention to use upper case letter for instances of monads (see https://github.com/FineCinnamon/Katz/pull/31/files#r108131007). As for merging as soon as you approve :D

PR #52, opened 2017-03-29
Implement OptionT as HK2

ffgiraldez:
use tipealias for HKX

### DIFF
--- a/katz/src/main/kotlin/katz/instances/IdMonad.kt
+++ b/katz/src/main/kotlin/katz/instances/IdMonad.kt
@@ -18,12 +18,12 @@ package katz
 
 object IdMonad : Monad<Id.F> {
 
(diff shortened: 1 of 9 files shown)


katz/src/main/kotlin/katz/instances/IdMonad.kt
```diff
@@ -18,12 +18,12 @@ package katz
 
 object IdMonad : Monad<Id.F> {
 
-    override fun <A, B> map(fa: HK<Id.F, A>, f: (A) -> B): HK<Id.F, B> =
+    override fun <A, B> map(fa: HK<Id.F, A>, f: (A) -> B): Id<B> =
             fa.ev().map(f)
 
-    override fun <A> pure(a: A): HK<Id.F, A> = Id(a)
+    override fun <A> pure(a: A): Id<A> = Id(a)
```
raulraja: Always take as args as generic as you can and return as concrete as you can is a good rule of thumb

PR #53, opened 2017-03-29
Add OptionT and OptionTMonad datatype

ffgiraldez:
Not all combinator are teststed

### DIFF
--- a/katz/src/main/kotlin/katz/instances/IorMonad.kt
+++ b/katz/src/main/kotlin/katz/instances/IorMonad.kt
@@ -22,7 +22,6 @@ class IorMonad<L>(val AA: Semigroup<L>) : Monad<HK<Ior.F, L>> {
 
     override fun <A> pure(a: A): HK2<Ior.F, L, A> = Ior.Right(a)
 
-
 }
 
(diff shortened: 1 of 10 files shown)


katz/src/main/kotlin/katz/instances/IorMonad.kt
```diff
@@ -16,12 +16,12 @@
 
 package katz
 
-class IorMonad<A>(val semigroup: Semigroup<A>) : Monad<HK<Ior.F, A>> {
-    override fun <B> pure(b: B): HK2<Ior.F, A, B> =
-            Ior.Right(b)
+class IorMonad<L>(val AA: Semigroup<L>) : Monad<HK<Ior.F, L>> {
+    override fun <A, B> flatMap(fa: HK<HK<Ior.F, L>, A>, f: (A) -> HK<HK<Ior.F, L>, B>): HK<HK<Ior.F, L>, B> =
```
raulraja: Ok to do it here or some other PR but the return types here should also be concrete instead of returning the HK version

ffgiraldez: we could create an issue to change that we have to return concrete types and create and use typealias like in Coproduct

PR #54, opened 2017-03-30
Progress toward Coproduct, FunctionK, and Inject, next Free :)

raulraja:
@pakoito This is ready and can't really test the properties in this clases until we have something like `discipline`. I'm gonna attempt to implement `Free` after this one and there we can add some test examples over Free algebras that use Coproduct, Inject and others.

### DIFF
--- a/katz/src/main/kotlin/katz/arrow/FunctionK.kt
+++ b/katz/src/main/kotlin/katz/arrow/FunctionK.kt
@@ -16,12 +16,18 @@
 
 package katz
 
-interface Functor<F> {
+interface FunctionK<F, G> {
 
-    fun <A, B> map(fa: HK<F, A>, f: (A) -> B): HK<F, B>
+    /**
+     * Applies this functor transformation from `F` to `G`
+     */
+    operator fun <A> invoke(fa: HK<F, A>): HK<G, A>
 
-    fun <A, B> lift(f: (A) -> B): (HK<F, A>) -> HK<F, B> =
-            { fa: HK<F, A> ->
-                map(fa, f)
+    fun <H> or(h: FunctionK<H, G>): FunctionK<CoproductFG<F, H>, G> =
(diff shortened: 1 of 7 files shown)


katz/src/main/kotlin/katz/arrow/FunctionK.kt
```diff
@@ -0,0 +1,33 @@
...
+package katz
+
+interface FunctionK<F, G> {
+
+    /**
+     * Applies this functor transformation from `F` to `G`
+     */
+    operator fun <A> invoke(fa: HK<F, A>): HK<G, A>
+
+    fun <H> or(h: FunctionK<H, G>): FunctionK<CoproductFG<F, H>, G> =
```
pakoito: I believe this will work better as an extension function, if there are no impediments. That way the interface is kept clean from default implementations.

katz/src/main/kotlin/katz/typeclasses/Functor.kt
```diff
@@ -17,5 +17,11 @@
 package katz
 
 interface Functor<F> {
+
     fun <A, B> map(fa: HK<F, A>, f: (A) -> B): HK<F, B>
+
+    fun <A, B> lift(f: (A) -> B): (HK<F, A>) -> HK<F, B> =
```
pakoito: I haven't checked the rest of the repository, so same as above in general I'm weary of default implementations on interfaces because they enable mixin-like composition which I find can lead to messy composition.

Maybe it's just personal bias based off gut feel, but as long as we have an easy replacement in extension functions I would push for it. Drawback: interop with Java would be with statics, but AFAIK that's not this project's concern.

raulraja: In this case `lift` is something that instances may override. Same with `map` and other similar combinators. In the case of `or` in `FunctionK` is different because that is a utility function frequently used to build interpreters

PR #56, opened 2017-04-03
Create and use type alias for HK concrete clases and small fixes

ffgiraldez:
Fix variance on FuntionK.
Suppress unused parameter Aplicative.kt due to dummyparamenter hack

Typealases created:
- EitherKind
- IdKind
- IorKind
- OptionKind

Also return concrete classes instead of HK counterpart on Monads

### DIFF
--- a/katz/src/main/kotlin/katz/arrow/FunctionK.kt
+++ b/katz/src/main/kotlin/katz/arrow/FunctionK.kt
@@ -16,7 +16,7 @@
 
 package katz
 
-interface FunctionK<F, G> {
+interface FunctionK<in F, out G> {
 
     /**
      * Applies this functor transformation from `F` to `G`
--- a/katz/src/main/kotlin/katz/typeclasses/Applicative.kt
+++ b/katz/src/main/kotlin/katz/typeclasses/Applicative.kt
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+@file:Suppress("UNUSED_PARAMETER")
+
 package katz
 
 interface Applicative<F> : Functor<F> {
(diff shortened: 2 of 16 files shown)


katz/src/main/kotlin/katz/typeclasses/Applicative.kt
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+@file:Suppress("UNUSED_PARAMETER")
```
raulraja: what is this for?

ffgiraldez: Suppress the compile warning for using dummy implicits to hack the type erasure
```
w: ../katz/katz/src/main/kotlin/katz/typeclasses/Applicative.kt: (50, 9): Parameter 'dummyImplicit' is never used
```

katz/src/main/kotlin/katz/arrow/FunctionK.kt
```diff
@@ -16,7 +16,7 @@
 
 package katz
 
-interface FunctionK<F, G> {
+interface FunctionK<in F, out G> {
```
raulraja: why can't it be invariant?

ffgiraldez: could be, but compiler warn me about they could have in/out variante in order to be more precise using their types so i add them.

Could let's keep with in/out that at this point don't hurt and remove a lint warning ,we added deteck, but need could need also a lint (if not implode with our generics 🤣 )

pakoito: AFAIK covariant is generally more flexible than invariant, right? Why wouldn't we want it?

raulraja: Not saying we don't just curious as they FunctionK in Scala does not need type variance over it's args

PR #59, opened 2017-04-03
Added test for Try.

aballano:
Fixed a small bug in the way.

Note: I did not use the `forAll` method because I don't see any benefit since we don't operate with the value itself but with the types. Also because adding this method basically makes very difficult to diagnose any test failure.

### DIFF
--- a/katz/src/main/kotlin/katz/data/Try.kt
+++ b/katz/src/main/kotlin/katz/data/Try.kt
@@ -35,16 +35,6 @@ sealed class Try<out A> {
     }
 
     /**
-     * Returns `true` if the `Try` is a `Failure`, `false` otherwise.
-     */
-    abstract val isFailure: Boolean
-
-    /**
-     * Returns `true` if the `Try` is a `Success`, `false` otherwise.
-     */
-    abstract val isSuccess: Boolean
-
-    /**
      * Returns the given function applied to the value from this `Success` or returns this if this is a `Failure`.
      */
     inline fun <B> flatMap(crossinline f: (A) -> Try<B>): Try<B> = fold({ Failure(it) }, { f(it) })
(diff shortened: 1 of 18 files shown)


katz/src/main/kotlin/katz/data/Try.kt
```diff
@@ -80,18 +80,21 @@ sealed class Try<out A> {
...
      */
-    class Success<out A>(val value: A) : Try<A>() {
-        override val isFailure: Boolean = true
-        override val isSuccess: Boolean = false
+    data class Success<out A>(val value: A) : Try<A>() {
+        override val isFailure: Boolean = false
+        override val isSuccess: Boolean = true
     }
+
+    data class PredicateException(override val message: String) : Throwable(message)
```
raulraja: let's make this a sealed hierarchy so consumers matching over the exceptions get full compiler support to deal with all potential datatypes.

aballano: Change done!

katz/src/main/kotlin/katz/data/Try.kt
```diff
@@ -74,26 +78,35 @@ sealed class Try<out A> {
...
 
     /**
      * The `Failure` type represents a computation that result in an exception.
      */
-    class Failure<out A>(val exception: Throwable) : Try<A>() {
-        override val isFailure: Boolean = false
-        override val isSuccess: Boolean = true
+    data class Failure<out A>(val exception: Throwable) : Try<A>() {
+        override val isFailure: Boolean = true
+        override val isSuccess: Boolean = false
```
pakoito: These booleans look like a possible scape hatch we don't really need. They could either be retrieved by a fold, or simply using when.

aballano: Fine for me, I think we put them there as a mirror of the Scala implementation, but I agree that I don't see the usage having other methods. Let's collect some more opinions before removing them, tho.

@FineCinnamon/katzmembers Please vote with :+1: or :-1: to remove those fields.

PR #61, opened 2017-04-06
Global typeclass instances

raulraja:
Runtime based default global instances that autoregister on class loads. 
This may be buggy but it's a step forward at getting instances from reified parameters when needed.
All methods provide an alternative where you can pass an instance by param if you wish not to use a global instance.

AS this uses runtime reflection and depends on class static initialization I expect bugs with usage, but won;'t find most cases until we give it a shot.

Runtime reflection for typeclasses. Kotlin,I feel dirty.

### DIFF
--- a/katz/src/main/kotlin/katz/data/Coproduct.kt
+++ b/katz/src/main/kotlin/katz/data/Coproduct.kt
@@ -1,0 +1,54 @@
+/*
+ * Copyright (C) 2017 The Katz Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package katz
+
+typealias CoproductF<F> = HK<Coproduct.F, F>
+typealias CoproductFG<F, G> = HK<CoproductF<F>, G>
+typealias CoproductKind<F, G, A> = HK<CoproductFG<F, G>, A>
+
+fun <F, G, A> CoproductKind<F, G, A>.ev(): Coproduct<F, G, A> = this as Coproduct<F, G, A>
+
+data class Coproduct<F, G, A>(val run: Either<HK<F, A>, HK<G, A>>) : CoproductKind<F, G, A> {
+
+    class F private constructor()
+
+    companion object {
+
+        fun <F, G, A> leftc(x: HK<F, A>): Coproduct<F, G, A> =
+                Coproduct(Either.Left(x))
+
+        fun <F, G, A> rightc(x: HK<G, A>): Coproduct<F, G, A> =
+                Coproduct(Either.Right(x))
+
+    }
+
+    fun <B> map(F: Functor<F>, G: Functor<G>, f: (A) -> B): Coproduct<F, G, B> =
+            Coproduct(run.bimap(F.lift(f), G.lift(f)))
+
+    fun <B> coflatMap(F: Comonad<F>, G: Comonad<G>, f: (Coproduct<F, G, A>) -> B): Coproduct<F, G, B> =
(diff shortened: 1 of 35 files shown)


katz/src/main/kotlin/katz/data/Coproduct.kt
```diff
@@ -22,33 +22,25 @@ typealias CoproductKind<F, G, A> = HK<CoproductFG<F, G>, A>
 
 fun <F, G, A> CoproductKind<F, G, A>.ev(): Coproduct<F, G, A> = this as Coproduct<F, G, A>
 
-data class Coproduct<F, G, A>(val run: Either<HK<F, A>, HK<G, A>>) : CoproductKind<F, G, A> {
+data class Coproduct<F, G, A>(val CF: Comonad<F>, val CG: Comonad<G>, val run: Either<HK<F, A>, HK<G, A>>) : CoproductKind<F, G, A> {
 
     class F private constructor()
 
-    companion object {
```
ffgiraldez: `Coproduct` could not have same companion object like `Kleisli` in order to provide their Comonad instances from implicits?

katz/src/main/kotlin/katz/data/OptionT.kt
```diff
@@ -24,65 +24,69 @@ typealias OptionTF<F> = HK<OptionT.F, F>
...
+    inline fun <B> subflatMap(crossinline f: (A) -> Option<B>): OptionT<F, B> =
+            transform({ it.flatMap(f) })
 
     //TODO: add toRight() and toLeft() once EitherT it's available
-}
\ No newline at end of file
+}
+
+
+inline fun <reified F, A> optionT(value: HK<F, Option<A>>, MF : Monad<F> = monad<F>()): OptionT<F, A> = OptionT(MF, value)
```
ffgiraldez: why not use the same magic from `Kleisli` invoke operator here instead of this function?

katz/src/main/kotlin/katz/predef.kt
```diff
@@ -0,0 +1,33 @@
...
+//class Bootstrap() {
+//    companion object {
+//        init {
+//            println("called!")
+//            listOf(EitherMonad, IdMonad, NonEmptyListMonad, OptionMonad)
+//        }
+//    }
+//}
+
+object IntSemigroup : Semigroup<Int>, GlobalInstance<Semigroup<Int>>() {
```
ffgiraldez: not in this PR, but we should create an issue to archive default Semigroup for language types Long, Float, String....

katz/src/main/kotlin/katz/data/Kleisli.kt
```diff
@@ -1,30 +1,36 @@
...
+    fun <DD> local(f: (DD) -> D): Kleisli<F, DD, A> = Kleisli(MF, { dd -> run(f(dd)) })
 
-    fun <B> map(ft: Functor<F>, f: (A) -> B): Kleisli<F, D, B> = Kleisli { a -> ft.map(run(a), f) }
+    companion object {
 
-    fun <B> flatMap(m: Monad<F>, f: (A) -> Kleisli<F, D, B>): Kleisli<F, D, B> =
-            Kleisli { d ->
-                m.flatMap(run(d)) { a -> f(a).run(d) }
-            }
+        inline operator fun <reified F, D, A> invoke(MF : Monad<F> = monad<F>(), noinline run: (D) -> HK<F, A>): Kleisli<F, D, A> = Kleisli(MF, run)
```
pakoito: Raul ahora mismo

pakoito: And the Kotlin compiler

JMPergar:
giphy 5

katz/src/main/kotlin/katz/data/Coproduct.kt
```diff
@@ -22,33 +22,25 @@ typealias CoproductKind<F, G, A> = HK<CoproductFG<F, G>, A>
...
-
-        fun <F, G, A> leftc(x: HK<F, A>): Coproduct<F, G, A> =
```
raulraja: They are just syntax and we are gonna do a syntax pass over all datatypes, this were left unused once I tried to make them reified and could not use them inside the class. Also is too early for syntax and it's getting in the way of refactoring and doing big PRs that touch all datatypes like this one

katz/src/main/kotlin/katz/data/Coproduct.kt
raulraja: Done!

Issue #63, opened 2017-04-07
Create Traverse


JorgeCastilloPrz:
### Requirements
`Traverse` needs to be done in terms of `Foldable`, so we need to wait for #62 to be done.

ffgiraldez:
Resolved on #62

PR #66, opened 2017-04-08
Add Eval, Foldable and Traverse types

### LINKED ISSUE #62: Create Foldable and Eval types
`Eval` type is being created under this PR also since its needed for `Foldable`.

JorgeCastilloPrz:
Fixes #62 and #63 

* Added `Eval` type.
* Added `Foldable` typeclass. Some of its methods are calculated in terms of `Eval`.
* Added method `map2Eval` to `Applicative`.

### TODO
* Write tests. I would like to get a big review from clever people on this (mainly on `Eval` type) in order to know whether the direction is ok before starting to write tests.


katz/src/main/kotlin/katz/data/Eval.kt
```diff
@@ -0,0 +1,229 @@
+/*
+ * Copyright (C) 2017 The Katz Authors
```
raulraja: We should modify all of our copyright headers to include that most of the datatypes like this one is a port originating from Scala cats. We don't want people to think we created this from scratch since the exercise is mainly about porting most of the cats one that are useful for what we are trying to achieve. I don't want people in the Scala community upset about not being properly credited in the copyrights so such a fact should be state clearly everywhere in the project including header files, README and LICENSE files. 

I'm not an attorney but I think it's important that those that have come up with the original encoding of these datatypes in both cats and Scalaz get properly credited.

We should follow similar practices as this:

https://github.com/typelevel/cats/blob/master/COPYING
https://github.com/typelevel/cats/blob/master/AUTHORS.md

Also I think we should remove copyright headers from all the files as a lot of the code on them is not really our copyright. Copyright assignment and contribution should be in the License and Copying files only to avoid mistakes like automatically adding a header to a file whose code is entirely translated from cats or scalaz.

katz/src/main/kotlin/katz/data/Eval.kt
```diff
@@ -0,0 +1,229 @@
...
+ * even when applied to a Now instance.
+ *
+ * It is not generally good style to pattern-match on Eval instances.
+ * Rather, use .map and .flatMap to chain computation, and use .value
+ * to get the result when needed. It is also not good style to create
+ * Eval instances whose computation involves calling .value on another
+ * Eval instance -- this can defeat the trampolining and lead to stack
+ * overflows.
+ */
+sealed class Eval<A> {
```
raulraja: The way to properly test this class is stack safe is to create in `tailrec` function or loop multiple instances of this class with flatMap and ensure it does not cause an stack overflow

katz/src/main/kotlin/katz/typeclasses/Foldable.kt
```diff
@@ -0,0 +1,179 @@
...
+     * .map and .flatMap.
+     *
+     * For more detailed information about how this method works see the documentation for `Eval[_]`.
+     */
+    fun <A, B> foldR(fa: HK<F, A>, lb: Eval<B>): ((A, Eval<B>) -> Eval<B>) -> Eval<B>
+
+    /**
+     * The size of this Foldable.
+     *
+     * This is overriden in structures that have more efficient size implementations (e.g. Vector, Set, Map).
```
raulraja: this may not be true in Kotlin

katz/src/main/kotlin/katz/typeclasses/Foldable.kt
```diff
@@ -0,0 +1,179 @@
...
+    /**
+     * Traverse `F[A]` using `Applicative[G]`.
+     *
+     * `A` values will be mapped into `G[B]` and combined using
+     * `Applicative#map2`.
+     *
+     * For example:
+     *
+     * {{{
+     * scala> import cats.implicits._
```
raulraja: We should not include examples from scala cats or anything similar

katz/src/main/kotlin/katz/typeclasses/Foldable.kt
```diff
@@ -0,0 +1,179 @@
...
+ * usually be used by the associated `Foldable[_]` instance.
+ *
+ * Foldable[F] is implemented in terms of two basic methods:
+ *
+ *  - `foldLeft(fa, b)(f)` eagerly folds `fa` from left-to-right.
+ *  - `foldRight(fa, b)(f)` lazily folds `fa` from right-to-left.
+ *
+ * Beyond these it provides many other useful methods related to folding over F[A] values.
+ */
+interface Foldable<F> {
```
raulraja: Since we need also `Traverse` for `sequence` and other useful combinators we should attempt to include that as part of this PR if you think it's appropriate, otherwise a separate PR once this is integrated is fine.

katz/src/main/kotlin/katz/typeclasses/Foldable.kt
JorgeCastilloPrz: Fixed

katz/src/main/kotlin/katz/typeclasses/Foldable.kt
JorgeCastilloPrz: Fixed

katz/src/main/kotlin/katz/data/Eval.kt
JorgeCastilloPrz: Fixed. Can you check the license format I have added please ?

katz/src/main/kotlin/katz/typeclasses/Foldable.kt
```diff
@@ -0,0 +1,205 @@
...
+IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY DIRECT, INDIRECT,
+INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT
+NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
+THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+------------------------------------------------------------------------
+ */
```
pakoito: Is this strictly necessary? I'd guess most pats of it can be classified as *concept* and there's no license relationship across languages. I may be wrong, IANAL.

katz/src/main/kotlin/katz/typeclasses/Foldable.kt
```diff
@@ -0,0 +1,205 @@
...
+            foldL { gb, a -> g.flatMap(gb) { f(it, a) } }
+        }
+    }
+
+    /**
+     * Monadic folding on F by mapping A values to G<B>, combining the B values using the given Monoid<B> instance.
+     *
+     * Similar to foldM, but using a Monoid<B>.
+     */
+    fun <G, A, B> foldMapM(g: Monad<G>, bb: Monoid<B>, fa: HK<F, A>): ((A) -> HK<G, B>) -> HK<G, B> {
```
pakoito: If I remember correctly the current nomenclature we were using required for Monad instances to be capitalized. I may need to pop a link in here from another PR to support this statement :D.

katz/src/main/kotlin/katz/typeclasses/Foldable.kt
JorgeCastilloPrz: Added base `sequence` combinator to `Foldable` and `Traverse` type with its own `sequence` implementation. I couldn't add other ones like `traverseU`, `flatTraverse` or `flatSequence` since we don't have the types `FlatMap` or `Unapply`.

katz/src/main/kotlin/katz/typeclasses/Foldable.kt
JorgeCastilloPrz: This was requested by @raulraja

katz/src/main/kotlin/katz/typeclasses/Foldable.kt
JorgeCastilloPrz: Just for `Monads` ?

JorgeCastilloPrz:
I have added `Traverse`.

katz/src/main/kotlin/katz/typeclasses/Foldable.kt
raulraja: We should not include licenses or copyrights in files only in the LICENSE and COPYING files in the repo.

katz/src/main/kotlin/katz/typeclasses/Foldable.kt
raulraja: All instances should be capitalized. For example `MG: Monad<G>`

katz/src/main/kotlin/katz/typeclasses/Traverse.kt
```diff
@@ -0,0 +1,95 @@
...
+
+------------------------------------------------------------------------
+ */
+
+package katz
+
+/**
+ * Traverse, also known as Traversable. Traversal over a structure with an effect.
+ */
+interface Traverse<F> : Functor<F>, Foldable<F> {
```
ffgiraldez: should be declared as `Typeclass` also

katz/src/main/kotlin/katz/typeclasses/Foldable.kt
```diff
@@ -0,0 +1,212 @@
...
+ * Data structures that can be folded to a summary value.
+ *
+ * Foldable<F> is implemented in terms of two basic methods:
+ *
+ *  - `foldLeft(fa, b)(f)` eagerly folds `fa` from left-to-right.
+ *  - `foldRight(fa, b)(f)` lazily folds `fa` from right-to-left.
+ *
+ * Beyond these it provides many other useful methods related to folding over F<A> values.
+ */
+interface Foldable<F> {
```
ffgiraldez: should be declared also as `Typeclass`

katz/src/main/kotlin/katz/data/Option.kt
```diff
@@ -31,6 +31,7 @@ sealed class Option<out A> : OptionKind<A> {
     companion object : OptionMonad, GlobalInstance<Monad<Option.F>>() {
         inline fun <reified A : Any> fromNullable(a: A?): Option<A> = if (a != null) Option.Some(a) else Option.None
         operator fun <A> invoke(a: A): Option<A> = Option.Some(a)
+        inline fun <reified T: Any> empty() = Option.fromNullable<T>(null)
```
ffgiraldez: why need this?

JorgeCastilloPrz: This is a way to create and bind a type to the Option without having the initial value ready yet. This method is also available in Scala Options and helps you on type inference for some given cases

JorgeCastilloPrz:
Thanks for your feedback guys. I'll be working on changes starting tomorrow.

katz/src/main/kotlin/katz/typeclasses/Foldable.kt
JorgeCastilloPrz: 👍

katz/src/main/kotlin/katz/typeclasses/Traverse.kt
JorgeCastilloPrz: This one is extending `Foldable` which is already extending `Typeclass`.

ffgiraldez: I know but check Monad for instance, extend Functor( Functor extends TypeClass) and also TypeClass

JorgeCastilloPrz:
Your feedback has been applied @ffgiraldez @raulraja

katz/src/main/kotlin/katz/data/Option.kt
raulraja: Isn't that just the same as returning `None`? Also this is part of the interface for `MonadFilter` which is one of the MTL typeclasses which we don't have yet.

JorgeCastilloPrz: I wouldn't say so. Option.None is now being created as an Option<Nothing> which is not the same that Option<A> with a null value. I was using this method for `Foldable#find`, but seems like I finally desisted and it's not being used anymore. So removing it for now.

katz/src/main/kotlin/katz/data/Eval.kt
```diff
@@ -1,79 +1,3 @@
...
-INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT
-NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
-DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
-THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
-(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF
-THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
-
-------------------------------------------------------------------------
- */
-
```
pakoito: Tuuuuuusa p'allá https://i.ytimg.com/vi/0s133HqAOzw/hqdefault.jpg

JorgeCastilloPrz: Is that a good or bad comment ? :D. This has been ported to `LICENSE.md`

pakoito: It's deep ruralesque to cheer for the forced removal of an unwanted subject or item.

katz/src/main/kotlin/katz/typeclasses/Traverse.kt
ffgiraldez: Check other typeclasses like Monad, all off them extends TypeClass no matter if any of their other interfaces already have it.

katz/src/main/kotlin/katz/data/Eval.kt
JorgeCastilloPrz: "deep ruralesque" 😂

katz/src/main/kotlin/katz/typeclasses/Traverse.kt
JorgeCastilloPrz: Which value does it add ?

JorgeCastilloPrz: I am adding it but not totally convinced of whether this useful or useless. Anyways, lets keep moving.

katz/src/main/kotlin/katz/data/Eval.kt
```diff
@@ -0,0 +1,200 @@
...
+ * even when applied to a Now instance.
+ *
+ * It is not generally good style to pattern-match on Eval instances.
+ * Rather, use .map and .flatMap to chain computation, and use .value
+ * to get the result when needed. It is also not good style to create
+ * Eval instances whose computation involves calling .value on another
+ * Eval instance -- this can defeat the trampolining and lead to stack
+ * overflows.
+ */
+sealed class Eval<A> {
```
ffgiraldez: Should be also HK<Eval.F, A>

PR #74, opened 2017-04-18
Add EitherT implementation

pakoito:
Guess who finally signed the paperwork? :D


pakoito:
Tests incoming!

katz/src/test/kotlin/katz/data/EitherTTest.kt
```diff
@@ -0,0 +1,89 @@
...
+                val ot: EitherT<NonEmptyList.F, Int, String> = EitherT(NonEmptyList.of(Either.Right(ignored)))
+                val mapped = ot.flatMap { EitherT(NonEmptyList.of(Either.Left(3))) }
+                val expected = EitherT.impure<NonEmptyList.F, Int>(3)
+
+                mapped == expected
+            }
+
+            forAll { ignored: String ->
+                val ot = EitherT.impure<NonEmptyList.F, Int>(3)
+                val mapped =  ot.flatMap{  EitherT<NonEmptyList.F, Int, Int>(NonEmptyList.of(Either.Right(2))) }
```
pakoito: I'm really unhappy about needing to hint the type here.

raulraja:
Regarding inference it will improve once we add syntax extensions, and also in that case I'm not sure if this is an actual limitation of emulating higher kinds. We can also use partial application of generics if the Kotlin compiler supports it.

pakoito:
I've updated with an important changes. The inference for pure() and impure() now requires all generic parameters instead of relying in Nothing.

It also adds tests for all the remaining APIs. I'd like to revisit the whole API for Either and EitherT at some point.

Could @raulraja please review it again?

katz/src/main/kotlin/katz/instances/EitherTMonad.kt
```diff
@@ -0,0 +1,29 @@
...
+    override fun <A> pure(a: A): EitherT<F, L, A> =
+            EitherT(MF, MF.pure(Either.Right(a)))
+
+    override fun <A, B> map(fa: EitherTKind<F, L, A>, f: (A) -> B): EitherT<F, L, B> =
+            fa.ev().map { f(it) }
+
+    override fun <A, B> flatMap(fa: EitherTKind<F, L, A>, f: (A) -> EitherTKind<F, L, B>): EitherT<F, L, B> =
+            fa.ev().flatMap { f(it).ev() }
+
+    override fun <A, B> tailRecM(a: A, f: (A) -> HK<EitherTF<F, L>, Either<A, B>>): EitherT<F, L, B> =
```
ffgiraldez: i think we should test at least this method for every monad in order to avoid stack overflows

pakoito: Done!

PR #76, opened 2017-04-21
Apparently using Kotlin version 1.1.3 fix inference errors during builds

ffgiraldez:
We suffer some weird build errors, specially with jitpack [here](https://jitpack.io/com/github/FineCinnamon/Katz/0.2.5/build.log) or [here](https://jitpack.io/com/github/FineCinnamon/Katz/0.2.5/build.log) are examples of them.

Initially we think the problem was [incremental builds](https://github.com/FineCinnamon/Katz/pull/72), but builds still failling. So I try use next version of Kotlin that are under development and [apparently it fix build errors ](https://jitpack.io/#FineCinnamon/Katz/test-dev-SNAPSHOT)

Question here, if we want to use this version of Kotlin during this stage of development that will affect our potentially users or not.


jrgonzalezg:
Two questions:
- Is that specific revision needed or it also works with the newest 1.13-dev-1450?
- Does this allow to re-enable incremental compilation? (It is currently disabled on gradle.properties)

Besides that i do not see a problem to switching to a 1.1.3 EAP version if it eases the development process. It is only a minor revision that should be mostly compatible and even if there were some temporary issues for users of the library it will probably get to a stable release in a couple of months.

jrgonzalezg:
Additionally it seems that detekt breaks the build because it modifies NonEmptyListMonad removed the return type of go and that causes a recursive problem on the type checker. This happens also on master though and may be related to the JitPack issue because using `gradle build` calls detekt as part of the build making it fail afterwards.

JorgeCastilloPrz:
Since we are on snapshot phases I would move forward and try newer versions of Kotlin to have the faster adaptation possible. Otherwise, we would be forced to update our Kotlin version after v1 release, since we should always target the most updated Kotlin release. 

So If this is really fixing the problem I vote for including it and just take that into account when we are ready to release v1 of Katz, since we could have to wait for that until Kotlin version is a full release.

ffgiraldez:
@jrgonzalezg not we could use 1.1.3-dev-1450 released yesterday 🕺 without problem, i will update it. also re-enable incremental compilation

I think i fix detekt issue, the problem it's NonEmptyListMonad have two `go` functions one without explicit returning type and other with `Unit` in order to work recursion. but detekt have a rule to remove Unit from return type. The solution it's update baseline from detekt in order to ignore go funcion.

jrgonzalezg:
You mention an updated README on last commit but it seems is not there. I think you can merge this directly after you recheck that.

JorgeCastilloPrz:
I am working on a new branch that I branched out from master, and I get the NonEmptyListMonad "Type checking has run into a recursive problem." error there also. So seems to not be very related to this PR changes.

ffgiraldez:
It's there, just now in order to use it we need to add coroutine & dev repositories

ffgiraldez:
@JorgeCastilloPrz add :Unit into 'go' function to fix the problem

PR #77, opened 2017-04-21
Trampoline

### LINKED ISSUE #78: Add Trampoline

JorgeCastilloPrz:
Fixes #78 

Trampoline is often used to emulate tail recursion. The idea is to have some step code that can be trampolined itself to emulate recursion. The difference with standard recursion would be that there is no need to rewind the whole stack when we reach the end of the stack, since the first value returned that is not a trampoline would be directly returned as the overall result value for the whole function chain. That means `Trampoline` emulates what **tail recursion** does.

There are languages that do not have `tailrec` modifier that use `Trampoline` to emulate the same behavior. But Trampoline is also useful for languages that also have trampolines out of the box:

*"As long as you use tail recursion with one function everything is fine as the compiler will optimize it for you and transform it in a loop, issues come when you are trying to use tail recursive calls between two functions or when you don’t have a tail recursive call."*

Trampolines are stack safe, since they do not need to maintain a function call stack (since they don't need to rewind it). So trampolines are useful also when you don't want to blow your stack up for too long recursive operations.

### DIFF
--- a/katz/src/test/kotlin/katz/data/TrampolineTest.kt
+++ b/katz/src/test/kotlin/katz/data/TrampolineTest.kt
@@ -1,0 +1,38 @@
+package katz
+
+import io.kotlintest.KTestJUnitRunner
+import io.kotlintest.matchers.shouldBe
+import io.kotlintest.properties.forAll
+import katz.Either.Left
+import katz.Either.Right
+import katz.data.Trampoline
+import org.junit.runner.RunWith
+
+@RunWith(KTestJUnitRunner::class)
+class TrampolineTest : UnitSpec() {
+    init {
+        "map should modify value" {
+            Trampoline.More { odd(10) }.runT() shouldBe false
+        }
+    }
+
+    fun odd(n: Int): Trampoline<Boolean> {
+        return when (n) {
+            0 -> Trampoline.Done(false)
+            else -> {
+                println(n)
+                Trampoline.More { even(n - 1) }
+            }
+        }
+    }
+
+    fun even(n: Int): Trampoline<Boolean> {
+        return when (n) {
+            0 -> Trampoline.Done(true)
+            else -> {
+                println(n)
+                Trampoline.More { odd(n - 1) }
+            }
+        }
+    }
+}
--- a/katz/src/test/kotlin/katz/data/TrampolineTest.kt
+++ b/katz/src/test/kotlin/katz/data/TrampolineTest.kt
@@ -11,7 +11,7 @@ import org.junit.runner.RunWith
 @RunWith(KTestJUnitRunner::class)
 class TrampolineTest : UnitSpec() {
     init {
-        "map should modify value" {
+        "trampoline over 10 should return false" {
             Trampoline.More { odd(10) }.runT() shouldBe false
         }
     }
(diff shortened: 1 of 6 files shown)


katz/src/test/kotlin/katz/data/TrampolineTest.kt
```diff
@@ -0,0 +1,38 @@
...
+import katz.Either.Left
+import katz.Either.Right
+import katz.data.Trampoline
+import org.junit.runner.RunWith
+
+@RunWith(KTestJUnitRunner::class)
+class TrampolineTest : UnitSpec() {
+    init {
+        "trampoline over 10 should return false" {
+            Trampoline.More { odd(10) }.runT() shouldBe false
```
pakoito: Where does runT come from?

katz/src/main/kotlin/katz/instances/NonEmptyListMonad.kt
```diff
@@ -11,18 +11,19 @@ interface NonEmptyListMonad : Monad<NonEmptyList.F> {
...
+                val x = NonEmptyList.fromList(v.tail)
+                when (x) {
+                    is Option.Some<NonEmptyList<Either<A, B>>> -> go(buf, f, x.value)
+                    is Option.None -> Unit
                 }
-                is Either.Left<*> -> go(buf, f, NonEmptyList.fromListUnsafe(f(v.head.a as A).ev().all + v.tail))
             }
+            is Either.Left<*> -> go(buf, f, NonEmptyList.fromListUnsafe(f(v.head.a as A).ev().all + v.tail))
+        }
+    }
```
pakoito: The changes to this file are just formatting, right?

katz/src/test/kotlin/katz/data/TrampolineTest.kt
JorgeCastilloPrz: Inherited from Trampoline, it's a sealed class. What's up?

katz/src/main/kotlin/katz/instances/NonEmptyListMonad.kt
JorgeCastilloPrz: No. Removed the need to return something on the body. I followed Fernando's advice to avoid compilation errors.

JorgeCastilloPrz: @ffgiraldez?

JorgeCastilloPrz:
I guess I should redefine `Trampoline` as a type alias for `Free<Function0, A>` now. Is that correct @raulraja ?

raulraja:
Yeah that would give you stack safe semantics and AFAIK is how is also done
in Scala. I plan on adding some tagless final encodings alternatives to
free to show another way to build interpretable program and we can test the
trampolining capabilities there over stack unsafe monads.

JorgeCastilloPrz:
I'm trying to work on the compiler plugin from now on on my spare time, so if anybody has a deeper idea on how to translate `Trampoline` to be declared in terms of `Free` it would be nice if he can take the ownership on this PR. Otherwise I'll do it when I'm finished with the plugin / annotation processor , whatever it is, since I still don't know :D

PR #82, opened 2017-04-25
Small issues with reified from Java

aballano:
Two changes done:

* Removed unnecessary reified from Option's companion object in order for the method to be called from Java.
* Also added @JvmStatic to some other companion object methods to generate easier to use methods.


aballano:
@ffgiraldez I added it to those 3 only since the rest only use the invoke operator. Please let me know if I missed any method in a companion object apart from those.

ffgiraldez:
@aballano any class under `katz/data` folder with a companion object should have it, `NonEmptyList` is just an example.

aballano:
Done!

PR #83, opened 2017-04-27
Add Yoneda and CoYoneda, along with their Functor instances

### LINKED ISSUE #80: Create CoYoneda construct
¡Me lo pido!

pakoito:
Add two central pieces to the Free machinery: Yoneda and CoYoneda. They include a minimal implementation, along with a Functor instance. I've also written a minimal test suite for each.

For consistency purposes I've swapped the parameter order in the Free class.

Closes #80

### DIFF
--- a/katz/src/main/kotlin/katz/free/Yoneda.kt
+++ b/katz/src/main/kotlin/katz/free/Yoneda.kt
@@ -4,8 +4,7 @@ typealias YonedaKind<U, A> = HK2<Yoneda.F, U, A>
 
 typealias YonedaF<U> = HK<Yoneda.F, U>
 
-// FIXME using YonedaKind throws a compiler error, but not the expanded form
-interface Yoneda<FU, A> : HK<HK<Yoneda.F, FU>, A> {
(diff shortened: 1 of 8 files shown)


katz/src/main/kotlin/katz/free/Yoneda.kt
```diff
@@ -0,0 +1,29 @@
+package katz
+
+typealias YonedaKind<U, A> = HK2<Yoneda.F, U, A>
+
+typealias YonedaF<U> = HK<Yoneda.F, U>
+
+// FIXME using YonedaKind throws a compiler error, but not the expanded form
+interface Yoneda<FU, A> : HK<HK<Yoneda.F, FU>, A> {
```
pakoito: A compiler bug, someone call the JB cops :D

pakoito: The bug went away when I tried to compile again to get a debug trace. I've reported it anyway to https://youtrack.jetbrains.com/issue/KT-17619

pakoito: The bog came back so I reverted it.

PR #87, opened 2017-04-29
Add Cokleisli

### LINKED ISSUE #86: Add CoKleisli type

pakoito:
Closes #86


pakoito:
Incomplete: missing instances and tests

katz/src/test/kotlin/katz/data/OptionMonoidTest.kt
```diff
@@ -8,7 +8,7 @@ import org.junit.runner.RunWith
 class OptionMonoidTest : UnitSpec() {
     init {
         "should semigroup with the instance passed" {
-            forAll { value: Int ->
+            forAll { _: Int ->
```
jrgonzalezg: Why the forAll if the value is not used?

pakoito: Because it's the case combining two None

PR #90, opened 2017-05-04
Sanogueralorenzo tweaks to applicative tuples

### LINKED ISSUE #89: Tweaks to Applicative Tuples
Rename swap() to reverse() and implement for all tuples.

https://github.com/FineCinnamon/Katz/pull/85#discussion_r114894091

Convert toT into plus.

https://github.com/FineCinnamon/Katz/pull/85#discussion_r114894514

sanogueralorenzo:
Closes #89

### DIFF
--- a/katz/src/main/kotlin/katz/data/Either.kt
+++ b/katz/src/main/kotlin/katz/data/Either.kt
@@ -61,8 +61,8 @@ sealed class Either<out A, out B> : EitherKind<A, B> {
      *
      * Example:
      * ```
-     * Left("left").swap()   // Result: Right("left")
-     * Right("right").swap() // Result: Left("right")
+     * Left("left").reverse()   // Result: Right("left")
+     * Right("right").reverse() // Result: Left("right")
      * ```
      */
     fun swap(): Either<B, A> =
--- a/katz/src/main/kotlin/katz/data/Either.kt
+++ b/katz/src/main/kotlin/katz/data/Either.kt
@@ -61,8 +61,8 @@ sealed class Either<out A, out B> : EitherKind<A, B> {
      *
      * Example:
      * ```
-     * Left("left").reverse()   // Result: Right("left")
-     * Right("right").reverse() // Result: Left("right")
+     * Left("left").swap()   // Result: Right("left")
+     * Right("right").swap() // Result: Left("right")
      * ```
      */
     fun swap(): Either<B, A> =
--- a/katz/src/main/kotlin/katz/data/Ior.kt
+++ b/katz/src/main/kotlin/katz/data/Ior.kt
@@ -165,13 +165,13 @@ sealed class Ior<out A, out B> : IorKind<A, B> {
 
     /**
      * If this is a [Left], then return the left value in [Right] or vice versa,
-     * when this is [Both] , left and right values are reverse
+     * when this is [Both] , left and right values are swap
      *
      * Example:
      * ```
-     * Left("left").reverse()   // Result: Right("left")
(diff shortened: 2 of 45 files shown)


katz/src/main/kotlin/katz/data/Either.kt
```diff
@@ -61,8 +61,8 @@ sealed class Either<out A, out B> : EitherKind<A, B> {
      *
      * Example:
      * ```
-     * Left("left").swap()   // Result: Right("left")
-     * Right("right").swap() // Result: Left("right")
+     * Left("left").reverse()   // Result: Right("left")
+     * Right("right").reverse() // Result: Left("right")
      * ```
      */
     fun swap(): Either<B, A> =
```
pakoito: The name of the method is still swap. The documentation should be reverted to say swap too. Let's keep revert only for tuples!

katz/src/main/kotlin/katz/data/Ior.kt
```diff
@@ -165,13 +165,13 @@ sealed class Ior<out A, out B> : IorKind<A, B> {
...
+     * when this is [Both] , left and right values are reverse
      *
      * Example:
      * ```
-     * Left("left").swap()   // Result: Right("left")
-     * Right("right").swap() // Result: Left("right")
-     * Both("left", "right").swap() // Result: Both("right", "left")
+     * Left("left").reverse()   // Result: Right("left")
+     * Right("right").reverse() // Result: Left("right")
+     * Both("left", "right").reverse() // Result: Both("right", "left")
```
pakoito: Same as above, if possible.

katz/src/test/kotlin/katz/data/EitherTest.kt
```diff
@@ -56,7 +56,7 @@ class EitherTest : UnitSpec() {
             }
         }
 
-        "swap should interchange values" {
+        "reverse should interchange values" {
```
pakoito: Revert to swap, same as the method.

pakoito:
The plan would be to overload the operator plus for Tuple2-9 plus any object. To avoid collisions, keep `infix fun <A, B> A.toT(b: B): Tuple2<A, B> = Tuple2(this, b)` with that name.

https://kotlinlang.org/docs/reference/operator-overloading.html#binary-operations

katz/src/main/kotlin/katz/data/WriterT.kt
```diff
@@ -8,10 +8,10 @@ data class WriterT<F, W, A>(val MF: Monad<F>, val value: HK<F, Tuple2<W, A>>) :
 
     companion object {
         inline fun <reified F, reified W, A> pure(a: A, MM: Monoid<W> = monoid(), MF: Monad<F> = monad()) =
-                WriterT(MF.pure(MM.empty() toT a), MF)
+                WriterT(MF.pure(MM.empty() plus a), MF)
```
raulraja: Can this be used as `+`?

katz/src/main/kotlin/katz/typeclasses/Applicative.kt
```diff
@@ -21,20 +21,42 @@ interface Applicative<F> : Functor<F>, Typeclass {
 }
 
 data class Tuple2<out A, out B>(val a: A, val b: B) {
```
ffgiraldez: all those TupleX stuff it's ok to be inside Applicative file or it's better move that stuff in another file?

raulraja: They should be moved out

PR #91, opened 2017-05-06
Add comonadic comprehensions 

### LINKED ISSUE #88: Implement comonadic comprehensions
Taken from the roadmap in #69. Requires #86 to be merged first.

pakoito:
Closes #88

### DIFF
--- a/katz/src/main/kotlin/katz/typeclasses/Comonad.kt
+++ b/katz/src/main/kotlin/katz/typeclasses/Comonad.kt
@@ -16,6 +16,9 @@ interface Comonad<F> : Functor<F>, Typeclass {
     fun <A, B> coflatMap(fa: HK<F, A>, f: (HK<F, A>) -> B): HK<F, B>
 
     fun <A> extract(fa: HK<F, A>): A
+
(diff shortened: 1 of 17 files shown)


katz/src/main/kotlin/katz/typeclasses/Comonad.kt
```diff
@@ -8,7 +16,51 @@ interface Comonad<F> : Functor<F>, Typeclass {
...
+        throw exception
+    }
+
+    internal lateinit var returnedMonad: A
+
+    operator suspend fun <B> HK<F, B>.not(): B = bind { this }
 
+    suspend fun <B> HK<F, B>.bind(): B = bind { this }
+
+    suspend fun <B> bind(m: () -> HK<F, B>): B = suspendCoroutineOrReturn { c ->
```
raulraja: is `bind` a good name for this? bind has the semantics of flatMap, whereas this is 'extract'. what about `cobind`?

pakoito: This should be extract then. Fixed!

PR #93, opened 2017-05-07
WIP for Global instances of any arity

pakoito:
This is a broken work in progress, in case anyone wants inspiration.

The limitation so far is that it doesn't support nested monads of 2+ arity, and EitherT instances of a concrete type, defaulting to Nothing or Any instead.


pakoito:
This PR is a PoC that it couldn't be done with this approach. The new processor should deal with it better: https://github.com/kategory/implicits-processor

PR #98, opened 2017-05-18
Add vector type

nomisRev:
I had some issues with type erasure when using `Array<T>`. Any ideas?


katz/src/main/kotlin/katz/instances/NonEmptyVectorMonad.kt
```diff
@@ -0,0 +1,32 @@
...
+import java.util.Vector
+
+interface NonEmptyVectorMonad : Monad<NonEmptyVector.F> {
+    override fun <A> pure(a: A): HK<NonEmptyVector.F, A> = NonEmptyVector.of(a)
+
+    override fun <A, B> flatMap(fa: HK<NonEmptyVector.F, A>, f: (A) -> HK<NonEmptyVector.F, B>): HK<NonEmptyVector.F, B> =
+            fa.ev().flatMap { f(it).ev() }
+
+    @Suppress("UNCHECKED_CAST")
+    private tailrec fun <A, B> go(buf: Vector<B>, f: (A) -> HK<NonEmptyVector.F, Either<A, B>>, v: NonEmptyVector<Either<A, B>>): Unit =
```
ffgiraldez: we had trouble with tailrec in the past with nested when expressions in a tailrec funciton.
One solution, if I not remember wrong, was move the tailrec function outside the class into an extension one. in [Free](https://github.com/FineCinnamon/Katz/blob/master/katz/src/main/kotlin/katz/free/Free.kt#L31) we have it in that way with a test to ensure that.

please could you add a test to ensure NonEmptyVectorMonad it's stack safe?

nomisRev: Shall I do the same for `NonEmptyListMonad`?

ffgiraldez: i will no say no to free work ;). this PR it's short enough to add it.

raulraja:
We should find out if something like this already exists in the Kotlin std lib or elsewhere https://github.com/scala/scala/blob/v2.12.0/src/library/scala/collection/immutable/Vector.scala#L1

pakoito:
@raulraja the deps don't seem to be too cumberson. Maybe worth porting over?

nomisRev:
Closing this for now. Let's come back to this when we have a Vector construct.

PR #105, opened 2017-06-06
allow specify reader monad explicitly.

ffgiraldez:
Using actual Reader implementation
```
FATAL EXCEPTION: main
 Process: com.github.jorgecastillo.kotlinandroid, PID: 32610
 java.lang.NoSuchMethodError: No interface method getTypeName()Ljava/lang/String; in class Ljava/lang/reflect/Type; or its super classes (declaration of 'java.lang.reflect.Type' appears in /system/framework/core-oj.jar)
    at katz.InstanceParametrizedType.toString(Typeclass.kt:88)
    at java.lang.String.valueOf(String.java:2683)
    at java.lang.StringBuilder.append(StringBuilder.java:129)
    at katz.TypeClassInstanceNotFound.<init>(Typeclass.kt:172)
    at katz.TypeclassKt.instance(Typeclass.kt:185)
    at katz.Reader.ask(Reader.kt:27)
```

something happens with implicits, but in order to not block I detect a defect in Reader API that does not allow pass the monad explicitly. this change fix that.

### DIFF
--- a/katz/src/main/kotlin/katz/data/Reader.kt
+++ b/katz/src/main/kotlin/katz/data/Reader.kt
@@ -14,13 +14,13 @@ fun <D, A> ReaderT<Id.F, D, A>.runId(d: D): A =
 
 object Reader {
 
-    inline operator fun <D, A> invoke(noinline run: (D) -> A, MF: Monad<Id.F> = monad<Id.F>()): ReaderT<Id.F, D, A> =
+    operator fun <D, A> invoke(run: (D) -> A, MF: Monad<Id.F> = Id): ReaderT<Id.F, D, A> =
             Kleisli(run.andThen { Id(it) }, MF)
 
-    fun <D, A> pure(x: A, MF: Monad<Id.F> = monad<Id.F>()): ReaderT<Id.F, D, A> =
+    fun <D, A> pure(x: A, MF: Monad<Id.F> = Id): ReaderT<Id.F, D, A> =
             Kleisli.pure<Id.F, D, A>(x, MF)
 
-    fun <D> ask(MF: Monad<Id.F> = monad<Id.F>()): ReaderT<Id.F, D, D> =
+    fun <D> ask(MF: Monad<Id.F> = Id): ReaderT<Id.F, D, D> =
             Kleisli.ask<Id.F, D>(MF)
 
 }
--- a/katz/src/main/kotlin/katz/data/Reader.kt
+++ b/katz/src/main/kotlin/katz/data/Reader.kt
@@ -7,20 +7,20 @@ infix fun <A, B, C> ((A) -> B).andThen(g: (B) -> C): (A) -> C =
         { a: A -> g(this(a)) }
 
 fun <D, A> ((D) -> A).reader(): ReaderT<Id.F, D, A> =
-        Reader(this)
+        Reader(this, Id)
 
 fun <D, A> ReaderT<Id.F, D, A>.runId(d: D): A =
         this.run(d).value()
 
 object Reader {
 
-    operator fun <D, A> invoke(run: (D) -> A): ReaderT<Id.F, D, A> =
-            Kleisli(run.andThen { Id(it) })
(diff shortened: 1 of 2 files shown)


JorgeCastilloPrz:
Tested it under a sample repo and this PR's code is working ok as long as you pass an explicit monad to it. 

We need to find the reason for the crash when we want implicits to work here (by not passing any monad to `ask` method).

This is the PR where I am using the `Reader` as it is in `0.3.2`.

katz/src/main/kotlin/katz/data/Reader.kt
```diff
@@ -7,20 +7,20 @@ infix fun <A, B, C> ((A) -> B).andThen(g: (B) -> C): (A) -> C =
...
+        Reader(this, Id)
 
 fun <D, A> ReaderT<Id.F, D, A>.runId(d: D): A =
         this.run(d).value()
 
 object Reader {
 
-    operator fun <D, A> invoke(run: (D) -> A): ReaderT<Id.F, D, A> =
-            Kleisli(run.andThen { Id(it) })
+    inline operator fun <D, A> invoke(noinline run: (D) -> A, MF: Monad<Id.F> = monad<Id.F>()): ReaderT<Id.F, D, A> =
```
pakoito: There is only one possible instance of this monad, so I'd be explicit about it instead.

MF: Monad<Id.F> = Id

katz/src/main/kotlin/katz/data/Reader.kt
```diff
@@ -7,20 +7,20 @@ infix fun <A, B, C> ((A) -> B).andThen(g: (B) -> C): (A) -> C =
...
+            Kleisli(run.andThen { Id(it) }, MF)
 
-    fun <D, A> pure(x: A): ReaderT<Id.F, D, A> =
-            Kleisli.pure<Id.F, D, A>(x)
+    fun <D, A> pure(x: A, MF: Monad<Id.F> = monad<Id.F>()): ReaderT<Id.F, D, A> =
```
pakoito: As above

katz/src/main/kotlin/katz/data/Reader.kt
```diff
@@ -7,20 +7,20 @@ infix fun <A, B, C> ((A) -> B).andThen(g: (B) -> C): (A) -> C =
...
+            Kleisli.pure<Id.F, D, A>(x, MF)
 
-    fun <D> ask(): ReaderT<Id.F, D, D> =
-            Kleisli.ask<Id.F, D>()
+    fun <D> ask(MF: Monad<Id.F> = monad<Id.F>()): ReaderT<Id.F, D, D> =
```
pakoito: As above

katz/src/main/kotlin/katz/data/Reader.kt
pakoito: Inline shouldn't be required.

JorgeCastilloPrz:
Updates on this. I have added `val id = Id(1)` before the `Reader.run` execution for the whole chain, and that makes me able to remove the explicit monad argument from all the readers. It completely relies on previous reference, and I see that as a usage problem cc @pakoito

pakoito:
@JorgeCastilloPrz do fill in an issue and we'll try and find a solution

JorgeCastilloPrz:
#118 @pakoito

Can we resolve conflicts and merge this @ffgiraldez? This is improving the Reader itself, independently of the detected implicits bug.

raulraja:
IMO this is ready to merge once conflicts are resolved.

ffgiraldez:
i completely forgot this PR, i will do ASAP

pakoito:
I've duplicated this fix on a new PR: #126

JorgeCastilloPrz:
I am using kategory 0.3.3 on a pet project and this is not happening anymore, unless you don't reference the implicits previously. So I would say it was already fixed on that one.

ffgiraldez:
Resolved in #126

PR #107, opened 2017-06-11
Add Effects package with class AndThen

pakoito:
AndThen is an internal construct used by IO to compose effects without blowing up the stack.

I'm working on IO, Effect, and the rest. This is the first diff to get the Effects package rolling.

### DIFF
--- a/katz/src/main/kotlin/katz/effects/internal/AndThen.kt
+++ b/katz/src/main/kotlin/katz/effects/internal/AndThen.kt
@@ -41,7 +41,7 @@ internal sealed class AndThen<in A, out B> : (A) -> B {
         fun processSuccess(f: (Any?) -> Any?): Unit =
                 try {
                     success = f(success)
-                } catch(throwable: Throwable) {
+                } catch (throwable: Throwable) {
                     failure = throwable
                     isSuccess = false
                 }
@@ -51,7 +51,7 @@ internal sealed class AndThen<in A, out B> : (A) -> B {
                     /* If we've made it this far then failure isn't null */
                     success = f(failure!!)
                     isSuccess = true
-                } catch(throwable: Throwable) {
+                } catch (throwable: Throwable) {
                     failure = throwable
                 }
 
--- a/katz/src/main/kotlin/katz/effects/internal/AndThen.kt
+++ b/katz/src/main/kotlin/katz/effects/internal/AndThen.kt
@@ -1,0 +1,134 @@
+package katz.effects.internal
+
+/**
+ * Abandon all hope
+ * Ye who enter here
+ */
+internal sealed class AndThen<in A, out B> : (A) -> B {
+    internal data class Single<in A, out B>(val f: (A) -> B) : AndThen<A, B>()
+    internal data class ErrorHandler<in A, out B>(val fa: (A) -> B, val fe: (Throwable) -> B) : AndThen<A, B>()
+    internal data class Concat<in A, E, out B>(val left: AndThen<A, E>, val right: AndThen<E, B>) : AndThen<A, B>()
+
+    fun <C> andThen(g: AndThen<B, C>): AndThen<A, C> =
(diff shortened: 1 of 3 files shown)


katz/src/main/kotlin/katz/effects/internal/AndThen.kt
```diff
@@ -0,0 +1,135 @@
+package katz.effects.internal
+
+/**
+ * Abandon all hope
+ * Ye who enter here
+ */
+internal sealed class AndThen<in A, out B> : (A) -> B {
```
ffgiraldez: what we should do with this [andThen()](https://github.com/FineCinnamon/Katz/blob/master/katz/src/main/kotlin/katz/data/Reader.kt#L6)?

katz/src/main/kotlin/katz/effects/internal/AndThen.kt
```diff
@@ -0,0 +1,135 @@
...
+
+    override operator fun invoke(a: A): B =
+            runLoop(a, null, true)
+
+    /**
+     * Raises an error continuation.
+     * As Kotlin lacks a way of signaling superclass, the return type of the function has to be hinted by the caller.
+     */
+    @Suppress("UNCHECKED_CAST")
+    fun error(throwable: Throwable, fe: Function1<Throwable, /* B */ *>): B =
```
ffgiraldez: have you tried defining `error` as extension function?

katz/src/main/kotlin/katz/effects/internal/AndThen.kt
```diff
@@ -0,0 +1,135 @@
...
+                    isSuccess = true
+                } catch (throwable: Throwable) {
+                    failure = throwable
+                }
+
+        fun <A, B, E> AndThen<A, B>.rotateAccum(_right: AndThen<B, E>): AndThen<A, E> {
+            var me: AndThen<Any?, Any?> = this as AndThen<Any?, Any?>
+            var right: AndThen<Any?, Any?> = _right as AndThen<Any?, Any?>
+            var continued = true
+            while (continued) {
```
ffgiraldez: what's the reason to use while loops instead of tail recursion?

katz/src/main/kotlin/katz/effects/internal/AndThen.kt
```diff
@@ -0,0 +1,135 @@
...
+
+    @Suppress("UNCHECKED_CAST")
+    private fun runLoop(_success: A?, _failure: Throwable?, _isSuccess: Boolean): B {
+        var self: AndThen<Any?, Any?> = this as AndThen<Any?, Any?>
+        var success: Any? = _success
+        var failure = _failure
+        var isSuccess = _isSuccess
+        var continues = true
+
+        fun processSuccess(f: (Any?) -> Any?): Unit =
```
ffgiraldez: functions inside functions

katz/src/main/kotlin/katz/effects/internal/AndThen.kt
pakoito: I would keep both. This class is only for internal use.

katz/src/main/kotlin/katz/effects/internal/AndThen.kt
pakoito: Extension functions inside functions :D

katz/src/main/kotlin/katz/effects/internal/AndThen.kt
pakoito: It seems to work. Amazing! Thanks!

katz/src/main/kotlin/katz/effects/internal/AndThen.kt
pakoito: It's ported from Cats, I don't believe there's any perf gain to be had.

Issue #103, opened 2017-05-29
Add missing tests and laws for Foldable and Traverse instances


pakoito:
Continuation from #101.

Missing tests:

Validated
Coproduct
Either
OptionT
Ior
Try
Id

Laws:

https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/FoldableLaws.scala

PR #108, opened 2017-06-11
Fix StackOverflow in monad instances

pakoito:
After testing some of the Traverse instances, I noticed there was a common StackOverflow problem across some monads. This problem is caused by ap and map calling each other, in expectation that at least one of them is overrided. I've checked all monads and added a simple delegated map call.


pakoito:
I fucked up the stacked branches. Merge only once IO is done.

pakoito:
The error wasn't found before because of #103

katz/src/main/kotlin/katz/effects/IO.kt
```diff
@@ -0,0 +1,220 @@
...
+
+sealed class IO<out A> {
+    abstract fun <B> map(f: (A) -> B): IO<B>
+
+    fun <B> flatMap(f: (A) -> IO<B>): IO<B> =
+            flatMapTotal(
+                    AndThen { a: A ->
+                        try {
+                            f(a)
+                        } catch (error: Throwable) {
```
raulraja: Do we want to catch nay Throwable or just those that are non fatal? https://github.com/scala/scala/blob/v2.9.3/src/library/scala/util/control/NonFatal.scala#L31

katz/src/main/kotlin/katz/effects/internal/AndThen.kt
```diff
@@ -0,0 +1,135 @@
...
+    @Suppress("UNCHECKED_CAST")
+    fun error(throwable: Throwable, fe: Function1<Throwable, /* B */ *>): B =
+            try {
+                runLoop(null, throwable, false)
+            } catch (throwable: Throwable) {
+                fe(throwable) as B
+            }
+
+    @Suppress("UNCHECKED_CAST")
+    private fun runLoop(_success: A?, _failure: Throwable?, _isSuccess: Boolean): B {
```
raulraja: what is the meaning of the `_` convention? We don't have that anywhere else.

katz/src/main/kotlin/katz/effects/Duration.kt
```diff
@@ -0,0 +1,70 @@
+/**
```
raulraja: We should add attributions in the Copying or License files but I thought we decided not to have headers copyrights.

katz/src/main/kotlin/katz/effects/IO.kt
pakoito: This diff is wrong, I made a mistake and is based off the IO branch. The IO changes are here https://github.com/FineCinnamon/Katz/pull/109

Could you please move the comments there?

pakoito:
Deprecated, the issue has been already fixed.

PR #109, opened 2017-06-11
Add IO Monad

pakoito:
This PR adds an implementation of IO based off Cats. It includes a Duration util for blocking operations.


pakoito:
Missing: most tests

https://github.com/typelevel/cats-effect/blob/f975bc8a2e6589f0435b1314158998b65427cfa0/laws/shared/src/test/scala/cats/effect/IOTests.scala

https://github.com/typelevel/cats-effect/blob/master/laws/shared/src/test/scala/cats/effect/IOAsyncTests.scala

katz/src/main/kotlin/katz/instances/IOMonoid.kt
```diff
@@ -0,0 +1,14 @@
+package katz
+
+class IOMonoid<A>(val SM: Monoid<A>) : Monoid<HK<IO.F, A>> {
```
ffgiraldez: i think you could use `IO<A>` instead of HK definition here

katz/src/main/kotlin/katz/instances/IOMonoid.kt
```diff
@@ -0,0 +1,14 @@
...
+    override fun empty(): HK<IO.F, A> =
```
ffgiraldez: same here for returning type

katz/src/main/kotlin/katz/effects/IO.kt
```diff
@@ -0,0 +1,227 @@
...
+
+sealed class IO<out A> : HK<IO.F, A> {
+
+    class F private constructor()
+
+    abstract fun <B> map(f: (A) -> B): IO<B>
+
+    fun <B> flatMap(f: (A) -> IO<B>): IO<B> =
+            flatMapTotal(
+                    AndThen { a: A ->
```
ffgiraldez: it's necessary specify the Type here?

katz/src/main/kotlin/katz/effects/Duration.kt
```diff
@@ -0,0 +1,70 @@
...
+package katz
+
+import java.util.concurrent.TimeUnit
+
+data class Duration(val amount: Long, val timeUnit: TimeUnit) {
+    val nanoseconds: Long by lazy { timeUnit.toNanos(amount) }
+
+    companion object {
+        // Actually limited to 9223372036854775807 days, so unless you are very patient, it is unlimited ;-)
+        val Infinite = Duration(amount = Long.MAX_VALUE, timeUnit = TimeUnit.DAYS)
```
JorgeCastilloPrz: Is the constants naming style capitalized?

pakoito: Fixed!

katz/src/main/kotlin/katz/effects/Duration.kt
```diff
@@ -0,0 +1,70 @@
+/**
```
raulraja: We should have attribution in files like License or where they belong like we do in the typelevel cats case.

katz/src/main/kotlin/katz/instances/IOSemigroup.kt
```diff
@@ -0,0 +1,11 @@
+package katz
+
+class IOSemigroup<A>(val SM: Semigroup<A>) : Semigroup<HK<IO.F, A>> {
```
raulraja: should this use default params for the registered instance value if not provided at instantiation?

katz/src/main/kotlin/katz/effects/IO.kt
```diff
@@ -0,0 +1,228 @@
...
+                    when (result) {
+                        is Either.Right -> f(result.b).unsafeRunAsync(cb)
+                        is Either.Left -> f.error(result.a, { RaiseError(it) }).unsafeRunAsync(cb)
+                    }
+                } catch (throwable: Throwable) {
+                    cb(Either.Left(throwable))
+                }
+            }
+
+    override fun unsafeRunTimedTotal(limit: Duration): Option<A> =
```
raulraja: I think returning `Option` in any of the unsafe methods will be confusing to users. We should return the expected value and since those methods are already unsafe let any exceptions propagate if unfolding the Option returned a NoSuchElementException.

katz/src/main/kotlin/katz/instances/IOSemigroup.kt
pakoito: The companion object has invoke overloaded to provide a default value. n this case IO acts as a wrapper, and the SemiGroup depends mostly on the contents. We could have separate invokes for common types that are not globally registered like numbers and lists.

raulraja: makes sense

katz/src/main/kotlin/katz/effects/Duration.kt
```diff
@@ -0,0 +1,70 @@
...
+data class Duration(val amount: Long, val timeUnit: TimeUnit) {
+    val nanoseconds: Long by lazy { timeUnit.toNanos(amount) }
+
+    companion object {
+        // Actually limited to 9223372036854775807 days, so unless you are very patient, it is unlimited ;-)
+        @JvmStatic val INFINITE = Duration(amount = Long.MAX_VALUE, timeUnit = TimeUnit.DAYS)
+    }
+}
+
+val Long.days: Duration
```
JorgeCastilloPrz: interesting way to provide duration out from any numeric types.

pakoito: It's copypasterino, check the license atop of the file :D

katz/src/main/kotlin/katz/effects/internal/AndThen.kt
```diff
@@ -0,0 +1,131 @@
+package katz.effects.internal
+
+/**
+ * Abandon all hope
```
JorgeCastilloPrz: Could you add some docs here about the purpose of this class ? I am trying to learn something here and that could be helpful :D

katz/src/main/kotlin/katz/effects/Duration.kt
JorgeCastilloPrz: Ok heheheh. Didn't see it.
image

katz/src/main/kotlin/katz/effects/internal/AndThen.kt
pakoito: Done in the branch that adds AndThen!

katz/src/main/kotlin/katz/effects/IO.kt
pakoito: The reason why this return Option is that None represents executing code that returns null. Errors still throw. I don't believe we should stop null-capable execution so we have to choose between returning Option<A> or A?

pakoito: We've had a chat and decided on the API. unsafeRunTimedTotal returns Option, unsafeRun() returns A.

Issue #110, opened 2017-06-13
Add katz-rx module


pakoito:
There is a good chance of experimenting integration with existing apps and frameworks, and RxJava seems like a natural fit. We're looking for someone to take on the following subtasks:

* create a new module depending on katz and RxJava 2.X
* add instances of Monad, Foldable and Traversable for Observable
* introduce some tests and examples of using Observable using imperative binding notation

The estimated effort is around 500-700 lines of code. The maintainers compromise to help and support whomever is completing this ticket, as it'll help assess the difficulty of applying the concepts for newcomers.

pakoito:
Closed by #233

Syntax in #274

PR #111, opened 2017-06-15
Add kategory-rx module

wiyarmir:
Claiming the issue!

From @pakoito 's list:

- [x] create a new module depending on kategory and RxJava 2.X
- [ ] add instances of Monad, Foldable and Traversable for Observable
- [ ] introduce some tests and examples of using Observable using imperative binding notation

Will solve #110


katz-rx/src/main/kotlin/katz/rx/observable/instances/ObservableWrap.kt
```diff
@@ -11,13 +11,9 @@ data class ObservableWrap<A>(val value: Observable<A>) : ObservableWrapKind<A> {
 
     class F private constructor()
 
-    inline fun <B> map(crossinline f: (A) -> B): ObservableWrap<B> =
-            ObservableWrap(value.map({ f(it) }))
+    fun <B> map(f: (A) -> B): ObservableWrap<B> =
+            ObservableWrap(value.map(f::invoke))
```
pakoito: I'm surprised you need ::invoke here. Doesn't inline + `noinline f `work?

wiyarmir: is there a point to have `inline fun ...( noinline f)`? here `f` is the only lambda parameter so both keywords would cancel each other, am I right?

pakoito: It inlines the constructor in the caller method, we skip one function call.

wiyarmir: got it! thanks

katz-rx/src/main/kotlin/katz/rx/observable/instances/ObservableWrap.kt
```diff
@@ -0,0 +1,19 @@
+package katz
+
+import io.reactivex.Observable
+
+typealias ObservableWrapKind<A> = HK<ObservableWrap.F, A>
+
+fun <A> ObservableWrapKind<A>.ev(): ObservableWrap<A> = this as ObservableWrap<A>
+fun <A> ObservableWrapKind<A>.value(): Observable<A> = this.ev().value
+
+data class ObservableWrap<A>(val value: Observable<A>) : ObservableWrapKind<A> {
```
ffgiraldez: i don't like to much `ObservableWrap` always come to my mind a burrito xD
I propose `ObservableHK`or `ObservableKindAdapter` or some variation of HK with suffix Adapter instead of Wrap.

this comment it's not mandatory, just wanna open de discussion.

wiyarmir: seems fair enough, I'm terrible naming classes anyway 😄

katz-rx/src/main/kotlin/katz/rx/observable/instances/ObservableMonad.kt
```diff
@@ -0,0 +1,18 @@
+package katz
+
+import io.reactivex.*
+
+interface ObservableMonad : Monad<ObservableWrap.F> {
+    override fun <A> pure(a: A): ObservableWrap<A> = ObservableWrap(Observable.just(a))
```
pakoito: There is an implicit requirement for monads: they have to override the map or the ap function, as they call into each other recursively because each can be expressed in terms of each other. Given that we already have map in the data type, that'd be my suggestion :D

pakoito:
Done by #233

PR #112, opened 2017-06-15
Typeclasses laws

raulraja:
- [x] FunctorLaws
- [x] ApplicativeLaws
- [x] MonadLaws
- [ ] MonadErrorLaws (deferred to another PR since there is missing instances)


katz/src/main/kotlin/katz/instances/EitherMonad.kt
```diff
@@ -1,6 +1,6 @@
 package katz
 
-class EitherMonad<L> : Monad<EitherF<L>> {
+open class EitherMonad<L> : Monad<EitherF<L>> {
```
pakoito: Why does this class need to be open? Could it be an Interface instead?

katz/src/main/kotlin/katz/data/Kleisli.kt
```diff
@@ -3,14 +3,14 @@ package katz
 typealias KleisiTKind<F, A, B> = HK3<Kleisli.F, F, A, B>
 typealias KleisiF<F> = HK<Kleisli.F, F>
```
pakoito: I believe these two are misnamed too. Could you please use this diff to rename them to Kleisli?

katz/src/test/kotlin/laws/FunctorLaws.kt
```diff
@@ -0,0 +1,32 @@
+package katz
+
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+
+object FunctorLaws {
+
+    inline fun <reified F> laws(AP: Applicative<F> = applicative<F>()): List<Law> =
```
ffgiraldez: why use an Applicative here, if we are testing Functor laws here? should it be convenient use Functor<F> instead of Applicative<F> ins't it?

katz/src/test/kotlin/laws/MonadLaws.kt
```diff
@@ -0,0 +1,49 @@
...
+            forAll(genFunctionAToB<Int, HK<F, Int>>(genApplicative(Gen.int(), M)), Gen.int(), { f: (Int) -> HK<F, Int>, a: Int ->
+                M.flatMap(M.pure(a), f) == f(a)
+            })
+
+    inline fun <reified F> rightIdentity(M: Monad<F> = monad<F>()): Unit =
+            forAll(genApplicative(Gen.int(), M), { fa: HK<F, Int> ->
+                M.flatMap(fa, { M.pure(it) }) == fa
+            })
+
+    inline fun <reified F> kleisliLeftIdentity(M: Monad<F> = monad<F>()): Unit =
```
ffgiraldez: one question, 

why in `MonadLaws` have the particular case for `Kleisli` Monad and not split into `MonadLaws` and `KleisliLaws` that have all the MonadLaws plus those 2 specific test?

katz/src/test/kotlin/laws/FunctorLaws.kt
raulraja: We can't create instances in the generator unless we create a specific typeclass for `pure` or `apply` since that is defined in Applicative and that is what we need to lift values into `F[_]`

katz/src/test/kotlin/laws/MonadLaws.kt
raulraja: Because Kleisli applies to all monads.
https://en.wikipedia.org/wiki/Kleisli_category

> In category theory, a Kleisli category is a category naturally associated to any monad T. It is equivalent to the category of free T-algebras.

raulraja:
@kategory/maintainers I suggest we merge this now since master is already broken in many ways and this proofs some of the types that need immediate action

PR #114, opened 2017-06-18
Perform renames towards "kategory" as the lib name

wiyarmir:
s/katz/kategory/


build.gradle
```diff
@@ -22,7 +22,7 @@ buildscript {
         gradleVersion = '3.5'
         gradleVersionsPluginVersion = '0.15.0'
         javaVersion = '1.7'
-        katzVersion = '0.1-SNAPSHOT'
+        kategoryVersion = '0.1-SNAPSHOT'
```
ffgiraldez: we probably need to update here to 0.3.3-SNAPSHOT

wiyarmir: how have been releases numbered so far? just by manually tagging?

wiyarmir: sorry, I was thinking to extract version to gradle.properties as many maven exported projects do, but that's probably subject for another PR ;)

raulraja:
kategory.io has been registered and we can map it to the doc site as soon as there is one.

wiyarmir:
(closing and reopening to trigger travis build)

PR #120, opened 2017-06-25
Add fixes to Monads found by new Law checks

pakoito:
This PR adds some missing functions and equality check required for some monads to be lawful.


pakoito:
Note that the Eq instance in the laws cannot be a default value because of this bug: https://youtrack.jetbrains.com/oauth?state=%2Fissue%2FKT-18660

kategory/src/main/kotlin/kategory/typeclasses/Eq.kt
```diff
@@ -0,0 +1,18 @@
+package kategory
+
+interface Eq<in F> : Typeclass {
```
raulraja: :clap:

kategory/src/main/kotlin/kategory/typeclasses/Eq.kt
```diff
@@ -0,0 +1,18 @@
...
+    fun eqv(a: F, b: F): Boolean
+
+    fun neqv(a: F, b: F): Boolean =
+            !eqv(a, b)
+
+    companion object {
+        operator fun <F> invoke() = object : Eq<F> {
```
raulraja: I think this should be a default instance of `Eq[Any]` outside of this object.

kategory/src/test/kotlin/kategory/data/EitherTTest.kt
```diff
@@ -10,7 +10,7 @@ import org.junit.runner.RunWith
 class EitherTTest : UnitSpec() {
     init {
 
-        testLaws(MonadLaws.laws(EitherTMonad<Id.F, Int>()))
+        testLaws(MonadLaws.laws(EitherTMonad<Id.F, Int>(), Eq()))
```
raulraja: Probably better named as `EqAny`

kategory/src/test/kotlin/kategory/data/EvalTest.kt
```diff
@@ -8,7 +8,10 @@ import org.junit.runner.RunWith
 class EvalTest : UnitSpec() {
     init {
 
-        testLaws(MonadLaws.laws(Eval))
+        testLaws(MonadLaws.laws(Eval, object : Eq<HK<Eval.F, Int>> {
```
raulraja: We should have this instance in the core not just in tests and there should be an instance for each one of the datatypes we implement.

kategory/src/test/kotlin/kategory/data/Function0Test.kt
```diff
@@ -8,7 +8,10 @@ import org.junit.runner.RunWith
 class Function0Test : UnitSpec() {
     init {
 
-        testLaws(MonadLaws.laws(Function0))
+        testLaws(MonadLaws.laws(Function0, object : Eq<HK<Function0.F, Int>> {
```
raulraja: We should have this instance in the core not just in tests.

kategory/src/test/kotlin/kategory/laws/Law.kt
```diff
@@ -2,3 +2,5 @@ package kategory
 
 data class Law(val name: String, val test: () -> Unit)
 
+inline fun <reified A> A.equalUnderTheLaw(b: A, eq: Eq<A> = Eq()): Boolean =
```
raulraja: Can we have extension functions that are `inline reify... <F, ..>` Once we have instances for the datatypes as part of our effort to provide syntax all datatypes will include syntax for `eqv` and `neqv` which can be used directly here provided thre is a registeres `Eq` instance for the datatype being compared.

kategory/src/test/kotlin/kategory/laws/MonadLaws.kt
```diff
@@ -1,49 +1,47 @@
...
 object MonadLaws {
 
-    inline fun <reified F> laws(M: Monad<F> = monad<F>()): List<Law> =
-            ApplicativeLaws.laws(M) + listOf(
-                    Law("Monad Laws: left identity", { leftIdentity(M) }),
-                    Law("Monad Laws: right identity", { rightIdentity(M) }),
-                    Law("Monad Laws: kleisli left identity", { kleisliLeftIdentity(M) }),
-                    Law("Monad Laws: kleisli right identity", { kleisliRightIdentity(M) }),
-                    Law("Monad Laws: map / flatMap coherence", { mapFlatMapCoherence(M) }),
-                    Law("Monad / JVM: stack safe", { mapFlatMapCoherence(M) })
```
raulraja: Thanks!

PR #122, opened 2017-06-27
MonadError Laws

raulraja:
Includes tests and instances for Try and Either

### DIFF
--- a/kategory/src/main/kotlin/kategory/instances/EitherMonad.kt
+++ b/kategory/src/main/kotlin/kategory/instances/EitherMonad.kt
@@ -1,6 +1,6 @@
 package kategory
 
-class EitherMonad<L> : Monad<EitherF<L>> {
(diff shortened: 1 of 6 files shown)


kategory/src/main/kotlin/kategory/instances/EitherMonad.kt
```diff
@@ -1,6 +1,6 @@
 package kategory
 
-class EitherMonad<L> : Monad<EitherF<L>> {
+open class EitherMonad<L> : Monad<EitherF<L>> {
```
pakoito: I'd rather it used delegation rather than inheritance like in here: https://github.com/kategory/kategory/blob/master/kategory/src/main/kotlin/kategory/instances/Composed.kt#L39

raulraja: Sounds good, we can do a pass to redo all of them if this makes sense. Let's discuss by chat.

PR #128, opened 2017-07-16
Add laws for Foldable instances

pakoito:
Working towards #103 

In this PR the test for all Foldable laws are added. I'll add the last for Traverse on the next PR, so we can have both reviewed separately.

Reference: https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/TraverseLaws.scala

I have tested them with Id and Coproduct, which will get Traverse laws on the next PR instead.


kategory/src/main/kotlin/kategory/typeclasses/Foldable.kt
```diff
@@ -63,18 +64,16 @@ interface Foldable<F> : Typeclass {
      * Certain structures are able to implement this in such a way that folds can be short-circuited (not traverse the
      * entirety of the structure), depending on the G result produced at a given step.
      */
-    fun <G, A, B> foldM(MG: Monad<G>, fa: HK<F, A>, z: B, f: (B, A) -> HK<G, B>): HK<G, B> {
-        return foldL(fa, MG.pure(z), { gb, a -> MG.flatMap(gb) { f(it, a) } })
-    }
+    fun <G, A, B> foldM(MG: Monad<G>, fa: HK<F, A>, z: B, f: (B, A) -> HK<G, B>): HK<G, B> =
```
raulraja: should this contain the default param value `= monad()`?

kategory/src/main/kotlin/kategory/typeclasses/Foldable.kt
```diff
@@ -63,18 +64,16 @@ interface Foldable<F> : Typeclass {
...
 
     /**
      * Monadic folding on F by mapping A values to G<B>, combining the B values using the given Monoid<B> instance.
      *
      * Similar to foldM, but using a Monoid<B>.
      */
-    fun <G, A, B> foldMapM(MG: Monad<G>, bb: Monoid<B>, fa: HK<F, A>, f: (A) -> HK<G, B>) : HK<G, B> {
-        return foldM(MG, fa, bb.empty(), { b, a -> MG.map(f(a)) { bb.combine(b, it) } })
-    }
+    fun <G, A, B> foldMapM(MG: Monad<G>, bb: Monoid<B>, fa: HK<F, A>, f: (A) -> HK<G, B>) : HK<G, B> =
```
raulraja: Same for monoid

kategory/src/main/kotlin/kategory/typeclasses/Foldable.kt
pakoito: Sadly it's not possible

screen shot 2017-07-17 at 12 26 42 am

raulraja: what if it is expressed as an extension function?

pakoito: That might work. I'm starting to seriously consider that we'd have to express all functions bar the pure virtual ones as extensions.

pakoito: I've added them to this PR https://github.com/kategory/kategory/pull/129

PR #129, opened 2017-07-17
Add instance lookup for Foldable and Traverse methods

pakoito:
This PR converts some methods that require typeclass instances into inline extension functions that allow for implicit lookup using our helpers.

### DIFF
--- a/kategory/src/main/kotlin/kategory/instances/Traverse.kt
+++ b/kategory/src/main/kotlin/kategory/instances/Traverse.kt
@@ -11,18 +11,18 @@ interface Traverse<F> : Functor<F>, Foldable<F>, Typeclass {
      */
     fun <G, A, B> traverse(fa: HK<F, A>, f: (A) -> HK<G, B>, GA: Applicative<G>): HK<G, HK<F, B>>
 
-    /**
-     * Thread all the G effects through the F structure to invert the structure from F<G<A>> to G<F<A>>.
-     */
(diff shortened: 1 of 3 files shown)


JorgeCastilloPrz:
Why moving an instance related function to be an extension function makes any difference in terms of usage? I guess it shouldn't, since that's the idea of extension functions, so they are used exactly the same way a function from an instance of the same type would be used, isn't it?

So how is this change helping us to use the global instance helper classes over it? I would love a bit deeper description about what we are trying to do here, otherwise It's not possible to get reviewed.

pakoito:
@JorgeCastilloPrz We have a series of lookup methods for the different typeclasses, loosely based off implicits on Scala. To do the lookup it uses runtime Class objects. 

When using generics, accessing the types of the parameters functions are generalized for isn't possible due to JVM limitations: type erasure. At runtime all List<T> are simple List, same for Set, Monad, etc...

Kotlin adds a new tool to the JVM arsenal: reified generics. A reified generic is an indication that one or more of the parameters has to also be captured as a runtime value. This allows us to call the lookup functions with them. To enable reified generics it is required that the function is inlined. For a function to be inlined there are some requirements: in this case not being virtual (cannot be possibly overridden).

Functions in an interface are virtual by default, as per interface semantics (https://github.com/kategory/kategory/pull/128#discussion_r127619894), so our only option is making them into extension functions. They look the same, perform better, disable weird OOP overrides, and allow lookup.

PR #130, opened 2017-07-17
WIP. Reorganize all instances in a single file per datatype

raulraja:
WIP,  Do not merge.

This PR reorganizes all instances to be in a single file for all the instances of a given datatype.
Instances that can be materialized without args are aggregated as part of the dataype's companion. 
Others such as `OptionMonoid` are exposed in the companion's function as functions that you can invoke with the proper params to retrieve the instance.


kategory/src/main/kotlin/kategory/data/StateT.kt
```diff
@@ -18,6 +18,16 @@ class StateT<F, S, A>(
     companion object {
         inline operator fun <reified F, S, A> invoke(run: StateTFunKind<F, S, A>, MF: Monad<F> = monad<F>()): StateT<F, S, A> =
                 StateT(MF, run)
+
+        fun <F, S> instances(MF: Monad<F>): StateTInstances<F, S> = object : StateTInstances<F, S> {
```
pakoito: This could use inline + reify + default value for monad.

kategory/src/test/kotlin/kategory/data/EvalTest.kt
```diff
@@ -8,7 +8,10 @@ import org.junit.runner.RunWith
 class EvalTest : UnitSpec() {
     init {
 
-        testLaws(MonadLaws.laws(Eval, EvalEq()))
+        testLaws(MonadLaws.laws(Eval, object : Eq<HK<Eval.F, Int>> {
```
pakoito: The Eq instance for Eval should be valid even for the main library. Was this a merge conflict, or is there something wrong with the previous one?

raulraja: It was fixed to Int, that would just work on tests.

kategory/src/main/kotlin/kategory/data/StateT.kt
raulraja: I'll do a pass on all those methods before merging

kategory/src/test/kotlin/kategory/data/EvalTest.kt
pakoito: Gotcha! Also, woops :P Could you please make a generic version instead that wraps another Eq?

kategory/src/main/kotlin/kategory/instances/IOEq.kt
```diff
@@ -1,6 +0,0 @@
-package kategory
-
-class IOEq : Eq<HK<IO.F, Int>> {
-    override fun eqv(a: HK<IO.F, Int>, b: HK<IO.F, Int>): Boolean =
```
pakoito: My god, I did all of them wrong. Shame on me.

kategory/src/test/kotlin/kategory/data/EitherTTest.kt
```diff
@@ -165,27 +165,27 @@ class EitherTTest : UnitSpec() {
...
             }
         }
 
         "EitherTMonad#tailRecM should execute and terminate without blowing up the stack" {
             forAll { a: Int ->
-                val value: EitherT<Id.F, Int, Int> = EitherTInstances<Id.F, Int>(Id).tailRecM(a) { b ->
+                val value: EitherT<Id.F, Int, Int> = EitherT.monad<Id.F, Int>(Id).tailRecM(a) { b ->
                     EitherT.pure<Id.F, Int, Either<Int, Int>>(Either.Right(b * a))
-                }
+                }.ev().ev()
```
pakoito: These two `ev()` may be causing an exception and why Travis stopped working after the previous test.

```
kategory.EitherTTest > EitherTTest.EitherTMonad#flatMap should be consistent with EitherT#flatMap PASSED
:kategory:test FAILED 
```

Seems like something for the Kotlin team.

kategory/src/main/kotlin/kategory/data/WriterT.kt
```diff
@@ -10,29 +10,29 @@ data class WriterT<F, W, A>(val MF: Monad<F>, val value: HK<F, Tuple2<W, A>>) :
     class F private constructor()
 
     companion object {
-        inline fun <reified F, reified W, A> pure(a: A, MM: Monoid<W> = monoid(), MF: Monad<F> = monad()) =
+        inline fun <reified F, reified W, A> pure(a: A, MM: Monoid<W> = monoid(), MF: Monad<F> = kategory.monad()) =
```
pakoito: Why is the disambiguation needed here and not on the other instances?

raulraja: because providing default args makes the monad() method ambiguous with the one in kategory. For other types where different args are required they are not necesary because is able to determine if it comes from kategory or the same companion instance.

PR #132, opened 2017-07-19
Syntax for all typeclass targets, values and datatypes

raulraja:


kategory/src/main/kotlin/kategory/typeclasses/Foldable.kt
```diff
@@ -31,16 +31,6 @@ interface Foldable<in F> : Typeclass {
     fun <A, B> foldR(fa: HK<F, A>, lb: Eval<B>, f: (A, Eval<B>) -> Eval<B>): Eval<B>
 
     /**
-     * The size of this Foldable.
-     *
-     * This can be overriden in structures that have more efficient size implementations.
-     *
-     * Note: will not terminate for infinite-sized collections.
-     */
-    fun <A> size(ml: Monoid<Long>, fa: HK<F, A>): Long =
```
pakoito: What happened to size?

kategory/src/main/kotlin/kategory/typeclasses/MonadReader.kt
```diff
@@ -11,5 +11,11 @@ interface MonadReader<F, D> : Monad<F> {
     fun <A> reader(f: (D) -> A): HK<F, A> = map(ask(), f)
 }
 
+inline fun <reified F, A, reified D> HK<F, A>.local(FT: MonadReader<F, D> = monadReader(), noinline f: (D) -> D): HK<F, A> =
```
pakoito: Can you call local with syntactic sugar, and still get the default?

raulraja: Not sure I get what you mean by the `default`. Can you please clarify? thanks!

pakoito: Assuming Option has a monadReader, could you call `1.some().local { it }`? I thought defaults only worked if they were the last parameters.

raulraja: Yes, that compiles though it fails at runtime because Option has no MonadReader instance which is one of the challenges we face because our instance lookups are at runtime since Kotlin lacks implicits or any proper support for typeclasses.

pakoito: I'm on it, I'm on it. Soon :D

PR #137, opened 2017-07-21
Add Function1

pakoito:
Another breakthrough in language limitations. This PR adds the first kind projection using skolemization!

The PR adds a HK wrapper for Functions with 1 parameter, along with an instance of monad() and monadReader()

For a background on the buzzwords, check these links:

https://github.com/non/kind-projector
https://gist.github.com/jdegoes/97459c0045f373f4eaf126998d8f65dc#skolemization

### DIFF
--- a/kategory/src/main/kotlin/kategory/arrow/Function1.kt
+++ b/kategory/src/main/kotlin/kategory/arrow/Function1.kt
@@ -18,38 +18,12 @@ data class Function1<in A, out R>(val f: (A) -> R) : FunctionInject<R>, Function
     class F private constructor()
 
     companion object {
-        fun <P> monad() = object : Function1MonadReader<P> {}
+        fun <P> functor() = object : Function1Instances<P> {}
 
-        fun <P> monadReader() = object : Function1MonadReader<P> {}
+        fun <P> applicative() = object : Function1Instances<P> {}
+
+        fun <P> monad() = object : Function1Instances<P> {}
+
+        fun <P> monadReader() = object : Function1Instances<P> {}
     }
-}
-
-interface Function1MonadReader<P> : MonadReader<Function1.F, P> {
-
-    override fun ask(): HK<Function1.F, P> =
-            { a: P -> a }.k()
-
-    override fun <A> local(f: (P) -> P, fa: HK<Function1.F, A>): Function1<P, A> =
-            f.andThen { fa.ev().invoke(it) }.k()
-
-    override fun <A> pure(a: A): Function1<P, A> =
-            { _: P -> a }.k()
-
-    override fun <A, B> map(fa: HK<Function1.F, A>, f: (A) -> B): HK<Function1.F, B> =
-            f.compose { b: B -> fa.ev().invoke(b) }.k()
-
-    override fun <A, B> flatMap(fa: HK<Function1.F, A>, f: (A) -> HK<Function1.F, B>): Function1<P, B> =
-            Function1 { p -> f(fa.ev().invoke(p)).ev().invoke(p) }
-
-    override fun <A, B> tailRecM(a: A, f: (A) -> HK<Function1.F, Either<A, B>>): Function1<P, B> =
-            Function1 { p ->
-                tailrec fun loop(thisA: A): B =
-                        f(thisA).ev().invoke(p).fold({ loop(it) }, { it })
-
-                loop(a)
-            }
-
-    fun monad(): Monad<Function1.F> = this
-
-    fun monadReader(): MonadReader<Function1.F, P> = this
 }
--- a/kategory/src/main/kotlin/kategory/arrow/Function1.kt
+++ b/kategory/src/main/kotlin/kategory/arrow/Function1.kt
@@ -1,0 +1,55 @@
+package kategory
+
+typealias Function1F<P> = HK<Function1.F, P>
+
+fun <P, R> ((P) -> R).k(): Function1<P, R> =
+        Function1(this)
+
+@Suppress("UNCHECKED_CAST")
+fun <R> Function1F<R>.ev(): FunctionInject<R> =
+        this as FunctionInject<R>
+
+// We don't we want an inherited class to avoid equivalence issues, so a simple HK wrapper will do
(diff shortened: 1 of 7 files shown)


kategory/src/main/kotlin/kategory/arrow/FunctionBijections.kt
```diff
@@ -0,0 +1,9 @@
+package kategory
+
+interface FunctionInject<out R> {
+    operator fun <P> invoke(p: P): R
+}
+
+interface FunctionSurject<in P> {
```
ffgiraldez: where is needed this interface?

kategory/src/main/kotlin/kategory/instances/Function1Instances.kt
```diff
@@ -0,0 +1,31 @@
...
+    override fun ask(): HK<Function1.F, P> =
+            { a: P -> a }.k()
+
+    override fun <A> local(f: (P) -> P, fa: HK<Function1.F, A>): Function1<P, A> =
+            f.andThen { fa.ev().invoke(it) }.k()
+
+    override fun <A> pure(a: A): Function1<P, A> =
+            { _: P -> a }.k()
+
+    override fun <A, B> map(fa: HK<Function1.F, A>, f: (A) -> B): HK<Function1.F, B> =
```
ffgiraldez: return type could be `Function1<F,B>` ?

kategory/src/main/kotlin/kategory/instances/Function1Instances.kt
```diff
@@ -0,0 +1,31 @@
+package kategory
+
+interface Function1Instances<P> :
+        Functor<Function1.F>,
+        Applicative<Function1.F>,
+        Monad<Function1.F>,
+        MonadReader<Function1.F, P> {
+
+    override fun ask(): HK<Function1.F, P> =
```
ffgiraldez: what the reason to not return `Function1<F,P>`? not expose to callers `FunctionInject`?

kategory/src/main/kotlin/kategory/arrow/FunctionBijections.kt
```diff
@@ -0,0 +1,9 @@
+package kategory
+
+interface FunctionInject<out R> {
```
ffgiraldez: we want those Bijections to be public?

kategory/src/main/kotlin/kategory/arrow/Function1.kt
```diff
@@ -0,0 +1,29 @@
...
+typealias Function1F<R> = HK<Function1.F, R>
+
+fun <P, R> ((P) -> R).k(): Function1<P, R> =
+        Function1(this)
+
+@Suppress("UNCHECKED_CAST")
+fun <R> Function1F<R>.ev(): FunctionInject<R> =
+        this as FunctionInject<R>
+
+// We don't we want an inherited class to avoid equivalence issues, so a simple HK wrapper will do
```
raulraja: We don't ~~we~~ want

kategory/src/test/kotlin/kategory/data/Function1Test.kt
```diff
@@ -0,0 +1,25 @@
+package kategory
+
+import io.kotlintest.KTestJUnitRunner
+import io.kotlintest.matchers.shouldBe
+import org.junit.runner.RunWith
+
+@RunWith(KTestJUnitRunner::class)
+class Function1Test : UnitSpec() {
+    init {
+        testLaws(MonadLaws.laws(Function1.monad<Int>(), object : Eq<HK<Function1.F, Int>> {
```
raulraja: Since this a new instance of `MonadReader` and so far we have Kleisli and this one I think it's time to add the missing `MonadReader` laws and apply them both here and in `Kleisli`, fine if addressed in a separate PR.

kategory/src/test/kotlin/kategory/data/Function1Test.kt
```diff
@@ -0,0 +1,25 @@
...
+            override fun eqv(a: HK<Function1.F, Int>, b: HK<Function1.F, Int>): Boolean =
+                    a.ev()(1) == b.ev()(1)
+        }))
+
+        "Function1Monad.binding should for comprehend over all values of multiple Function0" {
```
raulraja: We may be able to add generic tests for Monad and Applicative to tests comprehensions and the applicative builder works for all monads and applicative instances instead of hardcoding these on each datatype test.

kategory/src/test/kotlin/kategory/data/Function1Test.kt
raulraja: Also just realized we are missing `Category` and `MonoidK` as typeclasses, also addressable in a separate PR. `Function1` should also have an instance for `MonoidK`

kategory/src/main/kotlin/kategory/arrow/FunctionBijections.kt
pakoito: Yes, we have very few private things on the repo. It's all tools, really.

kategory/src/main/kotlin/kategory/arrow/FunctionBijections.kt
pakoito: Nowhere at the moment. It will, as it's dual of the Inject.

kategory/src/main/kotlin/kategory/instances/Function1Instances.kt
pakoito: The compiler doesn't have enough information about the generic F in this context, so it fails.

kategory/src/main/kotlin/kategory/arrow/FunctionBijections.kt
ffgiraldez: ok, make sense

kategory/src/main/kotlin/kategory/instances/Function1Instances.kt
pakoito: Same as above, the parameter info isn't available to the compiler. It infers Function1<B, B>.

PR #141, opened 2017-07-22
Simplify Function0 and Function1 

pakoito:
After a chat with @raulraja I realized I had overblown the implementation of Function1. There was a simpler solution all along using extension functions. Who would have thought.

This PR adds a new function invoke() that's an alias for ev().invoke(). This fixes an existing problem with the inference for ev() because one of the parameters couldn't be decided by the compiler and had to be injected manually.

### DIFF
--- a/kategory/src/main/kotlin/kategory/arrow/Function1.kt
+++ b/kategory/src/main/kotlin/kategory/arrow/Function1.kt
@@ -14,10 +14,10 @@ fun <P, R> Function1F<R>.invoke(p: P): R =
 
 // We don't want an inherited class to avoid equivalence issues, so a simple HK wrapper will do
 data class Function1<in A, out R>(val f: (A) -> R) : Function1F<R> {
+    class F private constructor()
+
     operator fun invoke(a: A): R =
             f(a)
-
-    class F private constructor()
 
     companion object {
         fun <P> functor() = object : Function1Instances<P> {}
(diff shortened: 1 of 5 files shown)


kategory/src/main/kotlin/kategory/arrow/Function1.kt
```diff
@@ -6,20 +6,19 @@ fun <P, R> ((P) -> R).k(): Function1<P, R> =
         Function1(this)
 
 @Suppress("UNCHECKED_CAST")
-fun <R> Function1F<R>.ev(): FunctionInject<R> =
-        this as FunctionInject<R>
+fun <P, R> Function1F<R>.ev(): (P) -> R =
+        (this as Function1<P, R>).f
```
pakoito: Look at this function. The P is only part of the output, so it has to be specified at callsite as the compiler inference doesn't necessarily pick it up.

kategory/src/main/kotlin/kategory/arrow/Function1.kt
```diff
@@ -6,20 +6,19 @@ fun <P, R> ((P) -> R).k(): Function1<P, R> =
...
+
+fun <P, R> Function1F<R>.invoke(p: P): R =
+        this.ev<P, R>().invoke(p)
```
pakoito: Invoke to the rescue! It hides the intermediate ev() step from the callsites and provides inference for the compiler.

PR #146, opened 2017-07-25
MonadState

### LINKED ISSUE #145: Add MonadState construct

JorgeCastilloPrz:
#fixes #145

* Added MonadState
* Added MonadStateLaws
* Test laws

### DIFF
--- a/kategory/src/main/kotlin/kategory/data/StateT.kt
+++ b/kategory/src/main/kotlin/kategory/data/StateT.kt
@@ -9,6 +9,9 @@ typealias StateTFunKind<F, S, A> = HK<F, StateTFun<F, S, A>>
 fun <F, S, A> StateTKind<F, S, A>.ev(): StateT<F, S, A> =
         this as StateT<F, S, A>
 
+fun <F, S, A> StateTKind<F, S, A>.runM(initial: S): HK<F, Tuple2<S, A>> =
+        (this as StateT<F, S, A>).run(initial)
+
 class StateT<F, S, A>(
         val MF: Monad<F>,
         val runF: StateTFunKind<F, S, A>
@@ -61,7 +64,7 @@ class StateT<F, S, A>(
                     MF.map(runF) { sfsa ->
                         sfsa.andThen { fsa ->
                             MF.flatMap(fsa) {
-                                fas(it.b).ev().run(it.a)
+                                fas(it.b).runM(it.a)
                             }
                         }
                     }
--- a/kategory/src/main/kotlin/kategory/data/StateT.kt
+++ b/kategory/src/main/kotlin/kategory/data/StateT.kt
@@ -28,6 +28,8 @@ class StateT<F, S, A>(
         inline fun <reified F, S> applicative(MF: Monad<F> = monad<F>()): StateTInstances<F, S> = instances(MF)
 
         inline fun <reified F, S> monad(MF: Monad<F> = monad<F>()): StateTInstances<F, S> = instances(MF)
+
+        inline fun <reified F, reified S> monadState(MF: Monad<F> = monad<F>()): StateTInstances<F, S> = instances(MF)
     }
 
     fun <B> map(f: (A) -> B): StateT<F, S, B> =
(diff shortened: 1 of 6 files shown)


kategory/src/main/kotlin/kategory/data/StateT.kt
```diff
@@ -16,8 +16,8 @@ class StateT<F, S, A>(
     class F private constructor()
 
     companion object {
-        inline operator fun <reified F, S, A> invoke(run: StateTFunKind<F, S, A>, MF: Monad<F> = monad<F>()): StateT<F, S, A> =
-                StateT(MF, run)
+        operator fun <F, S, A> invoke(run: (S) -> HK<F, Tuple2<S, A>>, MF: Monad<F>): StateT<F, S, A> =
```
pakoito: For all smart constructors we generally use inline + reified + default monad call. This way we can replace with the lookup later

pakoito: For StateTFunKind, if it's not going to be used does it make sense to just remove the typealias too?

kategory/src/test/kotlin/kategory/laws/MonadStateLaws.kt
```diff
@@ -0,0 +1,36 @@
...
+import io.kotlintest.properties.forAll
+
+object MonadStateLaws {
+
+    inline fun <reified F> laws(M: MonadState<F, Int> = monadState<F, Int>(), EQ: Eq<HK<F, Int>>): List<Law> =
+            MonadLaws.laws(M, EQ) + listOf(
+                    Law("Monad State Laws: idempotence", { monadStateGetIdempotent(M, Eq.any()) }),
+                    Law("Monad State Laws: set twice", { monadStateSetTwice(M, Eq.any()) }),
+                    Law("Monad State Laws: set get", { monadStateSetTwice(M, Eq.any()) }),
+                    Law("Monad State Laws: get set", { monadStateSetTwice(M, Eq.any()) })
```
pakoito: For some HKs, like IO for the case of Monad, this Eq instance will not work as their wrapped values, their equality identity, isn't available until they're run. The Eq instance would need to be passed as a parameter.

This is okay for tests, you can see an example here: https://github.com/kategory/kategory/blob/master/kategory/src/test/kotlin/kategory/effects/IOTest.kt#L15

pakoito: In retrospect this looks like a typo more than anything else :D

kategory/src/main/kotlin/kategory/data/StateT.kt
JorgeCastilloPrz: This one was tweaked by @raulraja because I had problems to invoke the StateT from `StateTInstances`. Feels like the type F cannot be reified here, otherwise it wouldn't be able to be used from the overriden `get()` and `set` methods inside `StateTInstances` because of:

captura de pantalla 2017-07-26 a las 8 42 48

JorgeCastilloPrz: It's still being used inside `applyF` and extension function `stateT` from `StateT`

kategory/src/test/kotlin/kategory/laws/MonadStateLaws.kt
JorgeCastilloPrz: Nono, it's not a typo. I was just ignoring the EQ already passed as argument to compare equality on 2 StateTKind<Try.F, Int, Int> and just sending Eq.any() for all the laws internally. That's why they don't pass. They also need different implementations of EQ though, the one passed as an argument is not under the required constrains.

kategory/src/main/kotlin/kategory/data/StateT.kt
pakoito: Could you please try to define it using pure or the other constructor pure() uses instead?

pakoito: Then for consistency should it be used here too?

JorgeCastilloPrz:
@pakoito @raulraja I got all the laws passing now except for the "set get" one. I have added an example of flakyness under set or get methods. You can check `MonadStateLaws#monadStateSetGet`.

As you will be able to see, It passes when you run the laws for a single comparison. If you comment the single line and uncomment the ones on top wrapped inside the generators, then it fails, even if it's using the same exact value (hardcoded).

Can it be side effects not being properly isolated between test executions ?

JorgeCastilloPrz:
Looks good @pakoito . Thanks for the fixes. I have rebased master to put it up to date. Can we move on and integrate it ?

PR #150, opened 2017-07-26
SemigroupK

JorgeCastilloPrz:
#149 

`SemigroupK` is basically a semigroup but capable of playing with generic type constructors. A normal semigroup would play with concrete types like `Option<Int>`, `Int`, `List<String>`. All the types and their generics are defined. The `SemigroupK` is defined to play with `Option`, `List` ...etc, and leave their generic type to be defined by the caller.

You can find better documentation [here](https://github.com/typelevel/cats/blob/master/docs/src/main/tut/typeclasses/semigroupk.md).

* Added SemigroupK
* Added laws
* Tested laws
* Used a dummy SemigroupK implementation to test laws until we have more relevant instances.
* Added semigroupK instances to all the types needed.


kategory/src/main/kotlin/kategory/typeclasses/SemigroupK.kt
```diff
@@ -0,0 +1,28 @@
...
+    fun <A> algebra(): Semigroup<HK<F, A>> = object : Semigroup<HK<F, A>> {
+        override fun combine(x: HK<F, A>, y: HK<F, A>): HK<F, A> =
+                combineK(x, y)
+    }
+}
+
+/**
+ * Dummy SemigroupK instance to be able to test laws.
+ */
+class OptionSemigroupK : SemigroupK<Option.F> {
```
pakoito: To the Option instances file!

JorgeCastilloPrz: 👍  Done

pakoito:
The PR is missing the important bit, combining two kinds into a semigroup.

https://github.com/typelevel/cats/blob/155f7f534993c30d6e757de990330ac796dad5da/core/src/main/scala/cats/SemigroupK.scala#L39

In Kategory this is a new instance like the ones in Composed.kt. Follow the example of ComposedTraverse: https://github.com/kategory/kategory/blob/master/kategory/src/main/kotlin/kategory/data/Composed.kt#L38

raulraja:
There is also some `SemigroupK` instances we can add for some of our existing data types: https://github.com/typelevel/cats/search?utf8=%E2%9C%93&q=SemigroupK&type=

kategory/src/main/kotlin/kategory/instances/OptionInstances.kt
```diff
@@ -81,3 +81,12 @@ interface OptionInstances :
             fa.ev().orElse({ f(Unit).ev() })
 
 }
+
+/**
+ * Dummy SemigroupK instance to be able to test laws for SemigroupK.
+ */
+class OptionSemigroupK : SemigroupK<Option.F> {
+    override fun <A> combineK(x: HK<Option.F, A>, y: HK<Option.F, A>): HK<Option.F, A> {
```
raulraja: When implementing typeclasses an IDEA automatically suggests: `HK<Option.F, A>` what the IDE really wants is for you to manually go and change that to `Option<A>` xD. This allows people using the instances not to have to upcast with `ev()` all over the place.

JorgeCastilloPrz: Tried doing it but then I can't compile anymore. The method is not recognized as a real override of:

`fun <A> combineK(x: HK<F, A>, y: HK<F, A>): HK<F, A>`

It feels like the typealias is not properly working for this case.

captura de pantalla 2017-07-27 a las 18 17 48

JorgeCastilloPrz:
I have tried declaring the compose function the same way you do @pakoito , but seems like the `cats` implementation does not apply equally here for some reason. The type `F` returned should be `ComposedType<F, G>` itself, but it is not being understood on this way since the function outside is passing it just as `F`, so could be literally any type, therefore it needs to be ComposedType<F, G> to be able to call `combineK` method for `x` and `y`.

I have pushed it as it is now so you can take a look.

captura de pantalla 2017-07-27 a las 18 22 48

pakoito:
@JorgeCastilloPrz You need the trickity tricks, else you'll never align types

` F().combineK(x.lower(), y.lower()).lift()``

Committing now along with combineKC

JorgeCastilloPrz:
Thanks to both. I tried lowering both arguments but didn't realize that I needed to lift everything again to match the return type.

JorgeCastilloPrz:
Sounds ok to me, but what happens for all the classes / types / whatever that we still don't have and also contain instances of `SemigroupK`, for example? Should I add those files with no more than my instance inside or just leave those `SemigroupK` instances undone?

raulraja:
Yeah, I ment just for the ones we have and then when we add new datatypes do the same but the other way around. For that given datatype all instances for all currently available typeclasses should be added. This way we ensure when adding something new eveything possible to that feature is included. 
As for adding typeclasses that depend upon other typeclasses we need to resolve those deps first so if for example say we were implementing Monad for the first time we should have added Functor and Applicative before we attempt to introduce Monad.

JorgeCastilloPrz:
I have added `semigroupK` instances for `StateT`, `Either`, `EitherT`, `WriterT`,  `NonEmptyList`,  `OptionT`. I think there are no other instances left for the types we currently have.

Can you take a look again @pakoito @raulraja ? 

Thanks.

kategory/src/main/kotlin/kategory/data/EitherT.kt
```diff
@@ -52,6 +52,17 @@ data class EitherT<F, A, B>(val MF: Monad<F>, val value: HK<F, Either<A, B>>) :
         }
 
         inline fun <reified F, A> foldable(FF: Traverse<F> = traverse<F>(), MF: Monad<F> = monad<F>()): Foldable<EitherTF<F, A>> = traverse(FF, MF)
+
+        inline fun <reified F, L> semigroupK(MF: Monad<F> = monad<F>()): SemigroupK<EitherTF<F, L>> =
+                object : SemigroupK<EitherTF<F, L>> {
```
pakoito: This should be in the instance rather than in the companion object.

kategory/src/main/kotlin/kategory/data/NonEmptyList.kt
```diff
@@ -76,8 +76,12 @@ class NonEmptyList<out A> private constructor(
         fun monad(): Monad<NonEmptyList.F> = this
 
         fun <A> semigroup(): Semigroup<NonEmptyList<A>> = object : NonEmptyListSemigroup<A> {}
+
+        fun semigroupK(): SemigroupK<NonEmptyList.F> = object : SemigroupK<NonEmptyList.F> {
+            override fun <A> combineK(x: HK<F, A>, y: HK<F, A>): NonEmptyList<A> = x.ev().plus(y.ev())
```
pakoito: Same as above

kategory/src/main/kotlin/kategory/data/StateT.kt
```diff
@@ -22,6 +22,9 @@ class StateT<F, S, A>(
         inline operator fun <reified F, S, A> invoke(noinline run: StateTFun<F, S, A>, MF: Monad<F> = monad<F>()): StateT<F, S, A> =
                 StateT(MF, MF.pure(run))
 
+        fun <F, S, A> invokeF(runF: StateTFunKind<F, S, A>, MF: Monad<F>): StateT<F, S, A> =
```
pakoito: Could this be inlined and a defualt monad instance given?

JorgeCastilloPrz: Sorry but can't. All the StateT class instance methods delegate their behavior on this one, so reifying the type here means we would need to make all of them inline reified also. I think that was the main reason for it to not be in the companion object, may be. But even tho, I think it's better placed inside of the comp obj.

JorgeCastilloPrz:
Fixed the instances stuff you requested @pakoito. I think the last one about the inline reified on the invokeF is not possible.

kategory/src/main/kotlin/kategory/data/StateT.kt
pakoito: Check as an extension function instead, just in case.

JorgeCastilloPrz: Same story. You can't use a reified type F even if you move it to extension function (tried it). In fact, that would be changing the main intention of the function, since it would require a valid `StateT` runtime instance to be executed, so it would not be a static utility method to create a `StateT` from nothing anymore.

PR #158, opened 2017-07-27
Fixes Function1 and Function0 stack safety and parametrization

raulraja:


kategory/src/main/kotlin/kategory/data/Trampoline.kt
```diff
@@ -15,9 +15,9 @@ interface TrampolineFunctions {
 
     fun <A> done(a: A): TrampolineF<A> = Free.pure<Function0.F, A>(a)
 
-    fun <A> suspend(a: () -> TrampolineF<A>): TrampolineF<A> = defer(a)
```
pakoito: Where did this function go?

PR #161, opened 2017-07-30
MonoidK

### LINKED ISSUE #151: Add MonoidK

JorgeCastilloPrz:
Fixes #151

* Added `MonoidK` and laws for it depending on `SemigroupK`.
* Added a bunch of concrete instance `monoidK()` companion obj functions for all the dependent classes we currently have (EitherT.monoidK(), Validated.monoidK(), WriteT.monoidK(), OptionT.monoidK() ...and many more). 
* Added law tests for all the concrete instances of both  `MonoidK` and `SemigroupK` (there were a lot of semigroupK ones not added on previous PR).
* Gave "the brasa a hierro" to @pakoito and @raulraja to get it done 😉 👍


kategory/src/main/kotlin/kategory/instances/IdInstances.kt
```diff
@@ -41,3 +41,14 @@ interface IdInstances :
             fa.ev().value
 
 }
+
+object IdSemigroupK : SemigroupK<Id.F>, GlobalInstance<SemigroupK<Id.F>>() {
+    override fun <A> combineK(x: HK<Id.F, A>, y: HK<Id.F, A>): HK<Id.F, A> = y
```
raulraja: Typeclass method implementations should always return the most concrete type close to the kind. In this case: `Id<A>` and same for all other methods. This removes the need of down-casting explicitly for users when using the instance directly.

kategory/src/main/kotlin/kategory/instances/IdInstances.kt
```diff
@@ -41,3 +41,14 @@ interface IdInstances :
...
+}
+
+class IdMonoidK(val a: Any) : MonoidK<Id.F>, GlobalInstance<MonoidK<Id.F>>() {
```
pakoito: My intuition says that this should be an instance of Monoid for the wrapped type. That way you can use IntMonoid, or ListMonoid or similar.

kategory/src/main/kotlin/kategory/instances/NumberInstances.kt
```diff
@@ -39,4 +39,4 @@ object FloatMonoid : Monoid<Float>, Semigroup<Float> by SGFloat, GlobalInstance<
     override fun empty(): Float = .0f
 }
 
-private val SGFloat: Semigroup<Float> = NumberSemigroup(Float::plus)
\ No newline at end of file
+private val SGFloat: Semigroup<Float> = NumberSemigroup(Float::plus)
```
raulraja: We should revisit providing these ambiguous instances until we have a proper way to have tagged types.
These instances are just for addition and questionably not precise some of the float ones in the JVM and as such we should either let the user provide them themselves or place them under `kategory.unsafe.*` or some other package. No need to do anything in this PR but thought I'd mentioning it since this `Semigroup<Float>` would not work for multiplication for example.

kategory/src/main/kotlin/kategory/instances/OptionInstances.kt
```diff
@@ -86,7 +86,16 @@ interface OptionInstances :
...
-    }
+    override fun <A> combineK(x: HK<Option.F, A>, y: HK<Option.F, A>): Option<A> =
+            x.ev().flatMap { y.ev() }
+}
+
+/**
+ * Dummy MonoidK instance to be able to test laws for MonoidK.
+ */
+class OptionMonoidK : MonoidK<Option.F>, GlobalInstance<ApplicativeError<Option.F, Unit>>() {
+    override fun <A> combineK(x: HK<Option.F, A>, y: HK<Option.F, A>): HK<Option.F, A> =
```
raulraja: `... : Option<A>`

kategory/src/test/kotlin/kategory/data/MonoidKTests.kt
```diff
@@ -0,0 +1,26 @@
+package kategory
+
+import io.kotlintest.KTestJUnitRunner
+import org.junit.runner.RunWith
+
+@RunWith(KTestJUnitRunner::class)
+class MonoidKTests : UnitSpec() {
```
pakoito: Should this class be part of OptionT tests?

raulraja: I think this just needs to be removed since we never test typeclasses directly but through their instances

kategory/src/test/kotlin/kategory/generators/Generators.kt
```diff
@@ -8,6 +8,12 @@ inline fun <reified F, A> genApplicative(valueGen: Gen<A>, AP: Applicative<F> =
                     AP.pure(valueGen.generate())
         }
 
+inline fun <reified F, A> genEmpty(MKF: MonoidK<F>): Gen<HK<F, A>> =
```
pakoito: GenEmpty always returns the same value. Couldn't it be replaced by just using the monoid inline?

kategory/src/test/kotlin/kategory/data/MonoidKTests.kt
JorgeCastilloPrz: This has to be removed. Forgot about it. MonoidK and SemigroupK laws are tested inside all their implementations test classes (I'm already doing that). These are basically repeated tests not adding value.

kategory/src/main/kotlin/kategory/instances/IdInstances.kt
```diff
@@ -41,3 +41,14 @@ interface IdInstances :
...
+    override fun <A> empty(): HK<Id.F, A> = Id(a as A)
+
+    override fun <A> combineK(x: HK<Id.F, A>, y: HK<Id.F, A>): HK<Id.F, A> = y
```
pakoito: Same here, the Monoid value depends on the Monoid that would get passed as a parameter.

kategory/src/main/kotlin/kategory/instances/IdInstances.kt
raulraja: I'm not sure this instance is valid and I could not find a matching one in cats. What does it mean to combine 2 Id's structurally, what is the empty value of an Id?

JorgeCastilloPrz: Brief summary: I couldn't generify the Id.F type there, since the type A is defined on methods, and cannot be ready on construction. You end up in a loop where you can raise a value of A on the method, since the A is being provided there and you can't ask for an Applicative of A in the class, since the A is still not defined. If you have any idea, just let me know!

pakoito: Not without using implicits. I don't believe we'll be able to provide the instance here.

kategory/src/main/kotlin/kategory/instances/IdInstances.kt
pakoito: Same as with MonoidK, I don't believe this implementation works as expected.

kategory/src/test/kotlin/kategory/laws/MonoidKLaws.kt
```diff
@@ -0,0 +1,21 @@
...
+
+object MonoidKLaws {
+
+    inline fun <reified F> laws(SGK: MonoidK<F>, AP: Applicative<F>, EQ: Eq<HK<F, Id.F>>, EQSemigroupK: Eq<HK<F, Int>>): List<Law> =
+            SemigroupKLaws.laws(SGK, AP, EQSemigroupK) + listOf(
+                    Law("MonoidK: Left identity", { monoidKLeftIdentity(SGK, EQ) }),
+                    Law("MonoidK: Right identity", { monoidKRightIdentity(SGK, EQ) }))
+
+    inline fun <reified F> monoidKLeftIdentity(SGK: MonoidK<F>, EQ: Eq<HK<F, Id.F>>): Unit =
+            forAll(genEmpty<F, Id.F>(SGK), { a ->
```
raulraja: Instead of explicitly using `Id.F` here can we have `F` be  generic arg to `MonoidKLaws` so callers can pass whatever `F` they wish?

kategory/src/main/kotlin/kategory/instances/IdInstances.kt
raulraja: Not sure this is possible even with implicits. What is the empty value of Id as a kind for all As? The empty value of a kind should satisfy all As. For example for `Option<A> = None`.  For `Id<A> = ???`?

pakoito: yes, we had a chat about it afterwards. Id probably doesn't have an instance at all.

JorgeCastilloPrz: These 2 IdInstances (`IdSemigroupK` and `IdMonoidK`) were created to able to provide 
`WriterT.semigroupK(Id, IdSemigroupK)` and `WriterT.monoidK(Id, IdMonoidK(0))` for law tests from `WriterTTests`, and also  `StateT.semigroupK(Id, IdSemigroupK),` from `StateTTests`. 

I needed an instance of `SemigroupK<Id.F>` and `MonoidK<Id.F>` there. I agree that it does not have a very clear sense and I would completely agree on switching to a different Monad for those law tests if you can provide an Idea of which one to use.

kategory/src/main/kotlin/kategory/instances/NumberInstances.kt
JorgeCastilloPrz: Not sure why this is appearing as a change on my PR, but this instances were added by you @raulraja 😝 . Im not even using it :)

raulraja: yeah, I just moved them to another file but my point as mentioned unrelated to this PR still stands I think those instances should be removed in a separate PR until we have a proper way to tag types so we can request `Monoid<F> @@ Addition` and a package for unsafe instances that require implicit imports.

kategory/src/main/kotlin/kategory/instances/IdInstances.kt
JorgeCastilloPrz: I have removed both instances and started using OptionInstances and OptionTInstances to test the laws, since both support monoid and semigroup. It's pushed.

kategory/src/test/kotlin/kategory/data/MonoidKTests.kt
JorgeCastilloPrz: Removed.

kategory/src/test/kotlin/kategory/generators/Generators.kt
JorgeCastilloPrz: Fixed.

kategory/src/main/kotlin/kategory/instances/IdInstances.kt
JorgeCastilloPrz: This instance got removed since both Id instances added on this PR were added just for law testing purposes and they are not under use anymore, since I have switched to Option and OptionT instances which both support monoidK and semigroupK.

kategory/src/main/kotlin/kategory/instances/IdInstances.kt
JorgeCastilloPrz: Removed, explained on previous comments.

JorgeCastilloPrz: This class was removed. Not needed anymore.

kategory/src/main/kotlin/kategory/instances/OptionInstances.kt
JorgeCastilloPrz: fixed

JorgeCastilloPrz:
@pakoito @raulraja I fixed all the requested issues, can you take a look again ? Thx.

kategory/src/test/kotlin/kategory/data/WriterTTest.kt
```diff
@@ -10,6 +10,16 @@ class WriterTTest : UnitSpec() {
     init {
 
         testLaws(MonadLaws.laws(WriterT.monad(NonEmptyList, IntMonoid), Eq.any()))
+        testLaws(SemigroupKLaws.laws<WriterF<Option.F, Int>>(
```
pakoito: MonoidKLaws already include SemigroupKLaws.

kategory/src/main/kotlin/kategory/instances/OptionInstances.kt
```diff
@@ -86,7 +86,17 @@ interface OptionInstances :
...
+/**
+ * Dummy MonoidK instance to be able to test laws for MonoidK.
+ */
+class OptionMonoidK : MonoidK<Option.F>, GlobalInstance<ApplicativeError<Option.F, Unit>>() {
+    override fun <A> combineK(x: HK<Option.F, A>, y: HK<Option.F, A>): Option<A> =
+            x.ev().flatMap { y.ev() }
+
+    override fun <A> empty(): HK<Option.F, A> = Option.None
+}
+
```
pakoito: The test for this instance are missing. The ones for OptionT are there tho :D

JorgeCastilloPrz:
Done guys :+1:

PR #162, opened 2017-07-30
StackSafe monad comprehensions.

raulraja:
The following PR brings the ability to comprehend over stack unsafe monads in a stack safe way.
The underlying implementation lifts bindings to `Free` and carries on with Free values through the computations.
Given a recursive definitions such as
```kotlin
fun <F> stackSafeTestProgram(M: Monad<F>, n: Int, stopAt: Int): Free<F, Int> = M.bindingStackSafe {
  val v = M.pure(n + 1).bind()
  val r = if (v < stopAt) stackSafeTestProgram(M, v, stopAt).bind() else M.pure(v).bind()
  yields(r)
}
```
one could run this over stack unsafe monads instances such as those in Option, Try, etc... without blowing the Stack
```kotlin
stackSafeTestProgram(Option, 0, 50000).run(Option) // Option(50000)
```
Users may choose `M.binding { ...}` for regular unsafe behavior on stack unsafe M's or `M.bindingStackSafe { ...}` to achieve stack safety automatically via Free.

This instance is available for all monads. New test show the IO instance contains a stack unsafe step that should be fixed before merging this in.

Additionally this PR includes new generic test cases for all monads and some cleanup to easy running values on `TrampolineF`


pakoito:
I incorrectly committed a fix to unsafeStep() directly to master: https://github.com/kategory/kategory/commit/2ba944b59a33e73f3060ba878121cf25d8c1eb31

A branch update should fix it.

kategory/src/main/kotlin/kategory/free/Free.kt
```diff
@@ -85,3 +85,5 @@ fun <M, S, A> Free<S, A>.foldMap(f: FunctionK<S, M>, MM: Monad<M>): HK<M, A> =
 
 fun <S, A> A.free(): Free<S, A> =
         Free.pure<S, A>(this)
+
+fun <F, A> Free<F, A>.run(M: Monad<F>): HK<F, A> = this.foldMap(FunctionK.id(), M)
```
pakoito: I'd also provide a version for the HKs like the one in https://github.com/kategory/kategory/pull/157

kategory/src/main/kotlin/kategory/typeclasses/Monad.kt
```diff
@@ -72,5 +72,56 @@ fun <F, B> Monad<F>.binding(c: suspend MonadContinuation<F, *>.() -> HK<F, B>):
...
+        throw exception
+    }
+
+    internal lateinit var returnedMonad: Free<F, A>
+
+    operator suspend fun <B> HK<F, B>.not(): B = this.bind()
+
+    suspend fun <B> HK<F, B>.bind(): B = bind { Free.liftF(this) }
+
+    suspend fun <B> Free<F, B>.bind(): B = bind { this }
```
pakoito: How does name collision work for an HK instance of Free, rather than the child Free class? I'd consider making the extfun based on the HK instead.

raulraja: Unfortunately Kotlin does not allow for disambiguation via generic constrains choosing the most concrete type otherwise I could have done:
```kotlin
suspend fun <B, FK : FreeKind<F, B>> FK.bind(): B = bind { this.ev() }
```
Removing the Free version to keep only the one based on `HK` would not allow for recursive programs because recursive programs have to be typed and return `Free<F,A>` from the `binding` block as in https://github.com/kategory/kategory/blob/a2bdb99c48123bd92afef0b8eda871c1bc52348c/kategory/src/test/kotlin/kategory/laws/MonadLaws.kt#L85

Binding over `FreeKind<F,A>` would just give you back the same value you bound wrapped in Free and would fail at compile time which IMO is the best we can do now. If you have other ideas I can try something else in a separate PR to improve syntax in bindings.

PR #166, opened 2017-08-02
Iranon list wrapper

Atternatt:
WIP for the ListKW class that wraps a std-lib's List from Kotlin


.gitignore
```diff
@@ -119,3 +119,6 @@ reports/
 
 # End of https://www.gitignore.io/api/android,intellij
 /.idea/misc.xml
```
Atternatt: ops sorry i forgot to remove this file, it's just a plugin that allows pushing apk's to slack, i have to remove it

kategory/src/main/kotlin/kategory/data/ListKW.kt
```diff
@@ -0,0 +1,49 @@
+package kategory
+
+typealias ListKindW<A> = HK<ListKW.F, A>
+
+fun <A> ListKindW<A>.ev(): ListKW<A> =
+        this as ListKW<A>
+
+class ListKW<A> private constructor(private val list: List<A>) : ListKindW<A>, Collection<A> by list {
```
raulraja: nice! `Collection<A> by list`

kategory/src/main/kotlin/kategory/data/ListKW.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    class F private constructor()
+
+    fun <B> map(f: (A) -> B): ListKW<B> = ListKW(list.map(f))
+
+    fun <B> flatMap(f: (A) -> ListKW<B>): ListKW<B> = ListKW(list.flatMap { f(it).list })
+
+    operator fun plus(list: List<@UnsafeVariance A>): ListKW<A> = ListKW(this.list + list)
```
raulraja: Was this unsafe in List too? and why is it unsafe? Perhaps these can be moved to extension functions.

raulraja: We should also add other possible combinators from the kotlin list such as `groupBy`, etc..

kategory/src/main/kotlin/kategory/data/ListKW.kt
```diff
@@ -0,0 +1,49 @@
...
+
+        fun functor(): Functor<ListKW.F> = this
+
+        fun applicative(): Applicative<ListKW.F> = this
+
+        fun monad(): Monad<ListKW.F> = this
+
+        fun <A> semigroup(): Semigroup<ListKW<A>> = object : ListKWSemigroup<A> {}
+
+        fun semigroupK(): SemigroupK<ListKW.F> = object : ListKWSemigroupK {}
```
raulraja: There is also an instance for `Monoid` because you can express `empty` as an `emptyList()`

kategory/src/main/kotlin/kategory/instances/ListKWInstances.kt
```diff
@@ -0,0 +1,38 @@
+package kategory
+
+/**
+ * Created by marc on 1/8/17.
```
raulraja: :fire: remove all IDEA generated comments

kategory/src/test/kotlin/kategory/data/ListKWTest.kt
```diff
@@ -0,0 +1,36 @@
...
+import org.junit.runner.RunWith
+
+/**
+ * Created by marc on 2/8/17.
+ */
+@RunWith(KTestJUnitRunner::class)
+class ListKWTest : UnitSpec() {
+    init {
+
+        testLaws(kategory.MonadLaws.laws(kategory.NonEmptyList, kategory.Eq.any()))
```
raulraja: You should add your instances here instead of the ones from `NonEmptyList`

kategory/src/test/kotlin/kategory/data/ListKWTest.kt
```diff
@@ -0,0 +1,36 @@
...
+        testLaws(kategory.SemigroupKLaws.laws(
+                kategory.NonEmptyList.semigroupK(),
+                kategory.NonEmptyList.applicative(),
+                kategory.Eq.any()))
+
+        "map should modify values" {
+            kategory.ListKW.listOfK(14).map { it * 3 } shouldBe kategory.ListKW.listOfK(42)
+        }
+
+        "flatMap should modify entity" {
```
raulraja: these types of test are not necessary since the MonadLaws test `flatMap`, `map`, etc...

kategory/src/main/kotlin/kategory/data/ListKW.kt
Atternatt: copy paste mistake, I can remove it. Why do we need to create a extension function when we can create the function in the class? I understand extensions as is, added behaviour over a class the we can't control directly.

kategory/src/test/kotlin/kategory/data/ListKWTest.kt
Atternatt: copypast error, i need rto replace for ListKW

kategory/src/test/kotlin/kategory/data/ListKWTest.kt
Atternatt: I was following the example in NonEmptyListTest, so what kind of test should I need to implement?

kategory/src/main/kotlin/kategory/data/ListKW.kt
raulraja: That was in case the unsafe variance was related in any way to encoding Contravariant positions but this is not the case. If you want to see an example of that `Option#getOrElse`

kategory/src/test/kotlin/kategory/data/ListKWTest.kt
raulraja: I would include test for delegation that assert each one of the delegated methods produce the same result as they do in the kotlin stdLib List.

kategory/src/test/kotlin/kategory/data/ListKWTest.kt
```diff
@@ -1,24 +1,24 @@
...
-
-        testLaws(kategory.MonadLaws.laws(kategory.NonEmptyList, kategory.Eq.any()))
+        val applicative = ListKW.applicative()
+        testLaws(kategory.MonadLaws.laws(ListKW, kategory.Eq.any()))
         testLaws(kategory.SemigroupKLaws.laws(
                 kategory.ListKW.semigroupK(),
-                kategory.ListKW.applicative(),
+                applicative,
                 kategory.Eq.any()))
+        testLaws(TraverseLaws.laws(ListKW.traverse(), applicative, { applicative.pure(it) }, Eq.any()))
```
pakoito: The second parameter, `applicative` , has to be of another type for simplicity. Try Option instead.

kategory/src/main/kotlin/kategory/instances/ListKWInstances.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    override fun <A, B> map(fa: HK<ListKW.F, A>, f: (A) -> B): HK<ListKW.F, B> =
+            fa.ev().map(f)
+
+    override fun <A, B, Z> map2(fa: HK<ListKW.F, A>, fb: HK<ListKW.F, B>, f: (Tuple2<A, B>) -> Z): HK<ListKW.F, Z> =
+            fa.ev().flatMap { a -> fb.ev().map { b -> f(Tuple2(a, b)) } }
+
+    override fun <A, B> foldL(fa: HK<ListKW.F, A>, b: B, f: (B, A) -> B): B =
+            fa.ev().fold(b, f)
+
```
raulraja: This extra empty lines is what detekt is complaining about

kategory/src/main/kotlin/kategory/instances/ListKWInstances.kt
```diff
@@ -0,0 +1,76 @@
+package kategory
+
+interface ListKWInstances :
+        Functor<ListKW.F>,
+        Applicative<ListKW.F>,
+        Monad<ListKW.F>,
+        Traverse<ListKW.F> {
+
+    override fun <A> pure(a: A): ListKW<A> =
+            ListKW.listOfK(a)
```
raulraja: Detekt is asking for these to be in the same line. This is something I dislike but until we fix the detekt rules this should be :
```kotlin
override fun <A> pure(a: A): ListKW<A> = ListKW.listOfK(a)
```

kategory/src/main/kotlin/kategory/data/ListKW.kt
```diff
@@ -0,0 +1,56 @@
+package kategory
+
+typealias ListKindW<A> = HK<ListKW.F, A>
+
+fun <A> ListKindW<A>.ev(): ListKW<A> =
+        this as ListKW<A>
+
+data class ListKW<A> constructor(val list: List<A>) : ListKindW<A>, Collection<A> by list {
```
raulraja: I initially thought it was a good idea to do `Collection<A> by list` but after using this implementation I realized all the delegated combinators don't return a `ListKW` but a `List`.
so if we use `map` we get a `ListKW` but if we use `mapIndexed` we get back a `List` which is not consistent. Either we remove this delegation or we add manual delegation to all the methods in List and wrap their results in `ListKW`

Atternatt: ok, i'll try to replace as many methods as i can

raulraja:
@Atternatt Awesome job, thanks!

pakoito:
@Atternatt Awesome! Welcome to the kontributors :D

PR #168, opened 2017-08-06
Stack-safe CoYoneda

### LINKED ISSUE #167: Stack-safe coyoneda
https://github.com/typelevel/cats/pull/1602/files

pakoito:
This PR introduces the stack-safe implementation of CoYoneda that's also used in cats. Thanks to @edmundnoble for the implementation.

At higher numbers for the test (i.e. 20k), it causes a StackOverflow due to some function checks by JetBrains. See the following issue until it's fixed:  https://youtrack.jetbrains.com/issue/KT-19499

Closes #167

### DIFF
--- a/kategory/src/main/kotlin/kategory/free/CoYoneda.kt
+++ b/kategory/src/main/kotlin/kategory/free/CoYoneda.kt
@@ -7,14 +7,21 @@ typealias CoYonedaF<U, P> = HK2<CoYoneda.F, U, P>
 fun <U, A, B> CoYonedaKind<U, A, B>.ev(): CoYoneda<U, A, B> =
         this as CoYoneda<U, A, B>
 
-data class CoYoneda<FU, P, A>(val function: (P) -> A, val pivot: HK<FU, P>) : CoYonedaKind<FU, P, A> {
+data class CoYoneda<FU, P, A>(val pivot: HK<FU, P>, internal val ks: List<(Any?) -> Any?>) : CoYonedaKind<FU, P, A> {
     class F private constructor()
 
-    fun lower(FF: Functor<FU>): HK<FU, A> =
-            FF.map(pivot, function)
+    private val transform: (P) -> A = {
+        var curr: Any? = it
+        ks.forEach { curr = it(curr) }
+        curr as A
+    }
 
+    fun lower(FF: Functor<FU>): HK<FU, A> =
+            FF.map(pivot, transform)
+
+    @Suppress("UNCHECKED_CAST")
     fun <B> map(f: (A) -> B): CoYoneda<FU, P, B> =
-            CoYoneda({ f(function(it)) }, pivot)
+            CoYoneda(pivot, ks + f as (Any?) -> Any)
 
     fun toYoneda(FF: Functor<FU>): Yoneda<FU, A> =
             object : Yoneda<FU, A> {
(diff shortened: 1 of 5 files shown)


kategory/src/main/kotlin/kategory/free/CoYoneda.kt
```diff
@@ -7,23 +7,36 @@ typealias CoYonedaF<U, P> = HK2<CoYoneda.F, U, P>
 fun <U, A, B> CoYonedaKind<U, A, B>.ev(): CoYoneda<U, A, B> =
         this as CoYoneda<U, A, B>
 
-data class CoYoneda<FU, P, A>(val function: (P) -> A, val pivot: HK<FU, P>) : CoYonedaKind<FU, P, A> {
+private typealias AnyFunc = (Any?) -> Any?
+
+data class CoYoneda<FU, P, A>(val pivot: HK<FU, P>, internal val ks: List<AnyFunc>) : CoYonedaKind<FU, P, A> {
```
raulraja: Curious as to why this one uses `FU` instead of `F` as we are doing in others.

kategory/src/main/kotlin/kategory/free/CoYoneda.kt
```diff
@@ -7,23 +7,36 @@ typealias CoYonedaF<U, P> = HK2<CoYoneda.F, U, P>
...
     fun toYoneda(FF: Functor<FU>): Yoneda<FU, A> =
             object : Yoneda<FU, A> {
                 override fun <B> apply(f: (A) -> B): HK<FU, B> = map(f).lower(FF)
             }
 
     companion object {
+        @Suppress("UNCHECKED_CAST")
         inline fun <reified U, A, B> apply(fa: HK<U, A>, noinline f: (A) -> B): CoYoneda<U, A, B> =
-                CoYoneda(f, fa)
+                usafeApply(fa, listOf(f as AnyFunc))
```
raulraja: `unsafeApply`

pakoito:
I've used this PR to add law tests for CoYoneda an Yoneda.

kategory/src/test/kotlin/kategory/free/CoYonedaTest.kt
```diff
@@ -1,17 +1,39 @@
...
+    }
+
+    val AP: Applicative<CoYonedaF<Id.F, Int>> = object : Applicative<CoYonedaF<Id.F, Int>> {
+        override fun <A> pure(a: A): CoYonedaKind<Id.F, Int, A> =
+                CoYoneda.apply(Id(0), { a })
+
+        override fun <A, B> map(fa: CoYonedaKind<Id.F, Int, A>, f: (A) -> B): CoYonedaKind<Id.F, Int, B> =
+               F.map(fa, f)
+
+        override fun <A, B> ap(fa: CoYonedaKind<Id.F, Int, A>, ff: CoYonedaKind<Id.F, Int, (A) -> B>): CoYonedaKind<Id.F, Int, B> =
```
raulraja: Instead of implementing an impossible instance here we should make the generator parametric and not require `Applicative`. Instead of `genApplicative` we can require something that provides `(A) -> HK<F, A>` which to most instances it would be `Applicative#pure`.
In cases like this one it would just be `a -> CoYoneda.apply(Id(0), { a })` so there is no need to be forced to implement `Applicative#ap`.

pakoito: Okay, I'll see if I can change the Functor laws. Shouldn't take long.

raulraja: Where is structural duck typing when we need it! :duck:

PR #172, opened 2017-08-08
Update detekt to M13

pakoito:
I've updated detekt to M13:

- Add new rules for ExpressionBodySyntax.
- Add MaxLineLength to 160.
- Suppress max methods.
- Enable unsafe cast, labeled expressions to be required to be added to the baseline.
- Apply auto-fix rules for ExpressionBodySyntax and ExpressionBodySyntaxLineBreaks.

After applying those changes I've added a new baseline.


pakoito:
Please note that ExpressionBodySyntaxLineBreaks autofix doesn't respect MaxLineLength: https://github.com/arturbosch/detekt/issues/286

PR #176, opened 2017-08-09
[WIP] Add missing methods in Functor enhancement

victorg1991:
Hey, I've implemented some of the methods, how should I test them? 

I have some questions regarding the others unimplemented methods:

[this](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Functor.scala#L17) is the same as the map function, right? 🤔 

[this](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Functor.scala#L32) Can I do something like B >: A in kotlin?

[this](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Functor.scala#L65-L75) what is this? Should I implement this too?

Thanks :D

### DIFF
--- a/kategory/src/main/kotlin/kategory/typeclasses/Functor.kt
+++ b/kategory/src/main/kotlin/kategory/typeclasses/Functor.kt
@@ -13,12 +13,11 @@ interface Functor<F> : Typeclass {
 
     fun <A, B> fproduct(fa: HK<F, A>, f: (A) -> B): HK<F, Tuple2<A, B>> = map(fa, { a -> Tuple2(a, f(a)) })
 
-    fun <A, B> `as`(fa: HK<F, A>, b: B): HK<F, B> = map(fa, { _ -> b})
+    fun <A, B> `as`(fa: HK<F, A>, b: B): HK<F, B> = map(fa, { _ -> b })
 
-    fun <A, B> tupleLeft(fa: HK<F, A>, b: B): HK<F, Tuple2<B, A>> = map(fa, { a -> Tuple2(b, a)})
+    fun <A, B> tupleLeft(fa: HK<F, A>, b: B): HK<F, Tuple2<B, A>> = map(fa, { a -> Tuple2(b, a) })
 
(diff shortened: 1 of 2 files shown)


raulraja:
@victorg1991 

https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Functor.scala#L17
That is for Invariant which is the invariant functor and we do not have that in Kategory yet. Feel free to include it with laws if you want to or it will be addressed down the road

https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Functor.scala#L17
You may be able to achieve contravariant semantics as extensions functions like in some of the `getOrElse` definitions in Katyegory for Try and Option

https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Functor.scala#L65-L75
`compose` is to obtain a functor from the union of two functors. You can implement Compose and we have Compose datatypes already in Kategory but we don;t have `Contravariant`. same I mentioned for Invariant, our Functor is currently just covariant but you may add those or they will get addressed down the road since we have other priorities right now.

pakoito:
@victorg1991 @raulraja ComposedFunctor requieres an instancia in https://github.com/kategory/kategory/blob/master/kategory/src/main/kotlin/kategory/data/Composed.kt

It won't be too complicated to implement, having the other Composed as reference :D

victorg1991:
Hey! Thanks for the help :) @pakoito @raulraja 

I've added some changes, I've implemented the Composed Functor as a data class, but I've seen that there is another PR which aims to change the ComposedFoldable and ComposedTransverse current implementation to an interface, should I do the same?

I prefer to not include the invariant in his PR because I'm fairly newbie with all of these x)

I'm not sure how to write tests for this :( any help?

victorg1991:
Okey it was compiling in my machine but not with the last changes. I will fix it

raulraja:
@victorg1991 The Functor Laws we have without invariant already cover most cases. You can add simple asserts for your the new methods. Take any Functor for example `Option` and assert calls to those methods expecting the results to be what they are supposed to be based on what the methods do.

victorg1991:
Updated with tests :)

pakoito:
@victorg1991 Reaaaaaaaaalllyyy reeeaaaaallllyyy nice! And the tests too.

PR #180, opened 2017-08-10
Add Traverse law for FoldMap derived: required Const

pakoito:
Along with #179, these add the last two missing laws for traverse compared to Cats.

It required porting Const, which is a nice HK2 whose bias is towards a non-existing value Tag.

### DIFF
--- a/kategory/src/main/kotlin/kategory/data/Const.kt
+++ b/kategory/src/main/kotlin/kategory/data/Const.kt
@@ -1,7 +1,6 @@
 package kategory
 
-fun <A, T> ConstKind<A, T>.value(): A =
-        this.ev().value
+fun <A, T> ConstKind<A, T>.value(): A = this.ev().value
 
 typealias ConstF<A> = HK<ConstHK, A>
 
@@ -10,8 +9,7 @@ typealias ConstF<A> = HK<ConstHK, A>
     @Suppress("UNCHECKED_CAST")
     fun <U> retag(): Const<A, U> = this as Const<A, U>
 
-    inline fun <F, U> traverse(f: (T) -> HK<F, U>, FA: Applicative<F>): HK<F, Const<A, U>> =
-            FA.pure(retag())
+    inline fun <F, U> traverse(f: (T) -> HK<F, U>, FA: Applicative<F>): HK<F, Const<A, U>> = FA.pure(retag())
 
     companion object {
         fun <T, A> pure(a: A): Const<A, T> = Const(a)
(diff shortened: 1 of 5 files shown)


kategory/src/test/kotlin/kategory/laws/TraverseLaws.kt
```diff
@@ -17,14 +17,14 @@ class TIF {
 }
 
 object TraverseLaws {
+    // FIXME(paco): cf is only required because AP::pure will crash the inliner
```
raulraja: Is this related to being a method reference lifted to a function and not an actual function?

kategory/src/main/kotlin/kategory/data/Const.kt
```diff
@@ -0,0 +1,39 @@
+package kategory
+
+fun <A, T> ConstKind<A, T>.value(): A = this.ev().value
```
raulraja: I wonder if we can use Kotlin's delegation facilities to delegate all methods automatically in the codegen phase so `ConstKind` receives everything declared in `Const` and same for all other @higherkind data types.

kategory/src/main/kotlin/kategory/data/Const.kt
```diff
@@ -0,0 +1,39 @@
...
+
+        fun <A> instances(MA: Monoid<A>): ConstInstances<A> = object : ConstInstances<A> {
+            override fun MA(): Monoid<A> = MA
+        }
+
+        inline fun <reified A> applicative(MA: Monoid<A> = kategory.monoid<A>()): Applicative<ConstF<A>> = instances(MA)
+
+        inline fun <reified A> traverse(MA: Monoid<A> = kategory.monoid<A>()): Traverse<ConstF<A>> = instances(MA)
+
+        fun <A, T> semigroup(MA: Monoid<A>): Semigroup<ConstKind<A, T>> = object : ConstMonoid<A, T> {
```
raulraja: `object : ConstMonoid<A, T> {...}` Seems like a common operation, perhaps we should provide a class representation for it to avoid users having to implement abstract members.

kategory/src/test/kotlin/kategory/laws/TraverseLaws.kt
```diff
@@ -17,14 +17,14 @@ class TIF {
...
     inline fun <reified F> laws(FF: Traverse<F> = traverse<F>(), AP: Applicative<F> = applicative<F>(), crossinline cf: (Int) -> HK<F, Int>, EQ: Eq<HK<F, Int>>): List<Law> =
             FoldableLaws.laws(FF, cf, Eq.any()) + FunctorLaws.laws(AP, EQ) + listOf(
                     Law("Traverse Laws: Identity", { identityTraverse(FF, AP, cf, EQ) }),
                     // TODO (#136)
                     // Law("Traverse Laws: Sequential composition", { sequentialComposition(FF, AP, cf, EQ) })
-                    Law("Traverse Laws: Parallel composition", { parallelComposition(FF, cf, EQ) })
-                    // TODO (#136)
-                    // Law("Traverse Laws: FoldMap derived", { foldMapDerived(FF, AP, cf, EQ) })
```
raulraja: Are these comments still necessary?

pakoito: They go away once everything gets merged together.

kategory/src/test/kotlin/kategory/laws/TraverseLaws.kt
pakoito: No there's a bug even if you use a lambda. It fails to find and capture the context of the lambda O.O

raulraja: :(

kategory/src/main/kotlin/kategory/data/Const.kt
pakoito: companion object constructor with invoke is what we normally use. Can be written.

kategory/src/main/kotlin/kategory/data/Const.kt
pakoito: We decided not to do it because it required creating lots of code overhead that would bloat the library.

raulraja: If it's doable we can have this as an optional arg for the best of both worlds in user defined datatypes.
```kotlin
@higherkind(delegation = DelegationStrategy.ALL)
```

PR #182, opened 2017-08-10
Add analyze to FreeApplicative

### LINKED ISSUE #136: Const and Nested combined types + missing Traverse laws
They're necessary to complete the Traverse laws after #133

Nested: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/Nested.scala

And the traverse law that requires Nested: https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/TraverseLaws.scala#L16

https://github.com/kategory/kategory/pull/133/files#diff-8ad20c0b396ad4072dcdf3c16dafafaaR23

--

Const: https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/Const.scala

Add traverse law that requires Const:
https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/TraverseLaws.scala#L61-L68

https://github.com/kategory/kategory/pull/133/files#diff-8ad20c0b396ad4072dcdf3c16dafafaaR26

pakoito:
Now that Const is in the repo via #180, the missing method in FreeApplicative can be added.

Please note that it requires to have merged #179 and #180 beforehand.

Closes #136

### DIFF
--- a/kategory/src/main/kotlin/kategory/free/FreeApplicative.kt
+++ b/kategory/src/main/kotlin/kategory/free/FreeApplicative.kt
@@ -56,8 +56,10 @@ inline fun <reified F, A> FreeApplicativeKind<F, A>.foldK(FA: Applicative<F> = a
 
     fun <G> flatCompile(f: FunctionK<F, FreeApplicativeF<G>>, GFA: Applicative<FreeApplicativeF<G>>): FreeApplicative<G, A> = foldMap(f, GFA).ev()
 
-    // TODO(paco): requires Const
-    // final def analyze[M: Monoid](f: FunctionK[F, λ[α => M]]): M
+    fun <M> analyze(f: FunctionK<F, ConstF<M>>, MM: Monoid<M>): M =
+            foldMap(object : FunctionK<F, ConstF<M>> {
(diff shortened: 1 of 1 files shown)


kategory/src/main/kotlin/kategory/data/Const.kt
```diff
@@ -0,0 +1,39 @@
...
+
+        fun <A, T> monoid(MA: Monoid<A>): Monoid<ConstKind<A, T>> = object : ConstMonoid<A, T> {
+            override fun MA(): Monoid<A> = MA
+        }
+    }
+}
+
+fun <A, T> ConstKind<A, T>.combine(that: ConstKind<A, T>, SG: Semigroup<A>): Const<A, T> = Const(SG.combine(this.value(), that.value()))
+
+fun <A, T, U> ConstKind<A, T>.ap(ff: ConstKind<A, (T) -> U>, SG: Semigroup<A>): Const<A, U> = ff.ev().retag<U>().combine(this.ev().retag(), SG)
```
raulraja: We gotta go back to impl in the next line, I keep scrolling right on GH and everywhere to be able to read the impls.

PR #189, opened 2017-08-13
Modularize Kategory into core, test, effects

pakoito:
This is the first step to split the project into logical units to avoid unnecessary imports, and bloating on Android.

So far the split contains the following modules:

* kategory: The core classes
* kategory-test: Laws, UnitTest scaffolding, generators, and utils like SideEffect
* kategory-effects: IO, Future, Effect, and other monads for side-effects. This will be the base for any integration modules like AsyncTask, Rx, or Kovenant


raulraja:
What is the proposed package and imports strategy for modules like `effects`?
Would user do `import kategory.effects.*` or just `import kategory.*`?

pakoito:
@raulraja I'm for keeping them the same, to have access to extfun for integrations without doing an import dance, nor imposing our style on 3rd parties

PR #191, opened 2017-08-13
`kategory-docs` module and proposed layout including `Functor` and `Applicative`

raulraja:
Let's agree on an standard format for the datatypes and typeclasses section so we can write the rest.
See the included `Functor` example.


kategory-docs/typeclasses/applicative/README.md
```diff
@@ -0,0 +1,138 @@
...
+Kategory provides [`ApplicativeLaws`](/docs/typeclasses/laws#applicativelaws) for internal verification of lawful instances and third party apps creating their own Applicative instances.
+
+#### Creating your own `Applicative` instances
+
+[Kategory already provides Applicative instances for most common datatypes](#datatypes) both in Kategory and the Kotlin stdlib. 
+See [Deriving and creating custom typeclass] to provide your own Applicative instances for custom datatypes.
+
+### Data types
+
+Thw following datatypes in Kategory provide instances that adhere to the `Applicative` typeclass.
```
ffgiraldez: typo `The`

wiyarmir:
Looks like detekt is not very good friends with gradle API? 

`> Could not get unknown property 'kotlinxCollectionsImmutableVersion' for object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.`

Does it run correctly in your machine @raulraja ? Mine still has Windows only issues with detekt

raulraja:
@wiyarmir never seen that error before.

pakoito:
@raulraja @wiyarmir That one is my fault. We weren't using the lib so I removed it from all dependencies. Maybe it carried over while copying the new module.

kategory-docs/typeclasses/applicative/README.md
```diff
@@ -0,0 +1,138 @@
...
+Examples of type constructors that can implement instances of the Applicative typeclass include `Option`, `NonEmptyList`,
+`List` and many other datatypes that include a `pure` and either `ap` function. `ap` may be derived for monadic types that include a `Monad` instance via `flatMap`.
+
+### Applicative Builder examples
+
+Often times we find ourselves in situations where we need to compute multiple independent values resulting from operations that do not depend on each other.
+
+In the following example we will define 3 invocations that may as well be remote or local services each one of them returning different results in the same computational context of `Option`
+
+```kotlin:ank
```
JorgeCastilloPrz: so is `ank` parsing this snippets and compiling them to fail CI if docs snippets are not compiling for any reason ?

raulraja: yes ank does all of that , the generated docs will be in markdown format which jekyll's support so you can push them to ghpages or anywhere else.

kategory-docs/docs/typeclasses/applicative/README.md
```diff
@@ -0,0 +1,140 @@
+---
+layout: docs
+title: Applicative
+permalink: /docs/typeclasses/applicative/
+---
+
+## Functor
```
pakoito: This header has to be Applicative.

pakoito:
I believe we should add, very close to the top, which other typeclasses do each inherit from. It's hard to miss that Applicative inherits from functor, so it gets map by default.

kategory-docs/docs/typeclasses/applicative/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+```kotlin:ank
+Option.pure(1) // Option(1)
+```
+
+#### ap
+
+Apply a function inside the type constructor's context
+
+`fun <A, B> ap(fa: HK<F, A>, ff: HK<F, (A) -> B>): HK<F, B>`
```
pakoito: This is the most generic example of the API. We should display some implementation examples below, i.e. the same for Option or Either. Not the implementation, just the signature with the generics pre-applied.

kategory-docs/docs/typeclasses/applicative/README.md
```diff
@@ -0,0 +1,140 @@
...
+Computation happens when `.value()` is invoked.
+
+```kotlin:ank
+Option(1).map2(Eval { Option("x") }, { z: Tuple2<Int, String> ->  z.a + z.b }).value() // Option("1x")
+```
+
+
+### Laws
+
+Kategory provides [`ApplicativeLaws`](/docs/typeclasses/laws#applicativelaws) for internal verification of lawful instances and third party apps creating their own Applicative instances.
```
pakoito: We should flesh out the explanation for laws, or at least use the term test or test suite to refer to them, as that relates more to coding concepts.

kategory-docs/docs/typeclasses/applicative/README.md
raulraja: There are examples below of usage. Adding a concrete form using for example `OptionKind` it's going to generate more confusion since we will have to explain here what kinds are. This can be better addressed in the derivation section or a section where we explain how we use higher kinds and how we have emulated them.

PR #199, opened 2017-08-14
Add support for CoroutineContext to execute binding in separate threads

### LINKED ISSUE #197: Add `bind` combinator for monadic comprehensions
`bind(coroutineContext)` is a special binding combinators that ensure bindings happen in the proper coroutine context since some framework such as Realm or Rx require effects happening in certain threads. This combinator can delegate to `async/await` internally passing through the coroutine context and invoking `resume` on completion.

Down the road integrations for RX, Realm etc can provide specializations and CorotuineContext instances specific to those frameworks.

pakoito:
This PR adds support for passing a CoroutineContext to both `binding` and `bindInContext`, which would then execute the block in that specific Context.

I've added a law to test it, along with some helpers to run resources-bound tests.  

Closes #197

### DIFF
--- a/kategory-test/src/main/kotlin/kategory/concurrency/CoroutineDispatchers.kt
+++ b/kategory-test/src/main/kotlin/kategory/concurrency/CoroutineDispatchers.kt
@@ -1,0 +1,15 @@
+package kategory
+
+import kotlinx.coroutines.experimental.CoroutineDispatcher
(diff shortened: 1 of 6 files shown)


pakoito:
Closes #197

raulraja:
👏 👏 👏

Issue #194, opened 2017-08-13
Remove all instances of `! (not)` in monadic binds


raulraja:
Take care of #39 leaving only `monad.bind()` and `bind { monad }`

PR #201, opened 2017-08-14
Remove instances of ! from tests

npatarino:
As it says in #194 the "!" mark it's already removed.

### DIFF
--- a/kategory-effects/src/test/kotlin/kategory/effects/data/IOTest.kt
+++ b/kategory-effects/src/test/kotlin/kategory/effects/data/IOTest.kt
@@ -209,11 +209,10 @@ class IOTest : UnitSpec() {
         "IO.binding should for comprehend over IO" {
             val result = IO.binding {
                 val x = IO.pure(1).bind()
-                val y = !IO { x + 1 }
-                val z = bind { IO { y + 1 } }
-                yields(z)
+                val y = bind { IO { x + 1 } }
+                yields(y)
             }.ev()
-            result.unsafeRunSync() shouldBe 3
+            result.unsafeRunSync() shouldBe 2
         }
     }
 }
--- a/kategory/src/test/kotlin/kategory/data/NonEmptyListTest.kt
+++ b/kategory/src/test/kotlin/kategory/data/NonEmptyListTest.kt
@@ -39,36 +39,32 @@ class NonEmptyListTest : UnitSpec() {
 
         "NonEmptyListMonad.binding should for comprehend over NonEmptyList" {
             val result = NonEmptyList.binding {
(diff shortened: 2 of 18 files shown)


raulraja:
We need to also remove the `not()` method declarations from all MonadContinuations coroutines.

npatarino:
You are right. I'll do it. Thanks

npatarino:
I think I've removed all the instances. it's OK or I've missed some?

kategory/src/test/kotlin/kategory/data/NonEmptyListTest.kt
```diff
@@ -39,71 +39,63 @@ class NonEmptyListTest : UnitSpec() {
...
             forAll { a: Int, b: List<Int> ->
                 val nel: NonEmptyList<Int> = NonEmptyList(a, b)
-                val nel2 = NonEmptyList.of(1, 2)
-                val nel3 = NonEmptyList.of(3, 4, 5)
+                val nel2 = NonEmptyList.of(1, 2, 3)
                 val result: Int = NonEmptyList.cobinding {
-                    val x = !nel
-                    val y = nel2.extract()
-                    val z = extract { nel3 }
-                    x + y + z
```
raulraja: Why is `z` gone here?

kategory-effects/src/test/kotlin/kategory/effects/data/IOTest.kt
```diff
@@ -209,11 +209,10 @@ class IOTest : UnitSpec() {
         "IO.binding should for comprehend over IO" {
             val result = IO.binding {
                 val x = IO.pure(1).bind()
-                val y = !IO { x + 1 }
-                val z = bind { IO { y + 1 } }
-                yields(z)
+                val y = bind { IO { x + 1 } }
+                yields(y)
```
raulraja: why is `z` gone here?

kategory/src/test/kotlin/kategory/data/NonEmptyListTest.kt
npatarino: I've removed `nel3` just to not repeat the `.extract()` in x.
I've thought it has 3 `NonEmptyList` because there was 3 different ways to extract, `!`, `.extract()` and `extract { ... }`

kategory-effects/src/test/kotlin/kategory/effects/data/IOTest.kt
npatarino: This is because the same reason of https://github.com/kategory/kategory/pull/201#discussion_r132927084
Should I change it?

PR #208, opened 2017-08-15
Remove `ListKW` boilerplate

raulraja:
Removes `ListKW` method delegation boilerplate using Kotlin's class delegation machinery.
This technique can be used for other datatypes in the std lib to avoid manually implementing all combinators.
`@higherkind data class ListKW<A> constructor(val list: List<A>) : ListKWKind<A>, List<A> by list { .. }`
This means that `ListKW<A>` is now a `List<A>` and inherits all public methods as well making them visible. Whenever a concrete `ListKW<A>` is required one can just invoke `list.k()` to rewrap the instance.

This technique may be used when implementing `Set` wrappers and others.

### DIFF
--- a/kategory/src/main/kotlin/kategory/data/ListKW.kt
+++ b/kategory/src/main/kotlin/kategory/data/ListKW.kt
@@ -1,27 +1,8 @@
 package kategory
 
-@higherkind data class ListKW<A> constructor(val list: List<A>) : ListKWKind<A> {
-
-    fun <B> map(f: (A) -> B): ListKW<B> = ListKW(list.map(f))
-
-    fun <B> flatMap(f: (A) -> ListKW<B>): ListKW<B> = ListKW(list.flatMap { f(it).list })
-
-    operator fun plus(list: List<A>): ListKW<A> = ListKW(this.list + list)
-
-    operator fun plus(listKW: ListKW<A>): ListKW<A> = ListKW(this.list + listKW.list)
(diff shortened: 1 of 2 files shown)


kategory/src/main/kotlin/kategory/data/ListKW.kt
```diff
@@ -1,28 +1,9 @@
...
-    operator fun plus(listKW: ListKW<A>): ListKW<A> = ListKW(this.list + listKW.list)
-
-    operator fun get(position: Int): Option<A> = if (list.isEmpty() || position < 0 || position > list.size) Option.None else Option.Some(list[position])
-
-    fun <B> fold(b: B, f: (B, A) -> B): B = list.fold(b, f)
-
-    fun drop(n: Int): ListKW<A> = ListKW(this.list.drop(n))
-
-    fun first(): A = this.list.first()
+@higherkind data class ListKW<A> constructor(val list: List<A>) : ListKWKind<A>, List<A> by list {
```
Atternatt: but then we'll need to call .k() for uncontrolled methods like .mapIndexed() as you told on the previouscomment right? it's fin for you guys?

raulraja: Yes, until we have full wrapper generation with the annotation processor yin code gen phase. 
Since List is already Iterable all of those functions should be already available.

PR #210, opened 2017-08-17
MonadWriter

### LINKED ISSUE #163: MonadWriter
Add MonadWriter.

JorgeCastilloPrz:
Fixes #163

**Disclaimer:** `MonadWriter` was moved to `FunctorListen` inside `cats-mtl` module. The reason to keep implementing `MonadWriter` here is that I think we should definitely keep "old" constructs (the one valids until cats `pre-1.0.0`) for the time being, until we have a first release. I would iterate towards `kategory-mtl` module with the same renaming / refactor iterations applied to cats probably after release, to avoid more noise.

### DIFF
--- a/kategory/src/main/kotlin/kategory/instances/WriterTInstances.kt
+++ b/kategory/src/main/kotlin/kategory/instances/WriterTInstances.kt
@@ -27,6 +27,21 @@ interface WriterTInstances<F, W> :
 
 }
 
+interface WriterTMonadWriter<F, W> : MonadWriter<WriterTKindPartial<F, W>, W> {
+
+    fun MF(): Monad<F>
+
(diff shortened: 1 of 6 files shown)


kategory/src/main/kotlin/kategory/instances/WriterTInstances.kt
```diff
@@ -27,6 +27,8 @@ interface WriterTInstances<F, W> :
 
 }
 
+interface WriterTMonadWriter<F, W> : MonadWriter<WriterTKindPartial<F, W>, W>
```
raulraja: Since this does not imposes new constrains I think it's fine to implement inline to be consistent how Functor, Applicative and Monad are implemented and there would be no need for the extra interface `WriterTMonadWriter`

JorgeCastilloPrz: Fixed 👍

PR #214, opened 2017-08-19
Comonad laws

pakoito:
This diff adds laws for Comonad, along with tests for all instances.

Turns out Cofree was correctly implemented after all. Huh.

### DIFF
--- a/kategory-test/src/main/kotlin/kategory/laws/TraverseLaws.kt
+++ b/kategory-test/src/main/kotlin/kategory/laws/TraverseLaws.kt
@@ -17,29 +17,39 @@ class TIF {
 }
 
 object TraverseLaws {
-    // FIXME(paco): cf is only required because AP::pure will crash the inliner
-    inline fun <reified F> laws(FF: Traverse<F> = traverse<F>(), AP: Applicative<F> = applicative<F>(), crossinline cf: (Int) -> HK<F, Int>, EQ: Eq<HK<F, Int>>): List<Law> =
-            FoldableLaws.laws(FF, cf, Eq.any()) + FunctorLaws.laws(AP, EQ) + listOf(
-                    Law("Traverse Laws: Identity", { identityTraverse(FF, AP, cf, EQ) }),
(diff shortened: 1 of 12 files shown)


kategory-test/src/main/kotlin/kategory/laws/TraverseLaws.kt
```diff
@@ -17,29 +17,39 @@ class TIF {
...
+    // FIXME(paco): this implementation will crash the inliner
+    /*
+    inline fun <reified F> laws(TF: Traverse<F> = traverse<F>(), AF: Applicative<F> = applicative<F>(), EQ: Eq<HK<F, Int>>): List<Law> =
+        FoldableLaws.laws(TF, { AF.pure(it) }, Eq.any()) + FunctorLaws.laws(AF, EQ) + listOf(
+                Law("Traverse Laws: Identity", { identityTraverse(TF, AF, { AF.pure(it) }, EQ) }),
+                Law("Traverse Laws: Sequential composition", { sequentialComposition(TF, { AF.pure(it) }, EQ) }),
+                Law("Traverse Laws: Parallel composition", { parallelComposition(TF, { AF.pure(it) }, EQ) }),
+                Law("Traverse Laws: FoldMap derived", { foldMapDerived(TF, { AF.pure(it) }) })
+        )
+    */
```
JorgeCastilloPrz: Shouldn't this comment get removed ?

pakoito: No, it has to be re-enabled after the issue in the fixme is fixed

JorgeCastilloPrz: So, it's gonna be fixed prior to merging this or in a different PR ?

pakoito: It's an issue on the Kotlin compiler, I may beed to add a reference to it: KT-18660

PR #215, opened 2017-08-20
Reducible

### LINKED ISSUE #211: Add Reducible
Add Reducible typeclass, laws and tests for it.

JorgeCastilloPrz:
Fixes #211

* Added `Reducible` and `NonEmptyReducible` instance.
* Added `Reducible` laws.
* Added tests for `Reducible` and `ReducibleLaws`.

### DIFF
--- a/detekt.yml
+++ b/detekt.yml
@@ -84,7 +84,7 @@ formatting:
     active: true
     autoCorrect: true
   ExpressionBodySyntaxLineBreaks:
-      active: true
+      active: false
       autoCorrect: true
 
 style:
--- a/detekt.yml
+++ b/detekt.yml
@@ -84,7 +84,7 @@ formatting:
     active: true
     autoCorrect: true
   ExpressionBodySyntaxLineBreaks:
-      active: false
+      active: true
       autoCorrect: true
 
 style:
--- a/kategory-test/src/main/kotlin/kategory/generators/Generators.kt
+++ b/kategory-test/src/main/kotlin/kategory/generators/Generators.kt
@@ -24,18 +24,6 @@ fun <A> genFunctionAAToA(genA: Gen<A>): Gen<(A, A) -> A> =
             }
         }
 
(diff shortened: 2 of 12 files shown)


detekt.yml
```diff
@@ -84,7 +84,7 @@ formatting:
     active: true
     autoCorrect: true
   ExpressionBodySyntaxLineBreaks:
-      active: true
+      active: false
```
raulraja: :tada:

pakoito: I'm the waterparties here. The detek baseline will need to be regenerated and the long lines fixed if this is not enforced anymore :P

JorgeCastilloPrz: Could enable it again, but detekt was not allowing me to compile in any way (tried many workarounds) because of format requirement collisions on 2 different rules (I think it was this one and MaxLineLength). I didn't have any other choice than disabling this.

pakoito: ExpressionBodySyntaxLineBreaks is an auto-fix that's bugged. Instead of applying to every instance, it only applies to the first one per file. You have to run it multiple times to make it work.

Once those errors are gone, you'll get a lot of MaxLineLength ones. For that you just regenerate the baseline with `./gradlew detektEstablishAcceptedErrors` and push

pakoito: The train of thought is that autofix > MaxLineLength + regenerate baseline for consistency purposes. It's quite annoying for now, so reverting would mean fixing all the MaxLineLength back. No more than 30 minutes to do it tho.

JorgeCastilloPrz: Ok. I'll do what we agreed on in slack for this PR:
* Auto-fixed ExpressionBodySyntaxLineBreaks violations with `./gradlew detekt` (running it multiple times).
* When I just had a few `MaxLineLength` violations I regenerated baseline with `./gradlew detektEstablishAcceptedErrors`.

PR #216, opened 2017-08-20
[WIP] Add compatibility layer for KindedJ

pakoito:
This PR replaces https://github.com/kategory/kategory-annotations/pull/19 with https://github.com/kategory/kategory-annotations/pull/21 as a way to provide a compatibility layer between KindedJ higher kinds and our covariant higher kind.

For invariant datatypes a KindJ is generated.

For covariant and contravariant data types it add a Convert class that has operations to create variance-safe wrapper classes for the transformation. These wrapper classes have their own witness HK<ConverHK, F> and provide methods to give a view for either interface: KindedJ or Kategory. Examples are provided as tests.

Note that this is a WIP:
* Uses a commit release from jitpack until everything's ready
* Names are also up for discussion
* Tests are incomplete


kategory/build.gradle
```diff
@@ -5,14 +5,16 @@ test {
...
 dependencies {
+    compile 'com.github.KindedJ:KindedJ:5bee1acf6f'
     compile "org.jetbrains.kotlin:kotlin-stdlib-jre7:$kotlinVersion"
     compile "org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinxCoroutinesVersion"
-    compile "io.kategory:kategory-annotations:$kategoryAnnotationsVersion"
-    kapt "io.kategory:kategory-annotations-compile:$kategoryAnnotationsVersion"
-    kaptTest "io.kategory:kategory-annotations-compile:$kategoryAnnotationsVersion"
-    compileOnly "io.kategory:kategory-annotations-compile:$kategoryAnnotationsVersion"
-    testCompileOnly "io.kategory:kategory-annotations-compile:$kategoryAnnotationsVersion"
+    compile 'com.github.kategory:kategory-annotations:paco-kindedjgen-SNAPSHOT'
```
JorgeCastilloPrz: is this correct ? weren't we moving forward to `kategory.io` as our domain? do we have snapshots deployed under a different groupid ?

kategory/src/main/java/kategory/Convert.java
```diff
@@ -0,0 +1,55 @@
+package kategory;
+
+import org.jetbrains.annotations.NotNull;
+
+public class Convert {
+    @NotNull
+    public static <F, A> FromKindedJToKategory<F, A> fromKindedJ(@NotNull io.kindedj.HK<F, A> hk) {
```
JorgeCastilloPrz: what's the point on creating a standard for HKs if you end up needing to translate from and to it ? Should the standard itself be the annotation processor capable of generating the HK abstractions over the types as we are doing now?

pakoito:
Moved to #503

PR #221, opened 2017-08-21
Added traverse and foldable instance for NonEmptyList

### LINKED ISSUE #202: Add missing traverse instance for NonEmptyList

tonilopezmr:
Fixes  #202

Added `Foldable` and `Traverse` for NonEmptyList.
Added tests for `TraverseLaws`.

### DIFF
--- a/kategory/src/main/kotlin/kategory/data/NonEmptyList.kt
+++ b/kategory/src/main/kotlin/kategory/data/NonEmptyList.kt
@@ -67,4 +67,4 @@ package kategory
     }
 }
 
-fun <A> A.k(): NonEmptyList<A> = NonEmptyList.of(this)
+fun <A> A.nel(): NonEmptyList<A> = NonEmptyList.of(this)
--- a/kategory/src/main/kotlin/kategory/data/NonEmptyList.kt
+++ b/kategory/src/main/kotlin/kategory/data/NonEmptyList.kt
@@ -67,4 +67,4 @@ package kategory
     }
 }
 
-fun <A> A.nel(): NonEmptyList<A> = NonEmptyList.of(this)
+fun <A> A.k(): NonEmptyList<A> = NonEmptyList.of(this)
--- a/kategory/src/main/kotlin/kategory/data/NonEmptyList.kt
+++ b/kategory/src/main/kotlin/kategory/data/NonEmptyList.kt
@@ -60,6 +60,10 @@ package kategory
         fun <A> semigroup(): Semigroup<NonEmptyList<A>> = object : NonEmptyListSemigroup<A> {}
 
         fun semigroupK(): SemigroupK<NonEmptyListHK> = object : NonEmptyListSemigroupK {}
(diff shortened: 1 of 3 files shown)


kategory/src/main/kotlin/kategory/data/NonEmptyList.kt
```diff
@@ -60,7 +60,11 @@ package kategory
...
         fun semigroupK(): SemigroupK<NonEmptyListHK> = object : NonEmptyListSemigroupK {}
+
+        fun foldable(): Foldable<NonEmptyListHK> = this
+
+        fun traverse(): Traverse<NonEmptyListHK> = this
     }
 }
 
-fun <A> A.nel(): NonEmptyList<A> = NonEmptyList.of(this)
+fun <A> A.k(): NonEmptyList<A> = NonEmptyList.of(this)
```
pakoito: I'd personally revert this change. k() is reserved for kind wrappers for list or function. In this case you're creating a nel() from a, the same way we have left() and right() for Either, success() for Try or valid() for Validated.

kategory/src/main/kotlin/kategory/instances/NonEmptyListInstances.kt
```diff
@@ -32,6 +34,23 @@ interface NonEmptyListInstances :
         }
     }
 
+    override fun <A, B> foldL(fa: HK<NonEmptyListHK, A>, b: B, f: (B, A) -> B): B = fa.ev().tail.fold(f(b, fa.ev().head), f)
+
+    override fun <A, B> foldR(fa: HK<NonEmptyListHK, A>, lb: Eval<B>, f: (A, Eval<B>) -> Eval<B>): Eval<B> {
+        fun loop(fa_p: NonEmptyList<A>): Eval<B> = when {
```
pakoito: Is loop stack-safe here? I know in the past we've found differences in implementation from cats in this regard. I see ListKW has the same implementation, maybe it needs a law.

kategory/src/main/kotlin/kategory/data/NonEmptyList.kt
tonilopezmr: Okey I'm going to revert 👍

kategory/src/main/kotlin/kategory/instances/NonEmptyListInstances.kt
tonilopezmr: Yep, I have refactored `NonEmptyList.foldR` to use `ListKW.foldR`

pakoito: Yes, I mean, is ListKW.foldR stack-safe :D That's beyond the scope of this ticket. I'll approve as it is, we can investigate later.

tonilopezmr: 💯

PR #224, opened 2017-08-22
Add FunctorFilter

JorgeCastilloPrz:
Fixes: #223 

* Added `FunctorFilter`
* Added `FunctorFilterLaws`
* Added `ComposedFunctorFilter` and `OptionTFunctor` `FunctorFilter` instances.
* Tested laws using `OptionTFunctor`.

### DIFF
--- a/kategory-test/src/main/kotlin/kategory/generators/Generators.kt
+++ b/kategory-test/src/main/kotlin/kategory/generators/Generators.kt
@@ -55,3 +55,8 @@ fun genIntPredicate(): Gen<(Int) -> Boolean> =
                     { it % absNum == absNum - 1 })
             )
         }
+
+fun <B> genOption(genB: Gen<B>): Gen<Option<B>> =
+        object : Gen<Option<B>> {
+            override fun generate(): Option<B> = Option.pure(genB.generate())
+        }
--- a/kategory-test/src/main/kotlin/kategory/generators/Generators.kt
+++ b/kategory-test/src/main/kotlin/kategory/generators/Generators.kt
@@ -58,8 +58,8 @@ fun genIntPredicate(): Gen<(Int) -> Boolean> =
 
 fun <B> genOption(genB: Gen<B>): Gen<Option<B>> =
         object : Gen<Option<B>> {
+            val random = genIntSmall()
             override fun generate(): Option<B> {
-                val random = genIntSmall().generate()
-                return if (random % 20 == 0) Option.None else Option.pure(genB.generate())
+                return if (random.generate() % 20 == 0) Option.None else Option.pure(genB.generate())
             }
         }
--- a/kategory-test/src/main/kotlin/kategory/generators/Generators.kt
+++ b/kategory-test/src/main/kotlin/kategory/generators/Generators.kt
@@ -58,5 +58,8 @@ fun genIntPredicate(): Gen<(Int) -> Boolean> =
 
 fun <B> genOption(genB: Gen<B>): Gen<Option<B>> =
         object : Gen<Option<B>> {
-            override fun generate(): Option<B> = Option.pure(genB.generate())
+            override fun generate(): Option<B> {
+                val random = genIntSmall().generate()
+                return if (random % 20 == 0) Option.None else Option.pure(genB.generate())
+            }
         }
--- a/kategory-test/src/main/kotlin/kategory/laws/FunctorFilterLaws.kt
+++ b/kategory-test/src/main/kotlin/kategory/laws/FunctorFilterLaws.kt
@@ -6,10 +6,10 @@ import kategory.*
 
 object FunctorFilterLaws {
 
-    inline fun <reified F> laws(AP: Applicative<F> = applicative(), FFF: FunctorFilter<F> = functorFilter(), EQ: Eq<HK<F, Int>>): List<Law> =
-            FunctorLaws.laws(AP, EQ) + listOf(
-                    Law("Functor Filter: mapFilter composition", { mapFilterComposition(FFF, AP::pure, EQ) }),
-                    Law("Functor Filter: mapFilter map consistency", { mapFilterMapConsistency(FFF, AP::pure, EQ) })
+    inline fun <reified F> laws(FFF: FunctorFilter<F> = functorFilter(), crossinline cf: (Int) -> HK<F, Int>, EQ: Eq<HK<F, Int>>): List<Law> =
+            FunctorLaws.laws(FFF, cf, EQ) + listOf(
+                    Law("Functor Filter: mapFilter composition", { mapFilterComposition(FFF, cf, EQ) }),
+                    Law("Functor Filter: mapFilter map consistency", { mapFilterMapConsistency(FFF, cf, EQ) })
             )
 
     inline fun <reified F> mapFilterComposition(FFF: FunctorFilter<F> = functorFilter(), crossinline ff: (Int) -> HK<F, Int>, EQ: Eq<HK<F, Int>> = Eq.any()): Unit =
--- a/kategory-test/src/main/kotlin/kategory/laws/FunctorFilterLaws.kt
+++ b/kategory-test/src/main/kotlin/kategory/laws/FunctorFilterLaws.kt
@@ -1,0 +1,31 @@
+package kategory.laws
+
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+import kategory.*
+
+object FunctorFilterLaws {
+
+    inline fun <reified F> laws(AP: Applicative<F> = applicative(), FFF: FunctorFilter<F> = functorFilter(), EQ: Eq<HK<F, Int>>): List<Law> =
+            FunctorLaws.laws(AP, EQ) + listOf(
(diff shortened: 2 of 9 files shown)


kategory-test/src/main/kotlin/kategory/generators/Generators.kt
```diff
@@ -55,3 +55,8 @@ fun genIntPredicate(): Gen<(Int) -> Boolean> =
                     { it % absNum == absNum - 1 })
             )
         }
+
+fun <B> genOption(genB: Gen<B>): Gen<Option<B>> =
```
pakoito: This method could be implemented using genApplicative(). In any case, to properly test Option we'd need to generate at all possible cases. What I'd suggest is to randomise None to a low percent by capturing a Gen.int() and checking modulus 20 for 5% chance.

kategory/src/main/kotlin/kategory/arrow/Function1.kt
```diff
@@ -4,7 +4,7 @@ fun <I, O> ((I) -> O).k(): Function1<I, O> = Function1(this)
 
 operator fun <I, O> Function1Kind<I, O>.invoke(i: I): O = this.ev().f(i)
 
-@higherkind class Function1<I, out O>(val f: (I) -> O) : Function1Kind<I, O> {
+@higherkind open class Function1<I, out O>(val f: (I) -> O) : Function1Kind<I, O> {
```
pakoito: Who inherits from Function1?

kategory/src/main/kotlin/kategory/data/OptionT.kt
```diff
@@ -45,6 +45,13 @@ package kategory
         inline fun <reified F> monoidK(MF: Monad<F> = kategory.monad<F>()): MonoidK<OptionTKindPartial<F>> = object : OptionTMonoidK<F> {
             override fun F(): Monad<F> = MF
         }
+
+        inline fun <reified F> functorFilter(MF: Monad<F> = kategory.monad<F>(), FF: Functor<F> = kategory.functor()): FunctorFilter<OptionTKindPartial<F>> =
```
pakoito: Monad implies functor. You only need 1 parameter.

kategory/src/main/kotlin/kategory/instances/OptionTInstances.kt
```diff
@@ -52,3 +52,14 @@ interface OptionTSemigroupK<F> : SemigroupK<OptionTKindPartial<F>> {
 interface OptionTMonoidK<F> : MonoidK<OptionTKindPartial<F>>, OptionTSemigroupK<F> {
     override fun <A> empty(): HK<OptionTKindPartial<F>, A> = OptionT(F(), F().pure(Option.None))
 }
+
+interface OptionTFunctor<F> : FunctorFilter<OptionTKindPartial<F>> {
+
+    fun FF(): Functor<F>
+
+    fun MF(): Monad<F>
```
pakoito: Monad implies functor. One should be enough.

kategory-test/src/main/kotlin/kategory/laws/FunctorFilterLaws.kt
```diff
@@ -0,0 +1,31 @@
+package kategory.laws
+
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+import kategory.*
+
+object FunctorFilterLaws {
+
+    inline fun <reified F> laws(AP: Applicative<F> = applicative(), FFF: FunctorFilter<F> = functorFilter(), EQ: Eq<HK<F, Int>>): List<Law> =
```
pakoito: Some FunctorFilters may not be Applicatives. I'd suggest passing a constructor, and inlining AP::pure in the current callsites instead.

kategory/src/test/kotlin/kategory/data/OptionTTest.kt
```diff
@@ -29,6 +30,11 @@ class OptionTTest : UnitSpec() {
                 Eq.any(),
                 Eq.any()))
 
+        testLaws(FunctorFilterLaws.laws(
+                OptionT.applicative(Option),
+                OptionT.functorFilter(),
+                Eq.any()))
```
pakoito: I believe it may need an instance of Eq<OptionTPartial<Option, Int>>, as we've seen similar issues with other tests before.

kategory/src/main/kotlin/kategory/arrow/Function1.kt
JorgeCastilloPrz: Nobody. It was being inherited by `Lifted` in the beginning, but no need for it anymore.

kategory-test/src/main/kotlin/kategory/laws/FunctorFilterLaws.kt
JorgeCastilloPrz: What do you mean ? FunctorFilterLaws depend on FunctorLaws which need an Applicative to run, that's why it's being passed here, but the FunctorFilterLaws are being tested using the FunctorFilter passed as a second argument.

kategory/src/test/kotlin/kategory/data/OptionTTest.kt
JorgeCastilloPrz: Fixed 👍

kategory-test/src/main/kotlin/kategory/generators/Generators.kt
JorgeCastilloPrz: 👍 fixed. I made it randomized to generate None and Some on a 5% - 95% basis.

JorgeCastilloPrz:
Back to you @pakoito

kategory-test/src/main/kotlin/kategory/generators/Generators.kt
pakoito: You need to capture genIntSmall().generate() as a field inside the object, else you won't get adequate distributions as it restarts the seed every time.

kategory-test/src/main/kotlin/kategory/laws/FunctorFilterLaws.kt
pakoito: FunctorLaws has 2 constructor, one convenient one with Applicative in the case that you're testing anything that inherits from applicative, and another that takes a constructor function. In this case as we're testing something that doesn't inherit from applicative please use the second one.

kategory/src/main/kotlin/kategory/data/Composed.kt
```diff
@@ -146,3 +146,24 @@ interface ComposedApplicative<F, G> : Applicative<ComposedType<F, G>>, ComposedF
...
+    override fun F(): Functor<F>
+    override fun G(): FunctorFilter<G>
+
+    override fun <A, B> mapFilter(fga: HK<ComposedType<F, G>, A>, f: (A) -> Option<B>): HK<ComposedType<F, G>, B> =
+            F().map(fga.lower(), { G().mapFilter(it, f) }).lift()
+
+    companion object {
+        inline operator fun <reified F, reified G> invoke(FF: Functor<F> = functor(), FFG: FunctorFilter<G> = functorFilter()): ComposedFunctorFilter<F, G> =
+                object : ComposedFunctorFilter<F, G> {
+                    override fun F(): Functor<F> = FF
```
pakoito: FunctorFilter implies Functor, you don't need the second parameter.

kategory/src/test/kotlin/kategory/data/OptionTTest.kt
```diff
@@ -29,6 +30,14 @@ class OptionTTest : UnitSpec() {
                 Eq.any(),
                 Eq.any()))
 
+        testLaws(FunctorFilterLaws.laws(
+                OptionT.applicative(Option),
+                OptionT.functorFilter(),
+                object : Eq<HK<OptionTKindPartial<OptionHK>, Int>>{
+                    override fun eqv(a: HK<OptionTKindPartial<OptionHK>, Int>, b: HK<OptionTKindPartial<OptionHK>, Int>): Boolean =
+                            a.ev().value == b.ev().value
```
pakoito: Are we sure we can compare Option with ==, or do we need an Eq instance?

kategory-test/src/main/kotlin/kategory/laws/FunctorFilterLaws.kt
JorgeCastilloPrz: Fixed.

kategory/src/test/kotlin/kategory/data/OptionTTest.kt
JorgeCastilloPrz: We are doing the same for other laws being tested on this class. I've changed it now to use the already created `OptionTFIdEq` instance (on top part of this test class).

kategory/src/main/kotlin/kategory/data/Composed.kt
JorgeCastilloPrz: Well that's inherited from `ComposedFunctor`, following cats implementation.

JorgeCastilloPrz:
Back to you again @pakoito

kategory/src/main/kotlin/kategory/data/Composed.kt
pakoito: Derp, I see.

PR #228, opened 2017-08-23
Make Continuations implement Monad through delegation

pakoito:
Change for marginal API improvement:

```
val M = Kleisli.monad()

M.binding {
  val result = pure(Id(1)).andThen { pure(Id(it++)) }
  yields(result)
}
```


pakoito:
@pablisco Correct. This way you can have direct access to pure() as a generic constructor without having to type the name of the dataclass, i.e. Either.Left(bla), or capture the monad on the block again.

PR #231, opened 2017-08-25
Add wrapper and instances for kotlinx.coroutines.Job

pakoito:
Job is the basic Coroutine type for running cancellable arbitrary code blocks. Turns out it's a MonadError and AsyncContext too so it's been assimilated.


kategory-effects/src/main/kotlin/kategory/effects/instances/JobW.kt
```diff
@@ -0,0 +1,89 @@
...
+                    either.fold({}, { state.set(it) })
+                }.apply {
+                    this.invokeOnCompletion { t: Throwable? ->
+                        val a: A? = state.get()
+                        if (t == null && a != null) {
+                            f(a).thunk { either: Either<Throwable, B> ->
+                                ff(either)
+                            }
+                        }
+                    }
```
pakoito: There were two possible approaches to this impl. The first one requires a runBlocking { } block, the other uses invokeOnCompletion and a simple AtomicReference. Code-wise it seemed like the fastest approach.

I'd like @elizarov to weight in if we could get his advice.

kategory-effects/src/main/kotlin/kategory/effects/instances/JobW.kt
```diff
@@ -0,0 +1,89 @@
...
+        JobW { ff: (Either<Throwable, A>) -> Unit ->
+            val runCallback: (Either<Throwable, A>) -> Unit = { either: Either<Throwable, A> ->
+                ff(either)
+            }
+            this.thunk(runCallback).apply {
+                invokeOnCompletion { t: Throwable? ->
+                    if (t != null) {
+                        function(t).thunk(runCallback)
+                    }
+                }
```
pakoito: Same as above, as both approaches with runBlocking or invokeOnCompletion were available.

kategory-effects/src/main/kotlin/kategory/effects/data/IO.kt
```diff
@@ -95,6 +95,10 @@ import kategory.effects.data.internal.error
 
         fun monad(): Monad<IOHK> = this
 
+        fun monadError(): MonadError<IOHK, Throwable> = this
```
raulraja: Since `@deriving` is already in master all this instances can be autogenerated

kategory-effects/src/main/kotlin/kategory/effects/instances/JobW.kt
```diff
@@ -0,0 +1,89 @@
+package kategory
+
+import kotlinx.coroutines.experimental.CoroutineStart
+import kotlinx.coroutines.experimental.Job
+import kotlinx.coroutines.experimental.launch
+import java.util.concurrent.atomic.AtomicReference
+import kotlin.coroutines.experimental.CoroutineContext
+
+@higherkind data class JobW<out A>(val thunk: ((Either<Throwable, A>) -> Unit) -> Job) : HK<JobWHK, A> {
```
raulraja: Generate instances with @deriving

kategory-effects/src/main/kotlin/kategory/effects/instances/JobWInstances.kt
```diff
@@ -0,0 +1,44 @@
+package kategory
+
+import kotlin.coroutines.experimental.CoroutineContext
+
+interface JobWInstances :
```
raulraja: :fire:

pakoito: It cannot be derived because several methods depend on CoroutineContext, and MonadError has a preapplied Error type.

PR #232, opened 2017-08-25
Split AsyncLaws into its own package

pakoito:
This PR moves AsyncLaws from the test package of kategory-effects into its own module. This module will be required to test 3rd party integrations.

It's built on top of #231, so wait until that's merged.


kategory-effects-test/gradle.properties
```diff
@@ -0,0 +1,4 @@
+# Maven publishing configuration
+POM_NAME=Kategory-Effects-Test
+POM_ARTIFACT_ID=kategory-effects-test
```
ffgiraldez: we could remove kategory from artifact because we already use it in group id i like `io.kategory:effects-test` instead of  `io.kategory:kategory-effects-test`

pakoito: All other packages, including the processor, use the kategory- prefix. I've kept it for consistency.

PR #233, opened 2017-08-25
Add kategory-effects-rx2

pakoito:
Depends on #232

This PR adds support for RxJava 2's Observable. It's a WIP until I figure some things out, like whether we need to support Single, if tailrecM is correct as it's slow as fuck, and whether to provide a better DSL for binding w/ Schedulers.


pakoito:
Threading and Schedulers should be a parameter

pakoito:
Proposed API:

```
ObservableWFlatMapInstances.binding {
 val users = getUserList().bind(destination = Schedulers.newThread())
 val user = Observable.from(users).bind()
 val friends = getFriends(user.id).bind(destination = Schedulers.newThread())
 val result = updateInDatabase().bind(origin = DatabaseScheduler.instance())
 yield(result)
}
```

kategory-effects-rx2/build.gradle
```diff
@@ -0,0 +1,31 @@
...
+    testCompile "io.kotlintest:kotlintest:$kotlinTestVersion"
+    testCompile project(':kategory-effects-test')
+
+    compile "io.reactivex.rxjava2:rxjava:2.1.3"
+}
+
+build.dependsOn ':detekt'
+
+sourceCompatibility = javaVersion
+targetCompatibility = javaVersion
```
nomisRev: This has moved to root build.gradle while ago

PR #234, opened 2017-08-26
Improve Jobs

pakoito:
After reading through the Coroutines documentation (https://github.com/Kotlin/kotlinx.coroutines/blob/master/coroutines-guide.md#coroutine-basics) I realised there were a couple of mistakes on our JobW implementation. I've corrected them.

This PR adds more constructors for JobW, fixes Jobs being run twice, and makes the API a bit cleaner. It also fixes MonadErrorLaws to take separate equals for errors and successes.

PR #235, opened 2017-08-26
Add ApplicativeError laws

pakoito:
This PR adds the laws for ApplicativeError. They add another EQ to the call chain, it wasn't too painful. Depends on #234


pakoito:
Wrong branch to merge to

PR #242, opened 2017-09-02
Add MonadFilter

JorgeCastilloPrz:
* Add MonadFilter construct
* Add MonadFilterLaws
* Add some MonadFilter instances, like the ones from WriterT

### SIDE REFACTORS
* Separated all the `WriterT` instances.


JorgeCastilloPrz:
@pakoito , it would be helpful If you could take a look on how the `WriterT` instances separation has been done. Just in case you see something that could be improved on the approach.

kategory/src/main/kotlin/kategory/instances/WriterTInstances.kt
```diff
@@ -1,31 +1,104 @@
...
+interface WriterTFunctor<F, W> : Functor<WriterTKindPartial<F, W>> {
+    fun F0(): Functor<F>
+
+    override fun <A, B> map(fa: HK<WriterTKindPartial<F, W>, A>, f: (A) -> B): HK<WriterTKindPartial<F, W>, B> = fa.ev().map { f(it) }
+}
+
+interface WriterTMonadFilter<F, W> : WriterTMonad<F, W>, MonadFilter<WriterTKindPartial<F, W>> {
+    override fun F0(): MonadFilter<F>
+
+    override fun <A> empty(): HK<WriterTKindPartial<F, W>, A> = WriterT(F0(), F0().empty())
```
raulraja: This method an all others that may return HK repesentations can use just `WriterT<F, W, A>` as return type.

kategory/src/main/kotlin/kategory/typeclasses/MonadFilter.kt
```diff
@@ -0,0 +1,17 @@
+package kategory
+
+/**
+ * a Monad equipped with an additional method which allows us to create an "empty" value for the Monad
+ * (for whatever "empty" makes sense for that particular monad). This is of particular interest to us
+ * since it allows us to add a filter method to a Monad, which is used when pattern matching or
+ * using guards in for comprehensions.
```
raulraja: :fire: the comment is only relevant to Scala.

kategory/src/main/kotlin/kategory/data/WriterT.kt
```diff
@@ -2,7 +2,8 @@ package kategory
 
 @higherkind data class WriterT<F, W, A>(val MF: Monad<F>, val value: HK<F, Tuple2<W, A>>) : WriterTKind<F, W, A> {
 
-    companion object {
+    companion object : WriterTInstances {
```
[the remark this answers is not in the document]
JorgeCastilloPrz: Hi @raulraja. This is a different approach @pakoito and me were talking about yesterday. I wanted to add a couple of new instances for `WriterT` and had some issues. In cats, instances are normally dependent using trait mixins. `WriterTInstances` inherits behaviors from `WriterTInstances0`, which inherits behaviors from `WriterTInstances1`... etc, and the companion object inherits from `WriterTInstances` to be able to acquire those behaviors.

This is just mimicking that in the same exact way.

The PR is Work in Progress and was opened to let @pakoito (and also you, ofc) to validate this new style, since to me it's more clear and requires less code written inside the class itself (`WriterT`).

kategory/src/main/kotlin/kategory/instances/WriterTInstances.kt
```diff
@@ -1,31 +1,104 @@
...
 
-interface WriterTInstances<F, W> :
-        Functor<WriterTKindPartial<F, W>>,
-        Applicative<WriterTKindPartial<F, W>>,
-        Monad<WriterTKindPartial<F, W>>,
-        MonadWriter<WriterTKindPartial<F, W>, W> {
+interface WriterTInstances : WriterTInstances0 {
 
-    fun MM(): Monad<F>
+    fun <F, W> monadForWriterT(MF: Monad<F>, MW: Monoid<W>): Monad<WriterTKindPartial<F, W>> =
```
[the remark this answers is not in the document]
JorgeCastilloPrz: Not exactly. Those interfaces are named, instead of directly using the anonymous classes created inside the `WriterT` companion. Thanks to this I am able to apply inheritance relationship between all the interfaces to compose the instances behavior, inherited by the companion.

The companion methods are just shortcuts to this, so they can pass injected global monad instances to them and keep the inheritance relationship between all of them.

kategory/src/main/kotlin/kategory/typeclasses/MonadFilter.kt
JorgeCastilloPrz: What do you mean? Yes, it's copied from Scala, but It's completely generic comment (not related to scala implementation) that applies equally here.

kategory/src/main/kotlin/kategory/data/WriterT.kt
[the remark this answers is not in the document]
pakoito: I misunderstood your intentions yesterday. The companion object should not inherit from WriterTInstances, instances are provided on demand instead.

kategory/src/main/kotlin/kategory/typeclasses/MonadFilter.kt
raulraja: The codebase has several comments that are only scala related:
https://github.com/kategory/kategory/blob/master/kategory/src/main/kotlin/kategory/data/Option.kt#L86
https://github.com/kategory/kategory/blob/master/kategory/src/main/kotlin/kategory/data/Option.kt#L181
https://github.com/kategory/kategory/blob/master/kategory/src/main/kotlin/kategory/data/Either.kt#L160

You can see subtle mistakes in all those links of examples that are not valid Kotlin and there is many more.
In this case there are terms used that are not called that in Kotlin:
- `for comprehensions` : We don't have `for` and therefore are either monadic comprehensions or we come up with new names but `for comprehensions` is a Scala thing.
- `pattern matching`: I don't think Kotlin has true pattern matching as it is in other langs like Haskell or Scala and even if we adopt this term in Kotlin we have no ocurrences of such an instance of using filter in when statements anywhere AFAIK.

My point is that since our implementation is deviating from the Scala one so much lately we may as well come up with our own comments or not provide them at all and leave it for docs so we use the terminology used and familiar to the Kotlin community.

kategory/src/main/kotlin/kategory/data/WriterT.kt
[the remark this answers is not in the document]
raulraja: The separation of instances is great but the companion should not extend them because that breaks modularity since users can't change the companion's inheritance being a singleton once declared and compiled. For that reason we decided a few weeks back after discussion to not ever use the companions as the instance and instead obtain the default instances through auxiliary methods where each method will create the instance it refers to. The reason this is done in the companion sometimes in Scala is because the Scala compiler looks in the companion for implicit resolution but in Kotlin that is not the case.

Having the companion extend an interface that adds just factory methods to the scope just creates indirection but IMO does not solve anything in Kotlin and increases the number of methods and stack size it takes to get to an instance.

The reason why we were doing that in the companion before in Kategory was to autoregister the instances but that is being moved to the @implicit declarations that Eugenio is implementing.

kategory/src/main/kotlin/kategory/instances/WriterTInstances.kt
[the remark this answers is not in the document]
raulraja: See https://github.com/kategory/kategory/pull/242/files/2d736d0782b2d36ed4b730ba668030fff5d2ec19#r136718720 , I still see no point for an extra layer of indirectionin Kotlin because we have no implicit resolution that justifies factories in each instance declaration and the companion. I may be missing something about your point on inheritance relationship between the interfaces. If you want to elaborate on that piece I'll appreciate it to understand what the benefit of these methods are.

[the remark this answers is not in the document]
raulraja: In Scala those methods exist as implicits to enable the Lower Priority Implicits pattern but we don;t need that in Kotlin https://stackoverflow.com/questions/33544212/explain-the-lowpriorityimplicits-pattern-used-in-scala-type-level-programming
Scala needs those to ensure user's provided implicits don't cause ambiguity and collisions when the compiler resolves implicits since those in lower traits will have less precedence.

JorgeCastilloPrz:
@raulraja @pakoito ready for review now. I'll start working on the `MonadCombine` and it's instances now, since it's very related to this, but it's another story from the roadmap.

kategory-test/src/main/kotlin/kategory/laws/MonadFilterLaws.kt
```diff
@@ -0,0 +1,29 @@
+package kategory.laws
+
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+import kategory.*
+
+object MonadFilterLaws {
```
raulraja: :+1:

kategory/src/main/kotlin/kategory/data/Option.kt
```diff
@@ -7,7 +7,7 @@ package kategory
  * are either an instance of $some or the object $none.
  */
 @higherkind
-@deriving(Functor::class, Applicative::class, Monad::class, Foldable::class, Traverse::class)
+@deriving(Functor::class, Applicative::class, Monad::class, Foldable::class, Traverse::class, MonadFilter::class)
```
raulraja: :clap:

kategory/src/main/kotlin/kategory/data/WriterT.kt
```diff
@@ -13,58 +14,66 @@ package kategory
...
-
-                    override fun <A> pass(fa: HK<WriterTKindPartial<F, W>, Tuple2<(W) -> W, A>>): HK<WriterTKindPartial<F, W>, A> =
-                            WriterT(MM, MM.flatMap(fa.ev().content(), { tuple2FA -> MM.map(fa.ev().write(), { l -> Tuple2(tuple2FA.a(l), tuple2FA.b) }) }))
-
-                    override fun tell(w: W): HK<WriterTKindPartial<F, W>, Unit> = WriterT.tell(w)
-
-                    override fun MM(): Monad<F> = MM
-
-                    override fun SG(): Monoid<W> = SG
+        inline fun <reified F, reified W> functor(FF: Functor<F> = kategory.functor<F>()): Functor<WriterTKindPartial<F, W>> =
```
raulraja: These methods should use the most concrete type instead to avoid pushing the burden to users having to call `ev()`. So in this case instead of returning `Functor<WriterTKindPartial<F, W>>` we should just return `WriterTFunctor<F, W>` and the same applied to the rest of the factory methods.

PR #243, opened 2017-09-02
DeferredKW

pakoito:
This PR adds a wrapper for the kotlinx.coroutines type Deferred.

I've also sneaked in a new construtor for Eq that takes a function and is so much cleaner.

### DIFF
--- a/kategory-effects/src/main/kotlin/kategory/effects/data/DeferredKW.kt
+++ b/kategory-effects/src/main/kotlin/kategory/effects/data/DeferredKW.kt
@@ -60,13 +60,13 @@ typealias DeferredResult<A> = Either<Throwable, A>
         fun <A, B> tailRecM(coroutineContext: CoroutineContext, a: A, f: (A) -> DeferredKW<Either<A, B>>): DeferredKW<B> {
             tailrec fun go(a: A, f: (A) -> DeferredKW<Either<A, B>>): DeferredKW<B> =
                     f(a).attempt().let { result ->
-                        /* If you remove return here, tailrec stops working. Magic. */
-                        return when (result) {
+                        /* If you remove return here tailrec stops working. Jetbrains Please. */
+                        when (result) {
                             is Either.Left -> raiseError(coroutineContext, result.a)
                             is Either.Right -> {
                                 val next: Either<A, B> = result.b
                                 when (next) {
-                                    is Either.Left -> go(next.a, f)
+                                    is Either.Left -> return go(next.a, f)
                                     is Either.Right -> pure(coroutineContext, next.b)
                                 }
                             }
--- a/kategory-effects/src/main/kotlin/kategory/effects/data/DeferredKW.kt
+++ b/kategory-effects/src/main/kotlin/kategory/effects/data/DeferredKW.kt
@@ -7,71 +7,61 @@ import kotlin.coroutines.experimental.CoroutineContext
 
 typealias DeferredResult<A> = Either<Throwable, A>
 
-@higherkind data class DeferredKW<out A>(val thunk: () -> Deferred<DeferredResult<A>>) : DeferredKWKind<A> {
+@higherkind data class DeferredKW<out A>(val coroutineContext: CoroutineContext, val thunk: (CoroutineContext) -> Deferred<A>) : DeferredKWKind<A> {
 
-    fun <B> map(coroutineContext: CoroutineContext, f: (A) -> B): DeferredKW<B> =
-            flatMap(coroutineContext, { a: A -> pure(coroutineContext, f(a)) })
+    fun runDeferred(): Deferred<A> =
+            thunk(coroutineContext)
 
-    fun <B> flatMap(coroutineContext: CoroutineContext, f: (A) -> DeferredKW<B>): DeferredKW<B> =
-            DeferredKW {
+    fun <B> map(f: (A) -> B): DeferredKW<B> =
+            flatMap { a: A -> pure(coroutineContext, f(a)) }
+
+    fun <B> flatMap(f: (A) -> DeferredKW<B>): DeferredKW<B> =
+            DeferredKW(coroutineContext) {
                 this@DeferredKW.attempt().fold(
-                        { raiseError<B>(coroutineContext, it).thunk() },
-                        { f(it).thunk() })
+                        { raiseError<B>(coroutineContext, it).runDeferred() },
+                        { f(it).runDeferred() })
             }
 
     companion object {
         inline operator fun <A> invoke(coroutineContext: CoroutineContext, noinline f: suspend CoroutineScope.() -> A): DeferredKW<A> =
-                DeferredKW {
+                DeferredKW(coroutineContext) { context: CoroutineContext ->
                     Platform.onceOnly<Unit> { Unit }.let {
-                        async(coroutineContext, CoroutineStart.DEFAULT) {
-                            try {
-                                f().right()
-                            } catch (throwable: Throwable) {
-                                throwable.left()
-                            }
-                        }
-                    }
-                }
-
-        inline fun <A> unsafe(coroutineContext: CoroutineContext, noinline f: suspend CoroutineScope.() -> DeferredResult<A>): DeferredKW<A> =
-                DeferredKW {
-                    Platform.onceOnly<Unit> { Unit }.let {
-                        async(coroutineContext, CoroutineStart.DEFAULT) {
+                        async(context, CoroutineStart.DEFAULT) {
                             f()
                         }
                     }
                 }
 
         inline fun <A> async(coroutineContext: CoroutineContext, noinline fa: Proc<A>): DeferredKW<A> =
-                DeferredKW {
+                DeferredKW(coroutineContext) { context: CoroutineContext ->
                     Platform.onceOnly<Unit> { Unit }.let {
-                        async(coroutineContext, CoroutineStart.DEFAULT) {
-                            IO.async(fa).attempt().unsafeRunSync()
+                        async(context, CoroutineStart.DEFAULT) {
+                            IO.async(fa).unsafeRunSync()
                         }
                     }
                 }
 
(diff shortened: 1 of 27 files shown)


kategory-effects/src/main/kotlin/kategory/effects/data/DeferredKW.kt
```diff
@@ -0,0 +1,115 @@
...
+        inline fun <A> pure(coroutineContext: CoroutineContext, a: A): DeferredKW<A> =
+                DeferredKW(coroutineContext) { a }
+
+        inline fun <A> raiseError(coroutineContext: CoroutineContext, t: Throwable): DeferredKW<A> =
+                DeferredKW.unsafe(coroutineContext) { t.left() }
+
+        fun <A, B> tailRecM(coroutineContext: CoroutineContext, a: A, f: (A) -> DeferredKW<Either<A, B>>): DeferredKW<B> {
+            tailrec fun go(a: A, f: (A) -> DeferredKW<Either<A, B>>): DeferredKW<B> =
+                    f(a).attempt().let { result ->
+                        /* If you remove return here tailrec stops working. Jetbrains Please. */
```
pakoito: I'VE SPENT HOURS IN THIS IMPL AND WTF.

kategory-effects/src/main/kotlin/kategory/effects/data/DeferredKW.kt
```diff
@@ -0,0 +1,110 @@
+package kategory
+
+import kategory.effects.data.internal.Platform
+import kategory.effects.instances.DeferredKWInstances
+import kotlinx.coroutines.experimental.*
+import kotlin.coroutines.experimental.CoroutineContext
+
+typealias DeferredResult<A> = Either<Throwable, A>
+
+@higherkind data class DeferredKW<out A>(val coroutineContext: CoroutineContext, val thunk: (CoroutineContext) -> Deferred<A>) : DeferredKWKind<A> {
```
raulraja: We should probably set the context default value to `CommonPool` or whatever `async/await` uses as default

kategory-effects/src/main/kotlin/kategory/effects/data/DeferredKW.kt
```diff
@@ -0,0 +1,110 @@
...
+
+        inline fun <A> pure(coroutineContext: CoroutineContext, a: A): DeferredKW<A> =
+                DeferredKW.invoke(coroutineContext, { a })
+
+        inline fun <A> raiseError(coroutineContext: CoroutineContext, t: Throwable): DeferredKW<A> =
+                DeferredKW(coroutineContext) { throw t }
+
+        fun <A, B> tailRecM(coroutineContext: CoroutineContext, a: A, f: (A) -> DeferredKW<Either<A, B>>): DeferredKW<B> {
+            tailrec fun go(a: A, f: (A) -> DeferredKW<Either<A, B>>): DeferredKW<B> {
+                val result: DeferredResult<Either<A, B>> = f(a).attempt()
```
raulraja: We should not block to obtain the result in `tailRecM`. Given the right async context his monad is already stacksafe so an implementation of `tailRecM` here should just use the regular combinators and there should be no need to declare this function as `tailrec`.
Something like this should work for all stack-safe monads:
```kotlin
fun <F, A, B> tailRecM(a: A, f: A -> HK<F, Either<A, B>>): HK<F, B> = flatMap(f(a), {
    when (it) {
      is Left => tailRecM(it.a, f)
      is Right => pure(it.b)
    }
}
```

kategory-effects/src/main/kotlin/kategory/effects/data/DeferredKW.kt
raulraja: All instances can be derived here with `@deriving`.

kategory-effects/src/main/kotlin/kategory/effects/data/DeferredKW.kt
```diff
@@ -0,0 +1,110 @@
...
+        inline fun <A> async(coroutineContext: CoroutineContext, noinline fa: Proc<A>): DeferredKW<A> =
+                DeferredKW(coroutineContext) { context: CoroutineContext ->
+                    Platform.onceOnly<Unit> { Unit }.let {
+                        async(context, CoroutineStart.DEFAULT) {
+                            IO.async(fa).unsafeRunSync()
+                        }
+                    }
+                }
+
+        inline fun <A> pure(coroutineContext: CoroutineContext, a: A): DeferredKW<A> =
```
raulraja: `pure` lifts an already computed value so it should not need to call `async` or need a context at all. I think we need here an actual ADT with `Pure`, `Async` and `Failed` because the context is only needed in async computations. that would remove the context from the constructor and reduce the overhead. Same applies for raising errors. I'm becoming convinced we don;t need the coroutine context at all except in the `async` convenience method. 
Furthermore if `bindAsync` already allows you to selectively use one in the middle of corutines and in bteween flatMaps the context can be entirely removed from this datatype.

kategory-effects/src/main/kotlin/kategory/effects/data/DeferredKW.kt
```diff
@@ -0,0 +1,110 @@
...
+            }
+            return go(a, f)
+        }
+
+        inline fun instances(coroutineContext: CoroutineContext): DeferredKWInstances =
+                object : DeferredKWInstances {
+                    override fun CC(): CoroutineContext = coroutineContext
+                }
+
+        inline fun functor(coroutineContext: CoroutineContext): Functor<DeferredKWHK> = instances(coroutineContext)
```
raulraja: This would away once auto derived but if any needed to remain with the context as argument the return type of these factory style method should be the concrete InstanceImpl. When we use the concrete impls users can access the concrete return types without having to manually `ev()` every result obtained from using the combinators in the typeclass.

kategory/src/main/kotlin/kategory/data/Option.kt
```diff
@@ -43,14 +43,15 @@ sealed class Option<out A> : OptionKind<A> {
...
+     * @note Used only for performance instead of fold.
      */
     internal abstract val isEmpty: Boolean
 
     /**
      * Returns true if the option is an instance of $some, false otherwise.
+     * @note Used only for performance instead of fold.
      */
-    val isDefined: Boolean = !isEmpty
+    internal val isDefined: Boolean = !isEmpty
```
raulraja: if these are internal they can't be used elsewhere. Do we want this APIs to actually be here for any reason?
Also I realized writing the docs for `Option` that we don't need subtypes overrides at all because:
```kotlin
val isDefined: Boolean = this is Some
val isEmpty: Boolean = this is None
```

kategory/src/main/kotlin/kategory/data/Option.kt
```diff
@@ -136,8 +137,9 @@ sealed class Option<out A> : OptionKind<A> {
     /**
      * Returns false if the option is $none, true otherwise.
      * @note Implemented here to avoid the implicit conversion to Iterable.
```
raulraja: Scala garbage in comments :fire: I think we should remove all these comments form the original ports in all datatypes since our implementations have by now deviated a lot.

kategory/src/test/kotlin/kategory/data/CoproductTest.kt
```diff
@@ -6,9 +6,8 @@ import org.junit.runner.RunWith
 
 @RunWith(KTestJUnitRunner::class)
 class CoproductTest : UnitSpec() {
-    val EQ: Eq<HK3<CoproductHK, IdHK, IdHK, Int>> = object : Eq<HK3<CoproductHK, IdHK, IdHK, Int>> {
-        override fun eqv(a: CoproductKind<IdHK, IdHK, Int>, b: CoproductKind<IdHK, IdHK, Int>): Boolean =
-                a.ev().extract() == b.ev().extract()
+    val EQ: Eq<HK3<CoproductHK, IdHK, IdHK, Int>> = Eq { a, b ->
```
raulraja: Nice simplification :clap:

kategory-effects/src/main/kotlin/kategory/effects/data/DeferredKW.kt
pakoito: My concern is that you'd be doing work and for a pure or error you'll be reporting on the wrong thread.

For adoption it'll be important for people to control their threads, as in mobile the thread where you operate matter for several frameworks, like UI and Database.

kategory-effects/src/main/kotlin/kategory/effects/data/DeferredKW.kt
pakoito: That version it's orders of magnitude slower to the point of looking like a freeze, I'm not sure why :/

kategory/src/main/kotlin/kategory/data/Option.kt
pakoito: I'm not a fan of porting the original documentation either. cc @JorgeCastilloPrz to comment it in, as he's more anal about inline documentation.

kategory-effects/src/main/kotlin/kategory/effects/data/DeferredKW.kt
pakoito: I agree, let me address that alongside other corutine issues on a different PR. They need to be fixed for Monad and Comonad too.

pakoito: pure requires a CoroutineContext to create an async block, which makes autogen fail :(

kategory/src/main/kotlin/kategory/data/Option.kt
pakoito: Microperf, as checking the boolean is chaper than checking the types

kategory/src/main/kotlin/kategory/data/Option.kt
pakoito: Let's address this in other PR.

PR #250, opened 2017-09-03
Remove dependency on kotlinx.coroutines for effects

pakoito:
This diff fixes a bunch of small issues related to coroutines:

1- New constructor for EQ
2- Remove bindingInContext temporarily until we work on a non-blocking solution
3- Remove JobKW, DeferredKW and other effects-kotlinx
4- Remove all dependencies on effects-kotlinx except for testing
5- Add support for CoroutineContext on Comonad

====

ORIGINAL POST (OUTDATED)

More fixes towards making effects stable:

1.- Remove dependency on kotlinx.coroutines. This is a helper library that'll be JVM-bound.
2.- New module effects-kotlinX for DeferredKW 
2.- Remove bindingInContext. It didn't work unless blocking.
3.- Add support for CoroutineContext on Comonad

Depends on #246


kategory-effects-kotlinx/src/main/kotlin/kategory/effects/data/DeferredKW.kt
```diff
@@ -0,0 +1,96 @@
+package kategory
+
+import kategory.effects.data.internal.Platform
+import kotlinx.coroutines.experimental.*
+import kotlin.coroutines.experimental.CoroutineContext
+
+typealias DeferredResult<A> = Either<Throwable, A>
+
+@higherkind
+@deriving(Monad::class, AsyncContext::class)
```
raulraja: Should
```
@deriving(Functor::class, Applicative::class, Monad::class, AsyncContext::class)
```

kategory-effects-kotlinx/src/main/kotlin/kategory/effects/data/DeferredKW.kt
```diff
@@ -0,0 +1,96 @@
...
+        inline fun <A> pure(a: A): DeferredKW<A> =
+                DeferredKW.invoke { a }
+
+        inline fun <A> raiseError(t: Throwable): DeferredKW<A> =
+                DeferredKW { throw t }
+
+        fun <A, B> tailRecM(a: A, f: (A) -> DeferredKWKind<Either<A, B>>): DeferredKW<B> {
+            tailrec fun go(coroutineContext: CoroutineContext, a: A, f: (A) -> DeferredKWKind<Either<A, B>>): DeferredKW<B> {
+                val result: DeferredResult<Either<A, B>> = f(a).attempt(coroutineContext)
+                /* FIXME(paco): KT-20075 If you remove return here tailrec stops working. Jetbrains Please. */
```
raulraja: I think that makes sense because otherwise you would not know what the stop condition in the translated loop would be. what it would be good to have by default in Kotlin is to have the last statement ina  block always be the return value. Explicit returns suck.

kategory-effects-kotlinx/src/main/kotlin/kategory/effects/data/DeferredKW.kt
```diff
@@ -0,0 +1,96 @@
...
+@higherkind
+@deriving(Monad::class, AsyncContext::class)
+data class DeferredKW<out A>(val thunk: (CoroutineContext) -> Deferred<A>) : DeferredKWKind<A> {
+
+    fun <B> map(f: (A) -> B): DeferredKW<B> =
+            flatMap { a: A -> pure(f(a)) }
+
+    fun <B> flatMap(f: (A) -> DeferredKWKind<B>): DeferredKW<B> =
+            DeferredKW { context: CoroutineContext ->
+                this@DeferredKW.attempt(context).fold(
```
raulraja: I think `attempt` defeats the purpose of having this datatype. I recall reimplementing this in an fully async way. But if we cause to block what good does deffer provides over `async/await`?

kategory-effects-kotlinx/src/main/kotlin/kategory/effects/data/DeferredKW.kt
```diff
@@ -0,0 +1,96 @@
...
+
+import kategory.effects.data.internal.Platform
+import kotlinx.coroutines.experimental.*
+import kotlin.coroutines.experimental.CoroutineContext
+
+typealias DeferredResult<A> = Either<Throwable, A>
+
+@higherkind
+@deriving(Monad::class, AsyncContext::class)
+data class DeferredKW<out A>(val thunk: (CoroutineContext) -> Deferred<A>) : DeferredKWKind<A> {
```
raulraja: I would just not consider this until we have a non-blocking solution

PR #255, opened 2017-09-05
Add TraverseFilter

tonilopezmr:
Fixes: #253 

Added `TraverseFilter` and `TraverseFilterLaws`
Added `OptionTTraverseFilter`, `ComposedTraverseFilter`
Added `TraverseFilter` and `TraverseFilterLaws` for `OptionT`, `Option` and `Const`


kategory-test/src/main/kotlin/kategory/laws/TraverseFilterLaws.kt
```diff
@@ -0,0 +1,21 @@
+package kategory.laws
+
+import io.kotlintest.properties.forAll
+import kategory.*
+
+object TraverseFilterLaws {
+
+    inline fun <reified F> laws(TF: TraverseFilter<F> = traverseFilter<F>(), GA: Applicative<F> = applicative<F>(), crossinline cf: (Int) -> HK<F, Int>, EQ: Eq<HK<F, Int>>): List<Law> =
+            TraverseLaws.laws(TF, GA, cf, EQ) + FunctorLaws.laws(GA, cf, EQ) + listOf(
```
raulraja: The traverse laws already include the functor laws so no need to add them here again or tests will run twice

kategory/src/main/kotlin/kategory/instances/ConstInstances.kt
```diff
@@ -16,6 +17,9 @@ interface ConstInstances<A> :
 
     override fun <T, U> foldR(fa: HK<ConstKindPartial<A>, T>, lb: Eval<U>, f: (T, Eval<U>) -> Eval<U>): Eval<U> = lb
 
+    override fun <G, T, U> traverseFilter(fa: HK<ConstKindPartial<A>, T>, f: (T) -> HK<G, Option<U>>, GA: Applicative<G>):
+            HK<G, HK<ConstKindPartial<A>, U>> = fa.ev().traverseFilter(f, GA)
```
raulraja: while method args is ok to take the most generic representation of the HK or Kind aliases in return type we should return the most concrete compatible In this case `HK<ConstKindPartial<A>, U>` can just be `Const<A, U>` same for all other instance methods, return always the most concrete type so users don't have to use `ev()` to extract values from the HK representation.

kategory-test/src/main/kotlin/kategory/laws/TraverseFilterLaws.kt
```diff
@@ -0,0 +1,30 @@
...
+                    Law("TraverseFilter Laws: Identity", { identityTraverseFilter(TF, GA, cf, EQ) })
+            )
+
+    //def traverseIdentity[A, B](fa: F[A], f: A => B): IsEq[F[B]] = {
```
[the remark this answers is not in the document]
tonilopezmr: In the latet commits this is already removed 👍

kategory/src/main/kotlin/kategory/instances/ConstInstances.kt
tonilopezmr: I'm going to change also in `traverse` method 👍

kategory-test/src/main/kotlin/kategory/UnitSpec.kt
```diff
@@ -20,6 +20,7 @@ abstract class UnitSpec : StringSpec() {
             Option.monad()
             Option.foldable()
             Option.traverse()
+            Option.traverseFilter()
```
tonilopezmr: It's magic!

kategory-core/src/test/kotlin/kategory/data/OptionTTest.kt
```diff
@@ -2,17 +2,26 @@ package kategory
...
 import org.junit.runner.RunWith
 
 @RunWith(KTestJUnitRunner::class)
 class OptionTTest : UnitSpec() {
+
     val EQ_ID: Eq<HK<OptionTKindPartial<IdHK>, Int>> = Eq { a, b ->
         a.value() == b.value()
     }
 
+    val EQ_OPTION = object : Eq<HK<OptionTKindPartial<OptionHK>, Int>> {
```
pakoito: There's a new, easier constructor. See `Eq { a, b -> a.value() == b.value() }` and let the type inference do the rest!

kategory-core/src/test/kotlin/kategory/data/OptionTTest.kt
```diff
@@ -2,17 +2,26 @@ package kategory
...
+        override fun eqv(a: HK<OptionTKindPartial<OptionHK>, Int>, b: HK<OptionTKindPartial<OptionHK>, Int>): Boolean =
+                a.ev().value == b.ev().value
+    }
+
+    val EQ_NESTED_OPTION = object : Eq<HK<OptionTKindPartial<OptionHK>, HK<OptionTKindPartial<OptionHK>, Int>>> {
```
pakoito: It may be that you could make the three Eq instances into a single method parametrized in T.

tonilopezmr: I have added 

```kotlin
fun <T> EQ(): Eq<T> = Eq { a, b ->
        a.value() == b.value()
    }
```

but it doesn't work


```
kategory/kategory-core/src/test/kotlin/kategory/data/OptionTTest.kt: (12, 24): Unresolved reference. None of the following candidates is applicable because of receiver type mismatch: 
public fun <A, T> ConstKind<???, ???> /* = HK<HK<ConstHK, ???>, ???> */.value(): ??? defined in kategory
public fun <A> IdKind<???> /* = HK<IdHK, ???> */.value(): ??? defined in kategory
public inline fun <F, A> OptionTKind<???, ???> /* = HK<HK<OptionTHK, ???>, ???> */.value(): HK<???, Option<???>> defined in kategory
public inline fun <F, W, A> WriterTKind<???, ???, ???> /* = HK<HK2<[ERROR : Unknown type parameter 3], [ERROR : Unknown type parameter 4], [ERROR : Unknown type parameter 5]> /* = HK<HK<WriterTHK, ???>, ???> */, ???> */.value(): HK<???, Tuple2<???, ???>> defined in kategory
```

I don't know why ヽ(。_°)ノ

pakoito: It has to be Eq<HK<OptionTKindPartial<T>>. Let's chat about why in Slack.

kategory-test/src/main/kotlin/kategory/laws/TraverseFilterLaws.kt
```diff
@@ -0,0 +1,25 @@
...
+object TraverseFilterLaws {
+
+    inline fun <reified F> laws(TF: TraverseFilter<F> = traverseFilter<F>(), GA: Applicative<F> = applicative<F>(), crossinline cf: (Int) -> HK<F, Int>, EQ: Eq<HK<F, Int>>, EQ_NESTED: Eq<HK<F, HK<F, Int>>> = Eq.any()): List<Law> =
+            TraverseLaws.laws(TF, GA, cf, EQ) + listOf(
+                    Law("TraverseFilter Laws: Identity", { identityTraverseFilter(TF, GA, cf, EQ_NESTED) }),
+                    Law("TraverseFilter Laws: filterA consistent with TraverseFilter", { filterAconsistentWithTraverseFilter(TF, GA, cf, EQ_NESTED) })
+            )
+
+    inline fun <reified F> identityTraverseFilter(FT: TraverseFilter<F>, GA: Applicative<F> = applicative<F>(), crossinline cf: (Int) -> HK<F, Int>, EQ: Eq<HK<F, HK<F, Int>>> = Eq.any()) =
+            forAll(genConstructor(genIntSmall(), cf), { fa: HK<F, Int> ->
```
pakoito: If you have an applicative of F you shouldn't need a constructor function. Applicative::pure does the same job!

PR #259, opened 2017-09-08
Bring `kategory-annotations` to the `kategory` repo

raulraja:
Also fixes a bunch of instances issues and provides some  `implicit` objects where instances can be looked up without the need to manually register anything.
- [x] `kategory-annotations` migrations
- [ ] Derive individual instances for all auto `@deriving` datatypes (will be addresed in #260)


kategory-test/src/main/kotlin/kategory/UnitSpec.kt
```diff
@@ -8,29 +8,7 @@ import io.kotlintest.specs.StringSpec
  * Base class for unit tests
  */
 abstract class UnitSpec : StringSpec() {
-    companion object {
```
pakoito: Ho ho ho hoooo

detekt.yml
```diff
@@ -78,7 +80,7 @@ formatting:
     active: true
     autoCorrect: true
   OptionalUnit:
-    active: true
+    active: false
```
pakoito: Ideally we'd need the opposite option: flagging places where Unit isn't there. For us explicitness is key and we've been bitten by returns that should be Unit instead before.

kategory-core/src/main/kotlin/kategory/data/ListKW.kt
```diff
@@ -59,17 +59,17 @@ data class ListKW<out A> constructor(val list: List<A>) : ListKWKind<A>, List<A>
             return ListKW(buf)
         }
 
-        fun functor(): ListKWHKMonadInstance = ListKW.monad()
+        fun functor(): ListKWMonadInstance = ListKW.monad()
```
tonilopezmr: Why haven't you deleted these methods too like in `Id` or `NonEmptyList`?

detekt.yml
raulraja: Agreed, If detekt rules can be user created we can also provide some to encourage a more functional style.

kategory-core/src/main/kotlin/kategory/data/ListKW.kt
raulraja: I though I did, but will look over before merging

PR #264, opened 2017-09-10
Typeclasses implicit objects

### LINKED ISSUE #260: Add all possible implicit objects to manual instances

raulraja:
This allows Typeclasses instances of arbitrary arity to be discovered implicitly by providing a default factory based on naming conventions that can be used to lookup the instance.
It also introduces granularity in the typeclass definitions for better composition.

Fixes #260 

- [x] Function0
- [x] Eval
- [x] ListKW
- [x] Option
- [x] Try
- [x] Const
- [x] Coproduct
- [x] Either
- [x] EitherT
- [x] FreeApplicative
- [x] Free
- [x] Function1
- [x] Ior
- [x] Iterable
- [x] Kleisli
- [x] NonEmptyList
- [x] Number
- [x] SequenceKW
- [x] StateT
- [x] String 
- [x] Validated
- [x] WriterT
- [x] Yoneda

Example on `EitherTFunctorInstance`
```kotlin
interface EitherTFunctorInstance<F, L> : Functor<EitherTKindPartial<F, L>> {
    fun FF(): Functor<F>
    override fun <A, B> map(fa: EitherTKind<F, L, A>, f: (A) -> B): EitherT<F, L, B> = fa.ev().map({ f(it) }, FF())
}

object EitherTFunctorInstanceImplicits {
    @JvmStatic
    fun <F, L> instance(FF: Functor<F>): EitherTFunctorInstance<F, L> = object : EitherTFunctorInstance<F, L> {
        override fun FF(): Functor<F> = FF
    }
}
```
You can now safely invoke without prior registration:
```
functor<EitherTKindPartial<OptionHK, Throwable>>()
```
Note the `EitherT` `Functor` instance depends on the `Functor<F>`. Kategory can now automatically resolve constrains recursively as declared in the `EitherTFunctorInstanceImplicits#instance` arguments.

Once all instances are migrated I will introduce a new processor `@typeclass` which generates the implicit object automatically and also the companion helpers for the datatype that the instance is targeting in the same way now `@deriving` does.

As a final pass we will remove all calls that manually provide instances that can be resolved implicitly in reified contexts.


kategory-core/src/main/kotlin/kategory/arrow/FunctionK.kt
```diff
@@ -1,6 +1,6 @@
 package kategory
 
-interface FunctionK<in F, out G> {
+interface FunctionK<F, G> {
```
pakoito: Why have they lost their variance?

raulraja: Inferred types with the type-literal in java reflection not liking contravariant positions. Also these are usually mapped to root ADTs and typeclasses which are just invariant positions AFAIK.

PR #265, opened 2017-09-11
Prisms processor

nomisRev:


kategory-test/src/main/kotlin/kategory/laws/LensLaws.kt
```diff
@@ -1,4 +1,4 @@
-package kategory.optics
+package kategory.laws
```
pakoito: For the time being everything we have is on the package `kategory`

kategory-optics/src/test/kotlin/kategory/optics/PrismSpec.kt
```diff
@@ -12,13 +12,14 @@ import kategory.applicative
...
 import kategory.genTuple
+import kategory.laws.PrismLaws
 import kategory.left
 import kategory.right
 import kategory.some
 import org.junit.runner.RunWith
 
 @RunWith(KTestJUnitRunner::class)
-class PrismTest : UnitSpec() {
+class PrismSpec : UnitSpec() {
```
pakoito: NIT: We've been using Tests in our test classes

PR #271, opened 2017-09-13
Optional optic

nomisRev:
In OptionalLaws I made an non-reusable First tag. This could be replaced with a type tagging in Kategory. You don't get the same nice syntax as in Scala `Option<A> @@  First` but an type alias could always close the gap if required. Thoughts?

```
data class Tagged<A, T>(val value: A) : ReadOnlyProperty<Any?, A> {

    /**
     * Auto unwrapping by delegation
     */
    override fun getValue(thisRef: Any?, property: KProperty<*>): A = value

}

fun <A, T> A.tag(): Tagged<A, T> = Tagged(this)

sealed class UserToken

data class User(val name: String, val token: Tagged<String, UserToken>)

    val taggedString: Tagged<String, UserToken> = "someUserToken".tag()
    println("I was manually unwrapped: ${taggedString.value}")

    val string by taggedString
    println("I was auto unwrapped: $string")

    val user = User("John", taggedString)
    println("I can only be created with a UserToken tagged string: $user")

}
```

### DIFF
--- a/kategory-optics/src/main/kotlin/kategory/optics/Optional.kt
+++ b/kategory-optics/src/main/kotlin/kategory/optics/Optional.kt
@@ -1,0 +1,129 @@
+package kategory.optics
+
+import kategory.Applicative
+import kategory.Either
+import kategory.HK
+import kategory.Option
+import kategory.Tuple2
+import kategory.compose
+import kategory.identity
+import kategory.none
+import kategory.some
+import kategory.toT
+
+/**
+ * A [Optional] can be seen as a pair of functions `getOption: (A) -> Option<B>` and `set: (B) -> (A) -> A`
+ *
+ * An [Optional] can also be defined as a weaker [Lens] and [Prism]
+ *
+ * @param A the source of a [Optional]
+ * @param B the target of a [Optional]
+ * @property getOption from an `A` we can extract a `Option<B?`
+ * @property set replace the target value by `B` in an `A` so we obtain another modified `A`
+ * @constructor Creates a Lens of type `A` with target `B`.
+ */
+abstract class Optional<A, B> {
+
+    /** get the target of a [Optional] or nothing if there is no target */
+    abstract fun getOption(a: A): Option<B>
+
+    /** get the modified source of a [Optional] */
+    abstract fun set(b: B): (A) -> (A)
+
+    companion object {
+        operator fun <A, B> invoke(getOption: (A) -> Option<B>, set: (B) -> (A) -> (A)) = object : Optional<A, B>() {
+            override fun getOption(a: A): Option<B> = getOption(a)
+
+            override fun set(b: B): (A) -> A = set(b)
+        }
+
+        fun <A, B> void() = Optional<A, B>(
+                { none() },
+                { _ -> { b -> identity(b) } }
+        )
+
+    }
+
+    /** get the target of a [Optional] or return the original value while allowing the type to change if it does not match */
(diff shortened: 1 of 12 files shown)


kategory-optics/src/main/kotlin/kategory/optics/Optional.kt
```diff
@@ -0,0 +1,173 @@
+package kategory.optics
```
pakoito: package kategory

kategory-optics/src/test/kotlin/kategory/optics/IsoTest.kt
```diff
@@ -3,7 +3,18 @@ package kategory.optics
 import io.kotlintest.KTestJUnitRunner
```
pakoito: package kategory :D

kategory-test/src/main/kotlin/kategory/laws/OptionalLaws.kt
```diff
@@ -0,0 +1,98 @@
...
+                }, a).value().option
+
+                getOption.exists { b ->
+                    modifyFId.exists {
+                        EQB.eqv(b, it)
+                    }
+                }
+            })
+
+    @PublishedApi internal data class FirstOption<A>(val option: Option<A>)
```
pakoito: What does @PublishedApi do? It's the first time I see it 😖

pakoito: Oh is this to allow inlines to use internals? Fuck yeah, I've been looking for something like this forever. We can hide some mechanisms now :D

kategory-test/src/main/kotlin/kategory/laws/OptionalLaws.kt
```diff
@@ -0,0 +1,98 @@
...
+            Law("Optional Law: modify identity = identity", { modifyIdentity(optional, aGen, EQA) }),
+            Law("Optional Law: compose modify", { composeModify(optional, aGen, funcGen, EQA) }),
+            Law("Optional Law: consistent set with modify", { consistentSetModify(optional, aGen, bGen, EQA) }),
+            Law("Optional Law: consistent modify with modify identity", { consistentModifyModifyId(optional, aGen, funcGen, EQA) }),
+            Law("Optional Law: consistent getOption with modify identity", { consistentGetOptionModifyId(optional, aGen, EQB) })
+    )
+
+    fun <A, B> getOptionSet(optional: Optional<A, B>, aGen: Gen<A>, EQA: Eq<A>): Unit =
+            forAll(aGen, { a ->
+                EQA.eqv(
```
pakoito: Is it worth using equalsUnderTheLaw here? We can get consistent changes in the future for all laws, and works like this behind the scenes.

nomisRev:
@pakoito shall I make the changes here and merge or shall we wait on a decision on [polymorphic optics #277](https://github.com/kategory/kategory/pull/277)

kategory-test/src/main/kotlin/kategory/laws/OptionalLaws.kt
nomisRev: Yes, that should be `equalsUnderTheLaw`

pakoito:
@nomisRev IMO we should merge this first, and continue the discussion about polymorphic lenses in the other ticket. I'm still wrapping my head around the concept, as Optics are somewhat new to me :D

PR #277, opened 2017-09-14
[WIP] polymorphic optics

nomisRev:
Small proposal for polymorphic optics. I don't have a lot of experience with them and I am not 100% how everything is going to work when writing other optics.

I added a small sample. https://github.com/nomisRev/kategory/blob/simon-polymorphic-optics/kategory-optics/src/main/kotlin/kategory/optics/sample.kt

Btw as you can see in the sample modify for prism is curried and lens is not curried. Signatures aren't curried by default in Kategory, should we follow the same convention for Optics?

Edit: I based this upon the Optional branch so you can find the interesting stuff in the last commit.


pakoito:
@nomisRev I'm for uncurried signatures, I've reverted such things in the past. There are already libraries to do it, so I'm not too concerned.

About polymorphic Lenses, what seems to be the largest advantage?

nomisRev:
@pakoito hard to say since I don't have a lot of experience with them. I was hoping you guys might shed some light on this.

I don't think they're absolutely necessary for the optics module to be successful and as this poc showed it was quite easy to build in without breaking existing (generated) code. I only ran into 1 issue but I should ask in the Kotlin Slack. The abstract class with a static invoke method + typealias acts up a little. On the other hand it was also quite easy to build them in so having them from the start might be easier.

To come back to the example I provided in my first comment. You can use a polymorphic lens to modify a `Tuple2<Int, Double>` to `Tuple2<String, Double>`, so you can lift `Int::toString()` to `Tuple2<Int, Double> -> Tuple2<String, Double>`.

For lifting functions curried signatures work best, so perhaps we should carefully consider which should be curried and which not to ease lifting of functions.

kategory-optics/src/main/kotlin/kategory/optics/sample.kt
```diff
@@ -0,0 +1,35 @@
...
+)
+
+fun <A, B, C> firstTuple(): PLens<Tuple2<A, B>, Tuple2<C, B>, A, C> = PLens(
+        { it.a },
+        { c -> { ab -> c toT ab.b } }
+)
+
+fun main(args: Array<String>) {
+
+    val modify = successTry<Int, Double>().modify { it.toDouble() } (Try { "ffff".toInt() })
```
raulraja: :clap:

kategory-test/src/main/kotlin/kategory/laws/OptionalLaws.kt
```diff
@@ -92,7 +92,7 @@ object OptionalLaws {
         override fun empty(): FirstOption<A> = FirstOption(Option.None)
 
         override fun combine(a: FirstOption<A>, b: FirstOption<A>): FirstOption<A> =
-                if (a.option.isDefined) a else b
```
raulraja: IMO these should not be internal but part of the public API, nothing wrong with asking an Option if it's defined, the `fold` method is too verbose for beginners and can be just as abused as here to obtain Boolean representations.

raulraja:
@nomisRev we need to bring this up to date with master and it's ready to merge

Issue #266, opened 2017-09-11
Alias for applicative builders for Try and other evaluation-based datatypes


pakoito:
Target: provide an API that looks like

`Try.parallel({ op1() }, { op2() }, { op3() }).fold({ error -> , tuple3Success -> })`

pakoito:
Closed by @anstaendig

PR #278, opened 2017-09-15
#266: Alias constructors for applicative builders for Try, Eval, IO

anstaendig:
Issue: #266


anstaendig:
Sure, I just thought it was just fitting for `Eval` not the rest. Will do!

kategory-effects/src/main/kotlin/kategory/effects/data/IO.kt
```diff
@@ -114,7 +68,180 @@ sealed class IO<out A> : IOKind<A> {
...
+                op1: () -> A,
+                op2: () -> B,
+                op3: () -> C,
+                op4: () -> D): IO<Tuple4<A, B, C, D>> = applicative().tupled(
+                invoke(op1),
+                invoke(op2),
+                invoke(op3),
+                invoke(op4)).ev()
+
+        fun <A, B, C, D, E> merge(
```
raulraja: Regarding codestyle we should follow the same conventions as in https://github.com/kategory/kategory/blob/master/kategory-core/src/main/kotlin/kategory/typeclasses/Applicative.kt. Alternatively we can change the ones in Applicative to match these one. I'll leave up to you.

anstaendig: I don't see one clear way of formatting it in`Applicative.kt`.

pakoito:
Thank you @anstaendig for the PR! I hope it helped you understand Kategory and FP a bit better :D Do ping us on Slack if you want another ticket, we can ramp up the difficulty :P

Issue #282, opened 2017-09-16
Order typeclass


pakoito:
As seen in https://leanpub.com/fpmortals/read#leanpub-auto-objecty-things

pakoito:
Completed by @nomisRev!

PR #287, opened 2017-09-17
Order typeclass

nomisRev:
As mentioned in #282.

Syntax `<=`, `>=`, etc is not possible. This is achieved through `operator fun compareTo` and only allows 1 parameter.


pakoito:
Great start, very clean :D Up next, laws and a couple of instances for Numbers: https://github.com/typelevel/cats/blob/master/kernel-laws/src/main/scala/cats/kernel/laws/OrderLaws.scala

kategory-core/src/main/kotlin/kategory/typeclasses/Order.kt
```diff
@@ -0,0 +1,50 @@
...
+    fun min(a: F, b: F): F = if (lt(a, b)) a else b
+    fun sort(a: F, b: F): Tuple2<F, F> = if (gte(a, b)) a toT b else b toT a
+
+    companion object {
+
+        inline operator fun <F> invoke(crossinline compare: (F, F) -> Int): Order<F> = object : Order<F> {
+            override fun compare(a: F, b: F): Int = compare(a, b)
+        }
+
+        fun <F : Comparable<F>> fromComparable(): Order<F> = object : Order<F> {
```
raulraja: Can we have a `comparable.toOrder(): Order<F>` extension for all `Comparable<F>` ?

kategory-core/src/main/kotlin/kategory/typeclasses/Order.kt
```diff
@@ -0,0 +1,50 @@
+package kategory
+
+interface Order<F> : Eq<F>, Typeclass {
```
raulraja: What are the possible instances for this typeclass currently in the codebase, it would be nice to add them as part of this PR and include laws , test for instance discovery checking such as the ones existing in master for each of the datatypes that can support it.

nomisRev: Currently I think only Number instances make sense. I've added them and the rest of your review remarks. Could you verify if I put them in the correct places?

nomisRev:
@pakoito @raulraja reason I made PR so early is because I had doubts about the implementation. In Scalaz they use `Ordering = LT | EQ | GT` as a result of Compare. Because this required an additional adapter between the Int result of compare and compareTo I decided not to use this implementation. Thoughts?

https://github.com/scalaz/scalaz/blob/v7.1.5/core/src/main/scala/scalaz/Order.scala

kategory-core/src/main/kotlin/kategory/typeclasses/Order.kt
```diff
@@ -0,0 +1,50 @@
...
+     * Compare [a] with [b]. Returns an Int whose sign is:
+     * - negative if `x < y`
+     * - zero     if `x = y`
+     * - positive if `x > y`
+     */
+    fun compare(a: F, b: F): Int
+
+    override fun eqv(a: F, b: F): Boolean = compare(a, b) == 0
+
+    fun lt(a: F, b: F): Boolean = compare(a, b) < 0
```
JorgeCastilloPrz: shouldn't these functions be infix or have infix alternatives ? you should be able to compare two orderable elements by `a lt b`

nomisRev: infix is only allowed on single parameter so to achieve this we could make an overload like so

```kotlin
inline fun <reified F> F.lt(O: Order<F> = order(), b: F): Boolean = O.lt(this, b)

inline infix fun <reified F> F.lt(b: F): Boolean = order<F>().lt(this, b)
```

I have not introduced this because it was also not done for Eq. I did do a similar overload for a different reason here https://github.com/kategory/kategory/pull/304 which is still open with a question about the overload.

Biggest issue is that we do not have scoped implicits so there is no way to overwrite the typeclass for a certain infix (overloaded) call.

kategory-core/src/main/kotlin/kategory/instances/NumberInstances.kt
```diff
@@ -32,6 +40,14 @@ object DoubleSemigroupInstanceImplicits {
     @JvmStatic fun instance(): Semigroup<Double> = SGDouble
 }
 
+object DoubleOrderInstance : Order<Double> {
```
raulraja: You may have to add these implicit objects also to `NumberInstancesJava`. Unfortunately the implicits lookup system that relies on reflection for the time being sees `kotlin.Double` as `java.lang.Double` and we have a special hack to handle those.

kategory-core/src/main/kotlin/kategory/typeclasses/Order.kt
raulraja: looks good, left a comment that needs to be addressed at https://github.com/kategory/kategory/pull/287/files#r140316364

kategory-core/src/main/kotlin/kategory/typeclasses/Order.kt
raulraja: Users can still change it globally in the worst case, not ideal but is what we have right now.

kategory-core/src/main/kotlin/kategory/instances/NumberInstances.kt
raulraja: nevermind, you already did!

PR #290, opened 2017-09-18
Add simple Monad cancellation with exception

pakoito:
This PR adds a simple implementation for binding cancellation based off exceptions.

I haven't checked the memory usage properly yet. My expectation would be that the final returned monad contains just a raiseError instead of retaining all previous state.


kategory-core/src/main/kotlin/kategory/typeclasses/ContinuationUtils.kt
```diff
@@ -15,7 +15,7 @@ private var <T> Continuation<T>.completion: Continuation<*>?
     get() = completionField.get(this) as Continuation<*>
     set(value) = completionField.set(this@completion, value)
 
-internal var <T> Continuation<T>.stackLabels: List<Any>
+var <T> Continuation<T>.stackLabels: List<Any>
```
raulraja: Is this required? This is a dangerous API to expose to the public inside `binding`. I'd rather have the code duplicated

kategory-effects-rx2/src/main/kotlin/kategory/effects/data/FlowableKW.kt
```diff
@@ -1,13 +1,14 @@
...
+import kategory.*
 import io.reactivex.*
 
 fun <A> Flowable<A>.k(): FlowableKW<A> = FlowableKW(this)
 
 fun <A> FlowableKWKind<A>.value(): Flowable<A> = this.ev().flowable
 
 @higherkind
-@deriving(Functor::class, Applicative::class)
+@deriving(Functor::class, Applicative::class, Monad::class)
```
raulraja: Are these Foldable and Traversable? I'm assuming that if they have an empty value others could follow.

kategory-effects/src/main/kotlin/kategory/effects/typeclass/MonadErrorCancellableContinuations.kt
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Entry point for monad bindings which enables for comprehensions. The underlying impl is based on coroutines.
+ * A coroutines is initiated and inside `MonadErrorContinuation` suspended yielding to `flatMap` once all the flatMap binds are completed
+ * the underlying monad is returned from the act of executing the coroutine
+ *
+ * This one operates over MonadError instances that can support `Throwable` in their error type automatically lifting
+ * errors as failed computations in their monadic context and not letting exceptions thrown as the regular monad binding does.
+ *
+ * This operation is cancellable by calling invoke on the [Disposable] return.
```
raulraja: The type of excection raised to the context of the Monad should be documented as the method signature just states `Throwable`.

pakoito: Added!

kategory-core/src/main/kotlin/kategory/typeclasses/ContinuationUtils.kt
pakoito: It's going to be used by reimplementations of binding done for cancellable effects. I'll copypaste it then.

kategory-effects-rx2/src/main/kotlin/kategory/effects/data/FlowableKW.kt
pakoito: Does Fold/Traverse require termination? If so, then they aren't as termination isn't assured.

raulraja: SequenceKW can model infinite streams and has a Foldable and Traverse instance

pakoito:
Do not merge. I've realised of one change I'd like to make.

PR #293, opened 2017-09-19
Automated deployment - Part deux

pablisco:
Let's try without conditions


pablisco:
The only reason to put it as deploy was to have a clear separation. Also, I thought it may not work with the tags, but it probably still does.

pablisco:
Also, we don't want to deploy the snapshot again when a tag is pushed

deploy-scripts/deploy_release.sh
```diff
@@ -16,6 +16,6 @@ elif [ "$TRAVIS_BRANCH" != "$BRANCH" ]; then
 elif ! [ "$VERSION_NAME" =~ VERSION_PATTERN ]; then
   fail "Failed release deployment: wrong version. Expected '$VERSION_NAME' to have pattern 'X.Y.Z'"
```
ffgiraldez: this will fail the build, isn't it? is that what we want if tag something that it's not a release?

pablisco: The idea is that if we got to this point there is a problem with either the scripts or with the setup. For instance, it will fail if we have a release tag on a commit that doesn't have the right version. However, I can see this being a problem for a commit that has no tag but has the final version instead of the 'SNAPSHOT' one. gonna investigate this.

pablisco: Ok, I've changed this for snapshot deployments. For release, we want to fail if we don't have the right version pattern. On snapshot releases, we only get a warning.

PR #304, opened 2017-09-21
Fold optic

nomisRev:
I am having some issues with `ListKWMonoidInstance` but I am not sure why.
```kotlin
val intFold = Fold.fromFoldable<ListKWHK, Int>()
intFold.getAll<ListKWKind<Int>, Int>(object : ListKWMonoidInstance<Int> {}, a= listOf(1,2,3).k()) == listOf(1,2,3).k()
//true

intFold.getAll<ListKWKind<Int>, Int>(a= listOf(1,2,3).k()) == listOf(1,2,3).k()
// java.lang.ClassCastException: kategory.ListKW cannot be cast to java.lang.Number
```

Any1 can guide me where I should look?

Opinions on an overload with reified Monoid lookup? I tried an internal abstract method but then it's impossible for a user to make a custom Fold.

```kotlin
/**
 * Map each target to a type R and use a Monoid to fold the results
 */
abstract fun <R> foldMap(M: Monoid<R>, s: S, f: (A) -> R): R

inline fun <reified R> foldMap(s: S, noinline f: (A) -> R): R = foldMap(monoid(), s, f)
```

### DIFF
--- a/kategory-optics/src/main/kotlin/kategory/optics/Fold.kt
+++ b/kategory-optics/src/main/kotlin/kategory/optics/Fold.kt
@@ -1,0 +1,201 @@
+package kategory.optics
+
+import kategory.Const
+import kategory.Either
+import kategory.Foldable
+import kategory.HK
+import kategory.IntMonoid
+import kategory.ListKW
+import kategory.Monoid
+import kategory.Option
+import kategory.foldable
+import kategory.identity
+import kategory.left
+import kategory.monoid
+import kategory.none
+import kategory.right
+import kategory.some
+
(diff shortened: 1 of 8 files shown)


nomisRev:
I have also introduced "type tagging" for First & Last using `Const`. I have 2 monoids for them and I also needed an `addMonoid: Monoid<Boolean>`. Does this remain locally here? Or shall I move it somewhere in core?

nomisRev:
I will re-open this PR from the branch on kategory/kategory with @raulraja his fix for the instance lookup.

PR #308, opened 2017-09-22
Optic fold

nomisRev:
Opinions on an overload with reified Monoid lookup? I tried an internal abstract method but then it's impossible for a user to make a custom Fold.

```kotlin
abstract fun <R> foldMap(M: Monoid<R>, s: S, f: (A) -> R): R

inline fun <reified R> foldMap(s: S, noinline f: (A) -> R): R = foldMap(monoid(), s, f)
```

I have also introduced "type tagging" for First & Last using Const. I have 2 monoids for them and I also needed an addMonoid: Monoid<Boolean>. Does this remain locally here? Or shall I move it somewhere in core?


raulraja:
If we can differentiate in a nice way from the different monoid possibilities `add/product` etc we should bring it to core. As for the param order my preference is for whatever does not force you to use named params as I had to do in the previous PR to get it to compile. We should probably look at a reduced example that list pros and cons of both approaches and make a decission that is enforced across the entire code base.

nomisRev:
Here is a small summary of my proposal. Maybe this can serve as a sort of guideline, should I place it somewhere else? I could also just make an issue an go through the codebase to do an cleaning  sweep to make it more consistent.

Concerning the named params the following gist covers our problem I think https://gist.github.com/nomisRev/366a4c2c8d421d3359f910d4eb682dd5. Unless I missed something.

So for methods that take another function like `HK<F, A>.map(FT : Functor<F> = functor(), noinline f: (A) -> B): HK<F, B>` we want the default param at the beginning so you define only the function like so `intListKw.map { it + 1 }`. Otherwise we have to explicitly specify that we want the default value by wrapping it in parentheses `intListKw.map({ it + 1 })`.

For methods that take a value like `Foldable<F>.size(fa: HK<F, A>, MB: Monoid<Long> = monoid()): Long` the default param should be at the end so you can call it like `foldable<ListKWHK>().size(intList)` in order to avoid having to specify skipping the default param like `foldable<ListKWHK>().size(s= intList)`.

Last as a method that takes both a value and another functions like `fun <A, B> foldMap(fa: HK<F, A>, mb: Monoid<B> = monoid(), f: (A) -> B): B` we would apply both rules. The default param is after the value and before the function so we can call it `foldMap(intList) { (it + 1).toString() }`

So IMO the rules should be following:
  * Default params should always be before HOF params
  * Default params should always be after no default value params

nomisRev:
Concerning the `add/product` monoid differentiation isn't that the same issue for Numbers atm?

Scalaz uses type tagging to differentiate between Monoids for Booleans. There is no instance of `Monoid[Boolean]` but there are two instances of `Monoid[Boolean @@ Conjunction]` and `Monoid[Boolean @@ Disjunction]`. This could be a solution to our problem, definitely if we'll only have global instances this would offer a nice solution to to differentiate between instances although type tagging wouldn't be as nice as in Scala `Tagged<Boolean, Conjunction>`.

I am unsure perhaps there are better solutions out there.

raulraja:
Excellent work @nomisRev !

PR #368, opened 2017-09-24
Add bindingInContext() to execute blocks in arbitrary threads

pakoito:
This PR creates a new binding block that takes a CoroutineContext and runs a `() -> HK<F, A>` block on it.


kategory-test/src/main/kotlin/kategory/laws/MonadLaws.kt
```diff
@@ -81,10 +84,30 @@ object MonadLaws {
                 }.equalUnderTheLaw(M.pure(num + 2), EQ)
             })
 
+    inline fun <reified F> monadComprehensionsBindInContext(M: Monad<F> = monad<F>(), EQ: Eq<HK<F, Int>>): Unit =
+            forAll(Gen.int(), { num: Int ->
+                M.binding {
+                    val a = bindInContext(CommonPool) { M.pure(num) }
```
raulraja: Should these tests use a different pool on each binding to prove this works?

pakoito: I could make thread checks like the ones I did for the Rx module. Each gets assigned a separate thread, which is expected.

PR #369, opened 2017-09-25
Optic traversal

nomisRev:


kategory-optics/src/main/kotlin/kategory/optics/Traversal.kt
```diff
@@ -0,0 +1,337 @@
...
+ * @param B the modified target of a [PTraversal]
+ */
+abstract class PTraversal<S, T, A, B> {
+
+    abstract fun <F> modifyF(FA: Applicative<F>, s: S, f: (A) -> HK<F, B>): HK<F, T>
+
+    /**
+     * Modify polymorphically the target of a [PTraversal] with an Applicative function
+     */
+    inline fun <reified F> modifyF(s: S, crossinline f: (A) -> HK<F, B>): HK<F, T> = modifyF(applicative(), s, { b -> f(b) })
```
pakoito: Applicative should be a parameter with a default value, rather than being inlined. It's rarae, still we could have multiple instances of Applicative for an F.

kategory-optics/src/test/kotlin/kategory/optics/TraversalTest.kt
```diff
@@ -0,0 +1,74 @@
...
+
+        testLaws(
+                TraversalLaws.laws(
+                        traversal = fromTraversable(),
+                        aGen = Gen.create { Gen.list(Gen.int()).generate().k() },
+                        bGen = Gen.int(),
+                        funcGen = genFunctionAToB(Gen.int()),
+                        EQA = Eq.any(),
+                        EQB = Eq.any()
+                ) + TraversalLaws.laws(
```
pakoito: For now we're separating multiple XXXLaws.laws calls. It'll be a good idea to do a pass to make them like they are in this class and remove potential duplicates. For now I'd keep them separated.

kategory-optics/src/main/kotlin/kategory/optics/Traversal.kt
nomisRev: The reason for the inline overload here is because the modifyF method with an `Applicative<F>` param is abstract and thus cannot be inlined. So I overloaded it to provide a method that could be called with the global instance, I did the same for Fold (https://github.com/kategory/kategory/pull/308).

I am unsure if there is better solution or if we'd rather force the parameter instead of an overload like this.

kategory-optics/src/main/kotlin/kategory/optics/Traversal.kt
```diff
@@ -0,0 +1,337 @@
...
+ *
+ * [Traversal] is a generalisation of [kategory.Traverse] and can be seen as a representation of modifyF.
+ * all methods are written in terms of modifyF
+ *
+ * @param S the source of a [PTraversal]
+ * @param T the modified source of a [PTraversal]
+ * @param A the target of a [PTraversal]
+ * @param B the modified target of a [PTraversal]
+ */
+abstract class PTraversal<S, T, A, B> {
```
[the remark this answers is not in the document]
nomisRev: Currently all optics are build with abstract classes since the polymorphic PR. I could make them interfaces. I can actually not remember why I went for `abstract class` instead of interface.

kategory-optics/src/main/kotlin/kategory/optics/Traversal.kt
pakoito: try this signature: `inline fun <reified F> modifyF(s: S, crossinline f: (A) -> HK<F, B>, AF: Applicative<F> = applicative())`

kategory-optics/src/main/kotlin/kategory/optics/Traversal.kt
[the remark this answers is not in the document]
pakoito: We've been using interfaces for typeclasses and such, so if there's no real reason for them to be abstract classes then yes, please do the refactor.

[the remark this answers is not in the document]
nomisRev: I will! Then I can also make the modifyF method which is currently abstract internal and only expose the reified function.  👍

PR #380, opened 2017-10-01
Upgrade kotlin to last stable release

JorgeCastilloPrz:
* Upgrading Kotlin to the last stable release: `1.1.51`.


pakoito:
Can you please update all other dependencies in this pass? Skip detekt, as it broke us a couple of releases ago.

JorgeCastilloPrz:
Done @pakoito . No great things to do here, just rxjava and coroutines can be updated to a stable version. Other than that, tons of kotlin deps could be updated to the new 1.2.0-Betas but we might want to wait a bit for that. I just wanted to get the last stable one working on this PR.

PR #403, opened 2017-10-12
Getter docs

nomisRev:


kategory-docs/docs/docs/optics/getter/README.md
```diff
@@ -6,4 +6,46 @@ permalink: /docs/optics/getter/
...
+Or from any of the optics defined in `kategory-optics` that allow to safely getting its focus.
+
+```kotlin:ank:silent
+val headGetter: Getter<NonEmptyList<String>, String> = nelHead<String().asGetter()
+val tupleGetter: Getter<Tuple<String, Int>, String> = firstTuple2<String, Int>().asGetter()
+``` 
+
+## Composition
+
+Unlike a regular `get` function a `Getter` composes. Similar to a `Lens` we can compose `Getter`s to create telescopes and zoom into nested sctructures.
```
raulraja: sctructures/structures

kategory-docs/docs/docs/optics/getter/README.md
```diff
@@ -6,4 +6,46 @@ permalink: /docs/optics/getter/
...
+
+```kotlin:ank
+val firstBar: Getter<NonEmptyList<Foo>, Int> = (nelHead<Foo>() compose barGetter)
+firstBar.get(Foo(5).nel())
+```
+
+`Getter` can be composed with `Getter`, `Iso`, `Lens` and `Fold` and the composition resulta in the following optics.
```
raulraja: resulta/results

kategory-optics/src/test/kotlin/kategory/optics/instances/NonEmptyListInstancesTest.kt
```diff
@@ -0,0 +1,30 @@
...
+import kategory.Eq
+import kategory.LensLaws
+import kategory.Option
+import kategory.UnitSpec
+import kategory.genFunctionAToB
+import kategory.genNonEmptyList
+import org.junit.runner.RunWith
+
+@RunWith(KTestJUnitRunner::class)
+class NonEmptyListInstancesTest : UnitSpec() {
```
pakoito: For tests we don't usually match the instance folder, rather the dataclass one.

kategory-optics/src/main/kotlin/kategory/optics/instances/NonEmptyListInstances.kt
```diff
@@ -0,0 +1,11 @@
+package kategory.optics
+
+import kategory.NonEmptyList
+
+/**
+ * [Lens] to operate on the head of a [NonEmptyList]
+ */
+fun <T> nelHead(): Lens<NonEmptyList<T>, T> = Lens(
```
pakoito: It would be worth adding it as an extfun to NEL's companion object.

kategory-optics/src/test/kotlin/kategory/optics/instances/NonEmptyListInstancesTest.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    init {
+
```
pakoito: Could you please add a test for instance lookup? Check the ones in IdTest for example.

kategory-optics/src/main/kotlin/kategory/optics/instances/NonEmptyListInstances.kt
nomisRev: Should I do the same for the existing instances?

kategory-optics/src/test/kotlin/kategory/optics/instances/NonEmptyListInstancesTest.kt
nomisRev: These "instances" are no typeclasses so no lookup occurs.

kategory-optics/src/main/kotlin/kategory/optics/instances/NonEmptyListInstances.kt
pakoito: Only the ones you have predefined, if you think it adds value! Definitely not a blocker.

nomisRev: I am going to leave this open for now. I originally also wanted generated optics as extensions on their respective companion object but since this required the user to explicitly define companion objects everywhere I decided not to. I think adding them to companion objects in some cases and in some cases not is really confusing. Also then there is no way for the users to import all predefined optics `kategory.optics.*` or `kategory.optics.std.*`.

nomisRev:
fixes #403

PR #421, opened 2017-10-23
Eq instances

### LINKED ISSUE #383: Add missing Eq instances
Several Eq instances are missing, i.e. Option, Either...

https://github.com/kategory/kategory/pull/382#discussion_r142039693

nomisRev:
closes #383


kategory-core/src/main/kotlin/kategory/instances/NumberInstances.kt
```diff
@@ -26,6 +26,14 @@ object ByteOrderInstanceImplicits {
     @JvmStatic fun instance(): Order<Byte> = ByteOrderInstance
 }
 
+object ByteEqInstance: Eq<Byte> {
+    override fun eqv(a: Byte, b: Byte): Boolean = a == b
+}
+
+object ByteEqInstanceImplicits {
```
pakoito: Can it be generated using @instance()?

kategory-annotations-processor/src/main/java/kategory/instances/InstanceFileGenerator.kt
```diff
@@ -194,7 +194,7 @@ class InstanceFileGenerator(
     private fun genCompanionReifiedExtensions(i: Instance): String =
             """|
                 |inline fun ${i.expandedTypeArgs(reified = true)} ${i.receiverTypeName}.Companion.${i.companionFactoryName}(${i.args.map {
-                "${it.first}: ${it.second} = ${it.second.split(".").map { it.decapitalize() }.joinToString(".")}()"
+                "${it.first}: ${it.second} = ${it.second.substringBefore("<").split(".").map { it.decapitalize() }.joinToString(".")}<${it.second.substringAfter("<")}()"
```
pakoito: Can you please give a small overview of what this does?

kategory-core/src/test/kotlin/kategory/data/EitherTest.kt
```diff
@@ -2,9 +2,11 @@ package kategory
 
 import io.kotlintest.KTestJUnitRunner
 import io.kotlintest.matchers.shouldNotBe
+import io.kotlintest.properties.Gen
```
pakoito: Is Gen required, or just a leftover?

kategory-core/src/test/kotlin/kategory/data/MapKWTest.kt
```diff
@@ -56,6 +53,7 @@ class MapKWTest : UnitSpec() {
             }
         }
 
-        testLaws(TraverseLaws.laws(MapKW.traverse<String>(), MapKW.traverse<String>(), { a: Int -> mapOf<String, Int>("key" to a).k() }, EQ))
+        testLaws(TraverseLaws.laws(MapKW.traverse<String>(), MapKW.traverse<String>(), { a: Int -> mapOf<String, Int>("key" to a).k() }))
```
pakoito: WHERE IS YOUR EQ NOW? Love it.

pakoito:
Do we know which instances of Eq are missing?

pakoito:
Small detekt failure, it's supposed to autofix but it didn't.

kategory-core/src/main/kotlin/kategory/instances/NumberInstances.kt
nomisRev: I did some investigation and this is not possible due to APT limitations. If you try to get the type from the annotation it becomes a java primitive byte and so we don't now the correct type anymore.

https://area-51.blog/2009/02/13/getting-class-values-from-annotations-in-an-annotationprocessor/
```
element.getAnnotation(instanceAnnotationClass).let {
    try {
        elementUtils.getTypeElement(it.target.qualifiedName)
    } catch (mte: MirroredTypeException) {
        mte.typeMirror //byte
    }
}
```

nomisRev:
@pakoito TupleN instances are missing. I wrote them all but generation failed because there is no companion object for them. I still have them locally so I can easily add them them all if we create companion objects for TupleN instances.

kategory-core/src/test/kotlin/kategory/data/MapKWTest.kt
nomisRev: Should I do a cleanup run over all tests for this? I was planning to do it in the optics module in a separate PR since I also need to review the laws depending on `Gen`.

kategory-annotations-processor/src/main/java/kategory/instances/InstanceFileGenerator.kt
nomisRev: I also extracted it to a separate method to make it a little clearer.

Here we are transforming the typeclass type to a method call for resolution.
`kategory.Eq<A>` -> `kategory.eq<A>()` we were doing so by splitting on `.` and decapitalizing.

But this fails for types with fixed types like `Try<A>`.
`kategory.Eq<kotlin.Throwable>` -> `kategory.eq<kotlin.throwable>()`.

On second thought I currently disabled implicit composition of `Eq<Throwable>` for `Try<A>` because this is also doesn't work for resolution. I'll look if this can be added there as well.

pakoito:
@nomisRev Yeah, add empty companion objects to them, it should be trivial.

PR #427, opened 2017-10-29
Validated fixes

aballano:
* Some nits in the Typeclass file
* Added a couple of tests for Validated's `combineK`
* Added `Validated.toIor`
* Added a vanilla `combine` function that performs a slightly different combination than `combineK`, based on [cats' version](https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/data/Validated.scala#L266).

@kategory/maintainers Please take a look and let me know if I messed it up 😆

### DIFF
--- a/kategory-core/src/test/kotlin/kategory/data/ValidatedTest.kt
+++ b/kategory-core/src/test/kotlin/kategory/data/ValidatedTest.kt
@@ -27,7 +27,7 @@ class ValidatedTest : UnitSpec() {
         testLaws(ApplicativeLaws.laws(Validated.applicative(StringMonoidInstance), Eq.any()))
         testLaws(TraverseLaws.laws(Validated.traverse(), Validated.applicative(StringMonoidInstance), ::Valid, Eq.any()))
         testLaws(SemigroupKLaws.laws(
-                Validated.semigroupK(IntMonoid, IntMonoid),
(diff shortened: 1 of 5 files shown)


kategory-core/src/test/kotlin/kategory/data/ValidatedTest.kt
```diff
@@ -165,27 +170,65 @@ class ValidatedTest : UnitSpec() {
...
+
+        "Combine should combine Valid and Invalid Validated" {
+            val valid = Valid("Who")
+            val invalid = Invalid("Nope")
+
+            valid.combine(invalid) shouldBe(Invalid("Nope"))
+        }
+
+        "Combine should combine Invalid Validated" {
+            val invalid: Validated<String, String> = Invalid("Nope")
```
aballano: @kategory/maintainers One thing I couldn't avoid is to specify String here (or any other inplicit Semigroup that we have, like Int), as Nothing wouldn't work with the reified needed for the `combine` call with "implicit" Semigroups and Any or Unit are not provided global instances (shall we maybe create them and return something dummy there? 🤔 )

raulraja: This is fine AFAIK since it's in tests

PR #433, opened 2017-10-30
Add toOption() and toNullable() helpers

### LINKED ISSUE #432: Native kotlin null management to Option conversion
Would it make sense to have tools to allow a quick conversion from kotlin native way of handing nulls (`Object?`) to kategory's way of dealing with it (`Option<Object>`).

It's pretty trivial to do and can allow easy conversions:

```kotlin
fun <T : Any> T?.toOption(): Option<T> = this?.some() ?: none()
fun <T> Option<T>.toNative(): T? = this.getOrElse { null }
```

Arguably it could also be `fun <T : Any> T?.toOption(): Option<T> = Option.fromNullable(this)` for the first one, but really anything will do.

ColinHebert:
- Add `toOption()` extension for all objects to transform a nullable Kotlin object into a Kategory `Option`
- Add `toNullable()` extension for all options to transform an `Option` into a nullable Kotlin object

Fixes #432


pakoito:
Good job! For the sake of completion and because I'd like you to get more familiar with the project:

* You also need to add a section about it on the docs: https://github.com/kategory/kategory/tree/master/kategory-docs/docs/docs/datatypes/option

* And add a short test: https://github.com/kategory/kategory/blob/master/kategory-core/src/test/kotlin/kategory/data/TryTest.kt

kategory-core/src/main/kotlin/kategory/data/Option.kt
```diff
@@ -195,6 +195,10 @@ fun <B> Option<B>.getOrElse(default: () -> B): B = fold({ default() }, { it })
  */
 fun <A, B : A> OptionKind<B>.orElse(alternative: () -> Option<B>): Option<B> = if (ev().isEmpty) alternative() else ev()
 
+fun <A : Any> A?.toOption(): Option<A> = Option.fromNullable(this)
+
+fun <A> Option<A>.toNative(): A? = this.getOrElse { null }
```
aballano: I don't think we need a `toNative` as the clients can easily do the `getOrElse` themselves, no?

aballano: @pakoito 👀

pakoito: I'm okay with the method, I'd rather just call it toNullable instead, but that's a nit that won't block the review.

artem-zinnatullin: @pakoito not a nit, it'll be more consistent with `fromNullable()`!

pakoito: First @sockeqwe in the stars, now you checking out the PRs? Daaaamn it's been a great day :D

artem-zinnatullin: Of course I check how accurately do you copy https://github.com/gojuno/koptional (I'm kidding, happy to see progress on KΛTEGORY)

@sockeqwe in stars, but @artem-zinnatullin in watchers! 😸 

// I only star projects I'm afraid to not find later, it's not the case with K∀∀∀∀TEGORY.

raulraja: :+1: toOption()
:-1: toNative()
should be `toNullable()`

sockeqwe: 👀

ColinHebert:
I've added the extra tests and the documentation as well as renamed the function to `toNullable()`.

kategory-core/src/test/kotlin/kategory/data/OptionTest.kt
```diff
@@ -60,6 +60,31 @@ class OptionTest : UnitSpec() {
             Option.fromNullable(a) shouldBe None
         }
 
+        "toOption should work for both null and non-null values of nullable types" {
+            forAll { a: Int? ->
+                // This seems to be generating only non-null values, so it is complemented by the next test
```
pakoito: You can combine generators using `Gen.oneOf()`. Try combining Gen.int() and create one for null.

kategory-docs/docs/docs/datatypes/option/README.md
```diff
@@ -19,13 +19,25 @@ Kategory models the absence of values through the `Option` datatype similar to h
 import kategory.*
 import kategory.Option.*
 
-val someValue: Option<String> = Some("I am wrapped in something")
+val message: String? = "I am wrapped in something"
+
+val someValue: Option<String> = Some(message)
 someValue
+
+// Or with the toOption extention
```
pakoito: Only the last return of a block is evaluated and returned by ank. You need to split these examples into two ```kotlin:ank blocks, one that returns someValue and another for otherValue

kategory-core/src/test/kotlin/kategory/data/OptionTest.kt
```diff
@@ -60,6 +60,31 @@ class OptionTest : UnitSpec() {
...
+
+        "toNullable should return null for none values" {
+            val a: Option<Int> = Option.none()
+            a.toNullable() shouldBe null
+        }
+
+        "toNullable should work for both null and non-null values of nullable types" {
+            forAll { a: Int? ->
+                // This seems to be generating only non-null values, so it is complemented by the next test
+                val o: Option<Int> = Option.some(a)
```
pakoito: The compiler is complaining about this call: https://travis-ci.org/kategory/kategory/builds/296163693

```
e: /home/travis/build/kategory/kategory/kategory-core/src/test/kotlin/kategory/data/OptionTest.kt: (84, 45): Type inference failed. Expected type mismatch: inferred type is Option<Option.Companion> but Option<Int> was expected
e: /home/travis/build/kategory/kategory/kategory-core/src/test/kotlin/kategory/data/OptionTest.kt: (84, 50): Too many arguments for public fun <A> A.some(): Option<A> defined in kategory
```

kategory-core/src/test/kotlin/kategory/data/OptionTest.kt
```diff
@@ -60,6 +60,31 @@ class OptionTest : UnitSpec() {
...
+            a.toOption() shouldBe None
+        }
+
+        "toNullable should return null for none values" {
+            val a: Option<Int> = Option.none()
+            a.toNullable() shouldBe null
+        }
+
+        "toNullable should work for both null and non-null values of nullable types" {
+            forAll { a: Int? ->
```
pakoito: Look into what `Gen.applicative()` does into our codebase, using `Option.applicative()` as the parameter.

kategory-core/src/test/kotlin/kategory/data/OptionTest.kt
pakoito: You can run the tests locally using `./gradlew test`

pakoito: The cause is that `Some` is the constructor and `some` is an extension function to make the value into a Some. You're making the companion object into a Some, instead of a :D

raulraja:
Tests seem to be failing to compile https://travis-ci.org/kategory/kategory/builds/296319227#L1053

raulraja:
@ColinHebert are you still interested in finishing this PR? Let us know or we could take it from here as other users are awaiting for `toNullable()` and we think it's a good addition.

modules/docs/arrow-docs/docs/docs/datatypes/option/README.md
```diff
@@ -18,13 +18,25 @@ Arrow models the absence of values through the `Option` datatype similar to how
...
+// Or with the toOption extention
+val otherValue: Option<String> = message.toOption()
+otherValue
 ```
 
 ```kotlin:ank
 val emptyValue: Option<String> = None
 emptyValue
+
+// Or with the toOption extention
```
pakoito: These blocks only evaluate the last statement of each! For both emptyvalue and othervalue to be displayed these lines need to be wrapped in another ```kotlin:ank block

diesalbla:
@raulraja There are already are an instance method `orNull` of `Option`, and a companion object method `fromNullable`, that do the same. Is this still relevant?

raulraja:
Nope, these utilities already made it

PR #434, opened 2017-10-31
Release 0.3.11

nomisRev:
# Notes 
* Instances for Eval
* Iso instances
* Improved kategory processor (Iso, higherkind & instance)
* Eq instances
* Validated fixes
* Corrected Semigroup & Monoid API

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,6 +1,6 @@
 # Package definitions
 GROUP=io.kategory
-VERSION_NAME=0.3.11-SNAPSHOT
+VERSION_NAME=0.3.11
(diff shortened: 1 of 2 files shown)


gradle.properties
```diff
@@ -1,6 +1,6 @@
 # Package definitions
 GROUP=io.kategory
-VERSION_NAME=0.3.11-SNAPSHOT
+VERSION_NAME=0.3.11
```
wiyarmir: Shouldn't we bring it back to a SNAPSHOT version so it stays like that in master?

PR #438, opened 2017-11-02
Update deploy script

pablisco:
Instead of relying on the tag from github, let's try to determine where to deploy from the actual version from the gradle properties


pakoito:
LGTM. Could we trigger a release of the docs alongside a release of the lib?

pablisco:
Yeah, can look into it. Separate PR though?

Issue #213, opened 2017-08-19
Deduplicate law tests


pakoito:
For some data types and laws we have to run the same tests multiple times. For example, Id has both Monad and Comonad, and each have laws that require testing for Functor. This effectively means the same batch of tests gets run twice.

Deduplication should be simple if the tests are not immediately run, and are instead aggregated in a set where their name string is their identity.

pakoito:
The point where functions should be deduplicated is here: https://github.com/kategory/kategory/blob/9c81e7a0d0964502fbdd748b23169e80f2022c6e/kategory-test/src/main/kotlin/kategory/UnitSpec.kt#L12

Make testLaws take a vararg of List<Law> that removes duplicates based off Law#name: https://github.com/kategory/kategory/blob/9c81e7a0d0964502fbdd748b23169e80f2022c6e/kategory-test/src/main/kotlin/kategory/laws/Law.kt#L5

groodt:
We should be able to close this now?

PR #439, opened 2017-11-04
#213: Dedupe law tests

groodt:
Addresses issue #213 

I've only modified kategory-core. If the approach I've taken is desired, then I'll continue to apply to the other modules.

### DIFF
--- a/kategory-core/src/test/kotlin/kategory/data/ConstTest.kt
+++ b/kategory-core/src/test/kotlin/kategory/data/ConstTest.kt
@@ -20,8 +20,10 @@ class ConstTest : UnitSpec() {
             eq<Const<Int, String>>() shouldNotBe null
         }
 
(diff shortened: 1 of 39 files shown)


raulraja:
👏 👏 👏 👏

groodt:
Ok, the changes are in. Tests are all passing. Awesome! 👍

pakoito:
@groodt One last change: FlowableKW has multiple tests for different permutations of backpressure strategy & flat/concat/switch. These tests are not duplicates but are filtered out. Could you please move every permutation into one different call to testLaws()

groodt:
@pakoito Good spot! I think I've split them out now. Please take a look.

pakoito:
@groodt Sadly I don't believe we can dedupe any of the permutations. Could you please revert it to what it was before this PR, where every permutation is its own testLaws()?

groodt:
@pakoito Ok, I've reverted the changes to ```FlowableKWTests```

raulraja:
@groodt thanks for your contribution, this is awesome!

PR #442, opened 2017-11-05
Add typealiases for public APIs datatypes

### LINKED ISSUE #300: Add typealiases for public APIs currently nested in their datatypes.
Currently in order to call `Some` directly in code we need to first `import kategory.Option.*`.
Ideally one can use all the nested public datatypes without imports.
We should:
```
package kategory
typealias Some<A> = Option.Some<A>
```
And same for `Left`, `Right` and others that are part of the public API.
Also all current references inside Kategory that use the extended names such as `Option.Some` should be refactored to use the new type aliases.

tonilopezmr:
Added typealiases for the following datatypes:

* Option: `Some` and  `None`
* Validated: `Valid` and `Invalid`
* Try: `Failure` and `Success`
* Either: `Right` and `Left` 

Closes #300


kategory-core/src/main/kotlin/kategory/data/Option.kt
```diff
@@ -1,5 +1,8 @@
 package kategory
 
+typealias Some<A> = Option.Some<A>
```
artem-zinnatullin: Why don't you just declare `Some` and `None` as top level?

Same is applicable to `Left`, `Right` and so on

tonilopezmr: Well I think we need a consistent API with all types inside their container. For example, `Ior` and `Either` have `Left` and `Right` types, and you can call them as `Either.Right` and `Ior.Right` without conflicts, also you should be able to call `Some` as `Option.Some` or just `Some`.

artem-zinnatullin: What's the point of typealiases then? Name clashing can be avoided by fully qualified imports. Long imports can be avoided by short package names :)

raulraja: @artem-zinnatullin We want commonly used types available at the top level `kategory.*` package since many users will just `import kategory.*` to obtain syntax and type class instances. Users wanting just a subset of Kategory can still import a la carte types already and that will not change.

PR #444, opened 2017-11-06
StateT KDoc and doc

nomisRev:


nomisRev:
closes #317

kategory-core/src/main/kotlin/kategory/data/State.kt
```diff
@@ -1,13 +1,72 @@
 package kategory
 
+/**
+ * Convenience object to make working with `StateT<IdHK, S, A>` more elegant.
+ */
 object State {
```
raulraja: Can we `typealias State<S, A> = StateT<IdHK, S, A>` and use the alias in all return types? That we we can define many of these methods as extensions to the alias which would allow users to not have to see the underlying StateT impl.

kategory-docs/docs/docs/datatypes/statet/README.md
```diff
@@ -6,4 +6,142 @@ permalink: /docs/datatypes/statet/
...
+```kotlin:ank
+stackOperationsS2().runM(listOf())
+```
+
+Available Instances:
+
+```kotlin:ank
+import kategory.debug.*
+
+showInstances<StateTHK, Unit>()
```
raulraja: Does this actually show the right instances or do you need a `KindPartial` version of the type here?

kategory-docs/docs/docs/datatypes/statet/README.md
```diff
@@ -6,4 +6,142 @@ permalink: /docs/datatypes/statet/
...
+
+For our purposes here, we're going to transform a monad that serves as a container that represents branching as an an error (left) or state (right) where computation can be performed. Given that both `State<S, A>` and `Either<L, R>` would be examples of datatypes that provide instances for the `Monad` typeclasses.
+
+Because [monads don't compose](http://tonymorris.github.io/blog/posts/monads-do-not-compose/), we may end up with nested structures such as `Either<Error, State<Either<Error, State<S, Unit>>, Unit>>` when using `Either` and `State` together. Using Monad Transformers can help us to reduce this boilerplate.
+
+In the most basic of scenarios, we'll only be dealing with one monad at a time making our lives nice and easy. However, it's not uncommon to get into scenarios where some function calls will return `Either<Error, A>`, and others will return `State<S, A>`.
+
+So let's rewrite the example of [`State` docs]({{ '/docs/datatypes/state' | relative_url }}), but instead of representing the `Stack` as an optional `NonEmptyList` let's represent it as a `List`.
+
+```kotlin:ank:silent
```
JorgeCastilloPrz: do we lose syntax highlighting when using ank ? Is the website adding highlighting afterwards in any way ? seems like its lost here on github:
captura de pantalla 2017-11-06 a las 19 38 44

kategory-docs/docs/docs/datatypes/statet/README.md
```diff
@@ -6,4 +6,142 @@ permalink: /docs/datatypes/statet/
...
+typealias StackEmpty = StackError.StackEmpty
+
+fun popE(stack: Stack): Either<StackError, Tuple2<Stack, String>> =
+        if (stack.isEmpty()) StackEmpty.left()
+        else stack.first().let {
+            stack.drop(1) toT it
+        }.right()
+
+fun pushE(s: String, stack: Stack): Either<StackError, Tuple2<Stack, Unit>> =
+        (listOf(s, *stack.toTypedArray()) toT Unit).right()
```
JorgeCastilloPrz: why using this style and not just `listOf(s) + stack` ? seems like less noisy

kategory-docs/docs/docs/datatypes/statet/README.md
```diff
@@ -6,4 +6,142 @@ permalink: /docs/datatypes/statet/
...
+import kategory.*
+
+typealias Stack = List<String>
+
+fun pop(stack: Stack): Tuple2<Stack, String> = stack.first().let {
+    stack.drop(1) toT it
+}
+
+fun push(s: String, stack: Stack): Tuple2<Stack, Unit> =
+        listOf(s, *stack.toTypedArray()) toT Unit
```
JorgeCastilloPrz: why using this style and not just listOf(s) + stack ? seems like less noisy

kategory-docs/docs/docs/datatypes/statet/README.md
```diff
@@ -6,4 +6,142 @@ permalink: /docs/datatypes/statet/
...
+In the most basic of scenarios, we'll only be dealing with one monad at a time making our lives nice and easy. However, it's not uncommon to get into scenarios where some function calls will return `Either<Error, A>`, and others will return `State<S, A>`.
+
+So let's rewrite the example of [`State` docs]({{ '/docs/datatypes/state' | relative_url }}), but instead of representing the `Stack` as an optional `NonEmptyList` let's represent it as a `List`.
+
+```kotlin:ank:silent
+import kategory.*
+
+typealias Stack = List<String>
+
+fun pop(stack: Stack): Tuple2<Stack, String> = stack.first().let {
```
JorgeCastilloPrz: is this correct? shouldn't it be stack.last().let{} ? or aren't you supposed to return the `Tuple2(currentStack, poppedElement)`?

kategory-docs/docs/docs/datatypes/statet/README.md
```diff
@@ -6,4 +6,142 @@ permalink: /docs/datatypes/statet/
...
+```kotlin:ank:silent
+sealed class StackError {
+    object StackEmpty : StackError()
+}
+
+typealias StackEmpty = StackError.StackEmpty
+
+fun popE(stack: Stack): Either<StackError, Tuple2<Stack, String>> =
+        if (stack.isEmpty()) StackEmpty.left()
+        else stack.first().let {
```
JorgeCastilloPrz: again, is this correct or should be `stack.last()`?

kategory-docs/docs/docs/datatypes/statet/README.md
```diff
@@ -6,4 +6,142 @@ permalink: /docs/datatypes/statet/
...
+stackOperationsE(listOf())
+```
+
+As is immediately clear, this code while properly modelling the errors, has become more complex but our signature now represents a simple `Stack` as a `List` with an error domain.
+Let's refactor our manual state management in the form of `(S) -> Tuple2<S, A>` to `State`.
+
+So what we want is a return type that represents `Either` a `StackError` or a certain `State` of `Stack.` When working with `State` we don't pass around `Stack` anymore, so there is no parameter to check if the `Stack` is empty.
+
+```kotlin:ank:silent
+fun _popS(): Either<StackError, StateT<IdHK, Stack, String>> = TODO()
```
JorgeCastilloPrz: wouldn't be `State` and not `StateT` here ? Seems like a typo

kategory-docs/docs/docs/datatypes/statet/README.md
```diff
@@ -6,4 +6,142 @@ permalink: /docs/datatypes/statet/
...
+```
+
+The only thing we can do is handle this with `StateT`. We want to wrap `State` with `Either`. Left to right wrap `State` `T` (with) `Either` `StateT<EitherHK`.
```
JorgeCastilloPrz: Not sure if that "Left to right wrap... sentence is easy to understand". I had to read it a couple of times :D

JorgeCastilloPrz:
Just one relevant question: what's the difference on wrapping like:
* `EitherT<State<S, A>>`
vs
* `StateT<Either<L, R>>` ?

kategory-docs/docs/docs/datatypes/statet/README.md
```diff
@@ -6,4 +6,142 @@ permalink: /docs/datatypes/statet/
...
+stackOperationsS().runM(listOf("hello", "world", "!"))
+```
+```kotlin:ank
+stackOperationsS().runM(listOf())
+```
+
+While our code looks very similar to what we had before there are some key advantages. State management is now contained within `State` and we are dealing only with 1 monad instead of 2 nested monads so we can use monad bindings!
+
+```kotlin:ank
+fun stackOperationsS2() = StateT .monad<EitherKindPartial<StackError>, Stack>().binding {
```
JorgeCastilloPrz: pretty helpful to know that transformers enable monad comprenhensions after collapsing everything to a single type 👏

kategory-docs/docs/docs/datatypes/statet/README.md
raulraja: Ank uses ```kotlin:ank so it's not recognized as Kotlin on Github. Suggestions for improvements are welcome, ideally something that works both on Github and allows us to identify a regular `kotlin` block from a `kotlin:evaluable` type of block

kategory-docs/docs/docs/datatypes/statet/README.md
raulraja: they are the same, `type State<S, A> = StateT<IdHK, S, A>` This is the same funny bizness as in the Reader vs ReaderT. It needs to be standardize for simplest usage in the same way in both cases. I still consider this APIs around State, Reader and in general any simplification parametric to the `Id` monad a part that is broken in Kategory at this moment. If you all agree we should create an issue to fix those api's in the cases of `Reader, Writer, State, ..`

kategory-docs/docs/docs/datatypes/statet/README.md
raulraja: People needs to get used to the generics syntax of transformers stack in Kotlin. There is nothing to fear let us write:
```kotlin
StateT<EitherKindPartial<L>, Stack, String>
```
This means a state transformation considering a value may be `L` or a `String` where the internal state is tracked in a Stack. Several effects encoded in that program. Useful programs will most of the time become concrete over transformer stacks like this one if type classes and HKTs land in the land and we can start writing in tagless style

kategory-docs/docs/docs/datatypes/statet/README.md
JorgeCastilloPrz: This concrete part of the docs is stating the need for StateT when you have a stack like `Either<StackError, State<IdHk, Stack, String>>` so I guess it should explicitly say `State` and not `StateT` since both are different things, even if their usage is equal.

About creating the issue, +1. It's good to standardize to simplify the usage IMO

JorgeCastilloPrz: Then it should probably say `Either<StackError, State<Stack String>>` isn't it? just for clarity about the stack we have here. Since having a transformer as the inner class does not make a lot of sense semantically speaking to me.

JorgeCastilloPrz:
Approving to avoid being the blocker here

kategory-core/src/main/kotlin/kategory/data/State.kt
nomisRev: `typealias State<S, A>` conflicts with `object State`.

kategory-docs/docs/docs/datatypes/statet/README.md
nomisRev: Hmm. I am prepending items in `push` instead of appending them. So for this particular implementation it would be `first`. I could switch this around if this makes things more clear. That would simplify push to `fun push(s: String, stack: Stack): Tuple2<Stack, Unit> = (stack + s) toT Unit` but then I should change it in `State` docs as well.

kategory-docs/docs/docs/datatypes/statet/README.md
nomisRev: This is not possible see first comment on this PR.

nomisRev:
@JorgeCastilloPrz 
`EitherT<State<S, A>> (EitherT<StateTKindPartial<IdHK, Stack>, String, Int>)`
=> `State<Stack, Either<String, Int>>`

And `StateT<Either<L, R>>` => Either<L, State<R, ?>`.

So `EitherT` can be used to model operations that `L | R` results and `StateT` can be used to model stateful computation where the `State` can become invalid.

nomisRev:
@JorgeCastilloPrz @raulraja any suggestion to improve `State` / `StateT` API? I am also implementing `IndexedStateT` on another branch and it would make things even more confusing since I wanted to create an alias for `StateT` such that `StateT<F, S, A>  = IndexedStateT<F, S, S, A>` but that will make all of this even much worse.

kategory-docs/docs/docs/datatypes/statet/README.md
JorgeCastilloPrz: Ok no worries anyway

raulraja:
I'm not sure how to improve it since typealiases are just like placeholders and not newtypes. I think this is good to merge until we have a solid strategy for the other cases, then we can change them all at once so its consistent.

PR #446, opened 2017-11-06
Fix release upload archives

ffgiraldez:
Only snapshots build should be at  oss.jfrog.org, modify default release URL to point to Bintray organization project.

With this change, I was able to upload and publish to Bintray from a local machine, and maybe also work in Travis-CI when detecting a release build.

The only behaviour it's some modules that do not have real artifact appears in Bintray:
* kategory-docs
* kategory-effects-test
* kategory-test


ffgiraldez:
if we agree, I could try to remove those modules from Bintray upload process

PR #447, opened 2017-11-06
Parallelizes ANK snippet compilation on a thread per file.

raulraja:


ank-core/src/main/kotlin/io/kategory/ank/interpreter.kt
```diff
@@ -118,37 +126,58 @@ fun extractCodeImpl(source: String, tree: ASTNode): ListKW<Snippet> {
...
+        }
+    }
+    return runBlocking {
+        snippets.map { (file, codeBlocks) ->
+            async(IOPool) {
+                val start = System.currentTimeMillis()
+                val classLoader = URLClassLoader(classpath.map { URL(it) }.ev().list.toTypedArray())
+                val seManager = ScriptEngineManager(classLoader)
+                println(":runAnk -> START eval: $file on ${Thread.currentThread().name}")
+                val engineCache: Map<String, ScriptEngine> =
```
pakoito: Should the engine cache be local per script, or shared among all of them?

ank-core/src/main/kotlin/io/kategory/ank/interpreter.kt
```diff
@@ -118,37 +126,58 @@ fun extractCodeImpl(source: String, tree: ASTNode): ListKW<Snippet> {
...
-    val evaledSnippets = snippets.list.map { snippet ->
-        val result = Try {
-            val engine = cachedEngines.get(snippet.lang)
-            val resolvedEngine = if (engine != null) engine else {
-                seManager.getEngineByExtension(extensionMappings.getOrDefault(snippet.lang, "kts"))!!
+
+fun compileCodeImpl(snippets: Map<File, ListKW<Snippet>>, classpath: ListKW<String>): ListKW<CompiledMarkdown> {
+    println(":runAnk -> started compilation")
+    val IOPool = object : CoroutineDispatcher() {
+        private val ioBoundPool = Executors.newFixedThreadPool(snippets.size)
```
pakoito: One Thread per file?

ank-core/src/main/kotlin/io/kategory/ank/interpreter.kt
```diff
@@ -118,37 +126,58 @@ fun extractCodeImpl(source: String, tree: ASTNode): ListKW<Snippet> {
...
-    val cachedEngines: ConcurrentHashMap<String, ScriptEngine> = ConcurrentHashMap()
-    val evaledSnippets = snippets.list.map { snippet ->
-        val result = Try {
-            val engine = cachedEngines.get(snippet.lang)
-            val resolvedEngine = if (engine != null) engine else {
-                seManager.getEngineByExtension(extensionMappings.getOrDefault(snippet.lang, "kts"))!!
+
+fun compileCodeImpl(snippets: Map<File, ListKW<Snippet>>, classpath: ListKW<String>): ListKW<CompiledMarkdown> {
+    println(":runAnk -> started compilation")
+    val IOPool = object : CoroutineDispatcher() {
```
pakoito: I'm quite sure that kotlinx.coroutines provides a better abstraction for a weighted ThreadPool.

nomisRev:
Gets stuck on output. I get same result locally. Seems to work fine on `CommonPool`, I also tried `newSingleThreadExecutor` I wanted to monitor difference with `CommonPool` but on `newSingleThreadExecutor` it also hang on output.

ank-core/src/main/kotlin/io/kategory/ank/interpreter.kt
raulraja: I was told they were thread safe but those rx issues I mentioned to you in private show they are not so there needs to be one instance per file basically. Those engines are not thread safe.

raulraja:
@nomisRev it eventually completes, this concurrent execution does not hide the fact that the Kotlin Script Engine is broken when evaling corotuines and other code that does type fiddling more than it is used to. If you connect with a debugger and pause the execution you will find yourself inside a `while(true)` in the Script Engine impl that is trying to resolve variable scopes and never ending until it gives up eventually.

nomisRev:
@raulraja even while it is printing the output files?

PR #448, opened 2017-11-07
Autofold

nomisRev:
Code generation that implements extension `fold` functions for sealed classes. Given following snippet.

```kotlin
@autofold sealed class Maybe<A> {
    data class Just<A>(val value: A): Maybe<A>()
    object Empty: Maybe<kotlin.Nothing>()
}
```

It will generate following code

```kotlin
inline fun <A, B> Maybe<A>.fold(
        crossinline just: (Maybe.Just<A>) -> B,
        crossinline empty: (Maybe.Empty) -> B
): B = when (this) {
    is Maybe.Just -> `just`(this)
    is Maybe.Empty -> `empty`(this)
}
```

It does not unwrap `Just` to `(A) -> B` because if there would be another subtype `data class Another<A>(val value: A): Maybe<A>()` you would lose type safety over function parameters but on the other hand I might make the code gen smart enough to check that... Thoughts?


nomisRev:
@pakoito `@fold` conflicts with `kategory.optics.Fold`

nomisRev:
@pakoito well damn :D `2 type arguments expected. Use 'W1<*, *>' if you don't want to pass type arguments`

```
inline fun <A, B> `kategory`.`optics`.`Wrap`<A>.fold(
        crossinline w1: (`kategory`.`optics`.`W1`<A, B>) -> B
): B = when (this) {
    is `kategory`.`optics`.`W1` -> `w1`(this)
}
```

nomisRev:
@pakoito what would you suggest here? Currently it code gen bails with following message, since it impossible to write a fold method in this situation.

```
autofold.wrap.kt: (6, 8): 2 type arguments expected. Use 'W1<*, *>' if you don't want to pass type arguments
autofold.wrap.kt: (6, 41): Type mismatch: inferred type is Wrap<A> but W1<A, B> was expected
```

pakoito:
@nomisRev Add <*, *> to the when clause, but IIRC that means that the caller can assume the type of B and get a runtime error.

raulraja:
We don;t support derivation at the moment anywhere else for kinds of arity higher than one so I think this is good as is until we want to support those across the board which we can tackle at once.

nomisRev:
@pakoito I wouldn't add something that will lead to unsafe code.

pakoito:
@raulraja Doesn't this cause the same kind of conflict PartialFunction had? That's what I mean with unsafe.

pakoito:
@nomisRev @raulraja My suggestion going forward: fail the processor for those cases where child classes have more generic parameters than the sealed class.

nomisRev:
@pakoito it currently fails for all cases where generics cannot be figured out and not just
> where child classes have more generic parameters than the sealed class.
So no runtime errors can occur from using the generated fold methods. Currently it fails with the message of the invalid code but I will update it proper compile time error message before we merge.

raulraja:
@pakoito I think https://github.com/kategory/kategory/pull/448#issuecomment-343746443 addressed your concern. Can we merge this?

PR #449, opened 2017-11-07
Order docs

nomisRev:


nomisRev:
Closes #347

kategory-core/src/main/kotlin/kategory/typeclasses/Eq.kt
```diff
@@ -1,28 +1,79 @@
...
             override fun eqv(a: F, b: F): Boolean =
                     feqv(a, b)
         }
 
+        /**
+         * Retrieve an [Eq] that defines all instances as equal for type [F].
+         *
+         * @returns an [Eq] instance wherefore all instances of type [F] are equal.
+         */
         fun any(): Eq<Any?> = EqAny
```
raulraja: Do we really need this instance? It seems dangerous and once we have real type classes this is going to be candidate for all nullable types.

kategory-core/src/main/kotlin/kategory/typeclasses/Order.kt
```diff
@@ -1,32 +1,110 @@
...
     override fun eqv(a: F, b: F): Boolean = compare(a, b) == 0
 
+    /**
+     * Check if [a] is `lower than` [b]
+     *
+     * @param a object to compare with [b]
+     * @param b object to compare with [a]
+     * @returns true if [a] is `lower than` [b] and false otherwise
+     */
     fun lt(a: F, b: F): Boolean = compare(a, b) < 0
```
raulraja: should we add operator extension syntax for all these ops? `>`, `<` ....? https://kotlinlang.org/docs/reference/operator-overloading.html#operator-overloading

nomisRev: Operator overloading is only allowed for functions with 1 parameter. So currently there is no way to add syntax since we have to pass in `Order` as a parameter.

kategory-core/src/main/kotlin/kategory/typeclasses/Eq.kt
nomisRev: We are using it all over the place in testing etc. I think we can remove it but we need to go through whole codebase to check and remove it everywhere.

PR #310, opened 2017-09-24
Add wrapper for Map

victorg1991:
Fixes https://github.com/kategory/kategory/issues/254

I hope you like it :)

Issue #205, opened 2017-08-14
Create Wrapper class for SortedMap


Atternatt:

krishnarb3:
Hello, I would like to take this up. Can you explain what exactly is to be done, I am new to this project.

pakoito:
Hi @krishnarb3! Sure, the PR will be quite similar to the wrapper for regular maps. @victorg1991 completed that some weeks ago, so you can take a look at his job: https://github.com/kategory/kategory/pull/310/files

victorg1991:
@krishnarb3 If you have any question about it I'll be happy to answer it :D

krishnarb3:
Thanks both of you, I'm going through the operator and the PR. Will definitely ask you once I'm done 😁 @victorg1991

krishnarb3:
Also do we have an irc/slack channel/mailing list for kategory?

pakoito:
Yes, try https://gitter.im/kategory/Lobby!

Cotel:
@krishnarb3 Are you still working on this Issue? I would like to take it if you dont mind 😄

krishnarb3:
Sorry for the late response, i wasn't online for a while. 
Also amazing work on the PR, I'll look into some other issue that I can contribute to.
Cheers :D

pakoito:
Completed by #451 . Thanks @Cotel !!!

PR #451, opened 2017-11-07
Add wrapper for SortedMap

Cotel:
Attempt to solve #205 

It is almost the same code as in #310 except this time we are using `SortedMap`.

I'm trying to change the tests to be more like `ListKW` ones so this is still WIP 😄


kategory-core/src/main/kotlin/kategory/data/SortedMapKW.kt
```diff
@@ -0,0 +1,77 @@
...
+
+    fun <B> foldL(b: B, f: (B, A) -> B): B = this.map.values.fold(b, f)
+
+    fun <B> foldLeft(b: SortedMapKW<K, B>, f: (SortedMapKW<K, B>, Tuple2<K, A>) -> SortedMapKW<K, B>): SortedMapKW<K, B> =
+            this.map.foldLeft(b) { m: SortedMap<K, B>, (k, v) -> f(m.k(), Tuple2(k, v)) }.k()
+
+    fun <G, B> traverse(f: (A) -> HK<G, B>, GA: Applicative<G>): HK<G, SortedMapKW<K, B>> =
+        Foldable.iterateRight(this.map.iterator(), Eval.always { GA.pure(sortedMapOf<K, B>().k()) })({
+            kv, lbuf ->
+            GA.map2Eval(f(kv.value), lbuf) { (sortedMapOf(kv.key to it.a).k() + it.b).toSortedMap().k() }
```
pakoito: If we're going to sort it afterwards, is there any perf gain from using `(mapOf(kv.key to it.a).k() + it.b)` here instead?

kategory-core/src/main/kotlin/kategory/data/SortedMapKW.kt
```diff
@@ -0,0 +1,77 @@
...
+        }).value()
+
+    companion object
+}
+
+fun <K: Comparable<K>, A> SortedMap<K, A>.k(): SortedMapKW<K, A> = SortedMapKW(this)
+
+fun <K: Comparable<K>, A> Option<Tuple2<K, A>>.k(): SortedMapKW<K, A> = when (this) {
```
pakoito: I'd suggest using fold instead of when.

raulraja:
@Cotel just wanna say so far it's looking awesome, thanks so much for your contribution!

kategory-core/src/main/kotlin/kategory/data/SortedMapKW.kt
```diff
@@ -0,0 +1,77 @@
+package kategory
+
+import java.util.*
+
+@higherkind
+data class SortedMapKW<K: Comparable<K>, A>(val map: SortedMap<K, A>) : SortedMapKWKind<K, A>, SortedMap<K, A> by map {
```
raulraja: If we are going to follow here the `K, V` instead of `A, B` we should rename `A` for `V` so it denotes we refer to the value associated to a key otherwise use `A, B` which is the Kategory convention for applied types that are not in higher kind position.

kategory-core/src/test/kotlin/kategory/data/SortedMapKWTest.kt
```diff
@@ -0,0 +1,69 @@
...
+            functor<SortedMapKWHK>() shouldNotBe null
+            foldable<SortedMapKWHK>() shouldNotBe null
+            traverse<SortedMapKWHK>() shouldNotBe null
+            semigroup<SortedMapKWKind<String, Int>>() shouldNotBe null
+            monoid<SortedMapKWKind<String, Int>>() shouldNotBe null
+        }
+
+        val monoid = SortedMapKW.monoid<String, Int>(SG)
+
+        "Monoid Laws: identity" {
```
pakoito: This may be wrong in Map too. This should be tested using MonoidLaws.laws rather than in the file.

kategory-core/src/test/kotlin/kategory/data/SortedMapKWTest.kt
```diff
@@ -0,0 +1,69 @@
...
+        testLaws(TraverseLaws.laws(
+                SortedMapKW.traverse<String>(),
+                SortedMapKW.traverse<String>(),
+                { a: Int -> sortedMapOf("key" to a).k() },
+                EQ))
+
+    }
+
+
+}
```
pakoito: NIT: newline at the end of the file

pakoito:
@Cotel Weeee, thank you for the PR :D

kategory-core/src/test/kotlin/kategory/data/SortedMapKWTest.kt
victorg1991: @pakoito we only have MonoidkLaws :(

kategory-core/src/main/kotlin/kategory/data/SortedMapKW.kt
Cotel: I don't really know, but there's no need for the first map to be sorted if we are sorting it right after as you say. I'm changing this 👍

kategory-core/src/main/kotlin/kategory/data/SortedMapKW.kt
Cotel: I didn't know `A, B` was the convention. I will change 👌

Cotel:
Thank you for your feedback. I hope this to be the first of many contributions 😃

kategory-core/src/test/kotlin/kategory/data/SortedMapKWTest.kt
raulraja: We should then add Monoid and Semigroup laws before merging this. @Cotel wanna take care of those in this PR as well?

Cotel: I can try 💪

Cotel:
I've added an implementation for `MonoidLaws` and `SemigroupLaws`. But I am still wondering if `SortedMapKWTest` should test `MonoidKLaws` and `SemigroupKLaws` too 🤔

raulraja:
@Cotel SortedMap does not provide instances for `MonoidK` or `SemigroupK` so no need to test those. We need now a new issue ensuring all instances of `Monoid` and `Semigroup` provided by Kategory are being tested by the new laws. Awesome work!

PR #461, opened 2017-11-13
Fix #459 - rename apply to operator invoke

### LINKED ISSUE #459: Rename instances of apply to operator invoke
In some of the classes ported from cats, like Yoneda and Coyoneda, we're using Scala's apply instead of Kotlin's invoke.

Yoneda: https://github.com/kategory/kategory/search?utf8=%E2%9C%93&q=Yoneda.apply&type=

Coyoneda: https://github.com/kategory/kategory/blob/master/kategory-core/src/main/kotlin/kategory/free/Coyoneda.kt#L26-L28

voghDev:
Renamed `apply` function to `operator invoke` for `Yoneda` and `Coyoneda`. Also updated tests. Build is passing in my machine

There is another `unsafeApply` in `Coyoneda` that I haven't modified.

### DIFF
--- a/kategory-core/src/test/kotlin/kategory/free/CoYonedaTest.kt
+++ b/kategory-core/src/test/kotlin/kategory/free/CoYonedaTest.kt
@@ -18,7 +18,7 @@ class CoyonedaTest : UnitSpec() {
             functor<CoyonedaKindPartial<IdHK, Int>>() shouldNotBe null
         }
 
-        testLaws(FunctorLaws.laws(Coyoneda.functor(), { Coyoneda.apply(Id(0), { it }) }, EQ))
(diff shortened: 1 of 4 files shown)


kategory-core/src/test/kotlin/kategory/free/CoYonedaTest.kt
```diff
@@ -27,17 +27,17 @@ class CoyonedaTest : UnitSpec() {
                     if (n <= 0) acc
                     else loop(n - 1, acc.map { it + 1 })
 
-            val result = loop(loops, Coyoneda.apply(Some(0), { it })).lower(Option.functor())
+            val result = loop(loops, Coyoneda(Some(0), { it })).lower(Option.functor())
```
pakoito: You could also write `Coyoneda(Some(0)), { it }`

PR #467, opened 2017-11-17
(fix) Readme: Wrong URL to Maven repository for Kategory.

hastebrot:
Gradle had problems finding `kategory-effects` and `kategory-optics` artifacts.


pakoito:
I thought we'd merged this a while ago! Thanks, fixing now.

PR #468, opened 2017-11-17
Support for kotlinx.coroutines.Deferred

pakoito:
The last version we wrote was blocking on flatMap. I've wisened up, understood Coroutines better, and this version is as non-blocking as it gets :D


pakoito:
It's missing the docs. It shouldn't take long to fill them.

kategory-effects-kotlinx-coroutines/src/main/kotlin/kategory/effects/data/DeferredKW.kt
```diff
@@ -0,0 +1,106 @@
...
+
+fun <A> DeferredKWKind<A>.value(): Deferred<A> = this.ev().deferred
+
+@higherkind
+@deriving(
+        Functor::class,
+        Applicative::class,
+        Monad::class
+)
+data class DeferredKW<out A>(val deferred: Deferred<A>, val start: CoroutineStart = CoroutineStart.DEFAULT) : DeferredKWKind<A>, Deferred<A> by deferred {
```
raulraja: Why does this need to be a data class?  A simple class with an invoke would be more efficient since we don't need to compare structurally DeferredKW instances

pakoito: All datatypes are data classes as a convention, excepts for the ones that are sealed classes + data classes.

pakoito:
@raulraja Docs done. I'm not sure if I follow Semigroup, as IO and Rx don't have one :/

pakoito:
*muffled screams at ank*

PR #480, opened 2017-11-26
State API update

nomisRev:
As discussed in https://github.com/kategory/kategory/pull/457 update to State to improve API.

As you can see to the changes in the docs nothing broke of the existing API (`StateT.monad<IdHK, Stack>().binding` is also still valid).

Now you can specify return types as `State<S, A>` instead of `StateT<IdHK, S, A>` and you can use `State()` to access the methods that are defined on `StateT.Companion` with `F` fixed to `IdHK`.

If this is approved I will do the same for `Reader`

### DIFF
--- a/kategory-core/src/main/kotlin/kategory/data/State.kt
+++ b/kategory-core/src/main/kotlin/kategory/data/State.kt
@@ -1,7 +1,7 @@
 package kategory
 
 /**
- * Alias that represent stateful computation of the form `(S) -> Tuple2<S, A>`.
+ * Alias that represents stateful computation of the form `(S) -> Tuple2<S, A>`.
  */
 typealias StateFun<S, A> = StateTFun<IdHK, S, A>
 
--- a/kategory-core/src/main/kotlin/kategory/data/State.kt
+++ b/kategory-core/src/main/kotlin/kategory/data/State.kt
@@ -1,44 +1,111 @@
 package kategory
 
 /**
- * Convenience object to make working with `StateT<IdHK, S, A>` more elegant.
+ * Alias that represent stateful computation of the form `(S) -> Tuple2<S, A>`.
  */
-object State {
+typealias StateFun<S, A> = StateTFun<IdHK, S, A>
 
-    /**
-     * Construct a `StateT<IdHK, S, A>` from a function [run] `(S) -> Tuple2<S, A>`.
-     */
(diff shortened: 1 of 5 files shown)


kategory-core/src/main/kotlin/kategory/data/State.kt
```diff
@@ -1,72 +1,125 @@
 package kategory
 
 /**
- * Convenience object to make working with `StateT<IdHK, S, A>` more elegant.
+ * Alias that represent stateful computation of the form `(S) -> Tuple2<S, A>`.
```
pakoito: *represents

nomisRev:
@pakoito It currently is not showing them correctly. That's why I changed the `showInstances` type parameters.

https://github.com/kategory/kategory/pull/480/files#diff-eb40df2419c8c1dc5877441142d31222R118

https://github.com/kategory/kategory/pull/480/files#diff-6fec409c9abe386d4dad923a584b77f9R144

Since `State` takes multiple generics they need to be partially applied for the lookup to work.

kategory-core/src/main/kotlin/kategory/data/State.kt
```diff
@@ -1,72 +1,125 @@
...
+ */
+typealias State<S, A> = StateT<IdHK, S, A>
+
+/**
+ * Constructor for State.
+ * State<S, A> is an alias for IndexedStateT<IdHK, S, S, A>
+ *
+ * @param run the stateful function to wrap with [State].
+ */
+@Suppress("FunctionName")
```
raulraja: :clap:

kategory-core/src/main/kotlin/kategory/data/State.kt
```diff
@@ -1,72 +1,125 @@
...
+ *
+ * @param initial state to start stateful computation.
+ */
+fun <S, A> StateT<IdHK, S, A>.runS(initial: S): S = run(initial).a
+
+/**
+ * Alias for StateId to make working with `StateT<IdHK, S, A>` more elegant.
+ */
+@Suppress("FunctionName")
+fun State() = StateApi
```
raulraja: can this be declared as a `val` so we are able to do `State.monad()` vs `State().monad()`?

nomisRev: Sadly no. `object State` and `val State` both conflict with the `typealias`.

raulraja: 👍

PR #482, opened 2017-11-26
Add Google Analytics

### LINKED ISSUE #481: Add google analytics to kategory.io & docs
https://stackoverflow.com/a/43574656/1266326

https://github.com/jekyll/minima/blob/master/_includes/google-analytics.html

https://github.com/jekyll/minima/blob/master/_config.yml

wiyarmir:
Fixes #481 

GA id belongs to kategory.io gmail account.

Let's check whether events come in before merging. 

🔍 @pakoito @Guardiola31337


pakoito:
The event's won't come until the PR merges, and even then it takes 24h right?

pakoito:
@wiyarmir We need to check after merging whether it worked, because I cannot test jekyll to be production for our current implementation. Hope it is.

PR #487, opened 2017-11-27
Weak data type

pablisco:
This is work in progress but will be good if people have some early feedback.

The core idea of `Weak` is a data type that represents a value that can be lost at the time of usage. Backed by a `WeakReference` when we need to consume it we have to provide an alternative in case that the value has been collected by the GC.


kategory-core/src/main/kotlin/kategory/data/Weak.kt
```diff
@@ -0,0 +1,58 @@
...
+        fun <B> emptyWeak(): Weak<B> = EMPTY
+
+        operator fun <A> invoke(a: A): Weak<A> {
+            val reference = WeakReference(a)
+            return Weak { reference.get() }
+        }
+    }
+
+    fun <B> fold(fn: () -> B, f: (A) -> B): B
+        = provider()?.let(f) ?: fn()
```
pakoito: What happens if f returns a null? Does it still call fn()?

kategory-core/src/main/kotlin/kategory/data/Weak.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    inline fun <B> map(crossinline f: (A) -> B): Weak<B> = fold({ emptyWeak() }, { Weak(f(it)) })
```
pakoito: f(it).weak() for consistency

kategory-core/src/main/kotlin/kategory/data/Weak.kt
```diff
@@ -0,0 +1,58 @@
+package kategory
+
+import java.lang.ref.WeakReference
```
pakoito: This will not be portable, I have to keep an eye on these because we may have to move them around.

pakoito:
So far so good, I like it. Waiting to see tests and docs :D

kategory-core/src/main/kotlin/kategory/data/Weak.kt
pablisco: True, B should extend `Any` Do'h!

pakoito: No, no, it could still be `Any?`. Just use if instead of ternary operator.

pablisco: Ah! `provider()` is private so can't inline (may expose it, not sure). I'll use a `when` similarly to `Option`

kategory-core/src/main/kotlin/kategory/data/Weak.kt
```diff
@@ -0,0 +1,67 @@
...
+    inline fun <B> fold(fn: () -> B, f: (A) -> B): B = provider().let {
+        when (it) {
+            null -> fn()
+            else -> f(it)
+        }
+    }
+
+    inline fun <B> map(crossinline f: (A) -> B): Weak<B> = fold({ emptyWeak() }, { f(it).weak() })
+
+    inline fun <B> flatMap(crossinline f: (A) -> Weak<B>): Weak<B> = fold({ emptyWeak() }, { a -> f(a) })
```
pakoito: `crossinline f: (A) -> WeakKind<B>` and `{ a -> f(a).ev() }` should fix deriving.

arrow-data/src/main/kotlin/arrow/data/Weak.kt
```diff
@@ -0,0 +1,82 @@
...
+ * @param fallback provides a new value if we have lost the current one.
+ */
+fun <B> WeakKind<B>.getOrElse(fallback: () -> B): B = ev().fold({ fallback() }, { it })
+
+/**
+ * Returns this Weak instance if present or an alternative.
+ *
+ * @param fallback provides a new value if we have lost the current one.
+ */
+fun <A, B : A> WeakKind<B>.orElse(fallback: () -> Weak<B>): Weak<B> = ev().asOption().orNull()?.weak() ?: fallback()
```
pakoito: Can this be implemented with fold?

arrow-data/src/main/kotlin/arrow/data/Weak.kt
```diff
@@ -0,0 +1,82 @@
+package arrow.data
+
+import arrow.core.*
+import arrow.higherkind
+import java.lang.ref.WeakReference
```
pakoito: Could you please typealias WeakReference to WeakRef in a new object file called `Platform` inside a folder called internal. That way when we have to replace it it'll be more obvious.

pablisco:
I wonder if commenting will re-open the PR. If not I'll create a new one

PR #489, opened 2017-11-29
Bump Kotlin & Gradle

aballano:
What's new http://kotlinlang.org/docs/reference/whatsnew12.html

> * Array literals in annotations
> * Lateinit top-level properties and local variables
> * Checking whether a lateinit var is initialized
> * Inline functions with default functional parameters
> * Information from explicit casts is used for type inference
>     `val button = findViewById(R.id.button) as Button`
> * Smart cast improvements
> * Support for ::foo as a shorthand for this::foo


pakoito:
Can you also please bump Gradle?

wiyarmir:
Do we have to update this as well?
https://github.com/kategory/kategory/blob/master/kategory-docs/docs/docs/README.md

Or is it generated?

wiyarmir:
Forget my comment I remember now that is not generated 😄

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Mon Sep 11 16:05:40 EEST 2017
+#Wed Nov 29 11:44:15 CET 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.5-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.3-all.zip
```
wiyarmir: run `./gradlew wrapper` as well to update wrapper bins

aballano: I did that already, see the change above this file

PR #491, opened 2017-11-30
[WIP] Compile ank

nomisRev:
I enabled compilation of the `EitherT` documentation and it compiles. 🎉 
Currently all snippets get combined into one `kts` file and it gets compiled if one snippet has the compile flag present.

This closes https://github.com/kategory/kategory/issues/472


ank-core/src/main/kotlin/io/kategory/ank/ank.kt
```diff
@@ -6,6 +6,7 @@ import java.io.File
 
 const val AnkBlock = ":ank"
 const val AnkSilentBlock = ":ank:silent"
+const val AnkCompileBlock = ":ank:compile"
```
raulraja: we can reuse `ank:silent` or remove it in favor of `ank:compile` since essentially both produce the same output which is guaranteeing compilation and being added to the scope of the page scripts but not actually evaluating anything.

ank-core/src/main/kotlin/io/kategory/ank/interpreter.kt
```diff
@@ -142,17 +151,42 @@ fun compileCodeImpl(snippets: Map<File, ListKW<Snippet>>, classpath: ListKW<Stri
...
+
+                    tmpFile.printWriter().use {
+                        it.println(imports.joinToString("\n"))
+                        it.println(code.joinToString("\n"))
+                    }
+
+                    val args = listOf("-Xcoroutines=enable", "-cp", classPath, "-d", file.parentFile.absolutePath, tmpFile.absolutePath).toTypedArray()
+
+                    //If compilation fails compiler shuts down VM. However correct output is shown in Gradle.
+                    K2JVMCompiler.main(args)
```
raulraja: what happens if the code snippet is `java:ank:compile` ?

kategory-docs/docs/docs/datatypes/eithert/README.md
```diff
@@ -199,7 +199,7 @@ eitherTVal.value()
 
 So how would our function look if we implemented it with the EitherT monad transformer?
 
-```kotlin
+```kotlin:ank:compile
```
raulraja: :clap:

raulraja:
Added some concerns but overall this looks awesome!

ank-core/src/main/kotlin/io/kategory/ank/ank.kt
nomisRev: `silent` gets evaluated by `ScriptEngine` which means you can reference it in a regular `ank` blocks. Whatever is defined in `compile` cannot be referenced from a regular `ank` block.

ank-core/src/main/kotlin/io/kategory/ank/interpreter.kt
nomisRev: Need to check!

ank-core/src/main/kotlin/io/kategory/ank/ank.kt
raulraja: makes sense, let's keep them separate then.

ank-core/src/main/kotlin/io/kategory/ank/interpreter.kt
```diff
@@ -142,17 +151,56 @@ fun compileCodeImpl(snippets: Map<File, ListKW<Snippet>>, classpath: ListKW<Stri
                                     it.lang to seManager.getEngineByExtension(extensionMappings.getOrDefault(it.lang, "kts"))
                                 }
                                 .toMap()
+
+                if (codeBlocks.find { it.compile }.nonEmpty()) {
+                    codeBlocks.filter { it.lang != "kotlin" }
+                            .find { it.compile }
+                            ?.let {
+                                throw CompilationException(file, it, IllegalStateException("```${it.lang}:ank:compile is not supported. Use ```${it.lang}:ank or ```${it.lang}:ank:silent"))
```
nomisRev: I throw a `CompilationException` here but Gradle seems to get stuck instead of terminating.

raulraja: This is because you are within an async boundary not blocking. If you block the result of `async` it will blow up in the gradle thread.

raulraja:
@nomisRev is this still relevant?, if so it needs to get updated to the latest arrow structure in master

raulraja:
@nomisRev, closing this since Ank is no longer hosted in this repo

PR #503, opened 2017-12-10
Add integration layer with KindedJ

### LINKED ISSUE #218: Document interop layer with KindedJ

pakoito:
Closes #218


kategory-docs/docs/docs/integrations/kindedj/README.md
```diff
@@ -0,0 +1,82 @@
...
+```kotlin
+val idj = IdJ(1)
+
+val id: HK2<ConvertHK, IdJHK, A> = idj.fromKindedJ()
+```
+
+and convert it back using `convert()`:
+
+```kotlin
+val idj2: io.kindedj.Hk<IdJHK, A> = id.convert()
```
raulraja: This identifier is too broad in meaning and should be instead something like `toKindedJ()` that denotes the direction of the conversion with more clear semantics.

kategory-kindedj/src/test/java/kategory/kindedj/KategoryShow.kt
```diff
@@ -0,0 +1,7 @@
+package kategory.kindedj
+
+import kategory.HK as HK_K
+
+interface KategoryShow<in F> {
```
raulraja: We actually need the `Show` type class in case you want to leave this in non test sources and add an issue to create the instances down the road.

pakoito:
Ticket for Show in #504

PR #509, opened 2017-12-12
Merging Funktionale into Kategory repository

FPerezP:
This PR merges Funktionale repository into Kategory one keeping git history from both repositories.

*Conflict resolutions:*
_.gitignore_
I removed the Funktionale

_README.md_
Content from both repos have been kept.


pakoito:
If it still fails after the merge I can take a look tonight.

JorgeCastilloPrz:
Is this the idea ? To bring in Funktionale modules along with the Kategory ones to eventually merge all of them together? or how is it gonna work?

pakoito:
@JorgeCastilloPrz the first step is to import the files & the full story. Then we're going on lockdown & refactor until we're ready. So a bunch of fun stuff to do this xmas if you'd like to join along!

JorgeCastilloPrz:
Sounds good :+1:

pakoito:
That "tonight" may become "this weekend" :p

PR #510, opened 2017-12-12
Make IO#Map's max fusion stack size configurable

pakoito:
Optimize map fusion following this commit and the next afterwards: https://github.com/typelevel/cats-effect/pull/95/commits/174ec5430351affb5e7f6958427ff05c6a440554

Note that enviro variables aren't really a thing in Android, so we'll opt for Kotlin's multiplatform config or another similar mechanism.


raulraja:
@pakoito can you update your PR with latest from master so we can merge it? thanks!

pakoito:
I thought this got merged already 🙈

PR #517, opened 2017-12-20
Toward Arrow: Arrow renaming and funKTionale `Option` integration

raulraja:
This is a massive refactor that changes `kategory` -> `arrow` and integrates funktionale `Option` fushioning it with `kategory.Option` into `arrow.Option` and it's syntax facilities.
Relevant files are in https://github.com/kategory/arrow/commit/267ed2da650d8c0e5d33507b8ffc8da335777782


MarioAriasC:
Fixed

PR #518, opened 2017-12-20
Update Kotlin to 1.2.10 and use the correct stdlib artifact for 1.2

kevinherron:


pakoito:
We need @raulraja approval before merging, as we're in the middle of a large refactor. It should be okay tho :D

raulraja:
@kevinherron can you update your PR with latest from master so we can merge it? thanks!

kevinherron:
Updated 👍

raulraja:
@kevinherron thanks for your contribution!

PR #519, opened 2017-12-20
Introduce Try#getOrElse((Throwable)->B) and rename the old impl getOrDefault

kevinherron:
These extension functions can't exist under the same name because of type erasure.

@pakoito suggested in Slack that renaming the existing signature to `getOrDefault` and using `getOrElse` for the new signature was okay.


kategory-docs/docs/docs/datatypes/try/README.md
```diff
@@ -63,12 +63,26 @@ val lotteryTry = Try { getLotteryNumbers() }
...
+lotteryTry.getOrElse { ex: Throwable -> emptyList() }
+```
+
+`getOrElse` can generally be used anywhere `getOrDefault` is used, ignoring the exception if it's not needed:
 
 ```kotlin:ank
 lotteryTry.getOrElse { emptyList() }
 ```
 
+
```
pakoito: Unnecessary additional lines

pakoito:
We cannot currently merge as @raulraja is going through a large refactor and we'll need to integrate it later!

raulraja:

kevinherron:
Updated 👍

PR #520, opened 2017-12-22
fix gradle deprecation warnings

epeee:
- The ConfigurableReport.setDestination(Object) method has been deprecated and is scheduled to be removed in Gradle 5.0. Please use the method ConfigurableReport.setDestination(File) instead.
   The Task.leftShift(Closure) method has been deprecated and is scheduled to be removed in Gradle 5.0. Please use Task.doLast(Action) instead.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -71,7 +71,7 @@ subprojects { project ->
     task codeCoverageReport(type: JacocoReport) {
         reports {
             xml.enabled true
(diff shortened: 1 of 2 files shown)


pakoito:
Wow, great catch. I didn't even know how to fix it. Thank you very much. We'll merge it shortly, we have a large refactor ongoing :D

wiyarmir:
My bad on the left shift, I somehow interpreted the depreciation the other way around :smile:

PR #524, opened 2017-12-26
Remove extra '_docs/' for broken monad link

ersin-ertan:


raulraja:

PR #525, opened 2017-12-26
Toward Arrow: funktionale.try,... -> arrow.data.Try

raulraja:
This PR fusions funktionale `Try` with arrow `Try`


arrow-data/src/main/kotlin/arrow/data/Try.kt
```diff
@@ -240,15 +246,71 @@ sealed class Try<out A> : TryKind<A> {
...
+    @Deprecated(DeprecatedUnsafeAccess, ReplaceWith("getOrElse { ifEmpty }"))
+    abstract fun get(): A
+
+    @Deprecated(DeprecatedUnsafeAccess, ReplaceWith("map { body(it); it }"))
+    fun onSuccess(body: (A) -> Unit): Try<A> {
+        foreach(body)
+        return this
+    }
+
+    @Deprecated(DeprecatedUnsafeAccess, ReplaceWith("fold ({ Try { body(it); it }}, { Try.pure(it) })"))
```
MarioAriasC: Technically true, but hard to remember

raulraja: We are gonna move them to an `unsafe` syntax package if it makes sense so even if deprecated people can still use them if they want to. The `DeprecatedUnsafeAccess` constant has that in the description now with instructions to use the new package.

PR #528, opened 2017-12-27
Toward Arrow: funktionale.validation -> arrow.validation

raulraja:
Migrates funktionale validation module to arrow.


arrow-data/src/main/kotlin/arrow/data/Either.kt
```diff
@@ -27,6 +27,10 @@ typealias Left<A, B> = Either.Left<A, B>
      */
     internal abstract val isLeft: Boolean
 
+    fun isLeft(): Boolean = isLeft
```
pakoito: Not a fan, these are kept private just for perf purposes.

raulraja: As part of the funKTionale migration we are only deprecating api's from funkTionale that constains unsafe access such as `Option#get`, `Try#get`, Either non biased projections etc. We can reconsider these after the first release of arrow which is ment to be mostly binary compatible except for packages with funKTionale. If you are suggesting we mark this as deprecated that's something I'd agree with but @MarioAriasC would need to decide since it could impact potentially funKTionale user base and IMO it would slow down people moving from funKTionale to Arrow if most non FP codebases rely on the shape of this combinators that encourage boolean checks and statements over expressions.

PR #529, opened 2017-12-27
Update README.md

ersin-ertan:
- Undo yelling
- Add info about including the type parameter for Either.Left calls within examples on Line 228
- Add functor, applicative, monad examples

Could not get the monad example with left to work
```kotlin
Either.monad<Int>().binding {
   val a = Either.Left(1).bind()
   val b = Either.Right(1 + a).bind()
   val c = Either.Right(1 + b).bind()
   yields(a + b + c)
}
```


pakoito:
YELLING IS GREAT

The example will work, although quite slowly, because the scripting engine isn't a friend of coroutines.

ersin-ertan:
The problem wasn't with the coroutine, but rather the return type of `Either.Left(1).bind()` is `Nothing` thus unreachable code.

raulraja:

PR #544, opened 2018-01-02
Bintray org ref fixes

raulraja:


build.gradle
```diff
@@ -33,14 +33,14 @@ buildscript {
...
-        maven { url "https://dl.bintray.com/arrow/arrow/" }
+        maven { url "https://dl.bintray.com/arrow-kt/arrow-kt/" }
     }
 
     dependencies {
         classpath "com.github.ben-manes:gradle-versions-plugin:$gradleVersionsPluginVersion"
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.0'
-        classpath('io.arrow:ank-gradle-plugin:0.1.7')
+        classpath 'io.arrow:ank-gradle-plugin:0.1.7'
```
pakoito: Did we update the plugin with this release?

raulraja: yes but we are not sure if it's gonna work because that was uploading to a different org. Still need to figure that out.

raulraja: gonna push to try in master and will fix accordingly

PR #546, opened 2018-01-03
WIP An improved encoding for ListKW style wrappers

raulraja:
Notice this forwards automatically a lot more methods returning the proper type signature than what we have now.

Thoughts?

### DIFF
--- a/arrow-data/src/main/kotlin/arrow/data/ListA.kt
+++ b/arrow-data/src/main/kotlin/arrow/data/ListA.kt
@@ -9,7 +9,7 @@ import arrow.higherkind
 import arrow.typeclasses.Applicative
 
 @higherkind
-class ListA<A>(val value: List<A>): ArrayList<A>(value), ListAKind<A> {
+class ListA<A>(val value: List<A>) : ArrayList<A>(value), ListAKind<A> {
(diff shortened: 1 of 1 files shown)


pakoito:
Making it an ArrayList allows using mutable methods on it. If we'd like to have a private defensive copy of the list we need a constructor to do that, and make ListA only extend List.

raulraja:
@pakoito commits welcome! :kissing_heart:

pakoito:
I'm code-free until I'm back in London next week. Doctor's orders!

raulraja:
there is a bunch of :hankey: to implement when you just do `List`, tried that.

pakoito:
@raulraja use delegation with `by` which is what ListKW does. It just forwards the methods. Make a local copy instead of having the val property and it should be ready to go :D

raulraja:
delegation forwards to the internal value type, that's what we are trying yo avoid. I want the combinators to return `ListA<?>` instead of `List<?>`

nomisRev:
@raulraja close?

JorgeCastilloPrz:
In an effort to clean the PR list we've decided to close all the ones that are older than 2 months and consider them obsolete except for very specific circumstances that we will evaluate per PR. This one doesn't apply any more given the new encoding for Kinds that is coming with meta.

PR #555, opened 2018-01-07
Improve Eval.Compute.value performance

aedans:
Currently, Eval.Compute.value is implemented using a mutable reference to an immutable list. This PR replaces it with a mutable reference to a mutable list, which greatly increases performance.

Tested using the following factorial implementation:
```kotlin
fun factorial(i: Int): Eval<Int> = when (i) {
    0 -> Eval.One
    else -> Eval.defer { factorial(i - 1) }.map { it * i }
}
```

With the old implementation, factorial 0..2000 takes ~13 seconds on my laptop. With the new implementation, it takes ~0.5 seconds.


pakoito:
Great finding, I completely missed when porting from Scala that they have persistent collections -.-

PR #561, opened 2018-01-09
Adding utility method to Duration

enhan:
Overloading the * operator to multiply a duration. Not sure it is really relevant to have it in Arrow though.

### DIFF
--- a/arrow-effects/src/test/kotlin/arrow/effects/data/DurationTest.kt
+++ b/arrow-effects/src/test/kotlin/arrow/effects/data/DurationTest.kt
@@ -1,0 +1,36 @@
+package arrow.effects.data
+
+import arrow.effects.Duration
+import arrow.test.UnitSpec
+import io.kotlintest.KTestJUnitRunner
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+import org.junit.runner.RunWith
+import java.util.concurrent.TimeUnit
+
+/**
+ *
+ */
+@RunWith(KTestJUnitRunner::class)
+class DurationTest : UnitSpec() {
+
+    init {
+        "plus should be commutative" {
+            forAll(Gen.long(), TimeUnitGen(), Gen.long(), TimeUnitGen()) { i: Long, u: TimeUnit, j: Long, v: TimeUnit ->
+                val a = Duration(i, u)
+                val b = Duration(j, v)
+                a + b == b + a
+            }
+        }
+    }
+
+}
+
+private class TimeUnitGen : Gen<TimeUnit> {
+
+    companion object {
+        val units = TimeUnit.values()
+    }
+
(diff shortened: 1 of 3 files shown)


pakoito:
If you're adding *, please add + too

arrow-effects/src/test/kotlin/arrow/effects/data/DurationTest.kt
```diff
@@ -0,0 +1,36 @@
...
+                val a = Duration(i, u)
+                val b = Duration(j, v)
+                a + b == b + a
+            }
+        }
+    }
+
+}
+
+private class TimeUnitGen : Gen<TimeUnit> {
```
pakoito: Can you please move this to: https://github.com/arrow-kt/arrow/blob/master/arrow-test/src/main/kotlin/arrow/test/generators/Generators.kt

arrow-test/src/main/kotlin/arrow/test/generators/Generators.kt
```diff
@@ -146,4 +147,13 @@ fun <K, V> genMap(genK: Gen<K>, genV: Gen<V>): Gen<Map<K, V>> =
...
\ No newline at end of file
+        Gen.create { Gen.list(genK).generate().map { it to genV.generate() }.toMap().k() }
+
+class TimeUnitGen : Gen<TimeUnit> {
+
+    companion object {
+        val units = TimeUnit.values()
+    }
+
+    override fun generate(): TimeUnit = units[Gen.choose(0, units.size -1).generate()]
```
pakoito: A new Gen.choose() is created on every generate call. This makes the seed to be reset every time, which messes with the entropy. You should have it as a field of the class instead.

arrow-test/src/main/kotlin/arrow/test/generators/Generators.kt
```diff
@@ -146,4 +147,13 @@ fun <K, V> genMap(genK: Gen<K>, genV: Gen<V>): Gen<Map<K, V>> =
         Gen.create { Gen.list(genK).generate().map { it to genV.generate() }.toMap() }
 
 fun <K, V> genMapKW(genK: Gen<K>, genV: Gen<V>): Gen<MapKW<K, V>> =
-        Gen.create { Gen.list(genK).generate().map { it to genV.generate() }.toMap().k() }
\ No newline at end of file
+        Gen.create { Gen.list(genK).generate().map { it to genV.generate() }.toMap().k() }
+
+class TimeUnitGen : Gen<TimeUnit> {
```
pakoito: Following the code style of the other generators, instead of a class this should be a function genTimeUnit() that returns an anonymous `object: Gen<TimeUnit>`

pakoito:
Almost there :D I'll merge as soon as the changes are introduced!

enhan:
@pakoito thank you for your insightful comments. Arrow is making me learn Kotlin efficiently, great lib.

PR #562, opened 2018-01-12
Include package name for generated lens .kt files

frenchdonuts:
So we can generate lenses for classes with the same name in different packages.


nomisRev:
@frenchdonuts great contribution thank you!

raulraja:
@frenchdonuts 👏  
We should perhaps consider applying this same technique across all other file generators for the different annotations.

PR #565, opened 2018-01-12
Include package name for generated isos and prisms .kt files

frenchdonuts:


raulraja:
I'll review and merge these since they need come after https://github.com/arrow-kt/arrow/pull/564 once Bintray grants us the fwd to jcenter

Issue #504, opened 2017-12-10
Add Show typeclass


pakoito:
Entry level ticket for the Show typeclass. Show has a single method that does the same as Java's `toString()`. We need to define the typeclass, laws, and add a default instance similar to `Eq.any()`.

Cotel:
I'm taking this 😄

pakoito:
Boom! Fukken done. Thanks @Cotel

PR #569, opened 2018-01-13
Adding Show typeclass

Cotel:
Trying to solve #504 

I'm still figuring out which laws should Show follow 🤔 Any help is welcome.


pakoito:
Check that two separate instances that are `Eq` return the same `Show`.

pakoito:
I've added more information to the previous comment, feel free to check it!

arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Show.kt
```diff
@@ -0,0 +1,55 @@
...
+
+import arrow.*
+
+
+/**
+ * A type class used to get a textual representation for an instance of type [F] in a type safe way.
+ *
+ */
+@typeclass
+interface Show<in F> : TC {
```
raulraja: The convention in type params here is to use `A` instead of `F` since this type param referes to a monomorphic applied type and not to an unapplied higher kinded type constructor

arrow-data/src/test/kotlin/arrow/data/IdTest.kt
```diff
@@ -22,10 +19,12 @@ class IdTest : UnitSpec() {
             foldable<IdKind<Int>>() shouldNotBe null
             traverse<IdKind<Int>>() shouldNotBe null
             eq<Id<Int>>() shouldNotBe null
+            show<Id<Int>>() shouldNotBe null
         }
 
         testLaws(
             EqLaws.laws { Id(it) },
+            ShowLaws.laws { it },
```
pakoito: This tests `Show<Int>`.  The correct form is ShowLaws.laws { Id(it) }.

pakoito:
Could you please add ShowLaws tests for all Numbers?

arrow-test/src/main/kotlin/arrow/test/laws/ShowLaws.kt
```diff
@@ -0,0 +1,22 @@
...
+
+import arrow.typeclasses.Show
+import arrow.typeclasses.show
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+
+object ShowLaws {
+
+    inline fun <reified F> laws(S: Show<F> = show<F>(), noinline cf: (Int) -> F): List<Law> =
+            listOf(
```
pakoito: Add a dependency on Eq laws I believe.

arrow-test/src/main/kotlin/arrow/test/laws/ShowLaws.kt
```diff
@@ -0,0 +1,22 @@
...
+                    Law("Show Laws: equality", { equalShow(S, cf) })
+            )
+
+    inline fun <reified F> equalShow(S: Show<F> = show(), crossinline cf: (Int) -> F): Unit =
+            forAll(Gen.int(), { int: Int ->
+                val a = cf(int)
+                val b = cf(int)
+                S.show(a) == S.show(b)
```
pakoito: Check with the Eq instance that both a and b are also equal.

Cotel: Is it okay to place an `if` there? Like:

```kotlin
forAll(Gen.int(), { int: Int ->
  val a = cf(int)
  val b = cf(int)
  if (EQ.eqv(a,b) {
    S.show(a) == S.show(b)
  }
})
```

pakoito: Just return `EQ.eqv(a,b) && S.show(a) == S.show(b)`. The test has to fail for differences too.

arrow-test/src/main/kotlin/arrow/test/laws/ShowLaws.kt
```diff
@@ -1,22 +1,23 @@
...
+            EqLaws.laws(eq<F>(), cf) + listOf(
                     Law("Show Laws: equality", { equalShow(S, cf) })
             )
 
     inline fun <reified F> equalShow(S: Show<F> = show(), crossinline cf: (Int) -> F): Unit =
             forAll(Gen.int(), { int: Int ->
                 val a = cf(int)
                 val b = cf(int)
-                S.show(a) == S.show(b)
+                eq<F>().eqv(a, b) && S.show(a) == S.show(b)
```
pakoito: This eq has to come as a parameter to the laws.

arrow-test/src/main/kotlin/arrow/test/laws/ShowLaws.kt
```diff
@@ -1,22 +1,23 @@
...
+import arrow.typeclasses.eq
 import arrow.typeclasses.show
 import io.kotlintest.properties.Gen
 import io.kotlintest.properties.forAll
 
 object ShowLaws {
 
     inline fun <reified F> laws(S: Show<F> = show<F>(), noinline cf: (Int) -> F): List<Law> =
-            listOf(
+            EqLaws.laws(eq<F>(), cf) + listOf(
```
pakoito: This eq has to come as a parameter to the laws, as we'd like for users to override it.

arrow-data/src/test/kotlin/arrow/data/OptionTest.kt
```diff
@@ -58,6 +60,7 @@ class OptionTest : UnitSpec() {
 
         testLaws(
                 EqLaws.laws(eq(), { genOption(Gen.int()).generate() }),
+                ShowLaws.laws(show(), eq(), { it }),
```
pakoito: You're testing Int here! Use it.some() instead.

Cotel: I always forget this 😅

Cotel:
Tests for `Either` aren't working because of a StackOverflow. Do you know what is causing the failure?

pakoito:
@Cotel ¿?¿?¿ no, I'm confused

Cotel:
When running tests either locally or by travis I get this:

```
arrow.data.EitherTest > initializationError FAILED
    java.lang.StackOverflowError
```

This happens when I add my `showLaws` to the tests for `Either` but I have no idea of what is happening 😅

pakoito:
You can see better traces it running with the -i or --trace flags. Else you can just debug it using --debug-jvm

If not, start by removing lines from the file and see which one causes the error. Good chance it's the Eq law, so you'll need to see the stacktrace :D

Cotel:
Solved. I just had to write explicitly the types for `eq` and `show` in `showLaws`.

Should I implement Show instances for every data type or is there any one which should be excluded?

arrow-data/src/test/kotlin/arrow/data/EitherTest.kt
```diff
@@ -32,7 +32,7 @@ class EitherTest : UnitSpec() {
 
         testLaws(
             EqLaws.laws(eq<Either<String, Int>>(), { Right(it) }),
-            ShowLaws.laws { Right(it) },
+            ShowLaws.laws(show<Either<String, Int>>(), eq<Either<String, Int>>(), { Right(it) }),
```
pakoito: Either.show() and Either.eq() have to be shorter :D

pakoito:
@Cotel Exclude the ones without clear identities. Those are the ones that can suspend functions: IO, Yoneda, Coyoneda, Observable, Deferred, Flowable and Eval.

arrow-data/src/test/kotlin/arrow/data/EitherTest.kt
Cotel: What do you mean?

pakoito:
Keep up the good work :D

Cotel:
I don't know if I am missing some data type but I think this is it 😁

pakoito:
@Cotel add Const and the Tuple2-9 ones, and we're good :D

pakoito:
If you'd like to round it up, add it to the docs. You'll need: 

* a new README.md in a folder arrow-docs/docs/docs/typeclasses/show
* a new short description in typeclasses/general here: https://github.com/arrow-kt/arrow/edit/master/arrow-docs/docs/docs/typeclasses/intro/README.md
* and a new menu entry above Eq: https://github.com/arrow-kt/arrow/blob/master/arrow-docs/docs/_data/menu.yml#L132

Cotel:
I've seen in the Eq docs that there are some extension functions in `syntax`. Should I do something similar for Show?

pakoito:
If you'd like to have for eq and eqv, yes. I believe @raulraja is looking into autogenerating them tho :)

pakoito:
You got caught in the middle of a refactor BTW, apologies for the inconveniences, specially after all the good work you've done :D  

Everything moved to a new folder, and there are some renames you'll have to take into account: https://github.com/arrow-kt/arrow/pull/671

Cotel:
Merge done, I hope nothing fails 🤞

PR #570, opened 2018-01-13
Use newThread scheduler to avoid flaky tests on single-core CI machines

### LINKED ISSUE #512: "ObservableKWTest.Multi-thread Observables should run on their required threads" is flaky
kategory.effects.ObservableKWTest > ObservableKWTest.Multi-thread Observables should run on their required threads FAILED

pakoito:
Fix #512


pakoito:
That's 12 runs without failing. Let's merge.

PR #572, opened 2018-01-14
Safe onSuccess and onFailure functions

MarioAriasC:
The functions `onSuccess` and `onFailure` are present on many computation abstractions such as `Deferred`, `Promises` and `Observables.`

This PR changes current implementation for a safe one and deletes the `Deprecated` annotation


arrow-data/src/main/kotlin/arrow/data/Try.kt
```diff
@@ -119,20 +120,16 @@ sealed class Try<out A> : TryKind<A> {
     @Deprecated(DeprecatedUnsafeAccess, ReplaceWith("getOrElse { ifEmpty }"))
     abstract fun get(): A
 
-    @Deprecated(DeprecatedUnsafeAccess, ReplaceWith("map { body(it); it }"))
-    fun onSuccess(body: (A) -> Unit): Try<A> {
-        foreach(body)
-        return this
+    fun onSuccess(body: (A) -> Unit): Try<A> = map {
```
raulraja: Given `body` is not actually captured and evaluating it may throw an exception I'd recommend we change the implementation to something like:
```kotlin
fun onSuccess(body: (A) -> Unit): Try<A> = 
  flatMap { Try { body(it); it } }
```
This ensures that if `body` is ill behaved any thrown exceptions would be represented as a `Failure` and this function will not blow up at run-time. Here is an example that would make the current implementation fail with an exception:
```kotlin
Try { "x "}.onSuccess { println(it.toInt()) }
```
In my opinion these functions belong in `import arrow.syntax.try.unsafe.*` because the function itself is not referentially transparent because for different inputs it produces the same observable output and hidden chain of side effects.

sbuettner:
Any updates on this?

JorgeCastilloPrz:
In an effort to clean the PR list we've decided to close all the ones that are older than 2 months and consider them obsolete except for very specific circumstances that we will evaluate per PR. In this case, `Try` got deprecated and will be out in further versions, so these extensions wouldn't apply anymore.

PR #578, opened 2018-01-15
remove private val dummy from Left and Right

robstoll:
I removed the private property dummy from Left and Right, I do not really see why they would be required.

### DIFF
--- a/arrow-core/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core/src/main/kotlin/arrow/core/Either.kt
@@ -151,7 +151,7 @@ import arrow.legacy.*
      * The left side of the disjoint union, as opposed to the [Right] side.
      */
     @Suppress("DataClassPrivateConstructor")
-    data class Left<out A, out B>  @PublishedApi internal constructor(val a: A) : Either<A, B>() {
+    data class Left<out A, out B> @PublishedApi internal constructor(val a: A) : Either<A, B>() {
         override val isLeft = true
         override val isRight = false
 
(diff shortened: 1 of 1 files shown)


pakoito:
>I do not really see why

I see you're fighting the compiler MUA HA HA HA 😈 

But seriously, if you arrive to a good solution we'll take it :D

pakoito:
Detekt is our static analyzer. It's a nagger haha You have a couple of extra spaces on these lines:

	MultipleSpaces - [Left] at arrow-core/src/main/kotlin/arrow/core/Either.kt:154:35
	MultipleSpaces - [Right] at arrow-core/src/main/kotlin/arrow/core/Either.kt:167:35

robstoll:
I am doing the change on the web. I figured I am not going to checkout the project just for such a small change (and fighting :dagger:  the compiler wouldn't be fun anymore :grin:).
Happy with the current solution (I'll clean up the spaces of course)?

pakoito:
If it works, sure. It was a real problem with ambiguity inside and outside the package, so we use nulls everywhere. If it compilers & works let's go ahead with it.

robstoll:
Well, the current solution wouldn't help with internal problems but the user shouldn't notice anything. I am not fussed if you do not merge the patch in the end. I understand that it would be painfull to add the generics everywhere.

pakoito:
Yeah we don't want that, no. Too big of a burden. Some of the types are quite unreasonable, and the IDE hinting doesn't always collaborate.

robstoll:
let us see how many errors we get. if there are only a few we could still consider to fix them and merge it, otherwise I close the pull request

pakoito:
It's in the tests so it looks like it compiles 👍

robstoll:
It's even green, surprise surprise, haha

pakoito:
Well, cheers @robstoll 

Feel free to take a look at other instances of dummy around. They're easy to find :D

PR #580, opened 2018-01-16
Remove or refactor to instances of java.utils imports

pakoito:
So far the plan is to keep JVM imports just inside the effects and the tests, as both depend on Java libraries and constructs anyway. For the others we can provide Platform objects that can be later refactored to use JS or native equivalents. It's all going to be messy, but at least we have a headstart.


arrow-syntax/src/main/kotlin/arrow/syntax/function/memoization.kt
```diff
@@ -214,11 +214,9 @@ private data class MemoizeKey22<out P1, out P2, out P3, out P4, out P5, out P6,
 }
 
 private class MemoizedHandler<F, in K : MemoizedCall<F, R>, out R>(val f: F) {
-    private val m = ConcurrentHashMap<K, R>()
+    private val m = newConcurrentMap<K, R>()
     operator fun invoke(k: K): R = m[k] ?: run({
-        val r = k(f)
-        m.putIfAbsent(k, r)
-        r
+        m[k].let { it ?: k(f).also { m[k] = it } }
```
pakoito: @MarioArias I had to change the logic not to use `putIfAbsent`. This gets a nullable, and if it doesn't exist it executes the function and returns it. Is it the case that the function should always be called?

MarioAriasC: We lost thread safety which is of paramount importance in this case; we must use `putIfAbsent` or expose any other equivalent function via a proxy

pakoito: Okay, lemme proxy it!

pakoito: It's even cooler now :D

PR #583, opened 2018-01-17
Adding a cond method on Either companion object

enhan:
This mimic the [Scala stdlib](https://www.scala-lang.org/api/current/scala/util/Either$.html#cond[A,B](test:Boolean,right:=%3EB,left:=%3EA):scala.util.Either[A,B]). 
Shall we make the t parameter a thunk too ?

### DIFF
--- a/arrow-core/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core/src/main/kotlin/arrow/core/Either.kt
@@ -193,6 +193,8 @@ import arrow.legacy.*
             }
         }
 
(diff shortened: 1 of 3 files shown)


raulraja:
@enhan Since the parameter is always evaluated what would be the value of having it expressed as `() -> Boolean`?. IMO this is good as it stands but if you have a compeling reason as to why it should be passed as a function we can change it.
Great stuff and thanks for the contribution!

raulraja:
@enhan I'm gonna merge this, feel free to send another PR if you think it should be passed as a function.

enhan:
@raulraja let's keep it as is. I see no reason to make the t a thunk.

PR #591, opened 2018-01-20
[WIP] Execute suspended monads created using bindIn in the corresponding CoroutineContext

pakoito:
This diff attempts to fix a gotcha in effects, the one described here [Gotchas] http://arrow-kt.io/docs/patterns/monad_comprehensions/


pakoito:
Taking it step by step instead. For now a refactor was needed and done in #594

PR #596, opened 2018-01-21
[at]optionals annotation

jereksel:
An idea for [at]optional annotation that would generate Optional optic for nullable and Option members of data class.

For now it's just PoC and I'm wondering if that would be useful.


pakoito:
I like the idea and the principle. Wouldn't it be better if it was generated off the [at]lenses annotation?

raulraja:
@pakoito I was also wondering already if `@lenses` was already generating Optional instances. I think there is a point also in an extra annotation because `Optional` are their own type of Optics and I believe they fall under Prisms since they operate over something that either is there or is not. Perhaps the @prisms annotation when placed over data classes that contain Nullable `?` types or `Option` properties could generate these.

raulraja:
@nomisRev can these `Optional` instances be expressed polymorphically over `Option<A>` and `?A` for all types so the codegen phase just delegates to `optionOptional()` and `nullableOptional()`?

jereksel:
I've tried writing Optional for A?, but I failed: https://gitter.im/arrow-kt/Lobby?at=5a63c5715a9ebe4f75b25281

nomisRev:
> @nomisRev can these Optional instances be expressed polymorphically over Option<A> and ?A for all types so the codegen phase just delegates to optionOptional() and nullableOptional()?

@raulraja I am not sure I follow you.

Currently a `Lens` would be generated `Lens<A, B?>` or `Lens<A, Option<B>>`.

raulraja:
@nomisRev makes sense. I was thinking similar to how the Semigroup and Monoid instance is provided us as a long as there is `Semigroup<A>> or `Monoid<A>` if the case with Optional for `?` and `Option` was similar and we could just provide those instances in `arrow-optics`

nomisRev:
@jereksel I looked at the `Optional` you wrote for `A?` and it is indeed incorrect. When an `Optional` cannot find its focus it can also not set a new value.

i.e. `listHead<String>().set(emptyList(), "A")` a new head cannot be set if no head is present.

```
fun <A> nullableOptional(): Optional<A?, A> = Optional(
        getOrModify = { a -> a?.right() ?: a.left() },
        set = { a -> { if (it != null) a else null } }
)
```

Feel free to add a PR to add this instance to optics and ask help if you bump into any issues. While looking into this I also found some bugs in the Laws. I will fix them tonight after work.

arrow-annotations-processor/src/main/java/arrow/optics/OptionalFileGenerator.kt
```diff
@@ -0,0 +1,71 @@
...
+                }
+            }
+
+    fun fileHeader(packageName: String): String =
+            """package $packageName
+               |
+               |import arrow.syntax.either.*
+               |import arrow.syntax.option.toOption
+               |
+               |""".trimMargin()
```
nomisRev: Small nitpick. Can you remove these 2 enters?

arrow-annotations/src/main/java/arrow/optics.kt
```diff
@@ -14,3 +14,7 @@ annotation class prisms
 @Retention(SOURCE)
 @Target(CLASS)
 annotation class isos
+
+@Retention(SOURCE)
+@Target(CLASS)
+annotation class optional
```
nomisRev: Should be plural.

nomisRev:
@jereksel Awesome job btw! 👏

raulraja:
@jereksel could you add docs to http://arrow-kt.io/docs/optics/optional/ similar to how it's done in the lenses section to showcase the `@optional` usage?. Once we have those we are ready to merge this. Thanks for the awesome work!

jereksel:
@raulraja Done

raulraja:
@jereksel outstanding contribution, thanks so much! :clap:

Issue #601, opened 2018-01-22
gradle build fails with a nullpointer exception


christophsturm:
this is with a clean checkout of master
```
~/Projects/fq/arrow% ./gradlew build             

FAILURE: Build failed with an exception.

* Where:
Build file '/Users/christoph/Projects/fq/arrow/arrow-annotations-processor/build.gradle' line: 16

* What went wrong:
A problem occurred evaluating project ':arrow-annotations-processor'.
> java.lang.NullPointerException (no error message)

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 1s
```

pakoito:
This is a weird one because it should fail on CI and our local machines and it doesn't. Lemme take a look, meanwhile do `./gradlew clean test` and see what happens.

christophsturm:
same!
maybe it needs some properties set that I'm missing?

pakoito:
Okay, the line that fails is `testCompile files(Jvm.current().getToolsJar())` which tells me there's something wrong with your Java install.

christophsturm:
it seems that it works with jdk 1.8 but not with jdk 9.0.1

this works:
`JAVA_HOME=$(/usr/libexec/java_home -v 1.8) ./gradlew clean test `

this fails: 
`JAVA_HOME=$(/usr/libexec/java_home -v 9) ./gradlew clean test`

pakoito:
I don't believe we're ready to support Java 9 yet because of modules, so I'm okay with this.

christophsturm:
yeah, maybe a more friendly error message or document it in the readme.

pakoito:
I've added it to the setup part of the docs 👍

christophsturm:
what do you think about something like this:
https://github.com/tbroyer/gradle-errorprone-plugin/commit/8aa08a87f42848e6532133697e10276f9d06f4ae

pakoito:
If you'd like to PR it I'll add it!

christophsturm:
done!

PR #602, opened 2018-01-22
improve error message on incompatible jdk version

christophsturm:
without this change the build just fails with a NullPointerException.


christophsturm:
fixes #601

raulraja:
thanks @christophsturm !

PR #604, opened 2018-01-24
Gradle task to run REPL with all built projects

mkobit:
I tried this locally with Java 8, and it seemed to work alright.

This is a bit of a modified version of my original attempt at https://gist.github.com/mkobit/e06ddf93fcb9a67ded3426bd5a266bed.

To run execute `./gradlew repl --no-daemon --console=plain` and you will be prompted by the REPL.


pakoito:
This sounds brilliant. Do you know how to hot-reload any file?

mkobit:
Hah sorry, no ideas on that yet! @raulraja brought this up in the Gradle channel

I'll poke a round as that sounds like an awesome idea.

raulraja:
@mkobit this is awesome, thank you so much for this contribution!

PR #606, opened 2018-01-24
Add Either.getOrHandle

sbuettner:
Add Either.getOrElseGet that allows clients to resolve to Right using the value of Left.

### DIFF
--- a/arrow-core/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-core/src/main/kotlin/arrow/core/Either.kt
@@ -221,6 +221,18 @@ fun <A, B, C> Either<A, B>.flatMap(f: (B) -> Either<A, C>): Either<A, C> = fold(
 inline fun <B> Either<*, B>.getOrElse(crossinline default: () -> B): B = fold({ default() }, { it })
 
 /**
+ * Returns the value from this [Either.Right] or allows clients to transform [Either.Left] to [Either.Right] while providing access to
+ * the value of [Either.Left].
+ *
+ * Example:
+ * ```
+ * Right(12).getOrHandle { 17 } // Result: 12
+ * Left(12).getOrHandle { it + 5 } // Result: 17
+ * ```
+ */
+inline fun <A, B> Either<A, B>.getOrHandle(crossinline default: (A) -> B): B = fold({ default(it) }, { it })
+
+/**
  * * Returns [Either.Right] with the existing value of [Either.Right] if this is a [Either.Right] and the given predicate
  * holds for the right value.
  * * Returns `Left(default)` if this is a [Either.Right] and the given predicate does not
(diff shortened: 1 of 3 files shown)


arrow-core/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -220,6 +220,18 @@ fun <A, B, C> Either<A, B>.flatMap(f: (B) -> Either<A, C>): Either<A, C> = fold(
...
+ * Returns the value from this [Either.Right] or the the given argument if this is a [Either.Left] while providing access to
+ * the value of [Either.Left].
+ *
+ * Example:
+ * ```
+ * Right(12).getOrElseGet { 17 } // Result: 12
+ * Left(12).getOrElseGet { it + 5 } // Result: 17
+ * ```
+ */
+inline fun <A, B> Either<A, B>.getOrElseGet(crossinline default: (A) -> B): B = fold({ default(it) }, { it })
```
raulraja: Since We have already in `ApplicativeError` `handleError` which is a polymorphic version of this method I think a better name would be just `handle` since `handleError` in this context would be inaccurate given `A` is not always used to represent an error.

raulraja: Another name that could work here is just `getOrHandle`

sbuettner: The idea of this name comes from the functional Java lib vavr. I agree that an Either.left is not always representing an error. I could live with „getOrHandle“, „getOrResolve“ or maybe simply „resolve“.

pakoito: If handleError exists, let's reuse that name right?

sbuettner: @pakoito I would prefer `getOrHandle` as it doesnt imply that `Left` is an error.

sbuettner: @raulraja Changed to `getOrHandle`.

sbuettner:
@raulraja Any other concerns?

raulraja:
@sbuettner can you add an example to the Either docs that explains this new combinator? http://arrow-kt.io/docs/datatypes/either/

sbuettner:
@raulraja I added documentation. Looks like the git history is messed up. Maybe we can squash merge this PR.

pakoito:
Wait, what happened here hahahaha Could you please merge with Master?

sbuettner:
@pakoito I rewrote my PR and based it off the current master.

pakoito:
Cool, let's wait for @raulraja to give the thumbs up and merge afterwards. Thanks!

sbuettner:
@raulraja @pakoito Thanks for taking the time to review this PR.

PR #607, opened 2018-01-24
[Docs] Make constructing Option type less weird

abs-zero:


pakoito:
Awesome change! Thank you for noticing.

It'd be great if you could also fix the couple of examples in the `Option` docs: https://github.com/arrow-kt/arrow/edit/master/arrow-docs/docs/docs/datatypes/option/README.md

PR #608, opened 2018-01-24
Add more Optional converters

jereksel:
Add converters for A? and Option<A> to Optional. Also add Optional for list position and list predicate.


arrow-optics/src/test/kotlin/arrow/optics/OptionalTest.kt
```diff
@@ -22,6 +25,25 @@ class OptionalTest : UnitSpec() {
     init {
 
         testLaws(
```
nomisRev: These law test have to be separated into different `testLaws` calls or only one set will run. Will be fixed it in #611.

Also fixes in #611 will fix false negatives in laws.

arrow-optics/src/main/kotlin/arrow/optics/instances/OptionalInstances.kt
```diff
@@ -0,0 +1,30 @@
...
+
+fun <A> listElementPositionOptional(position: Int): Optional<List<A>, A> = Optional(
+        getOrModify = { l -> l.getOrNull(position)?.right() ?: l.left() },
+        set = { e -> { l -> l.mapIndexed { index: Int, value: A -> if (index == position) e else value } } }
+)
+
+fun <A> listElementOptional(predicate: Predicate<A>): Optional<List<A>, A> = Optional(
+        getOrModify = { l -> l.find(predicate)?.right() ?: l.left() },
+        set = { e -> { l ->
+            val location = l.first(predicate)
```
nomisRev: This instance is unlawful. Throws `NoSuchElementException` if element is not found.

arrow-optics/src/main/kotlin/arrow/optics/instances/OptionalInstances.kt
```diff
@@ -0,0 +1,30 @@
...
+            l.mapIndexed { index: Int, value: A ->  if (index == location) e else value }
```
nomisRev: Also location is of type `A` and you're comparing it to `index: Int` ;) What you want here is `indexOfFirst` which returns the index or `-1` if not found.

arrow-optics/src/test/kotlin/arrow/optics/OptionalTest.kt
```diff
@@ -22,6 +25,25 @@ class OptionalTest : UnitSpec() {
...
+
+            OptionalLaws.laws(
```
nomisRev: Other instances should also be added since it would've caught the unlawfulness of `listElementOptional` already ;)

PR #609, opened 2018-01-24
Remove Android sourceSets hack

jereksel:
It's not required anymore: https://github.com/JetBrains/kotlin/blob/master/ChangeLog.md#1220


raulraja:
@JorgeCastilloPrz can maybe help test this or determine if it's safe to remove that now.

JorgeCastilloPrz:
I've tested this on my sample Android repo, precisely on a module where `@instance` annotations are being used, and It still works perfectly after removing the block. 

https://github.com/JorgeCastilloPrz/ArrowAndroidSamples/blob/master/tagless-final/src/main/java/com/github/jorgecastillo/kotlinandroid/tagless/runtime/instances/Instances.kt#L14

Also, I've checked out a [commit](https://github.com/JorgeCastilloPrz/ArrowAndroidSamples/commit/9a4e45416a020c9c230983a38a5bcc5aa025ad88) where I was still using `@deriving` to derive typeclass instances, and also removed the block. It keeps compiling with no errors.

So I would say classes are being indexes properly without it, even before Arrow. It can be removed safely.

JorgeCastilloPrz:
If you could also remove it from the docs, that would be helpful.

raulraja:
thanks @jereksel !

PR #610, opened 2018-01-24
replace runA and runS methods in State with a named Tuple

tschuchortdev:
remove the `State.runA` and `State.runS` functions in favor of named getter functions for the returned tuple to improve readability (as discussed on Slack).

### DIFF
--- a/ank-core/build.gradle
+++ b/ank-core/build.gradle
@@ -9,7 +9,7 @@ kotlin {
 }
 
 dependencies {
-    compile "org.jetbrains.kotlin:kotlin-stdlib-jre8:$parent.ext.kotlinVersion"
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$parent.ext.kotlinVersion"
     compile "org.jetbrains.kotlinx:kotlinx-coroutines-core:$kotlinxCoroutinesVersion"
     compile "org.jetbrains:markdown:0.1.25"
     compile group: 'com.github.born2snipe', name: 'cli-progress', version: '0.4'
--- a/ank-gradle-plugin/build.gradle
+++ b/ank-gradle-plugin/build.gradle
@@ -2,7 +2,7 @@ apply plugin: 'java-gradle-plugin'
 apply plugin: 'kotlin'
 
 dependencies {
-    compile "org.jetbrains.kotlin:kotlin-stdlib-jre8:$parent.ext.kotlinVersion"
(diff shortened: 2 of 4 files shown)


raulraja:
Looks great @tschuchortdev ! Thanks for the contribution!. We need to also update to reflect these changes. You may verify that the docs compile with your changes with `./gradlew :arrow-docs:runAnk`.
Once that compiles we can approve and merge these changes to master.

arrow-data/src/main/kotlin/arrow/data/State.kt
```diff
@@ -37,22 +37,46 @@ typealias StateKindPartial<S> = StateTKindPartial<IdHK, S>
...
+ */
+typealias StateResult<S, A> = Tuple2<S, A> // typealias necessary here because data classes can't be extended
+
+/**
+ * extracts the final state of a monadic computation in `State` out of the returned `StateResult`
+ *
+ * @param S type of the state
+ * @param A type of the returned value
+ */
+val <S,A> StateResult<S,A>.state: S get() = this.a
```
pakoito: This extension will leak to Tuple2, I believe. Can anyone please double-check?

raulraja: It will leak for sure if someone imports `import arrow.data.*`, we can leave the type alias where it is and the extensions to `import arrow.syntax.state.*` so it requires an implicit import. Thoguhts @tschuchortdev ?

tschuchortdev: yeah, I think it will. But I don't really see a better solution since `Tuple2` is a data class that can't be extended, but is required in order to work with the `Applicative` functions. The only option is probably to move it to the syntax module like Raul said, so it has to be imported manually. 

I should also add a constructor function for `StateResult` with the right argument names.

pakoito: You could create a new type by wrapping the tuple and forwarding the accessor methods, if that was your intention. Is it worth creating this new object for the sake of the API?

raulraja: @pakoito I asked this return to remain a Tuple2 because we are building generic machinery for those. Creating a new type is fine if we add yet another extension to the result of `run()` that returns the state as a tuple for example `run().tupled()`

MarioAriasC:
`runA` and `runS` can be marked as deprecated and delete in a next iteration. Removing existing functions will break production code with no warning

MarioAriasC:
`runA` and `runS` can be marked as deprecated and delete in a next iteration. Deleting existing functions will break production code with no warning

PR #612, opened 2018-01-25
Replace rest of usages of `kotlin-stdlib-jre7` with `kotlin-stdlib-jdk7`

mkobit:
This is a followup to the missed libraries in ce99db67b595314942d585ef6aec3d27eef2e1c5.
See https://kotlinlang.org/docs/reference/whatsnew12.html#kotlin-standard-library-artifacts-and-split-packages


mkobit:
OK, I still may have missed a few. Let me double check again...

raulraja:
Thanks @mkobit !

PR #613, opened 2018-01-25
Move classes and tests to the directory under their package namespace

mkobit:
## Notes

This is just cleanup from what I noticed with this open in IntelliJ as it was producings warnings.

**This aligns the code directory to the package it was in. This may not be desired for all classes as the package could be incorrect**.

For example: `arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/data/DeferredKW.kt` has package `package arrow.effects`. 

It was moved to `arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredKW.kt` to reflect the package it was in (compared to `data`).


raulraja:
Thanks so much @mkobit! Maybe @pakoito can weight in since he split the tests the first time and if it's all good we can merge this.

mkobit:
I've updated the title to better reflect the change.

pakoito:
If I remember correctly this was because we had the entire lib in the base package, and it was simpler for imports. As that doesn't seem to be the case anymore we could just go ahead with this.

Personally, I'd like to see all tests for the datatypes of any module in a single package and maybe the rest (platform or internal tests) separate, but I wouldn't block on it.

raulraja:
Thanks for the contribution @mkobit !

PR #614, opened 2018-01-25
Change Left and Right field-backed properties isRight and isLeft to function-only backed ones

ricmf:
This reduces the memory overhead of Either and has no impact in performance.


pakoito:
CI failed because of some trailing spaces: https://travis-ci.org/arrow-kt/arrow/builds/334944985#L858

If you could please take a look and fix them, we can merge afterwards.

ricmf:
Ok, i'm fixing it!

pakoito:
In! Thanks @ricmf

Issue #628, opened 2018-01-30
Option.forall(<empty>) returns false, should be true


benfleis:
In general, by definition, `all(<empty>) → true` and `any(<empty>) → false`.

In arrow's Option impementation, `Option.forall` is `Option.exists`, giving these both a default of `false`.

PR coming shortly.

pakoito:
Done here. Thank you @benfleis!!

PR #629, opened 2018-01-30
update Option.forall to return true on None (ie, empty)

benfleis:
This fixes behavior noted to be incorrect in #628, and aligns with
typical definition where `all(<empty>) -> true`.


arrow-core/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -119,18 +119,18 @@ sealed class Option<out A> : OptionKind<A> {
...
-    }
-
     /**
      * Returns true if this option is empty '''or''' the predicate
      * $p returns true when applied to this $option's value.
      *
      * @param p the predicate to test
      */
-    inline fun forall(crossinline p: (A) -> Boolean): Boolean = exists(p)
+    inline fun forall(crossinline p: Predicate<A>): Boolean = fold({ true }, p)
```
benfleis: Note that I also put forall underneath exists -- sibling funcs now live next to one another.

arrow-data/src/test/kotlin/arrow/data/OptionTest.kt
```diff
@@ -137,8 +137,14 @@ class OptionTest : UnitSpec() {
 
         "exists" {
             some.exists { it.startsWith('k') } shouldBe true
+            some.exists { it.startsWith('j') } shouldBe false
             none.exists { it.startsWith('k') } shouldBe false
```
benfleis: I debated changing the predicate in both None cases to throw a runtime exception, to additionally assert that the predicate itself is never called. Is there already an existing mechanism for this type of assertion in arrow?

raulraja: @benfleis if you are referring to do that in the test you can use `kotlin-test` facilities https://github.com/kotlintest/kotlintest/blob/master/doc/reference.md#exceptions- to capture exceptions.

pakoito: the exception will propagate and fail the tests. If you'd like to fail the test manually use fail("Message").

benfleis: If I update none.exists as follows:

```kotlin
    none.exists { throw NotImplementedError("predicate must not run") } shouldBe false
```

it strengthens the test. I was curious whether that throw above (a la `TODO(...)`) is the standard. Thus, it should not throw -- this variant would simply make it apparent if None's implementation became wrong.

pakoito: I'd rather it was a simple `fail()` call as it's part of the framework. Won't block on it tho.

benfleis: ah, that's what I was looking for. Have to run, but will try to drop a tiny PR for this later.

benfleis:
@pakoito @raulraja  it appears to work, and one of you has to take it home (after rebase). correct?

pakoito:
@benfleis That's correct. You're on Master now :D Thank you for your contribution. How was it?

benfleis:
feels good, of course.

PR #630, opened 2018-01-30
Change monad comprehensions to accept any return value

pakoito:
This PR removes the need to use `yields()` in comprehensions to form a correct value. Instead we take the return value from the block and lift it using `pure()`.

A deprecation path has been added for yields() that replaces the call with the first parameter, as users were not meant to use the return value anyway.

### DIFF
--- a/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadContinuations.kt
+++ b/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadContinuations.kt
@@ -86,6 +86,6 @@ open class MonadContinuation<F, A>(M: Monad<F>, override val context: CoroutineC
     @Deprecated("Yielding in comprehensions isn't required anymore", ReplaceWith("b"))
     fun <B> yields(b: B): B = b
 
-    @Deprecated("Yielding in comprehensions isn't required anymore", ReplaceWith("b"))
+    @Deprecated("Yielding in comprehensions isn't required anymore", ReplaceWith("b()"))
     fun <B> yields(b: () -> B): B = b()
 }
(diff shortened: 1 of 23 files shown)


arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadContinuations.kt
```diff
@@ -83,7 +83,9 @@ open class MonadContinuation<F, A>(M: Monad<F>, override val context: CoroutineC
         COROUTINE_SUSPENDED
     }
 
-    infix fun <B> yields(b: B): HK<F, B> = yields { b }
+    @Deprecated("Yielding in comprehensions isn't required anymore", ReplaceWith("b"))
+    fun <B> yields(b: B): B = b
 
-    infix fun <B> yields(b: () -> B): HK<F, B> = pure(b())
+    @Deprecated("Yielding in comprehensions isn't required anymore", ReplaceWith("b"))
```
nomisRev: Doesn't this change the type of the result? After replace the type is `() -> B` instead of `B`.

raulraja: @nomisRev `B` also get's automatically lifted at the end of the comprehension into `HK<F, B>` via `wrapReturn` that delegates to `pure` https://github.com/arrow-kt/arrow/pull/630/files#diff-017320deefaadcf24021183b902598d1R51

nomisRev: @raulraja I meant in following situation.

```kotlin
fun produceString(): String = "test"

 val t: Either<Throwable, String> = Either.monad<Throwable>().binding {
        yields(::produceString)
}.ev()
```

Doesn't replace result in the following?

```kotlin
val t: Either<Throwable, () -> String> = Either.monad<Throwable>().binding {
    ::produceString
}.ev()
```

pakoito: >IntelliJ replace is not working for me at all atm. Not sure why. Can someone confirm?

Mine either :/

PR #631, opened 2018-01-30
Tests for annotation processor

### LINKED ISSUE #623: Research unit testing for annotation processor
Google's testing library doesn't work on kotlin, but kapt doesn't remove its stubs (they're in build/tmp/kapt3/stubs/debug). We should be able to use them with `compile-testing`.

This may not be required for now, but it will be for #605 

Please assign this to me

jereksel:
Closes #623


.gitignore
```diff
@@ -121,3 +121,6 @@ reports/
 /.idea/misc.xml
 
 .DS_Store
+
+build/
```
pakoito: Build is already part of the gitignore in the section about gradlew

raulraja:
Is there an advantage to have this as a top level module vs just in `src/main/kotlin`? inside `arrow-annotations-processor`? I'd prefer if this was just inside the current module unless we plan on exporting this as a library.

jereksel:
Project with processor will invoke that processor on itself. Because of that we can't have classes that fail tests. Can be checked here: https://github.com/jereksel/arrow/tree/feature/failing_apt

arrow-annotations-processor-test/src/test/java/arrow/ap/tests/APTest.kt
```diff
@@ -0,0 +1,77 @@
...
+    fun testProcessor(
+        vararg processor: AnnotationProcessor
+    ) {
+
+        processor.forEach {
+
+            val stubs = Paths.get("", "build", "tmp", "kapt3", "stubs", "main", "arrow", "ap", "objects").toFile()
+            val expectedDir = Paths.get("", "src", "test", "resources", "arrow", "ap", "objects").toFile()
+
+            val (name, source, dest, proc, error) = it
```
raulraja: you can destructure this arguments also in the function declaration:
```
processor.forEach { (name, source, dest, proc, error)  ->
  ...
}
```

PR #645, opened 2018-02-02
Remove ΛNK modules from ΛRROW repo

Guardiola31337:
- Removes ΛNK modules (`ank-core` and `ank-gradle-plugin`) from ΛRROW repo
- Removes always failing `bintrayUpload` task from `deploy_release.sh` (`uploadArchives` task uploads the artifacts to Bintray so this is not necessary)
- Adds `'io.arrow-kt:ank-core:0.6.1'` dependency to `arrow-docs`

👀 @raulraja @nomisRev @pakoito


build.gradle
```diff
@@ -94,7 +94,7 @@ subprojects { project ->
         key = project.hasProperty('bintrayApiKey') ? project.property('bintrayApiKey') : System.getenv('BINTRAY_API_KEY')
         configurations = ['archives']
         pkg {
-            repo = 'maven'
+            repo = 'arrow-kt'
```
pakoito: Will this affect current releases?

Guardiola31337: Nope, the repo name in Bintray is `arrow-kt` 👀 https://bintray.com/arrow-kt/arrow-kt
In any case, this is only used when running the `bintrayUpload` task which is now disabled https://github.com/arrow-kt/arrow/blob/9e4ced86a11fa167b9fb548a51e2e25f769bcea1/deploy-scripts/deploy_release.sh#L19

PR #646, opened 2018-02-02
Optics at typeclass 

nomisRev:


arrow-test/src/main/kotlin/arrow/test/generators/Generators.kt
```diff
@@ -162,4 +163,7 @@ fun <A> genSequenceKW(genA: Gen<A>): Gen<SequenceKW<A>> =
         Gen.create { Gen.list(genA).generate().asSequence().k() }
 
 fun genChars(): Gen<Char> =
-        Gen.oneOf(('A'..'Z') + ('a'..'z') + ('0'..'9') + "!@#$%%^&*()_-~`,<.?/:;}{][±§".toList())
\ No newline at end of file
+        Gen.oneOf(('A'..'Z') + ('a'..'z') + ('0'..'9') + "!@#$%%^&*()_-~`,<.?/:;}{][±§".toList())
+
+fun <A> genSetKW(genA: Gen<A>): Gen<SetKW<A>> =
+        Gen.create { Gen.set(genA).generate().k() }
```
pakoito: The seed resets on every call, messing with entropy. Capture `Gen.set(genA)` as a value outside the scope and you should be fine.

arrow-docs/docs/docs/optics/at/README.MD
```diff
@@ -0,0 +1,54 @@
...
+`At` provides a [Lens]({{ '/docs/optics/lens' | relative_url }}) for a structure `S` to focus in `A` at a given index `I`.
+
+### Example
+
+If for a structure `S` the focus `A` can be indexed by `I` then `At` can create an `Lens` with focus at `S` for a given index `I`.
+We can use that `Lens` to operate on that focus `S` to get, set and modify the focus at a given index `I`.
+
+A `MapKW<Int, String>` can be indexed by its keys `Int` but not for every index an entry can be found.
+
+```kotlin
```
raulraja: Does Ank not work in these snippets?

PR #647, opened 2018-02-03
Fix typo in MonadError docs

tkaden4:


tkaden4:
The docs definitely have room for improvement, especially around Free Algebras, but MonadError is good. I'm just getting started in the codebase, getting a feel for the internals now.

PR #650, opened 2018-02-04
Add Tuple2 higherkind and instances

aedans:
Adds Functor, Applicative, Monad, Comonad, Foldable, and Traverse instances for Tuple2, as well as higherkind.


arrow-core/src/main/kotlin/arrow/core/TupleN.kt
```diff
@@ -1,8 +1,30 @@
...
+    fun <C> coflatMap(f: (Tuple2Kind<A, B>) -> C): Tuple2<A, C> = a toT f(this)
+    fun extract() = b
+    fun <C> foldL(b: C, f: (C, B) -> C) = f(b, this.b)
+    fun <C> foldR(lb: Eval<C>, f: (B, Eval<C>) -> Eval<C>) = f(b, lb)
+
     fun reverse(): Tuple2<B, A> = Tuple2(b, a)
-    companion object
+
+    companion object {
+        fun <A> pure(a: A) = null toT a
```
pakoito: The type here has to be hinted to avoid confusion on the return.

arrow-data/src/test/kotlin/arrow/data/TupleTest.kt
```diff
@@ -28,16 +33,21 @@ class TupleTest : UnitSpec() {
...
-            EqLaws.laws { Tuple2(it, it) },
-            EqLaws.laws { Tuple3(it, it, it) },
-            EqLaws.laws { Tuple4(it, it, it, it) },
-            EqLaws.laws { Tuple5(it, it, it, it, it) },
-            EqLaws.laws { Tuple6(it, it, it, it, it, it) },
-            EqLaws.laws { Tuple7(it, it, it, it, it, it, it) },
-            EqLaws.laws { Tuple8(it, it, it, it, it, it, it, it) },
-            EqLaws.laws { Tuple9(it, it, it, it, it, it, it, it, it) },
-            EqLaws.laws { Tuple10(it, it, it, it, it, it, it, it, it, it) }
+                FunctorLaws.laws(Tuple2.functor(), { Tuple2.pure(it) }, Eq.any()),
```
pakoito: There is no need to repeat the laws for every typeclass in a hierarchy. MonadLaws internally call to Applicative and Functor laws. Traverse laws call to Foldable laws too.

arrow-weak/src/test/kotlin/arrow/weak/WeakTest.kt
```diff
@@ -27,7 +27,7 @@ class WeakTest : UnitSpec() {
         }
 
         testLaws(
-            EqLaws.laws { Weak(it) },
+            EqLaws.laws { Weak(it) }
```
pakoito: I believe Pablo just removed this module altogether for now, so that'd fix your merge.

arrow-instances/src/main/kotlin/arrow/instances/tuple.kt
```diff
@@ -1,9 +1,49 @@
...
 
-import arrow.typeclasses.Eq
-import arrow.typeclasses.Monoid
+import arrow.*
 import arrow.core.*
-import arrow.instance
+import arrow.typeclasses.*
+
+@instance(Tuple2::class)
+interface Tuple2FunctorInstance : Functor<Tuple2KindPartial<*>> {
```
pakoito: Star projections work well as input types, but not as output ones as users won't be able to work with the first value of the tuple without casting, as it's not carried over these transformations. Check how other HK2 types like Either do it.

aedans: Whoops, didn't notice that.

arrow-instances/src/main/kotlin/arrow/instances/tuple.kt
```diff
@@ -1,9 +1,52 @@
...
+interface Tuple2ApplicativeInstance<F> : Tuple2FunctorInstance<F>, Applicative<Tuple2KindPartial<F>> {
+    fun MF(): Monoid<F>
+    override fun <A, B> map(fa: Tuple2Kind<F, A>, f: (A) -> B) = fa.ev().map(f)
+    override fun <A, B> ap(fa: Tuple2Kind<F, A>, ff: Tuple2Kind<F, (A) -> B>) = fa.ev().ap(ff.ev())
+    override fun <A> pure(a: A) = MF().empty() toT a
+}
+
+@instance(Tuple2::class)
+interface Tuple2MonadInstance<F> : Tuple2ApplicativeInstance<F>, Monad<Tuple2KindPartial<F>> {
+    override fun <A, B> ap(fa: Tuple2Kind<F, A>, ff: Tuple2Kind<F, (A) -> B>) = fa.ev().ap(ff)
```
pakoito: NIT: Spaces between functions, expression body after = in a new line

arrow-instances/src/main/kotlin/arrow/instances/tuple.kt
```diff
@@ -1,9 +1,52 @@
...
-import arrow.typeclasses.Eq
-import arrow.typeclasses.Monoid
+import arrow.*
 import arrow.core.*
-import arrow.instance
+import arrow.core.Either.*
+import arrow.typeclasses.*
+
+@instance(Tuple2::class)
+interface Tuple2FunctorInstance<F> : Functor<Tuple2KindPartial<F>> {
```
pakoito: NIT: F is used for witness types, but I cannot suggest a better option :D

arrow-data/src/test/kotlin/arrow/data/TupleTest.kt
```diff
@@ -28,16 +33,18 @@ class TupleTest : UnitSpec() {
...
+                TraverseLaws.laws(Tuple2.traverse(), Tuple2.functor(), { 0 toT it }, Eq.any()),
+                EqLaws.laws { Tuple2(it, it) },
+                EqLaws.laws { Tuple3(it, it, it) },
+                EqLaws.laws { Tuple4(it, it, it, it) },
+                EqLaws.laws { Tuple5(it, it, it, it, it) },
+                EqLaws.laws { Tuple6(it, it, it, it, it, it) },
+                EqLaws.laws { Tuple7(it, it, it, it, it, it, it) },
+                EqLaws.laws { Tuple8(it, it, it, it, it, it, it, it) },
+                EqLaws.laws { Tuple9(it, it, it, it, it, it, it, it, it) },
+                EqLaws.laws { Tuple10(it, it, it, it, it, it, it, it, it, it) }
```
pakoito: What instance of Eq is resolved here?

arrow-instances/src/main/kotlin/arrow/instances/tuple.kt
raulraja: In the case of applied types not in higher kinded position the convention we are following is `A`, `B`,... if higher kinded then `F`, `G`... In this case it should be `A` since the left is also that in `Tuple2<A, B>`

pakoito:
@aedans apologies, this PR got caught up in some refactors we're doing. We'll help you merge once we're stable.

aedans:
No problem, just tell me when.

raulraja:
@aedans master is ready and this approved, let us know when conflicts are resolved and we will merge this in. Thanks!

pakoito:
We're stable now!! You will need to merge and apply the changes related to the renames we made on the codebase. They're listed here: https://github.com/arrow-kt/arrow/pull/671

PR #652, opened 2018-02-04
[WIP] Weak data type (part deux)

pablisco:
# Work in Progress

The core idea of Weak is a data type that represents a value that can be lost at the time of usage. Backed by a WeakReference when we need to consume it we have to provide an alternative in case that the value has been collected by the GC.

TODO:
- [ ] Create documentation on website
- [ ] Make tests more stable
- [ ] Add more test cases


JorgeCastilloPrz:
In an effort to clean the PR list we've decided to close all the ones that are older than 2 months and consider them obsolete except for very specific circumstances that we will evaluate per PR. This doesn't look bad but has been there almost 2 years already, so we can consider it obsolete. Note we can restore it if we want to get back to it at some point 👍

PR #655, opened 2018-02-05
Auto-generate Syntax for Type Classes

raulraja:
For a type class such as:
```kotlin
@typeclass
interface Functor<F> : TC {

    fun <A, B> map(fa: HK<F, A>, f: (A) -> B): HK<F, B>

    fun <A, B> lift(f: (A) -> B): (HK<F, A>) -> HK<F, B> =
            { fa: HK<F, A> ->
                map(fa, f)
            }

    fun <A> void(fa: HK<F, A>): HK<F, Unit> = map(fa, { _ -> Unit })

    fun <A, B> fproduct(fa: HK<F, A>, f: (A) -> B): HK<F, Tuple2<A, B>> = map(fa, { a -> Tuple2(a, f(a)) })

    fun <A, B> `as`(fa: HK<F, A>, b: B): HK<F, B> = map(fa, { _ -> b })

    fun <A, B> tupleLeft(fa: HK<F, A>, b: B): HK<F, Tuple2<B, A>> = map(fa, { a -> Tuple2(b, a) })

    fun <A, B> tupleRight(fa: HK<F, A>, b: B): HK<F, Tuple2<A, B>> = map(fa, { a -> Tuple2(a, b) })

}
```
It generates all it's syntax facilities:

```kotlin
interface FunctorSyntax<F> {

  fun functor(): Functor<F>

  fun <A, B> arrow.HK<F, A>.`as`(b: B): arrow.HK<F, B> =
    functor().`as`(this, b)

  fun <A, B> arrow.HK<F, A>.`fproduct`(f: kotlin.Function1<A, B>): arrow.HK<F, arrow.core.Tuple2<A, B>> =
    functor().`fproduct`(this, f)

  fun <A, B> arrow.HK<F, A>.`map`(f: kotlin.Function1<A, B>): arrow.HK<F, B> =
    functor().`map`(this, f)

  fun <A, B> arrow.HK<F, A>.`tupleLeft`(b: B): arrow.HK<F, arrow.core.Tuple2<B, A>> =
    functor().`tupleLeft`(this, b)

  fun <A, B> arrow.HK<F, A>.`tupleRight`(b: B): arrow.HK<F, arrow.core.Tuple2<A, B>> =
    functor().`tupleRight`(this, b)

  fun <A> arrow.HK<F, A>.`void`(): arrow.HK<F, kotlin.Unit> =
    functor().`void`(this)
}

```

Which makes it easy to include at call sites for example in abstract contexts like F bound algebras in tagless final:

```kotlin
class Service<F> @Inject constructor(
  override val functor: Functor<F>
): FunctorSyntax<F> by functor {
  fun <A, B> doStuff(fa: HK<F, A>, f: (A) -> B): HK<F, B> = fa.map(f) 
}
```
`fa.map(f)` here is valid because of the inherited `FunctorSyntax<F>`, otherwise we would have to use `functor` directly `functor.map(fa, f)`

### DIFF
--- a/arrow-annotations-processor/src/main/java/arrow/tc/TypeclassFileGenerator.kt
+++ b/arrow-annotations-processor/src/main/java/arrow/tc/TypeclassFileGenerator.kt
@@ -110,11 +110,9 @@ class TypeclassFileGenerator(
 
     private val typeclasses: List<Typeclass> = annotatedList.map { Typeclass(it.classOrPackageProto.`package`, it) }
 
-    private fun functionSignatures(typeClass: Typeclass): List<SyntaxFunctionSignature> {
-        return typeClass.target.classOrPackageProto.functionList.map {
+    private fun functionSignatures(typeClass: Typeclass): List<SyntaxFunctionSignature> = typeClass.target.classOrPackageProto.functionList.map {
             SyntaxFunctionSignature.from("arrow.HK", typeClass, it)
         }.filter { it.hkArgs == HKArgs.First }
-    }
 
     /**
      * Main entry point for higher kinds extension generation
--- a/arrow-annotations-processor/src/main/java/arrow/tc/TypeclassFileGenerator.kt
+++ b/arrow-annotations-processor/src/main/java/arrow/tc/TypeclassFileGenerator.kt
@@ -2,7 +2,12 @@ package arrow.tc
 
 import arrow.common.Package
 import arrow.common.utils.ClassOrPackageDataWrapper
+import arrow.common.utils.extractFullName
+import arrow.common.utils.removeBackticks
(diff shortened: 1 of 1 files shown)


raulraja:
@pakoito it is possible but the type classes are tedious to use without syntax in most cases. If we add it to the annotation as parameter I'd be inclined to make it opt out since we want people to use the syntax for a more pleasant coding experience. thoughts?

pakoito:
It's a weight vs discoverability question :D Opt-in for now, opt-out if anyone complains?

nomisRev:
> Opt-in for now, opt-out if anyone complains?
 Sounds good to me

PR #663, opened 2018-02-05
Generate all static and extensions methods in Type class Syntax

raulraja:
Auto derived from the Monoid syntax:
```kotlin
interface MonoidSyntax<A> {

  val monoid: Monoid<A>

  fun kotlin.collections.Collection<A>.`combineAll`(dummy: Unit = Unit): A =
    monoid.`combineAll`(this)

  fun `empty`(): A =
    monoid.`empty`()
}
```


pakoito:
One example, please

raulraja:
@pakoito added an example in the description, previously those methods were not considered because they did not have anything to do with HK's

Issue #605, opened 2018-01-24
Replace all optics annotation with one annotation.


nomisRev:
One annotation `@optics` with a parameter for granular control. Similar to `@deriving`, by default it will generate all optics applicable to the structure.

While doing so we should deprecate the existing annotations.

jereksel:
@nomisRev My idea for api: https://gist.github.com/jereksel/f302bc67e11072b5fe2fbab77c244e49

nomisRev:
@jereksel properties are allowed for kotlin annotations like so

```kotlin
annotation class optics(val targets: Array<OpticTarget> = [OpticTarget.ISO, OpticTarget.LENS, OpticTarget.PRISM, OpticTarget.OPTIONAL])

enum class OpticTarget {
    ISO, LENS, PRISM, OPTIONAL
}
```

API question would be for the API @raulraja @pakoito @jereksel thoughts?

If you annotate a `sealed class` with `@optics` does it also generate `LENS`, `ISO` and `OPTIONAL`, `PRISM` for its nested types.

```kotlin
@optics sealed class Json {
     data class JsString(val str: String): Json()

     sealed class JsNumber : JsNumber() {
          data class JsInt(val int: Int): JsNumber()
          data class JsLong(val long: Long): JsNumber()
     }

}
```

Would we generate 4 prisms: `Prism<Json, JsString>`, `Prism<Json, JsNumber>`, `Prism<JsNumber, JsInt>`, `Prism<JsNumber, JsLong>`.
And 3 lenses: `Lens<JsString, String>`, `Lens<JsInt, Int>`, `Lens<JsLong, Long>`
etc.

Or does an annotation only generate optics for its immediate targets. i.e. `JsString` and `JsNumber` in this case so 2 Prisms `Prism<Json, JsString>`, `Prism<Json, JsNumber>`.

jereksel:
API: targets in optics must be empty by default - and we should treat empty array as "everything that can be generated".

I think generating for every subclass makes more sense than generating only for direct subclass

PR #669, opened 2018-02-06
Property based optics (with annotation processor)

### LINKED ISSUE #654: Idea for propery-based syntax for lenses
This approach allows threat lenses as extension properties.  

Consider following code:
```kotlin

/** Lens with bound receiver */
interface BoundLens<T, FT> {
    fun get(): FT
    fun set(fv: FT): T
}

inline fun <T, FT, FFT> BoundLens<T, FT>.lens(
        crossinline get: FT.() -> FFT,
        crossinline set: FT.(FFT) -> FT
): BoundLens<T, FFT> {

    val value = get()
    return object : BoundLens<T, FFT> {
        override fun get() = value.get()
        override fun set(fv: FFT) = set(value.set(fv))
    }
}

fun <T> T.lensing() = object : BoundLens<T, T> {
    override fun get(): T = this@lensing
    override fun set(fv: T) = fv
}

inline fun <T, FT> BoundLens<T, FT>.modify(modifier: (FT)->FT) = set(modifier(get()))
```

This allows to define lens as follows:
```kotlin
data class A(val value: Int)

val <T> BoundLens<T, A>.value: BoundLens<T, Int>
    get() = lens(
            get = { value },
            set = { copy(value = it) }
    )

//usage
val a = A(24)
val newA = a.lensing().value.set(42)
``` 

Let's take more complicated example

```kotlin
data class Street(val number: Int, val name: String)
data class Address(val city: String, val street: Street)
data class Company(val name: String, val address: Address)
data class Employee(val name: String, val company: Company)
```

And write lenses:

```kotlin

val <T> BoundLens<T, Employee>.company: BoundLens<T, Company>
    get() = lens(
            get = { company },
            set = { copy(company = it) }
    )

val <T> BoundLens<T, Company>.address: BoundLens<T, Address>
    get() = lens(
            get = { address },
            set = { copy(address = it) }
    )

val <T> BoundLens<T, Address>.street: BoundLens<T, Street>
    get() = lens(
            get = { street },
            set = { copy(street = it) }
    )

val <T> BoundLens<T, Street>.name: BoundLens<T, String>
    get() = lens(
            get = { name },
            set = { copy(name = it) }
    )

```

Note that lenses above may are considered to be generated via annotating processing like original optics. So it shouldn't be written manually. 

But we gain pretty nice syntax: 
```kotlin

val employee = Employee("John Doe", Company("Kategory", Address("Functional city", Street(42, "lambda street"))))

val newEmployee = employee.lensing().company.address.street.name.modify { it.capitalize() }
```

Any thoughts?

jereksel:
Resolves #654


arrow-annotations-processor/src/main/java/arrow/optics/OpticsProcessor.kt
```diff
@@ -55,12 +58,20 @@ class OptikalProcessor : AbstractProcessor() {
...
+
         if (roundEnv.processingOver()) {
             val generatedDir = File(this.generatedDir!!, "").also { it.mkdirs() }
             LensesFileGenerator(annotatedLenses, generatedDir).generate()
             PrismsFileGenerator(annotatedPrisms, generatedDir).generate()
             IsosFileGenerator(annotatedIsos, generatedDir).generate()
             OptionalFileGenerator(annotatedOptional, generatedDir).generate()
+
+            LensesFileGenerator(annotatedBounded, generatedDir).generate()
+            OptionalFileGenerator(annotatedBounded, generatedDir).generate()
```
nomisRev: Why is `LensesFileGenerator` and `OptionalFileGenerator` added twice?

arrow-annotations-processor/src/main/java/arrow/optics/BoundLensGenerator.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    private fun processElement(annotatedOptic: AnnotatedOptic): Pair<AnnotatedOptic, List<String>> =
+            annotatedOptic to annotatedOptic.targets.map { variable ->
+                val sourceClassName = annotatedOptic.classData.fullName.escapedClassName
+                val sourceName = annotatedOptic.type.simpleName.toString().decapitalize()
+                val targetClassName = variable.fullName
+                val targetName = variable.paramName
+
+                """
+                    |val <T> $boundLens<T, $sourceClassName>.$targetName: $boundLens<T, $targetClassName> get() = this.compose($sourceName${targetName.toUpperCamelCase()}())
```
nomisRev: `get` can be `inline`

arrow-annotations-processor/src/main/java/arrow/optics/BoundLensGenerator.kt
```diff
@@ -0,0 +1,34 @@
+package arrow.optics
+
+import arrow.common.utils.fullName
+import me.eugeniomarletti.kotlin.metadata.escapedClassName
+import java.io.File
+
+class BoundLensGenerator(
```
nomisRev: `BoundSetterGenerator`

arrow-optics/src/main/kotlin/arrow/optics/bound.kt
```diff
@@ -0,0 +1,24 @@
...
+    fun <T> compose(other: Lens<A, T>)      = BoundSetter(value, setter + other)
+    fun <T> compose(other: Iso<A, T>)       = BoundSetter(value, setter + other)
+    fun <T> compose(other: Traversal<A, T>) = BoundSetter(value, setter + other)
+
+    fun modify(f: (A) -> A) = setter.modify(value, f)
+
+    fun set(a: A) = setter.set(value, a)
+}
+
+fun <T> T.setter() = BoundSetter(this, Setter.id())
```
nomisRev: `Setter.id()` 👍

arrow-annotations-processor/src/main/java/arrow/optics/BoundLensGenerator.kt
```diff
@@ -0,0 +1,34 @@
...
+        private val annotatedList: Collection<AnnotatedOptic>,
+        private val generatedDir: File
+) {
+
+    private val boundLens = "arrow.optics.BoundSetter"
```
nomisRev: `arrow.optics.syntax.BoundSetter`

arrow-optics/src/main/kotlin/arrow/optics/bound.kt
```diff
@@ -0,0 +1,24 @@
+package arrow.optics
```
nomisRev: We want to separate this DSL in a separate package. `arrow.optics.syntax`

arrow-annotations/src/main/java/arrow/optics.kt
```diff
@@ -18,3 +18,7 @@ annotation class isos
 @Retention(SOURCE)
 @Target(CLASS)
 annotation class optionals
+
+@Retention(SOURCE)
+@Target(CLASS)
+annotation class bounded
```
qradimir: I'm not sure about this name. It refers to implementation details and doesn't describe what would be generated from it. Maybe `@setters`, `@boundSetters` or `@syntaxSetters`?

nomisRev: Perhaps we could name it `@syntax` for now. Since will be enabled by importing `arrow.optics.syntax.*`.

In #605 we'll combine all code gens by one annotation with fine grained control if needed. I hope to implement this over the weekend so it might even be in before next release.

arrow-annotations-processor/src/main/java/arrow/optics/OpticsProcessor.kt
jereksel: Bounded requires that Lenses and Optionals are also generated - this is just in case when only class is only annotated with [at]bounded

modules/docs/arrow-docs/docs/docs/patterns/optics/README.md
```diff
@@ -6,4 +6,26 @@ permalink: /docs/patterns/optics/
 
 ## Optics
 
-TODO
\ No newline at end of file
+TODO
+
+### Generating optics
```
nomisRev: This doesn't really belong under patterns as this is rather a feature of `arrow-optics`. Shall we place this under `docs/optics/syntax` or `docs/optics/dsl/` and refer to it from `docs/optics/setter` @raulraja ?

raulraja: I think it's fine under `syntax` in the optics section.

arrow-optics/src/test/kotlin/arrow/optics/BoundedTest.kt
```diff
@@ -0,0 +1,35 @@
...
+
+//TODO: Rewrite to annotation processor test when multiple file support is implemented
+class BoundedTest: StringSpec() {
+
+    init {
+
+        "@syntax classes are generated properly" {
+
+            val employee = Employee("John Doe", Company("Kategory", Address("Functional city", Street(42, "lambda street"))))
+            val newEmployee = employee.setter().company.nullable.address.street.name.modify { it.capitalize() }
```
nomisRev: Can we try to immediately compose an `Optional` instead of requiring `nullable`? `nullable` is only noise, right? Thoughts?

nomisRev:
@jereksel Since there was no activity here I pushed some changes in order to close this PR.

I made some changes in the generators as well.

Earlier for a class with a nullable or `Option` field i.e. `data class Person(val name: String, val age: Option<Int>)`

```kotlin
fun personAge(): Lens<Person, Option<Int>> = TODO()
fun personAgeOptional(): Optional<Person, Int> = TODO()
```

I changed this to

```kotlin
fun personOptionAge(): Lens<Person, Option<Int>> = TODO()
fun personAge(): Optional<Person, Int> = TODO()
```

Rationale:

In the original generation `Optional` was treated as a special case (postfix), which doesn't make a lot of sense.
1. You want to be able to do simple composition. `personAge() compose ...`. When you have a `Lens` with an `Option` or nullable focus this will get hairy very quickly.
2. `Lens<S, Option<A>>` has a special meaning a http://arrow-kt.io/docs/optics/at/#example. Which is the ability to _delete_ info from a type `S`.

This also solves [this](https://github.com/arrow-kt/arrow/pull/669#discussion_r167328009).

This change will make for a much more elegant DSL and it will also improve usage of Optics.

Also I am unsure about how meaningfull `setter()` is to go into the DSL context. Any suggestions?

Thoughts? @raulraja @pakoito

jereksel:
@nomisRev Thanks a lot for finishing this PR.

raulraja:
@nomisRev the new way looks great!. Are we including this in 0.7.0? If so it needs to be ready for master before Monday which is when the release will be cut.

modules/docs/arrow-docs/docs/docs/optics/syntax/README.md
```diff
@@ -0,0 +1,52 @@
...
+permalink: /docs/optics/syntax/
+---
+
+## Syntax DSL
+
+In some cases the full power of Optics is not required and a familiar property-like syntax to work with values of immutable structures is desired.
+To avoid boilerplate Arrow can generate this property-like dsl using `@syntax` annotation.
+
+```kotlin
+@syntax
```
raulraja: I like everything here but I think `syntax` is way too ambiguous since any of the modules could have named their annotations also syntax to expand exten funs for example. I think we should come up with a better one before the release but in whatever case this is beyond awesome!

nomisRev:
@raulraja I think it would be nice to have this in 0.7.

I also want to try and include #605 so we might drop `@syntax` altogether and always generate it, but that means that isos, lenses, optionals and DSL would get generated for every data class unless you specify otherwise.

Then we might end with something like this, and then syntax or DSL makes sense again IMO.

```kotlin
annotation class optics(val targets: Array<OpticTarget> = [OpticTarget.ISO, OpticTarget.LENS, OpticTarget.PRISM, OpticTarget.OPTIONAL, OpticTarget.DSL])

enum class OpticTarget {
    ISO, LENS, PRISM, OPTIONAL, DSL
}
```

PR #670, opened 2018-02-06
Dagger Instances for Effects integrations

raulraja:
Provides compile time verifiable type class instances for apps and libs depending on Dagger2


JorgeCastilloPrz:
Shoulnd't we provide some docs along with this with further examples?

raulraja:
@JorgeCastilloPrz I will once we have the entire compile time support finished. There is more changes coming regarding Syntax extending other Syntax and I will write the docs once of all of it it's final and tested in https://github.com/JorgeCastilloPrz/ArrowAndroidSamples/tree/master/tagless-final

PR #671, opened 2018-02-07
The Renamening

pakoito:
This PR improves readability for the whole repo:

* The interface `HK` becomes `Kind`. `HK2` will be `Kind2`, `HK3` is `Kind3`, etc
* The witness types will go from being suffixed with `HK` to being prefixed with `For`
* The typealias of `HK<OptionHK, A>` (now `Kind<ForOption, A>`) goes from being `OptionKind<A>` to `OptionOf<A>`
* The partial typealias of `HK<EitherHK, A>` (now `Kind<ForEither, A>`) goes from being `EitherKindPartial<A>` to `EitherPartialOf<A>`
* The evidence/evaluation function will go from `ev()` to `reify()`
* All wrapped types (like map, set, or observable) go from being suffixed `KW` to just `K`. 

Here are a few examples of how these changes will look like: 

`HK<F, Int>` -> `Kind<F, Int>`
`HK<OptionHK, Int>` -> `Kind<ForOption, Int>`
`OptionKind<Int>` -> `OptionOf<Int>`
`Monad<OptionHK>` -> `Monad<ForOption>`
`OptionKind<ListKWKind<A>` -> `OptionOf<ListKOf<Int>>`
`HK<OptionHK, HK<ListKWHK, Int>>` -> `Kind<ForOption, Kind<ForListK, Int>>`
`HK<HK<EitherHK, Throwable>, Int>` -> `Kind<Kind<ForEither, Throwable>, Int>`
`HK2<EitherHK, Throwable, Int>` -> `Kind2<ForEither, Throwable, Int>`
`EitherKindPartial<Throwable>` -> `EitherPartialOf<Throwable>`
`Option.monad().binding { 1 }.ev()` -> `Option.monad().binding { 1 }.reify()`
`ObservableKW.pure(1)`-> `ObservableK.pure(1)`

The docs should reflect these changes.


pakoito:
There is a version mismatch between these changes and what Ank 0.6.1 accepts. Ank depends on Arrow 0.6.1, which has a bunch of binary incompatibilities with this version. As they're not co-located anymore keeping this changes synced is a problem. 👀 @Guardiola31337 @raulraja

raulraja:
We can't make an Ank release that depends on an Arrow snapshot so I believe we should proceed as follows:
1. Merge and release 0.6.2
2. Fix Ank to depend on 0.6.2 and release Ank
3. Fix Arrow to depend on the latest ank release
4. Announce the release officially once the docs are verified
//cc @Guardiola31337

PR #680, opened 2018-02-08
Fix edit button for new folder structure

pakoito:


israelperezglez:
@pakoito I can't compile the project but i think that the path is good ;)

Issue #687, opened 2018-02-09
Hash typeclass


pakoito:
We have equals in `Eq`, toString in `Show`. Let's do hashCode in `Hash`!

https://github.com/typelevel/cats/blob/master/kernel/src/main/scala/cats/kernel/Hash.scala

pt2121:
i'd like to take a stab at this

1Jajen1:
As per #795 I think I'll attempt this. Is the old branch still worth using, or should I start fresh. On that note: Are there some guides to contributing? Or is it best to just start and go along with it in a pr? :)

pakoito:
The contributing guide is...incomplete at best. You need:

* The typeclass
* An instance for at least a couple of datatypes, specially wrappers like Id, ListK, SetK, MapK, SequenceK...
* A set of Laws/Tests
* Apply the laws to the instances

And lastly, the docs. Ping me when you're ready for this
* An entry on the docs menu
* Its own doc

pakoito:
Closed by #1138

PR #690, opened 2018-02-11
add Hash typeclass [WIP]

pt2121:
addressing #687 

**WIP**

### DIFF
--- a/modules/core/arrow-data/src/test/kotlin/arrow/data/EitherTest.kt
+++ b/modules/core/arrow-data/src/test/kotlin/arrow/data/EitherTest.kt
@@ -28,11 +28,13 @@ class EitherTest : UnitSpec() {
             monadError<EitherPartialOf<Throwable>, Throwable>() shouldNotBe null
             semigroupK<EitherPartialOf<Throwable>>() shouldNotBe null
             eq<Either<String, Int>>() shouldNotBe null
+            hash<Either<String, Int>>() shouldNotBe null
             show<Either<String, Int>>() shouldNotBe null
         }
 
         testLaws(
             EqLaws.laws(eq<Either<String, Int>>(), { Right(it) }),
+            HashLaws.laws(hash<Either<String, Int>>(), eq<Either<String, Int>>(), { Right(it) }),
             ShowLaws.laws(show<Either<String, Int>>(), eq<Either<String, Int>>(), { Right(it) }),
             MonadErrorLaws.laws(Either.monadError(), Eq.any(), Eq.any()),
             TraverseLaws.laws(Either.traverse<Throwable>(), Either.applicative(), { Right(it) }, Eq.any()),
--- a/modules/core/arrow-instances/src/main/kotlin/arrow/instances/option.kt
+++ b/modules/core/arrow-instances/src/main/kotlin/arrow/instances/option.kt
@@ -63,6 +63,19 @@ interface OptionEqInstance<A> : Eq<Option<A>> {
 }
 
 @instance(Option::class)
+interface OptionHashInstance<A> : OptionEqInstance<A>, Hash<Option<A>> {
+
+    fun HS(): Hash<A>
+
+    override fun EQ(): Eq<A> = HS()
+
+    override fun hash(a: Option<A>): Int = when (a) {
+        None -> None.hashCode()
+        is Some -> HS().hash(a.t).hashCode()
+    }
+}
+
+@instance(Option::class)
 interface OptionShowInstance<A> : Show<Option<A>> {
     override fun show(a: Option<A>): String =
             a.toString()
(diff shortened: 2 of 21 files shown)


modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Hash.kt
```diff
@@ -0,0 +1,50 @@
...
+
+        /**
+         * Construct a [Hash] instance by using the universal `hashCode` function
+         * and the universal equality relation.
+         *
+         * @returns a [Hash] instance that is defined by the [A] `hashCode` method.
+         */
+        fun <F> fromUniversalHashCode(): Hash<F> = object : Hash<F> {
+            override fun hash(a: F): Int =
+                    a!!.hashCode()
```
pakoito: Instead of forcing non-null here, you can do it in the parameter like this `fun <F: Any> fromUniversalHashCode(): Hash<F>`

pakoito:
So far so good.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Hash.kt
```diff
@@ -0,0 +1,50 @@
...
+                    fhash(a)
+        }
+
+        /**
+         * Construct a [Hash] instance by using the universal `hashCode` function
+         * and the universal equality relation.
+         *
+         * @returns a [Hash] instance that is defined by the [A] `hashCode` method.
+         */
+        fun <F> fromUniversalHashCode(): Hash<F> = object : Hash<F> {
```
pakoito: We're calling these functions that reuse the Java functions `any()`.

raulraja:
Thanks @pt2121 ! These instances should also be declared in the exported instances for Dagger 2 in the dagger modules were relevant

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Hash.kt
pt2121: good to know. will update. thx!

modules/dagger/arrow-dagger/src/main/kotlin/arrow/dagger/instances/number.kt
```diff
@@ -23,6 +24,9 @@ class NumberInstances {
     @Provides
     fun byteEq(): Eq<Byte> = ByteEqInstance
 
+    @Provides
```
raulraja: thanks! :clap:

pt2121: yep sure 👍

modules/core/arrow-instances/src/main/kotlin/arrow/instances/option.kt
```diff
@@ -62,6 +62,12 @@ interface OptionEqInstance<A> : Eq<Option<A>> {
 
 }
 
+@instance(Option::class)
+interface OptionHashInstance<A> : OptionEqInstance<A>, Hash<Option<A>> {
```
pt2121: i'm fixing this. i think it should be more like:

```
@instance(Option::class)
interface OptionHashInstance<A> : Eq<Option<A>>, Hash<Option<A>> {

    fun HASH(): Hash<A>

    override fun hash(a: Option<A>): Int = when (a) {
       None -> None.hashCode()
       is Some -> HASH().hash(a.t).hashCode()
    }
}
```

but then i get:

```Object is not abstract and does not implement abstract member public abstract fun eqv(a: Option<A>, b: Option<A>): Boolean defined in arrow.instances.OptionHashInstance```

any idea?

Comment (raulraja) [reply modules/core/arrow-instances/src/main/kotlin/arrow/instances/option.kt]: what happens if you provide?
```kotlin
@instance(Option::class)
interface OptionHashInstance<A> : OptionEqInstance<A>, Hash<Option<A>> {

    fun HS(): Hash<A>

    override fun EQ(): Eq<A> = HS()

    override fun hash(a: Option<A>): Int = when (a) {
       None -> None.hashCode()
       is Some -> HASH().hash(a.t).hashCode()
    }
}
```

pt2121: nice, that works. thanks!

modules/core/arrow-data/src/test/kotlin/arrow/data/EitherTest.kt
```diff
@@ -28,11 +28,13 @@ class EitherTest : UnitSpec() {
             monadError<EitherPartialOf<Throwable>, Throwable>() shouldNotBe null
             semigroupK<EitherPartialOf<Throwable>>() shouldNotBe null
             eq<Either<String, Int>>() shouldNotBe null
+            hash<Either<String, Int>>() shouldNotBe null
             show<Either<String, Int>>() shouldNotBe null
         }
 
         testLaws(
             EqLaws.laws(eq<Either<String, Int>>(), { Right(it) }),
+//            HashLaws.laws(hash<Either<String, Int>>(), eq<Either<String, Int>>(), { Right(it) }),
```
pt2121: this failed

pt2121:
sorry, I am super busy with my work. gonna close this and give ppl a chance to take it on.

Issue #638, opened 2018-02-01
`mapkw` and `sortedmapkw` docs


pakoito:

pakoito:
@iferca is taking these! Good luck Israel :D

ShelbyCohen:
I'll take these!

PR #692, opened 2018-02-11
#638

iferca:
MapKW tests migrated to use `Laws`

MapKWTest test case migrated to use `Laws` and `testLaws` function. 
Some minor refactor and code cleanup to prevent deprecation warning related to yield function use not being required in comprehension. Adoption of new KindN and HK nomenclature


iferca:
I created this branch from master, fetched both origin and upstream and no differences except my changes with master. 
I see this message in the PR and the `Resolve conflicts` button is disabled to me: 

```
This branch has conflicts that must be resolved
Only those with write access to this repository can merge pull requests.
```

pakoito:
@iferca we moved the files to a new folder several days ago, and those changes aren't in this diff. Try rebasing on the latest master.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/MonadSuspendCancellableContinuations.kt
```diff
@@ -30,7 +30,7 @@ open class MonadSuspendCancellableContinuation<F, A>(SC: MonadSuspend<F>, overri
             invoke(f).bind()
 
     suspend fun <B> bindDeferIn(context: CoroutineContext, f: () -> B): B =
-            suspend { bindingCatch { yields(bindIn(context, f)) } }.bind()
+            suspend { bindingCatch { bindIn(context, f) } }.bind()
```
pakoito: Why this change?

iferca: I clean up some files to prevent the warning regarding `yields` not needed in monad comprehension, this is one of those. Do you agree on the `yields` use there is not needed?

pakoito: Oh I didn't see the bindIn was already there. Yes.

pakoito:
Thank you @iferca for the fix! It's working great now ^^

PR #697, opened 2018-02-12
Removing async laws duplication

Cotel:
Removing some unnecessary code in `ObservableKTest`


pakoito:
Closing because it's an intended duplication

pakoito:
Reopening because I remembered why this is now broken.

JorgeCastilloPrz:
In an effort to clean the PR list we've decided to close all the ones that are older than 2 months and consider them obsolete except for very specific circumstances that we will evaluate per PR. In this case, these law tests are not replicated anymore since those are covered implicitly by the concurrent ones [in here](https://github.com/arrow-kt/arrow/blob/c920e2ca415c0827b7fed3eff5be95d5bde26c7b/modules/fx/arrow-fx-rx2/src/test/kotlin/arrow/fx/ObservableKTests.kt#L51) (`Concurrent` extends `Async`).

PR #702, opened 2018-02-16
Optic kinds

nomisRev:


modules/optics/arrow-optics/src/main/kotlin/arrow/optics/Fold.kt
```diff
@@ -18,7 +18,8 @@ import arrow.typeclasses.monoid
  * @param S the source of a [Fold]
  * @param A the target of a [Fold]
  */
-interface Fold<S, A> {
+@higherkind
```
pakoito: I'm surprised that this works. I expected for @higherkind to be limited to classes.

nomisRev: I was surprised 2 :D

PR #704, opened 2018-02-18
Fix Syntax Diamond Inheritance Issues

raulraja:
Fix for Syntax derivation so we cna solve some of the diamond related issues caused by subtyping when generating Syntax type classes.
A type class such as `MonadErrorSyntax` know now it has to `override fun applicative(): Applicative` as this appears both in the `MonadSyntax` and `ApplicativeError` declarations.

Codegen looks like:

```kotlin
interface MonadErrorSyntax<F, E> : arrow.typeclasses.ApplicativeErrorSyntax<F, E>, arrow.typeclasses.MonadSyntax<F> {

  fun monadError(): MonadError<F, E>

  override fun applicativeError() : arrow.typeclasses.ApplicativeError <F, E> = monadError()

  override fun monad() : arrow.typeclasses.Monad <F> = monadError()

  override fun applicative() : arrow.typeclasses.Applicative <F> = monadError()

  override fun functor() : arrow.typeclasses.Functor <F> = monadError()

  fun <A> arrow.Kind<F, A>.`ensure`(dummy: Unit = Unit, error: kotlin.Function0<E>, predicate: kotlin.Function1<A, kotlin.Boolean>): arrow.Kind<F, A> =
    this@MonadErrorSyntax.monadError().`ensure`(this, error, predicate)
}

```

### DIFF
--- a/modules/core/arrow-annotations-processor/src/main/java/arrow/common/utils/ProcessorUtils.kt
+++ b/modules/core/arrow-annotations-processor/src/main/java/arrow/common/utils/ProcessorUtils.kt
@@ -38,6 +38,21 @@ interface ProcessorUtils : KotlinMetadataUtils {
 
     fun ProtoBuf.Function.overrides(o: ProtoBuf.Function): Boolean = false
 
+    fun ClassOrPackageDataWrapper.Class.declaredTypeClassInterfaces(
+            typeTable: TypeTable): List<ClassOrPackageDataWrapper> {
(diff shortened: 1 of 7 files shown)


raulraja:
Derivation already solves the issue by overriding all of those methods in the same way and providing the instances as arguments to the implicit object. It overrides it's parent always if they have compatible return types, similar technique to a manifestation of the same issue in other parts that are not extend clauses but constructor arguments that are forwarded to implement funs in the interface.

PR #710, opened 2018-02-21
Invariant F in Kinds

pakoito:
Forces F in Kind to be invariant. As Witness types are not expected to be inherited from, covariance is pointless.


pakoito:
Nevermind, nested types.

raulraja:
I did the same a couple of weeks ago xD

PR #711, opened 2018-02-21
Moves `Try` to `arrow-core` 

raulraja:
Since `Try` is a popular used data type along with other core ones like `Option` we are moving it to `arrow-core` with the intent of reducing the need for users to depend on `arrow-data` if they don't plan on using other more advanced data types.

### DIFF
--- a/modules/core/arrow-core/src/main/kotlin/arrow/core/Try.kt
+++ b/modules/core/arrow-core/src/main/kotlin/arrow/core/Try.kt
@@ -1,9 +1,7 @@
-package arrow.data
+package arrow.core
 
 import arrow.*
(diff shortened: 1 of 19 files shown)


pakoito:
You need to move the tests too: modules/core/arrow-data/src/test/kotlin/arrow/data/TryTest.kt

raulraja:
@pakoito all tests are in data currently since the test data module depends on everything else. That never got migrated. We need to migrate all of them to their respectively modules

raulraja:
The build is failing because Ank needs to be updated since it also uses Try. Let's merge and fix Ank with the new snapshot. //cc @Guardiola31337

PR #712, opened 2018-02-21
Enable docs after fixing ank with try changes (core module)

Guardiola31337:
- Enables docs after fixing ΛNK with `Try` changes (`core` module)

👀 @raulraja @pakoito

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ jdk:
   - oraclejdk8
 
 script:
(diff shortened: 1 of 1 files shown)


raulraja:
Build is broken because:
`Could not find io.arrow-kt:macroing-cel4j-artifact:0.6.2-SNAPSHOT.`

raulraja:
Included in #713

PR #713, opened 2018-02-21
NEL docs and Ank Fixes

raulraja:
Includes #712

### DIFF
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/UnitSpec.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/UnitSpec.kt
@@ -1,9 +1,11 @@
 package arrow.test
 
+import arrow.data.NonEmptyList
 import arrow.typeclasses.Eq
 import arrow.test.laws.Law
 import io.kotlintest.TestCase
 import io.kotlintest.specs.StringSpec
+import java.util.*
 
 
 /**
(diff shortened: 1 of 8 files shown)


modules/core/arrow-test/src/main/kotlin/arrow/test/UnitSpec.kt
```diff
@@ -1,9 +1,11 @@
 package arrow.test
 
+import arrow.data.NonEmptyList
 import arrow.typeclasses.Eq
 import arrow.test.laws.Law
 import io.kotlintest.TestCase
 import io.kotlintest.specs.StringSpec
+import java.util.*
```
pakoito: Is this necessary?

raulraja: We use `UUID` but we can remove the wildcard import for a concrete one.

pakoito: Oh this is in the tests. Nevermind then.

modules/docs/arrow-docs/docs/docs/datatypes/nonemptylist/README.md
```diff
@@ -6,4 +6,133 @@ permalink: /docs/datatypes/nonemptylist/
...
+
+```kotlin
+NonEmptyList.of(1, 2, 3, 4, 5).head
+```
+
+---
+
+## foldLeft
+
+Whe we fold with turn a __NonEmptyList< A >__ into __B__ by providing a __seed__ value and a __function__ that carries the state on each iteration over the elements of the list.
```
aballano: Whe we fold with turn a -> When we fold we turn a (?)

PR #736, opened 2018-03-21
Allow exception to propagate from fold Success function

stoyle:
Handling a failed Success as if it was the same as original Failure is not the way it is done in cats. I would argue this makes for a better fold function.


stoyle:
Hi guys! I guess you've made a really explicit choice to include this behavior. I hope you will consider changing it, it really confused me in our code base.

Cheers,
Alf

pakoito:
I believe it was me who made the change. I don't feel strongly for or against it. The version in this PR is more correct specially coming from Scala, but less user-friendly for the Java people :D

jasonab:
As a Java person, I agree with the change - it would be a real headache to track down that it was the success function that was causing the failure, not the previous function!

pakoito:
Merged it is!

stoyle:
That was quick, awesome and thanks :)

PR #745, opened 2018-03-25
[DO NOT MERGE] Typeclassless refactor - Part 2 - Rewritte all typeclasses

pakoito:
Do not merge!

This PR includes all the changes necessary to convert the repo to typeclassless. Please note it's just a part of the complete changeset, better suited for review.


modules/core/arrow-data/src/main/kotlin/arrow/data/EitherT.kt
```diff
@@ -21,66 +21,83 @@ data class EitherT<F, A, B>(val value: Kind<F, Either<A, B>>) : EitherTOf<F, A,
...
-                    MF.map(f(it).fix().value) { recursionControl ->
-                        when (recursionControl) {
-                            is Either.Left<L, Either<A, B>> -> Right(Left(recursionControl.a))
-                            is Either.Right<L, Either<A, B>> -> {
-                                val b: Either<A, B> = recursionControl.b
-                                when (b) {
-                                    is Either.Left<A, B> -> Left(b.a)
-                                    is Either.Right<A, B> -> Right(Right(b.b))
-                                }
+        fun <F, L, A, B> tailRecM(a: A, f: (A) -> EitherTOf<F, L, Either<A, B>>, MF: Monad<F>): EitherT<F, L, B> = MF.run {
```
raulraja: Should functions like this depending on nested constrains be defined outside of companions such as that what now is:
```kotlin
fun <F, L, A, B> tailRecM(a: A, f: (A) -> EitherTOf<F, L, Either<A, B>>, MF: Monad<F>): EitherT<F, L, B>
```
becomes
```
fun <F, L, A, B> Monad<F>.tailRecM(a: A, f: (A) -> EitherTOf<F, L, Either<A, B>>): EitherT<F, L, B>
```

modules/core/arrow-annotations-processor/src/main/java/arrow/derive/DerivingProcessor.kt
```diff
@@ -1,16 +1,17 @@
...
 import java.io.File
-import javax.annotation.processing.Processor
 import javax.annotation.processing.RoundEnvironment
 import javax.lang.model.SourceVersion
 import javax.lang.model.element.ElementKind
 import javax.lang.model.element.TypeElement
 
-@AutoService(Processor::class)
+// FIXME(paco) Enable Deriving once it's ready again
+//@AutoService(Processor::class)
```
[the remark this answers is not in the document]
pakoito: I could work on it next, everything else but the docs are in place already.

modules/core/arrow-data/src/main/kotlin/arrow/data/Ior.kt
```diff
@@ -28,7 +28,7 @@ typealias IorNel<A, B> = Ior<Nel<A>, B>
  * The isomorphic Either form can be accessed via the [unwrap] method.
  */
 @higherkind
-sealed class Ior<A, B> : IorOf<A, B> {
+sealed class Ior<out A, out B> : IorOf<A, B> {
```
raulraja: nice one

modules/core/arrow-data/src/main/kotlin/arrow/data/EitherT.kt
pakoito: Can't, because they're only available in the scope of EitherT. You'd have to do something like

```
val et = getEitherT()
et.run { MF.railrecM(...) }
```

to be able to use them. Typeclassless simply doesn't work with methods inside classes.

modules/core/arrow-data/src/main/kotlin/arrow/data/Ior.kt
pakoito: It is like that on Master, I just fixed it back after breaking it xD

modules/core/arrow-instances/src/main/kotlin/arrow/instances/listk.kt
```diff
@@ -0,0 +1,125 @@
...
+
+    override fun <A, B> Kind<ForListK, A>.foldRight(lb: Eval<B>, f: (A, Eval<B>) -> Eval<B>): Eval<B> =
+            fix().foldRight(lb, f)
+
+    override fun <A> Kind<ForListK, A>.isEmpty(): kotlin.Boolean =
+            fix().isEmpty()
+}
+
+@instance(ListK::class)
+interface ListKTraverseInstance : Traverse<ForListK> {
```
raulraja: Can we add `sequence()` to all Traverse data types via `traverse(::identity, GA)`?

pakoito: It's already in `Foldable` https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Foldable.kt#L103

modules/optics/arrow-optics/src/main/kotlin/arrow/optics/Iso.kt
```diff
@@ -69,22 +69,26 @@ interface PIso<S, T, A, B> : PIsoOf<S, T, A, B> {
     /**
      * Lift a [PIso] to a Functor level
      */
-    fun <F> mapping(FF: Functor<F>): PIso<Kind<F, S>, Kind<F, T>, Kind<F, A>, Kind<F, B>> = PIso(
```
nomisRev: Unrelated to this PR but upon seeing this I was wondering do we want to completely remove ` fun <F, A, B> Functor<F>.map(fa: Kind<F, A>. f: (A) -> B): Kind<F, B>`? Because in this case or [here](https://github.com/arrow-kt/arrow/pull/745/files#diff-5e328743e478d4b73e209941fff61b3dL86) I am in doubt that it is an improvement.

Maybe it is just something I need to get used to though but you're forced to use `with` or `run` now. Our `Functor` definition also looks really different from what people are used to for `Functor`.

Thoughts?

modules/core/arrow-data/src/test/kotlin/arrow/data/ReaderTest.kt
```diff
@@ -3,49 +3,49 @@ package arrow.data
...
 import org.junit.runner.RunWith
-import arrow.test.UnitSpec
 
 @RunWith(KTestJUnitRunner::class)
 class ReaderTest : UnitSpec() {
     init {
 
         "map should return mapped value" {
-            Reader<Int, Int> { it * 2 }.map { it -> it * 3 }.runId(2) shouldBe 12
+            { a: Int -> a * 2 }.reader().map { it -> it * 3 }.runId(2) shouldBe 12
```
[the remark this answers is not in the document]
pakoito: `fun Reader()` doesn't get picked up by the compiler for some reason. It gets the ReaderT constructor instead :(

modules/core/arrow-free/src/main/kotlin/arrow/free/Cofree.kt
```diff
@@ -11,50 +14,54 @@ import arrow.typeclasses.Traverse
 typealias CofreeEval<S, A> = Kind<S, Cofree<S, A>>
 
 @higherkind
-data class Cofree<S, A>(val FS: Functor<S>, val head: A, val tail: Eval<CofreeEval<S, A>>) : CofreeOf<S, A>, CofreeKindedJ<S, A> {
+data class Cofree<S, A>(val FS: Functor<S>, val head: A, val tail: Eval<CofreeEval<S, A>>) : CofreeOf<S, A>, CofreeKindedJ<S, A>, Functor<S> by FS {
```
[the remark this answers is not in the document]
pakoito: KindedJ generation only works for datatypes with invariant generics. For the rest you need the conversion layer :(

modules/optics/arrow-optics/src/main/kotlin/arrow/optics/Iso.kt
pakoito: Where is that extension function `Functor<F>.map` defined? AFAIK it was deleted.

>Maybe it is just something I need to get used to though but you're forced to use with or run now.

I need to write a couple of blog posts and docs about the new intuitions for typeclasses. Functor now works in a similar way to how it does in Haskell, where it's a set of functions parametrized to a single type and available at a scope. For people coming from Java we can now say that Typeclasses are like DSLs to create Kinds.

modules/core/arrow-free/src/main/kotlin/arrow/free/Cofree.kt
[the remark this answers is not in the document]
nomisRev: Alright, I had a feeling I was just missing the reason why :D

PR #747, opened 2018-03-26
Arrow Generic

raulraja:
This first PR is the first part of Arrow generic including meta facilities for derivation of the simpler type classes you can derive on product types. 

Subsequent PRs will include support for coproduct(sealed classes) and auto derivation of type classes we can use in libraries like Helios for auto deriving encoder and decoders at compile time.

## Arrow Generic

`arrow-generic` provides meta programming facilities over Product types like data classes, tuples, and heterogeneous lists; and soon Coproduct types like sealed classes.

### Features

`@product`

We refer to data classes, tuples, and heterogeneous lists as Product types because they all represent a container of typed values in which all those values need to be present.

That is to say that in the following data class both `balance` and `available` are properties of the `Account` class and both are typed and guaranteed to always be present within an `Account`

```kotlin
@product
data class Account(val balance: Int, val available: Int) {
  companion object
}
```

All `@product` annotated data classes must include a `companion object` so that codegen can be properly expanded as extension functions to the companion.

Because of such properties we can automatically derive interesting behaviors from our data classes by using the `@product` annotation:

#### Extensions

`@product` autmatically derives instances for `Semigroup` and `Monoid` supporting recursion in declared data types. In the example below we are able to `+` two `Account` objects because the instance `Int.semigroup()` is provided by Arrow.

##### + operator

```kotlin
import arrow.core.*
import arrow.generic.*

Account(1000, 900) + Account(1000, 900)
// Account(2000, 1800)
```

##### combineAll

`@product` enables also syntax over `List<Account>`  to reduce `(List<Account>) -> Account` automatically based also on the `Semigroup` instance, `@product` expects already defined instances for all contained data types which for most basic primitives Arrow already provides and for custom data types can be manually generated or automatically derived by Arrow with `@product`

```kotlin
listOf(Account(1000, 900), Account(1000, 900), Account(1000, 900)).combineAll()
// Account(3000, 2700)
```

##### tupled

`@product` enables `Account#tupled` and `Tuple2#toAccount` extensions automatically to go back and forth between the data class values to tuple representations such as `Tuple2` with the same arity and property types as those declared in the data class for all data classes with at least 2 properties.

```kotlin
Account(1000, 900).tupled()
//Tuple2(1000, 9000)
```

```kotlin
Account(1000, 900).tupledLabelled()
Tuple2(Tuple2("balance", 9000), Tuple2("available", 9000))
```

```kotlin:ank
Tuple2(1000, 900).toAccount()
//Account(1000, 900)
```

##### toHList

`@product` enables `Account#tpHList` and `HList2#toAccount` extensions automatically to go back and forth between the data class value to a heterogeneous list representation such as `HList2` with the same arity and property types as those declared in the data class regardless of the number of properties.

```kotlin:ank
Account(1000, 900).toHList()
HCons(head=1000, tail=HCons(head=900, tail=arrow.generic.HNil@2ca2159a))
```

```kotlin:ank
Account(1000, 900).toHListLabelled()
```

```kotlin:ank
hListOf(1000, 900).toAccount()
```

##### Applicative#mapTo___

`@product` allows us map independent values in the context of any `Applicative` capable data type straight to the data class inside the data type context

In the examples below we can observe how 2 different `Int` properties are returned inside a type constructor such as `Option`, `Try`, `Deferred` etc... and the automatically mapped to the shape of our `Account` data class removing all boilerplate from extracting the values from their context and returning and `Account value in the same context.

```kotlin
val maybeBalance: Option<Int> = Option(1000)
val maybeAvailable: Option<Int> = Option(900)

Option.applicative().mapToAccount(maybeBalance, maybeAvailable)
// Some(Account(balance=1000, available=900))
```

```kotlin
val tryBalance: Try<Int> = Try { 1000 }
val tryAvailable: Try<Int> = Try { 900 }

Try.applicative().mapToAccount(tryBalance, tryAvailable)
// Success(value=Account(balance=1000, available=900))
```

```kotlin:ank
import arrow.effects.*

val asyncBalance: DeferredK<Int> = async { 1000 }.k()
val asyncAvailable: DeferredK<Int> = async { throw RuntimeException("BOOM") }.k()

DeferredK.applicative().mapToAccount(asyncBalance, asyncAvailable)
// Deferred
```

##### Semigroup 

Combine and reduce a data class based on it's internal properties reduction and combination properties as defined by their `Semigroup` instance.

```kotlin:ank
with(Account.semigroup()) {
  Account(1000, 900).combine(Account(1000, 900))
}
```

##### Monoid 

Extends `Semigroup` by providing the concept of absent or empty value. It derives it's empty value based on the empty value of each one of it's contained properties.

```kotlin:ank
emptyAccount()
```

```kotlin:ank
Account.monoid().empty()
```

##### Eq 

Structural equality in terms of `Eq`, a type class that represents equality.

```kotlin:ank
with(Account.eq()) {
  Account(1000, 900).eqv(Account(1000, 900))
}
//true
```

##### Show 

`toString` as a type class: `Show`

```kotlin:ank
with(Account.show()) {
  Account(1000, 900).show()
}
// Account(balance=1000, available=900)
```

#### Creating instances for custom properties

Sometimes you may be in need of creating type class instances for custom properties that Arrow does not provide by default.

In the following example our `Car` data class contains a `maxSpeed: Speed` property for a custom type.

Arrow can auto derive `Semigroup`, `Monoid`, `Eq` and `Show` for `Car` as long as we also have instances for `Speed`.

```kotlin
data class Speed(val kmh: Int) {
  companion object
}

@product
data class Car(val mod: Int, val speed: Speed) {
  companion object
}
```

Once we attempt to compile this we would get an error similar to the one below:

```$xslt
:arrow-docs:compileKotline: /home/raulraja/workspace/arrow/arrow/modules/docs/arrow-docs/build/generated/source/kaptKotlin/main/product.arrow.generic.car.kt: (60, 119): Unresolved reference.
```

This is because `Speed` is a data class not flagged as `@product`. Let's fix that:

```kotlin
@product
data class Speed(val kmh: Int) {
  companion object
}

@product
data class Car(val mod: Int, val speed: Speed) {
  companion object
}
```

The reason the code compiles now is that Arrow was able to complete the instance for `Car` once we proved we had one for `Speed`. 

Now that `Speed` is also flagged as `@product` its `Semigroup`, `Monoid`, `Show` and `Eq` instances are available and visible in `Car`

```kotlin:ank
Speed(50) + Speed(50)
//Speed(100)
```

```kotlin:ank
Car(Speed(50)) + Car(Speed(50))
//Car(Speed(100))
```


modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/char.kt
```diff
@@ -1,7 +1,31 @@
...
+
+    override fun Char.gt(b: Char): Boolean = this > b
+
+    override fun Char.gte(b: Char): Boolean = this >= b
+
+    override fun Char.neqv(b: Char): Boolean = this != b
+}
+
+fun Char.Companion.eq(): Eq<Char> =
+        lazyOf(object : CharEqInstance {}).value
```
nomisRev: Why are these wrapped in `lazyOf` and then immediately extract the value?

pakoito: Unlike Scala, in Kotlin that `lazyOf` will be invoked on every function call. You need a separate field that's returned from the function.

JorgeCastilloPrz:
Couldn't we use compile time for validating scenarios where this error is shown? 
`:arrow-docs:compileKotline: /home/raulraja/workspace/arrow/arrow/modules/docs/arrow-docs/build/generated/source/kaptKotlin/main/product.arrow.generic.car.kt: (60, 119): Unresolved reference.`

It would be very handy to tell the user what's he missing exactly, like "You need to annotate`Speed` as `@product` in order to derive `Car` typeclass instances."

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/char.kt
JorgeCastilloPrz: What do you mean by a separate field ? Can't you just return the object as is?
```kotlin
fun Char.Companion.eq(): Eq<Char> = object : CharEqInstance {}
```

modules/core/arrow-annotations-processor/src/main/java/arrow/generic/ProductProcessor.kt
```diff
@@ -0,0 +1,84 @@
...
+    override fun getSupportedAnnotationTypes() = setOf(productAnnotationClass.canonicalName)
+
+    override fun onProcess(annotations: Set<TypeElement>, roundEnv: RoundEnvironment) {
+        annotatedProduct += roundEnv
+                .getElementsAnnotatedWith(productAnnotationClass)
+                .map(this::evalAnnotatedProductElement)
+
+        if (roundEnv.processingOver()) {
+            val generatedDir = File(this.generatedDir!!, "").also { it.mkdirs() }
+            ProductFileGenerator(annotatedProduct, generatedDir).generate()
```
JorgeCastilloPrz: I think it'd be good to generate each file under the target class package, like other processors do (Dagger, ButterKnife). Those do it bc they need package protected visibility to access injectable fields on targets, which is not needed here, but that's still handy for traceability. The need to browse a generated class is usual and it's good to have them right under the same package structure of the target class if you navigate through the `build/` dir on your project.

modules/core/arrow-annotations-processor/src/main/java/arrow/generic/ProductProcessor.kt
```diff
@@ -0,0 +1,84 @@
...
+            |It can only be used on $targetName.""".trimMargin()
+
+    private fun evalAnnotatedProductElement(element: Element): AnnotatedGeneric = when {
+        (element.kotlinMetadata as? KotlinClassMetadata)?.data?.classProto?.isDataClass == true -> {
+            val elementClassData = getClassData(element)
+            val paramNames = getConstructorParamNames(element)
+            val typeNames = getConstructorTypesNames(element)
+            val properties = paramNames.zip(typeNames).map { Target(it.second,it.first) }
+            if (properties.size > 22)
+                knownError("${element.enclosingElement}.${element.simpleName} up to 22 constructor parameters is supported")
```
JorgeCastilloPrz: Reading the `knownError()` implementation seems like you're throwing exceptions when something goes wrong. Why aren't you using the procesing environment `messager` for these error logs?

Messager plays well with compile time testing frameworks like `compile-testing` and `Truth` which allow you to assert over generated classes, errors, etc. [Some test samples here](https://github.com/JorgeCastilloPrz/corleone/blob/master/corleone/src/test/java/com/github/jorgecastilloprz/corleone/AnnotationValidationTest.java).

I'm not sure about how problematic can become to just fail by throwing.

modules/core/arrow-annotations-processor/src/main/java/arrow/generic/ProductProcessor.kt
```diff
@@ -0,0 +1,84 @@
...
+            else
+                AnnotatedGeneric(element as TypeElement, elementClassData, properties)
+        }
+
+        else -> knownError(productAnnotationError(element, productAnnotationName, productAnnotationTarget))
+    }
+
+    private fun getConstructorTypesNames(element: Element): List<String> = element.kotlinMetadata
```
JorgeCastilloPrz: shouldn't we extract all this utility functions to the commons package? getting the constructor type names seems quite important and it can probably be reused by many APs. Since we're moving on this direction where we use APs a lot it seems like a good idea to create our own infraestructure syntax to work with them. Same for the following methods.

modules/core/arrow-annotations/src/main/java/arrow/product.kt
```diff
@@ -0,0 +1,8 @@
+package arrow
+
+import kotlin.annotation.AnnotationRetention.SOURCE
+import kotlin.annotation.AnnotationTarget.CLASS
+
+@Retention(SOURCE)
+@Target(CLASS)
+annotation class product
```
JorgeCastilloPrz: I guess product refers to product types but the annotation should probably reflect what's it really doing for the end user. Why not calling it something like `@derive` or something similar?

modules/core/arrow-generic/build.gradle
```diff
@@ -0,0 +1,18 @@
+dependencies {
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jre7:$kotlinVersion"
```
JorgeCastilloPrz: shouldn't we move on to newer gradle versions and use `implementation` (for non allowing transitive deps) and `api` (for allowing transitive deps) ? these already old `compile` statements are throwing warnings at compile time on projects compiled with newer versions of gradle (or the wrapper).

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/number.kt
```diff
@@ -2,150 +2,234 @@ package arrow.instances
...
-
 //////////
 // Byte
 //////////
-
-private val SGByte: Semigroup<Byte> = NumberSemigroup({ one, two -> (one + two).toByte() })
-
-object ByteMonoidInstance : Monoid<Byte>, Semigroup<Byte> by SGByte {
-  override fun empty(): Byte = 0
+interface ByteSemigroupInstance : Semigroup<Byte> {
```
JorgeCastilloPrz: does it make sense to call instances to interfaces that you still need to instantiate ? why not returning the objects already implementing those anymore?

modules/core/arrow-annotations-processor/src/main/java/arrow/generic/ProductProcessor.kt
raulraja: I agree, if we want to do that we should create an issue and tackle it in a different PR because that is not just specific to the `ProductProcessor` or Generic but to all other processors as well for all other annotations in Arrow.

modules/core/arrow-annotations-processor/src/main/java/arrow/generic/ProductProcessor.kt
raulraja: Sounds good, but that was already there and other processors may depend on it so it should be tackle in a different issue. Hopefully `kotlin-meta` will make meta programming easier once we have an AST we control ourselves, the current state of meta-programming facilities in these processor is far from ideal and the proto APIs a royal pain in the ass to deal with.

modules/core/arrow-annotations/src/main/java/arrow/product.kt
raulraja: @product is ment to do a lot more than derivation. For example the `tupled` methods are not directly related to derivation. The idea is to have under generic `@product`, `@coproduct` and that add the features inside those annotations opt out. `@product(deriving = Show.class, isoExtensions = HList)` This would tell to just generate `Show` and the `HList` extension for advanced use cases.
At the moment granular support for a la carte derivation is not built in since this is just a first approach.

modules/core/arrow-generic/build.gradle
raulraja: Sounds good, but since all modules are coded now like that it should be done as a separate issue being unrelated to Generic.

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/number.kt
raulraja: This is the same convention that is used across `Arrow`. Originally the sufix `Instance` was added to have a naming convention by which to look up instances at runtime. Since 0.7.0 eliminates the runtime system I propose that we get rid of `Instances` sufixing names across the board. the `OptionFunctorInstance` becomes `OptionFunctor` and it's real instantiated instance is:
```
interface OptionFunctor {
  ...
  companion object {
    val defaultInstace: Functor<ForOption> = object : OptionFunctor {}  
  }
}

fun Option.Companion.functor(): Functor<ForOption> = 
   OptionFunctor.defaultInstance
```
This brings us simpler names and also proper singleton values for instances that require no further arguments for instantiation. If this looks good we should tackle this refactoring in a separate PR.

modules/core/arrow-core/src/main/kotlin/arrow/core/TupleN.kt
```diff
@@ -30,54 +30,82 @@ data class Tuple2<out A, out B>(val a: A, val b: B) : Tuple2Of<A, B> {
...
-  companion object
+@higherkind
+data class Tuple10<out A, out B, out C, out D, out E, out F, out G, out H, out I, out J>(val a: A, val b: B, val c: C, val d: D, val e: E, val f: F, val g: G, val h: H, val i: I, val j: J) : Tuple10Of<A, B, C, D, E, F, G, H, I, J> {
+    companion object
 }
 
-data class Tuple8<out A, out B, out C, out D, out E, out F, out G, out H>(val a: A, val b: B, val c: C, val d: D, val e: E, val f: F, val g: G, val h: H) {
-  fun reverse(): Tuple8<H, G, F, E, D, C, B, A> = Tuple8(h, g, f, e, d, c, b, a)
+@higherkind
+data class Tuple11<out A, out B, out C, out D, out E, out F, out G, out H, out I, out J, out K>(val a: A, val b: B, val c: C, val d: D, val e: E, val f: F, val g: G, val h: H, val i: I, val j: J, val k: K) : Tuple11Of<A, B, C, D, E, F, G, H, I, J, K>
```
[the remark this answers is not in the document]
raulraja: true dat, good catch

raulraja:
@nomisRev Can't do anything yet with HList since we have not implemented folds or anything, for now the only thing you can do is create the structure and that is why we have not added the docs yet. Alternatively I can remove the HList part from the docs until we have more meat there but I don't think it's worth it. HList will be documented when it provides some interesting functionality.

modules/core/arrow-annotations-processor/src/main/java/arrow/generic/ProductProcessor.kt
raulraja: Done

raulraja:
@JorgeCastilloPrz @nomisRev addressed all your comments

modules/core/arrow-annotations-processor/src/main/java/arrow/generic/ProductProcessor.kt
JorgeCastilloPrz: Agree. Created #765

modules/core/arrow-annotations-processor/src/main/java/arrow/generic/ProductProcessor.kt
JorgeCastilloPrz: OK #766

modules/core/arrow-generic/build.gradle
JorgeCastilloPrz: Nevermind my comment, that's part of the android gradle plugin, not part of gradle itself.

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/number.kt
JorgeCastilloPrz: Sounds good to me, but we would probably want to discuss that on an issue apart to avoid blocking here and move towards 0.7.0 faster for monday. (since that doesn't sound blocker at all anyways) There you go #767

modules/core/arrow-annotations-processor/src/main/java/arrow/generic/ProductProcessor.kt
jereksel: Throwing works fine with tests:
https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-annotations-processor-test/src/test/java/arrow/ap/tests/OptionalTest.kt#L20
https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-annotations-processor-test/src/test/java/arrow/ap/tests/APTest.kt#L48

But printing error with messager automatically throws exception (https://docs.oracle.com/javase/7/docs/api/javax/annotation/processing/Messager.html), so I guess that's the correct way of reporting errors.

modules/core/arrow-annotations-processor/src/main/java/arrow/common/utils/ProcessorUtils.kt
```diff
@@ -13,121 +13,151 @@ import javax.lang.model.element.VariableElement
 
 interface ProcessorUtils : KotlinMetadataUtils {
 
-  fun KotlinMetadata.asClassOrPackageDataWrapper(classElement: TypeElement): ClassOrPackageDataWrapper? {
-    val `package` = elementUtils.getPackageOf(classElement).toString()
-    return when (this) {
-      is KotlinClassMetadata -> data.asClassOrPackageDataWrapper(`package`)
-      is KotlinPackageMetadata -> data.asClassOrPackageDataWrapper(`package`)
-      else -> null
+    fun getConstructorParamNames(element: Element): List<String> = element.kotlinMetadata
```
pakoito: Better as an extfun or extension field actually :D Same for the others in this file.

modules/core/arrow-generic/src/main/kotlin/arrow/generic/Generic.kt
```diff
@@ -0,0 +1,10 @@
+package arrow.generic
+
+interface Generic<T, Repr> {
```
pakoito: As per the signatures, this interface models any isomorphism, not just generic representation. What about renaming it to `Cast` or `Transform` or `As`.

modules/core/arrow-generic/src/main/kotlin/arrow/generic/ProductTypeClass.kt
```diff
@@ -0,0 +1,62 @@
...
+            typeClass.project({ cg }, gen::to, gen::from)
+}
+
+/**
+ * A type class abstracting over the `product` operation of type classes over
+ * types of kind `*`, as well as deriving instances using an isomorphism.
+ * Refines ProductTypeClass with the addition of runtime `String` labels
+ * corresponding to the names of the product elements.
+ */
+interface LabelledProductTypeClass<C> {
```
pakoito: Labeled?

modules/core/arrow-generic/src/main/kotlin/arrow/generic/instances/tuples.kt
```diff
@@ -0,0 +1,2 @@
+package arrow.generic.instances
+
```
pakoito: This file seems to be empty

modules/core/arrow-generic/src/test/kotlin/arrow/generic/ProductTest.kt
```diff
@@ -0,0 +1,154 @@
...
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+import org.junit.runner.RunWith
+
+@product
+data class Person(val name: String, val age: Int, val related: Option<Person>) {
+    companion object
+}
+
+fun personGen(): Gen<Person> = with(Gen) {
```
pakoito: Oh, I didn't know about this one. Nice!

modules/docs/arrow-docs/build.gradle
```diff
@@ -2,7 +2,11 @@ apply plugin: 'ank-gradle-plugin'
 apply plugin: 'org.ajoberstar.git-publish'
 
 dependencies {
-    compile 'io.arrow-kt:ank-core:0.7.0-20180405.163901-2'
+    compile ('io.arrow-kt:ank-core:0.7.0-20180405.163901-2') {
+        exclude group: 'io.arrow-kt', module: 'arrow-core'
```
pakoito: Smart. Good chance we can avoid breaking by using these

modules/docs/arrow-docs/docs/_data/menu.yml
```diff
@@ -257,6 +257,9 @@ options:
       - title: Each
         url: /docs/optics/each/
 
+  - title: Generic
+    url: /docs/generic/
+
```
pakoito: Can you please add one nested option called `Intro` or something similar? We can fill it in tomorrow.

pakoito: I'm not sure clicking this menu option will go to `/docs/generic/` as described below. I believe it needs a submenu.

pakoito:
Added a couple of projects. The PR needs to be reformatted to 2 spaces, to be consistent with the rest of the repo. Once that's ready, let's merge!

raulraja:
@pakoito doesn't detekt auto formats to two spaces? What's the command to format it?

pakoito:
@raulraja that's intellij. Right click on a project folder and select `Format Code`. Make sure not to optimise or mess with the imports.

modules/core/arrow-generic/src/main/kotlin/arrow/generic/Generic.kt
raulraja: I'm not sure any of those names convey the fact that is bidirectional and preserves structure. In any case `Iso` would be a better name but we already have an Iso in Optics. This is not yet in use and I may discard it, I'll move it to type classes if we end up using it for the generic derivation of arbitrary type classes.

PR #751, opened 2018-03-29
simplify settings.gradle by tutteli's settings plugin

robstoll:
For this to work, we have to turn the settings.gradle.kts into a regular settings.gradle (IMO its worth it regarding how much better it reads). Afterwards we have to:
- rename rootProject to arrow (this way we can benefit of a common prefix)
- apply the ch.tutteli.gradle.settings plugin
  - simplify includes with the include extension


robstoll:
No worries and sorry for the missing module

PR #759, opened 2018-04-03
Turns out FunctionK is not a typeclass

pakoito:
Rolling manual FreeEq instances instead.


pakoito:
*sigh* we have to do the Ank Update Dance again. I'll merge, Ank updates, we update. It should all be good.

PR #768, opened 2018-04-06
Semigroup `+` syntax

raulraja:
Adds `+` to all `A` for which a `Semigroup` instance exists

```kotlin
operator fun A.plus(b: A): A =
  this.combine(b)
```


pakoito:
@JorgeCastilloPrz 

`data class ListK<out A>(val list: List<A>) : ListKOf<A>, List<A> by list`

The rule for extfuns is that members take precedence over extfuns. In this case there are two extfuns for plus, the one you're already in and the one in the standard library. It always picks up the one in the scope, causing a stack overflow. You have to import the one on the standard library with an alias to disambiguate.

JorgeCastilloPrz:
I see. Thanks for clarifying.

PR #774, opened 2018-04-08
Arrow 0.7 War Room, day 2

pakoito:
Tasklist, prioritised (X means completed): https://pastebin.com/ptLHjwQu

## Code
 
- [x] Move invoke functions inside typeclasses
- [x] Make sure that all typeclass requirements are the first parameter in data types
- [ ] Make sure that all typeclasses are implemented as extfuns on Kind, except for the constructors
- [ ] All existing KDocs match their method signature
- [x] Get https://github.com/arrow-kt/arrow/pull/669 in
- [x]  2 spaces everywhere
- [x] Remove unnecessary labels this@blabla
- [ ] BONUS: Add KDocs at will
- [ ] BONUS IF YOU WANT TO MAKE PACO HAPPY: everything that's an expression body starts on the next line EXCEPT methods that start with `TYPECLASS.run {`. Those stay on the same line.
- [x] Relocate `Const` to avoid `free` dependencies on `optics` and `mtl` modules
 
## Docs
 
- [x] Redo the typeclass intro doc
- [x] Redo the glossary
- [ ] Write a section on Dependency Injection for new typeclass encodign
- [ ] List typeclasses available for platform types
- [x] All datatypes list the typeclasses they implement an instance for
- [ ] All typeclasses list the datatypes they're implemented for
- [x] Re-read all the typeclasses and see if they make sense with the new encoding
- [ ] Re-read all code examples in datatypes and see if they make sense with the new encoding
- [ ] Syntax sections only where necessary: Option, Either, Try, IO...
- [x] Review examples is Kleisli, as they apparently don't compile
- [ ] Add shapeless license?
- [ ] BONUS: List which module (core/data) each datatype belongs to
- [ ] BONUS: easy docs, i.e. ListK, MapK, SortedMapK
- [ ] BONUS: Docs on funKtionale helpers
- [x] Uberto's findings (COMPLETED): https://pastebin.com/kwbUEuDW

### DIFF
--- a/modules/docs/arrow-docs/docs/docs/datatypes/basic/README.md
+++ b/modules/docs/arrow-docs/docs/docs/datatypes/basic/README.md
@@ -1,0 +1,54 @@
+---
+layout: docs
+title: Basic Types
+permalink: /docs/datatypes/basic/
+---
+
+## Basic Types
+
+Arrow provides [typeclass]({{ '/docs/patterns/glossary/' | relative_url }}) instances for several platform types.
+These instances are available in the module `arrow-instances`.
+
+### Numbers
+
+- [`Show`]({{ '/docs/typeclasses/show/' | relative_url }})
+
+- [`Eq`]({{ '/docs/typeclasses/eq/' | relative_url }})
+
+- [`Order`]({{ '/docs/typeclasses/order/' | relative_url }})
+
+- [`Semigroup`]({{ '/docs/typeclasses/semigroup/' | relative_url }})
+
+- [`Monoid`]({{ '/docs/typeclasses/monoid/' | relative_url }})
+
+### String
+
+- [`Show`]({{ '/docs/typeclasses/show/' | relative_url }})
+
+- [`Eq`]({{ '/docs/typeclasses/eq/' | relative_url }})
+
+- [`Order`]({{ '/docs/typeclasses/order/' | relative_url }})
+
+- [`Semigroup`]({{ '/docs/typeclasses/semigroup/' | relative_url }})
+
+- [`Monoid`]({{ '/docs/typeclasses/monoid/' | relative_url }})
+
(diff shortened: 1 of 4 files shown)


modules/docs/arrow-docs/docs/docs/patterns/dependencyinjection/README.md
```diff
@@ -4,6 +4,479 @@ title: Dependency Injection
...
+
+We have the same benefits and pitfalls we had in methods inside a class.
+Once inside the `run` blocks, we're able to call all other scoped extension functions declared for the enclosing typeclass.
+This disambiguation only happens in one of the layers, who's responsible for the ambiguity in the first place!
+
+Cool, once we have all the pieces in place, explained and understood, let's mention one last use case: general dependency injection!
+
+## Using DI to inject any object
+
+Nnow that you know how to define new types to declare multiple dependencies, and add fields to prevent collision between typeclasses in the same hierarchy,
```
nomisRev: Now

modules/docs/arrow-docs/docs/docs/datatypes/basic/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+- [`Eq`]({{ '/docs/typeclasses/eq/' | relative_url }})
+
+- [`Order`]({{ '/docs/typeclasses/order/' | relative_url }})
+
+- [`Semigroup`]({{ '/docs/typeclasses/semigroup/' | relative_url }})
+
+- [`Monoid`]({{ '/docs/typeclasses/monoid/' | relative_url }})
+
+### String
```
raulraja: We should probably think of a way to automate this since maintaining instances like this is going to become a pain and can send users the wrong message of certain instances not being available for certain data types.

PR #793, opened 2018-04-14
Fix parameter order in extension functions in arrow-instances-core

pakoito:


pakoito:
@raulraja sequence is there on Traverse https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Traverse.kt#L23 

Or you mean for the instances?

raulraja:
I mean as extensions to the kinded versions of the actual data types. `ListK` for example has no `sequence`. All data types that can implement `traverse` and `sequence` should expose it as an operation over their concrete self without relying on the `Traverse` instance to derive it from `traverse` because in many cases people want to `sequence` now just simple concrete values like `ListK<DeferredK<A>> -> DeferredK<ListK<A>>` and they are using `traverse` because is the only combinator `ListK` exposes directly.

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/id.kt
```diff
@@ -101,8 +102,8 @@ interface IdFoldableInstance : Foldable<ForId> {
     fix().foldRight(lb, f)
 }
 
-fun <A, G, B> Id<A>.traverse(f: (A) -> Kind<G, B>, GA: Applicative<G>): Kind<G, Id<B>> = GA.run {
-  f(fix().value).map({ Id(it) })
+fun <A, G, B> Id<A>.traverse(GA: Applicative<G>, f: (A) -> Kind<G, B>): Kind<G, Id<B>> = GA.run {
```
nomisRev: `IdOf<A>` ?

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/try.kt
```diff
@@ -99,7 +100,7 @@ interface TryFoldableInstance : Foldable<ForTry> {
     fix().foldRight(lb, f)
 }
 
-fun <A, B, G> Try<A>.traverse(f: (A) -> Kind<G, B>, GA: Applicative<G>): Kind<G, Try<B>> = GA.run {
+fun <A, B, G> Try<A>.traverse(GA: Applicative<G>, f: (A) -> Kind<G, B>): Kind<G, Try<B>> = GA.run {
```
nomisRev: `TryOf<A>`?

PR #803, opened 2018-04-17
Improve Optics DSL with support for sum types and enable Each/At

nomisRev:
Quite a lot of stuff here :)

This will enable generation of DSL for sum types.

i.e. A (part of) JSON ADT

```kotlin
@optics sealed class Json 
@optics data class JsString(val string: string): Json()
@optics sealed class JsNumber : Json()  
@optics data class JsInt(val int: Int) : Json() 
@optics data class JsDouble(val double: Double) : Json() 
@optics data class JsArray(val value: List<Json>) : Json()

val jsArrayEach: Each<JsArray, Json> = object : Each<JsArray, Json> {
  override fun each(): Traversal<JsArray, Json> = jsArrayIso() compose ListTraversal()
}

val json: Json = JsArray(listOf(
  JsInt(1),
  JsString("Hello"),
  JsInt(2),
  JsInt(3),
  JsDouble(4.0)
))

json.setter().jsArray.every(jsArrayEach).jsInt.int.modify { it + 1 }
//JsArray(value=[JsInt(int=2), JsString(value=Hello), JsInt(int=3), JsInt(int=4), JsDouble(double=4.0)])
```

This also already shows the power of `every` which relies on `Each`. The name `every` comes from a discussion a while back for Helios, back then we still had lookups. Also in Helios we can hide `Each` from the user. I am however unsure if we should still name it `every` over `each` since the use of `Each` is much more prominent now and I'd prefer not to rename the `Each` typeclass. Thoughts?

Use of `At`:

```kotlin
@optics data class Db(val content: MapK<Keys, String>)

sealed class Keys
object One : Keys()
object Two : Keys()
object Three : Keys()
object Four : Keys()

val db = Db(mapOf(
  One to "one",
  Two to "two",
  Three to "three",
  Four to "four"
).k())

db.setter().content.at(MapK.at(), One).some.modify(String::reversed)
//Db(content=MapK(map={arrow.optics.One@a09ee92=eno, arrow.optics.Two@30f39991=two, arrow.optics.Three@452b3a41=three, arrow.optics.Four@4a574795=four}))
```

Finally I also did an attempt to improve error messages in the code gen.

@pakoito I did not add any way to enable `each.run {  data.list.every() } ` since I thought it didn't belong in the `Each`typeclass.


modules/core/arrow-annotations-processor/src/main/java/arrow/common/utils/ProcessorUtils.kt
```diff
@@ -47,13 +47,14 @@ interface ProcessorUtils : KotlinMetadataUtils {
   }
 
   fun getClassOrPackageDataWrapper(classElement: TypeElement): ClassOrPackageDataWrapper {
-    val metadata = classElement.kotlinMetadata ?: knownError("These annotations can only be used in Kotlin")
+    val metadata = classElement.kotlinMetadata ?: knownError("Arrow's annotations can only be used on Kotlin classes. Not valid for $classElement")
+
     return metadata.asClassOrPackageDataWrapper(classElement)
-      ?: knownError("This annotation can't be used on this element")
+      ?: knownError("Arrow's annotation can't be used on $classElement")
```
JorgeCastilloPrz: can't you get the potential binary name for this element (class name) from the `TypeElement` to include it in the error instead of the TypeElement instance reference? (`ProcessingEnvironment.getElementUtils().getBinaryName(TypeElement)`).

modules/core/arrow-annotations-processor/src/main/java/arrow/optics/BoundSetterGenerator.kt
```diff
@@ -41,16 +41,16 @@ class BoundSetterGenerator(
...
         |/**
         | * @receiver [${annotatedOptic.sourceClassName.removeBackticks()}] the instance you want to bind the dsl on.
         | * @return [$boundSetter] an intermediate optics that is bound to the instance.
         | */
         |fun ${annotatedOptic.sourceClassName}.setter() = $boundSetter(this, arrow.optics.PSetter.id())
         |""".trimMargin()
 
-  fun processBoundSetter(sourceClassName: String, targetName: String, targetClassName: String, sourceName: String) = """
-      |inline val <T> $boundSetter<T, $sourceClassName>.$targetName: $boundSetter<T, $targetClassName>
+  private fun processBoundSetter(sourceClassName: String, targetName: String, targetClassName: String, sourceName: String) = """
```
JorgeCastilloPrz: maybe reorder to `sourceName, sourceClassName, targetName, targetClassName` ? The args feel quite disordered now

modules/core/arrow-annotations-processor/src/main/java/arrow/optics/OpticsProcessor.kt
```diff
@@ -37,43 +39,19 @@ class OpticsProcessor : AbstractProcessor() {
...
-                logE("Only data classes can have DSL target", element); emptyList()
-              }
-
-            }
-          else -> targets
-        }
-
-        normalizedTargets.forEach { target ->
-
+        element.normalizedTargets().forEach { target ->
```
JorgeCastilloPrz: nice cleaning

modules/core/arrow-annotations-processor/src/main/java/arrow/optics/OpticsProcessor.kt
```diff
@@ -145,39 +109,84 @@ class OpticsProcessor : AbstractProcessor() {
...
     }
 
-    else -> {
-      logE("Isos can only be generated for data classes", element)
-      null
-    }
+    else -> null.also { logE(element.isoErrorMessage, element) }
   }
 
+  private val Element.otherClassTypeErrorMessage
```
JorgeCastilloPrz: Not a blocker, but could be better to move these to a separated `ErrorMessages.kt`or `ErrorSyntax.kt` file, IMO. After all, these functions don't seem highly tied / related to the processor class itself.

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/either.kt
```diff
@@ -74,7 +74,7 @@ fun <G, A, B, C> EitherOf<A, B>.traverse(f: (B) -> Kind<G, C>, GA: Applicative<G
 interface EitherTraverseInstance<L> : EitherFoldableInstance<L>, Traverse<EitherPartialOf<L>> {
 
   override fun <G, A, B> Kind<EitherPartialOf<L>, A>.traverse(AP: Applicative<G>, f: (A) -> Kind<G, B>): Kind<G, Kind<EitherPartialOf<L>, B>> =
-    fix().traverse(f, AP)
+    traverse(f, AP)
```
JorgeCastilloPrz: is this due to typeclass methods defined as extfuns on top of kinds now?

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/try.kt
```diff
@@ -99,7 +99,7 @@ interface TryFoldableInstance : Foldable<ForTry> {
     fix().foldRight(lb, f)
 }
 
-fun <A, B, G> Try<A>.traverse(f: (A) -> Kind<G, B>, GA: Applicative<G>): Kind<G, Try<B>> = GA.run {
+fun <A, B, G> TryOf<A>.traverse(f: (A) -> Kind<G, B>, GA: Applicative<G>): Kind<G, Try<B>> = GA.run {
```
JorgeCastilloPrz: Ok, I guess it is.

modules/docs/arrow-docs/docs/docs/optics/syntax/README.md
```diff
@@ -12,14 +12,10 @@ To avoid boilerplate Arrow can generate this property-like dsl using `@optics` a
...
 
-@optics
-data class Street(val number: Int, val name: String)
-@optics
-data class Address(val city: String, val street: Street)
-@optics
-data class Company(val name: String, val address: Address)
-@optics
-data class Employee(val name: String, val company: Company?)
+@optics data class Street(val number: Int, val name: String)
```
JorgeCastilloPrz: Is it mandatory to annotate each one of the implementations of the sum type (sealed class) to generate the optics for each ? Shouldn't we be able to generate optics for all the implementations on the sealed hierarchy if you just annotate the parent sealed class itself?

modules/docs/arrow-docs/docs/docs/optics/syntax/README.md
```diff
@@ -33,6 +29,35 @@ val employee = Employee("John Doe", Company("Kategory", Address("Functional city
...
+when (networkResult) {
+  is HttpError -> networkResult.copy(f(networkResult.message))
+  else -> networkResult
+}
+```
+
+We can rewrite this code with our generated dsl.
+
+```kotlin:ank
+networkResult.setter().networkError.httpError.message.modify(f)
```
JorgeCastilloPrz: why is it `setter().networkError.httpError` and not `setter().httpError` ? How `networkError` is required here if we just want to set the new modified message value when the type is `HttpError` ?

modules/optics/arrow-optics/src/main/kotlin/arrow/optics/docs_wip.kt
```diff
@@ -0,0 +1,76 @@
+package arrow.optics
```
JorgeCastilloPrz: why is this docs helper file (or that's how it looks like) inside of the optics module and not the docs one?

modules/docs/arrow-docs/docs/docs/optics/syntax/README.md
nomisRev: That is because of the nested hierarchy of the sealed classes.

We need to compose `Prism<NetworkResult, NetworkError>` with `Prism<NetworkError, HttpError>` to get `Prism<NeworkResult, HttpError>`.

modules/docs/arrow-docs/docs/docs/optics/syntax/README.md
nomisRev: Yes, it is. Otherwise, it might get complex really quickly.

Given the network example, what is generated?

```
@optics sealed class NetworkResult
data class Success(val content: String): NetworkResult()
sealed class NetworkError: NetworkResult()
data class HttpError(val message: String): NetworkError()
object TimeOut: NetworkError()
```

Does it generate `Prism<NetworkResult, Success>`, `Prism<NetworkResult, NetworkError>` and everything applicable to `Success`? Or does it also generate everything for `HttpError` and Prisms for `NetworkError`?

What happens if I pass arguments to `@optics`? i.e. what happens if I change it to `@optics([LENS])`? Does it fail because I cannot generate `Lens` for NetworkResult? Or does it generate a `Lens` for `Success` and doesn't complain? What happens in same scenario when `sealed class` only contains `object` and thus cannot generate any `Lens`?

We will create a ton of edge cases which I think will be extremely confusing for the users.

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/try.kt
nomisRev: Actually, it's already fixed in #793.

modules/core/arrow-annotations-processor/src/main/java/arrow/common/utils/ProcessorUtils.kt
nomisRev: Actually for

```kotlin
@instance(String::class)
interface StringEqInstance : Eq<String> {
  override fun String.eqv(b: String): Boolean = this == b
}
```

it prints `e: error: Arrow's annotations can only be used on Kotlin classes. Not valid for java.lang.String`. Not sure if that's what you wanted?

`toString()` is correctly implemented for `Element` and it thus prints the class names.

JorgeCastilloPrz: no, I was expecting to get the same but for the annotated class itself, like `arrow.StringEqInstance`. Are you calling it for the `classElement` argument?

modules/docs/arrow-docs/docs/docs/optics/syntax/README.md
JorgeCastilloPrz: Fair enough

modules/docs/arrow-docs/docs/docs/optics/syntax/README.md
JorgeCastilloPrz: Yeah I've noticed now that there's a second level of inheritance there.

modules/core/arrow-annotations-processor/src/main/java/arrow/common/utils/ProcessorUtils.kt
```diff
@@ -47,7 +47,8 @@ interface ProcessorUtils : KotlinMetadataUtils {
   }
 
   fun getClassOrPackageDataWrapper(classElement: TypeElement): ClassOrPackageDataWrapper {
-    val metadata = classElement.kotlinMetadata ?: knownError("Arrow's annotations can only be used on Kotlin classes. Not valid for $classElement")
+    val metadata = classElement.kotlinMetadata
```
JorgeCastilloPrz: what's the code style we're using for this repo?

nomisRev: Not sure. I reformatted project with 2 spaces in my last commit and it moved this on two lines. Since I was unsure what the project style is for this case I left it so.

modules/core/arrow-annotations-processor/src/main/java/arrow/common/utils/ProcessorUtils.kt
nomisRev: Yes, that makes sense. We need to improve error message.

I tried to write an `Each` instance for `String` and got `This annotation can't be used on this element`. So I made a small change to improve what we currently have.

We should, however, improve error message for all processors but that doesn't belong in this PR. I would go as far as provide clear custom messages for each error case.

In this case something like.

```
e: error: @instance cannot be used for java.lang.String (or even better kotlin.String)

interface StringEqInstance : Eq<String>
                  ^

Cannot generate instance method for arrow.data.instances.StringEqInstance.
```

nomisRev:
@raulraja I added scoped syntax as discussed on Slack. https://github.com/arrow-kt/arrow/pull/803/files#diff-43ed86ff2970e9668758930fe702b695R81

I initially didn't add any scoped syntax for `Each` and `At` because it creates a dependency on `BoundSetter`. Which can be argued that it might not belong in a typeclass. However, we decided that the benefit outweighs the possible downside.

raulraja:
Awesome cleanup :clap:

JorgeCastilloPrz:
Looks good @nomisRev :)

Issue #766, opened 2018-04-06
Do not throw on processors


JorgeCastilloPrz:
Some processors are throwing errors instead of using the messager for that. Ideallly they should use the messager which we're already using on different parts. Explanations [here](https://github.com/arrow-kt/arrow/pull/747#discussion_r179699753)

nomisRev:
Actually, in [AbstractProcessor](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-annotations-processor/src/main/java/arrow/common/utils/AbstractProcessor.kt) we catch `KnownException` and use the messager.

Still not ideal but better than not using it I guess. I was looking into how we could offer better error messages but this requires quite some changes.

i.e. `@autofold` for an untyped `Maybe`

```kotlin
sealed class Maybe
data class Just<B>(val b: B): Maybe()
object Nothing: Maybe()
```

You would want an error message like.

```
@autofold cannot create a fold method for sealed class Maybe
sealed class Maybe
                                 ^ contains less generic information than a variant 

is Just<A> -> just(this)
              ^  Cannot check for an instance of an erased type: Just<A>
```

I've accomplished above error message for autofold on a local branch but it relies on `knownError` and the implementation is still quite dirty.

Some bigger changes are required in our meta code for testing, maintainability and error messages IMHO. I see it growing/evolving at a increasing speed with `optics`, typeclasses and `generic`.

raulraja:
Closing old issues related to old kapt processors. thanks @JorgeCastilloPrz @nomisRev

PR #806, opened 2018-04-19
Add error message to autofold

### LINKED ISSUE #462: Error message autofold
Improve error messages for sealed classes that fold can not be generated for.

```kotlin
sealed class Test<A, B>
data class ATest<C, D>(val unit: Unit): Test<Nothing, Nothing>()
```

```kotlin
sealed class Test2<A>
data class ATest<A, B>(val unit: Unit): Test<A>()
```

nomisRev:
This closes #462 

@JorgeCastilloPrz @raulraja As discussed in #766 I am not sure how we should proceed with throwing `KnownException`. I did a small spike with the scenario handled in this PR and had to rely on `MonadError` and `Traverse` for it to be somewhat manageable.


nomisRev:
@jereksel I tried adding some tests for the processor but ran into some issues. If I understand correctly I should.

1. Add annotated targets to `:arrow-annotation-processor-test-models`
2. Add expected generated result to `arrow-annotation-processor/src/test/resources`
3. Add test under `arrow-annotation-processor/src/test`

Since `@autofold` generates extension functions I have a dependency on `:arrow-annotation-processor-test-models` in resources. This was not working for me.

Any ideas? Am I missing something?

jereksel:
@nomisRev Try removing "build" directory in models. When you add new classes kapt often doesn't transform them.

raulraja:
I think it's ok to throw exceptions since this should fail fast at compile time but I don't have an strong opinion since I'm actually more interested in removing the entire kapt API we use all together for something that wraps the Proto stuff which is tedious as hell to deal with. I'm also realizing that `arrow-generic` will provide also this same feature when we have `@coproduct sealed ...` since fold is one of the ops in `@coproduct` and soon will be in `@product` because of the `HList` <-> `data class` Iso.

Maybe then we should consider merging these processors and annotations to be part of `arrow-generic`

nomisRev:
@jereksel I doubled checked everything but still get `FileNotFoundException`. Although I can see the java stub under `.../models/build/tmp/kapt3/stubs/main/arrow/ap/objects/autofold`.

Test fail with following error `java.io.FileNotFoundException: /var/folders/y6/3f_jn1qd2xg6kkgsnmf2_p680000gp/T/1524998582721-0/autofold (Is a directory)`

I am unsure why it's trying to look in that location. Any ideas?

jereksel:
@nomisRev That means that AP passed fine, but checking of output file failed.

nomisRev:
@jereksel I have no clue how to make it find the directory. The same thing is done for Optics. I'd like to close this PR, if we cannot fix tests I prefer to remove them for now.

PR #811, opened 2018-04-22
Direct type class `syntax` over concrete data types

raulraja:
The following PR adds a `syntax` DSL for all data types that arrow exposes type class instances for.

This DSL allows users to not worry about knowing if `binding`, `flatMap` etc are in `Monad`, or what any other combinator they plan on using belongs to a certain type class.

It eliminates the need to explicitly refer to the instances inside the data type companion.

Until now users where forced to do things like:
```kotlin
Option.applicative().map(Option(1), Option(1), Option(1), { (a, b, c) -> a + b + c }) //Option(3)
Option.monad().binding {
  val a = Option(1).bind()
  val b = Option(a + 1).bind()
  a + b
} 
// Option(3)
Option.functor()...
Option.traverse()...
```
Instead with the syntax DSL most used type classes that the data type provides instances for are bundled as the single `this` scope of the contained `syntax` function.
The scope contains all extension functions and static functions that the type classes define and are available to use within the user provided block.
```kotlin
import arrow.instances.*

Option syntax { //`this` is Monad, Applicative, Functor, Traverse, etc...
  binding { ... }
  map(....)
  traverse(...)
}

Either<String>() syntax { //`this` is Monad, Applicative, Functor, Traverse, etc...
  binding { ... }
  map(....)
  traverse(...)
}
```

The docs contain now instructions for users wanting to provide a similar DSL for their custom data type and it looks like this:

> If you defined your own instances over your own data types and wish to use a similar `syntax` DSL you can do so for both types with a single type argument such as `Option`:

```kotlin
object OptionContext : OptionMonadErrorInstance, OptionTraverseInstance {
  override fun <A, B> Kind<ForOption, A>.map(f: (A) -> B): Option<B> =
    fix().map(f)
}

infix fun <A> Option.Companion.syntax(f: OptionContext.() -> A): A =
  f(OptionContext)
```

> Or for types that require partial application of some of their type arguments such as `Either<L, R>` where `L` needs to be partially applied

```kotlin
class EitherContext<L> : EitherMonadErrorInstance<L>, EitherTraverseInstance<L> {
  override fun <A, B> Kind<EitherPartialOf<L>, A>.map(f: (A) -> B): Either<L, B> =
    fix().map(f)
}

class EitherContextPartiallyApplied<L> {
  infix fun <A> syntax(f: EitherContext<L>.() -> A): A =
    f(EitherContext())
}

fun <L> Either(): EitherContextPartiallyApplied<L> =
  EitherContextPartiallyApplied()
```


modules/docs/arrow-docs/docs/docs/patterns/glossary/README.md
```diff
@@ -86,6 +88,93 @@ Either.monadError<Throwable>()
 ListK.traverse()
 ```
 
+# Syntax
```
pakoito: For syntax to match the other headers, it needs to be ###

modules/docs/arrow-docs/docs/docs/patterns/glossary/README.md
```diff
@@ -86,6 +88,93 @@ Either.monadError<Throwable>()
...
+class EitherContextPartiallyApplied<L> {
+  infix fun <A> syntax(f: EitherContext<L>.() -> A): A =
+    f(EitherContext())
+}
+
+fun <L> Either(): EitherContextPartiallyApplied<L> =
+  EitherContextPartiallyApplied()
+```
+
 If you're defining your own instances and would like for them to be discoverable in their corresponding datatypes' companion object, you can generate it by annotating them as `@instance`, and Arrow's [annotation processor](https://github.com/arrow-kt/arrow#additional-setup) will create the extension functions for you.
```
pakoito: This block belongs to the Instances section, before the Syntax one.

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/either.kt
```diff
@@ -110,3 +110,16 @@ interface EitherShowInstance<L, R> : Show<Either<L, R>> {
...
+    fix().map(f)
+}
+
+class EitherContextPartiallyApplied<L> {
+  infix fun <A> syntax(f: EitherContext<L>.() -> A): A =
+    f(EitherContext())
+}
+
+fun <L> Either(): EitherContextPartiallyApplied<L> =
+  EitherContextPartiallyApplied()
```
pakoito: Can we hide the partial implementation?

```kotlin
  interface ContextPartiallyApplied<T> {
    infix fun <A> syntax(f: T.() -> A): A
  }

fun <L> Either(): ContextPartiallyApplied<EitherContext<L>> =
  object: ContextPartiallyApplied {
    infix fun <A> syntax(f: EitherContext<L>.() -> A): A =
      f(EitherContext())
  }
```

pakoito:
To align with the existing documentation I'd rather the DSL objects were suffixed `Syntax` rather than `Context`.

pakoito:
Another suggestion I can think of is to move the functions to the companion object of the witness types. That way we can get consistent syntax across types, i.e. `ForEither() syntax { }` and `ForOption() syntax`.

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/either.kt
raulraja: yep

raulraja:
@pakoito Many types use now `syntax` that are not kinded and have no witness. For example:
```kotlin
String syntax {
  "a".combine("b")
}
// ab
```

ersin-ertan:
Proposal to change `syntax` to `instances` because:
- `instances` plural, when paired with a data type named as a singular ex.`Option instances` will hopefully be enough to avoid naming clashes with the OOP concept of instantiation(though one may still wrongly assume that a list of instances is being returned)
- `instances` acts like special version of Kotlins scoped function `with(...)`, allowing you to call the functions of the data type's available instances, so instead of writing `withAvailableInstancesOf(Option) { ... }`, the word `instances` provides enough context for the representation

pakoito:
What about `extensions`? `ForOption() extensions { ... }`

nomisRev:
I'm a little unsure about instances because when I open a `ObservableK instances` block I'd expect to also have access to `effects` instances

raulraja:
@nomisRev that has access to the Effect instance.
```kotlin
ForObservableK instances { 
  // `this` is also an `Effect<F>`
}
```
https://github.com/arrow-kt/arrow/pull/811/files#diff-088c779bf125135ce8dbe42b7af6d477R119

nomisRev:
Ah yeah of course. That data type only lives in effects... My bad :D

I meant if I manually add a instance. Or typeclasses/instances defined in Optics, Helios,..

raulraja:
@nomisRev For those cases we are providing instructions to users as part of this PR that tells them how to create the same DSL including their custom instances: 
https://github.com/arrow-kt/arrow/pull/811/files#diff-f6d7de9e73be8613678763d3780567e2R152

raulraja:
@pakoito @ersin-ertan @nomisRev 
Seems most people I talked to prefer `syntax` or `extensions` over `instances`. also @pakoito proposed prefixing all types with `For` to work around missing companions for example in `Boolean`.

Does this look ok to everyone and a fair compromise for now?

```kotlin
ForOption extensions { ... }
ForIO extensions { ... }
ForBoolean extensions { ... }
```

nomisRev:
Looks good to me! Also thanks for pointing out the entry in docs 👍

Issue #815, opened 2018-04-23
[v-0.0.7] arrow.syntax.tuples.plus operator fun no more available in version 0.0.7


jtonic:
Hi,

After the migration from 0.6.1 to 0.7.0 I noticed that the arrow.syntax.tuples.plus is no longer available.
More or less its signature looked like:

`operator fun Tuple2<A,B>.plus(c: C): Tuple3<A,B,C> {...}`

And I think it was available for all Tuples (even if I used it only for Tuple2, and Tuple3).

It would be better to have it back in arrow, I found it very useful.

Kind regards

PR #818, opened 2018-04-24
Add "plus" method for Tuples

jereksel:
implements #815


modules/core/arrow-syntax/src/main/kotlin/arrow/syntax/collections/tuple.generator.kts
```diff
@@ -0,0 +1,48 @@
...
+  val constructor = (letters.dropLast(1).map { "this.$it" } + newLetter).joinToString()
+
+  fileContent.append(
+          "operator fun $diamond1 Tuple$i$diamond2.plus($newLetter: ${newLetter.toUpperCase()}): Tuple${i+1}$diamond1 = Tuple${i + 1}($constructor)"
+  )
+  fileContent.append("\n")
+
+}
+
+File("tuple.kt").writeText(fileContent.toString())
```
pakoito: This is too useful to make it throwaway code, we can modify it for other functions. It cannot be shipped with the library tho, can you please move it to...

...to where?

Can you create a folder in the root of the project called scripts and drop it there? Will it even work? How did you invoke it?

raulraja: Those scripts can live in the annotation processor which is not a runtime dependency

raulraja:
I think the generator can be moved to the annotation processor and what is generated either to Tuple in core or in a different syntax place that is not `collections` unless there is other tuple stuff there already.

jereksel:
I don't think it should be in annotation processor - it doesn't process any annotation. I'll create a directory in root and write readme file how to use them - in future we could create gradle task for that.

PR #819, opened 2018-04-25
Safe implementations for `Try` functions: `onSuccess` and `onFailure`

hpoul:
based on #572 by @MarioAriasC - but with a single commit, and mergeable without conflict.

@raulraja still open to merge non-deprecated versions of `onSuccess` / `onFailure`?

### DIFF
--- a/modules/core/arrow-core/src/main/kotlin/arrow/core/Try.kt
+++ b/modules/core/arrow-core/src/main/kotlin/arrow/core/Try.kt
@@ -108,20 +108,14 @@ sealed class Try<out A> : TryOf<A> {
   @Deprecated(DeprecatedUnsafeAccess, ReplaceWith("getOrElse { ifEmpty }"))
   abstract fun get(): A
 
-  @Deprecated(DeprecatedUnsafeAccess, ReplaceWith("map { body(it); it }"))
-  fun onSuccess(body: (A) -> Unit): Try<A> {
-    foreach(body)
-    return this
-  }
+  fun onSuccess(body: (A) -> Unit): Try<A> =
+          flatMap { Try { body(it); it } }
 
-  @Deprecated(DeprecatedUnsafeAccess, ReplaceWith("fold ({ Try { body(it); it }}, { Try.just(it) })"))
-  fun onFailure(body: (Throwable) -> Unit): Try<A> = when (this) {
-    is Success -> this
-    is Failure -> {
-      body(exception)
-      this
-    }
-  }
+  fun onFailure(body: (Throwable) -> Unit): Try<A> =
+          fold({
+            body(it)
+            Try.raise(it)
+          }, { Try.just(it) })
 
   fun toOption(): Option<A> = fold({ None }, { Some(it) })
 


raulraja:
@hpoul I think this functions are fine if we have a proper strategy to explicitly namespace unsafe functions like this one despite the branch name :)
`onSuccess` and `onError` are unsafe because they both perform effects which are then ignored which is fine if someone wants to do that but being Arrow a library of functional programming we want such utility functions to not be invoked unless the user is aware of what is about to do.

I think a good compromise is to have an `unsafe` package on each module where such utility functions can live. Users will be forced to import these a la carte.
```
import arrow.core.unsafe.*
Try { ...}.onSuccess { println("succeded") } //this is fine
```
```
import arrow.core.*
Try { ...}.onSuccess { println("succeded") } //not ok unless unsafe is imported.
```
Thoughts?

hpoul:
@raulraja I'm not sure i'm completely sold on the `unsafe` idea—what I'm not grasping is how this is less safe than letting users directly call `Try { ... }.fold( ..., ...)`? I guess it just needs a definition what the methods do in case of (another) exception?

raulraja:
Fold is defined such as a return value is expected. If the user chooses Unit is their responsibility but with such a return type all you can do is perform side effects.
```
fun <B> fold(onError: (Throwable) -> B, onSuccess: (A) -> B): B
```
```
fun onSuccess(f: (A) -> Unit): Try<A>
```
When Arrow provides a combinator such as `onSuccess` and `onFailure` where the only useful thing you can returns is `Unit` we are encouraging a programming model that is against good practices in Functional Programming. That is because functions that return Unit all they can do is perform explicit effects without a controlled context. Arrow provides 3 other data types with contexts suitable for effect control and those are `IO`, `DeferredK` and `ObservableK`.

The moment you make a function in your system use `onSuccess` and `onError` you make your entire program downstream to the edge impure, for example:

```kotlin
fun load(user: UserId): Try<UserId>=
  db.load(user).onSuccess { sendNukes() }
```

There is nothing in the function signature above that indicates that upon loading a user nukes will be sent. You can tell the function is impure right away because you would have to mock some context to get to what `sendNukes` did in the outside world outside of the domain of the function. 

IMO It's fine if the author wants that but we can't encourage methods that create impurity downstream for all dependent callers of load because it's not a style we want to encourage and if we allow them in the library it has to be explicitly acknowledge by the user that those methods are unsafe. This is the same case as an unsafe system like java in terms of null handling and warts in other data types such as `Option.get` that blow up at run time. They are convenient methods for some styles of programming but we have the responsibility to provide an API in arrow-core that is as pure as possible.

I also feel this does not bring much to the table beyond what you can already do. Both the `Success` and `Failure` constructors are public types and encouraged to be used for people that require these kinds of things.
```
when (op) { is Success -> println("success") }
```
IMO If we accept these methods in Arrow they should be explicit opt in by users and not magically appear in the regular namespace which attempts to remain as pure as possible.

hpoul:
@raulraja thanks, I guess I just have to rethink how i'm doing stuff.. I actually only need `onFailure` to basically log stuff.. you're right that this is obviously impure.. do you have a hint how to do this, maybe with IO? `Try { ... }.onFailure { e -> logger.warning("Something went wrong, I just care enough to log about it", e) }`

(although thinking about it.. `Try` is some weird thing anyway, it's hard to think about any pure use case, because exceptions make only sense when the function itself is somehow communicating with the outside world.. so your example is already discouraged before reaching the `onSuccess` method since `db.load(user)` isn't a pure function anyway?)

raulraja:
You can use `IO` instead of `Try` since `IO` also supports async stuff whereas `Try` is only useful in sync contexts. Also `Try` is just an specialization of `Either<Throwable, A>`. `IO` provides instances of `MonadError` and therefore you already have available: `raiseError(e: Throwable)` handleError, etc... also all effects explicitly encapsulated by `IO` are suspended and defered until you are ready to execute them. You have them the guarantee that calling a function that returns `IO<A>` you will never execute its effects until you explicitly do so in a single place in your application. 

I propose that if you have interest in learning more about IO and Effect control you come up with a small contained example expressed in terms of Try and onSuccess and I or @pakoito  can suggest what it would look like built atop IO so we can go over the differences. 🍻 

Also in case you haven't run into these already some links related to this topic:
http://arrow-kt.io/docs/effects/io/
http://arrow-kt.io/docs/patterns/error_handling/

sbuettner:
@raulraja Is `IO` really the right choice when I just want a `Try` to catch possible exceptions of sync code? How would you model @hpoul´s example of `Try { ... }.onFailure { e -> log...}` without too much ceremony?

JorgeCastilloPrz:
In an effort to clean the PR list we've decided to close all the ones that are older than 2 months and consider them obsolete except for very specific circumstances that we will evaluate per PR. In this case, `Try` is being deprecated and will be removed in further releases, so this changes are obsolete at this point.

PR #825, opened 2018-04-29
Optics as extension on Companion

nomisRev:
Since other generators require `companion object` I propose the same requirement for optics generation. If so then we can generate optics as extension properties on the companion object instead of global functions. This will improve discoverability and place make for an easier to use API.

```kotlin
val optic = (employeeCompany() compose
  companyAddress() compose
  addressStreet() compose
  streetName())
```

will become

```kotlin
val optic = (Employee.company compose
  Company.address compose
  Address.street compose
  Street.name)
```

Thoughts?


jereksel:
This will require declaring empty `compation object` for annotated classes - I don't like that.

nomisRev:
> This will require declaring an empty companion object for annotated classes - I don't like that.

Yes it does and neither do I but we already require it for `@instance` and `@product` too.

Concerning the syntax. I think `Person.name` is a big win over a global function `personName()` both for discoverability as user-friendliness. Searching global function by name convention can be really tedious and IDE support is much more friendly when searching companion objects.

That it looks like accessing a static field is fine for me as you could see it as a composition of functional references. Any other suggestions?

We could also make them `fun` instead of `val`, would that be a better improvement?  I wanted to stay away from pre- or post fixes such as `Person.nameLens` or `Person.lensName`.

jereksel:
So let's do it another way - instead of 
```
val optic = (Employee.company compose
  Company.address compose
  Address.street compose
  Street.name)
```

Maybe we should go with something like
```
val optic = (Employee.company compose
  address compose
  street compose
  name)
```

nomisRev:
Snippet 1 & 2 are the same but with a different set of imports.

I thought about snippet 3 but if we do that it would be a DSL on top of snippet 1 but not replace traditional composition. Just like we do with bound setter.

raulraja:
I like the lenses in the companion more than functions on the package but the concern I have here is  name collisions.
`Employee.name` seems to me like a overly generic namespace that resembles a property but not a lens and it's potentially a namespace anyone could claim on its own for any other purpose regarding manipulation of properties in a data class.

raulraja:
@nomisRev alright, I'm sold :+1:

raulraja:
I also like `val optic = Employee.company.address.street.name` as @jereksel  suggested and if we can do both styles it would be awesome.

nomisRev:
@raulraja I'd really love a DSL like `val optic = Employee.company.address.street.name` but it needs quite a bit of code gen to allow for it to work.

However I think this is a much stronger feature than the current bounded syntax and am not sure if the additionally generated code is an issue given the power this offers. Also with this style syntax I personnaly don't feel the need anymore to support bounded syntax. Thoughts? @raulraja @jereksel @pakoito @JorgeCastilloPrz 

For example to make the following snippet compile
```kotlin
@optics
data class Address(val city: String, val street: Street) {
  companion object
}

@optics
data class Company(val name: String, val address: Address) {
  companion object
}

@optics
data class Employee(val name: String, val company: Company?) {
  companion object
}
val address: Optional<Employee, Address> = Employee.company.address
```

We need to generate following code

```kotlin
/** For Employee::company **/
inline val <S> Iso<S, Employee>.company: Optional<S, Company> get() = this + Employee.company
inline val <S> Lens<S, Employee>.company: Optional<S, Company> get() = this + Employee.company
inline val <S> Optional<S, Employee>.company: Optional<S, Company> get() = this + Employee.company
inline val <S> Prism<S, Employee>.company: Optional<S, Company> get() = this + Employee.company
inline val <S> Setter<S, Employee>.company: Setter<S, Company> get() = this + Employee.company
inline val <S> Traversal<S, Employee>.company: Traversal<S, Company> get() = this + Employee.company
inline val <S> Fold<S, Employee>.company: Fold<S, Company> get() = this + Employee.company

/** For Company::address **/
inline val <S> Iso<S, Company>.address: Lens<S, Address> get() = this + Company.address
inline val <S> Lens<S, Company>.address: Lens<S, Address> get() = this + Company.address
inline val <S> Optional<S, Company>.address: Optional<S, Address> get() = this + Company.address
inline val <S> Prism<S, Company>.address: Optional<S, Address> get() = this + Company.address
inline val <S> Getter<S, Company>.address: Getter<S, Address> get() = this + Company.address
inline val <S> Setter<S, Company>.address: Setter<S, Address> get() = this + Company.address
inline val <S> Traversal<S, Company>.address: Traversal<S, Address> get() = this + Company.address
inline val <S> Fold<S, Company>.address: Fold<S, Address> get() = this + Company.address
```

raulraja:
@nomisRev I think that as long as those val's are not initialized until usage it would be fine regardless of the amount of code. If you think the new DSL would supersede the bounded syntax one then great!. We are in a privileged situation now that won't last much longer as we approach 1.0 and whatever makes the barrier of entry to lenses easier the better.

nomisRev:
@raulraja The to-generate-code is quite trivial so I should be able to add that over the weekend.

Since the generated code are all `inline val` there should be no overhead. I am unsure what happens with `inline val` if they're not used.

Reason I think the new DSL supersedes the bounded syntax is that we're making actual Optics here instead of bounded instances so they live up to their potential. 

Additionally, I think the bounded syntax is not worth the overhead `employee.setter().company.address.street.name.modify(String::toUpperCase)` over

`Employee.company.address.street.name.modify(employee, String::toUpperCase)`

modules/optics/arrow-optics/src/main/kotlin/arrow/optics/dsl/at.kt
```diff
@@ -0,0 +1,84 @@
+package arrow.optics.dsl
```
pakoito: How can this file be accessed, if not from the menu?

nomisRev: You mean `DSL` docs?

nomisRev:
Re-enabling grained control for DSL will be done in #848

nomisRev:
@arrow-kt/maintainers ready for review :)

modules/core/arrow-annotations-processor/models/src/main/java/arrow/ap/objects/iso/IsoCompanion.kt
```diff
@@ -0,0 +1,8 @@
+package arrow.ap.objects.iso
+
+import arrow.core.Option
+import arrow.optics.OpticsTarget
+import arrow.optics.optics
+
+@optics([OpticsTarget.ISO])
+data class IsoCompanion(val field: String, val nullable: String?, val option: Option<String>)
```
pakoito: Companion object?

nomisRev: Test model to test for a compiler error. https://github.com/arrow-kt/arrow/pull/825/files#diff-8bcf2e9e0361c1d797accab0aede2539R32

modules/core/arrow-annotations-processor/models/src/main/java/arrow/ap/objects/iso/IsoSealed.kt
```diff
@@ -6,5 +6,6 @@ import arrow.optics.optics
 
 @optics([OpticsTarget.ISO])
 sealed class IsoSealed(val field: String, val nullable: String?, val option: Option<String>) {
+  companion object {}
```
raulraja: No need for `{}`

modules/docs/arrow-docs/docs/docs/optics/dsl/README.md
```diff
@@ -73,14 +75,14 @@ import arrow.data.*
 val jane = Employee("Jane Doe", Company("Kategory", Address("Functional city", Street(42, "lambda street"))))
 val employees = Employees(listOf(john, jane).k())
 
-employees.setter().employees.every(ListK.each()).company.address.street.name.modify(String::capitalize)
+Employees.employees.every(ListK.each()).company.address.street.name.modify(employees, String::capitalize)
```
raulraja: Is it possible to provide `ListK.each()` automatically based on the property type internally so the user can just do?
```kotlin
Employees.employees.every().company.address.street.name.modify(employees, String::capitalize)
```
Also if we detect types for which we now a wrapper exists for example `List <-> ListK`, is it possible to preserve the user's type `List` while providing internally a `Each` instance for its wrapper type?

modules/optics/arrow-optics/src/main/kotlin/arrow/optics/syntax/BoundSetter.kt
```diff
@@ -1,75 +0,0 @@
...
-import arrow.optics.typeclasses.*
-
-/**
- * A [BoundSetter] is an optic that allows to see into a bound structure and set or modify its focus.
- * It is used to enable a DSL to work with values of immutable structures where the full power of Optics is not required.
- *
- * @param S the source of a [BoundSetter]
- * @param A the focus of a [BoundSetter]
- */
-class BoundSetter<S, A>(val value: S, val setter: Setter<S, A>) {
```
raulraja: :fire: :clap:

modules/docs/arrow-docs/docs/docs/optics/dsl/README.md
nomisRev: @raulraja I tried to find an encoding for this but we need true typeclasses AFAIK.

We can do this only for one `F` due to type erasure and I am unsure if we should favor `List`/`ListK` by adding
`fun <T, A> Lens<T, ListK<A>>.every(): Traversal<T, A> = this compose ListK.traversal()`.

Because then we cannot do it for any other `F` like
`fun <T, A> Lens<T, SetK<A>>.every()` or `fun <T, K, A> Lens<T, MapK<K, A>>.every()`

PR #826, opened 2018-04-29
Add generated files to sources jar

jereksel:


pakoito:
Great find! Merging as soon as it greens :D

PR #831, opened 2018-05-04
Add Katalyst Core

aedans:
Replaces the arrow-recursion module with [katalyst](https://github.com/aedans/Katalyst) 0.7.0.


modules/recursion-schemes/arrow-recursion/src/test/kotlin/arrow.recursion/laws/RecursiveLaws.kt
```diff
@@ -0,0 +1,14 @@
+package arrow.recursion.laws
+
+import arrow.recursion.typeclasses.Corecursive
+import arrow.recursion.typeclasses.Recursive
+import arrow.test.laws.Law
+import io.kotlintest.matchers.shouldEqual
+
+object RecursiveLaws {
```
pakoito: We have all laws in the `arrow-laws` module which should depend on all `arrow` libraries, if you prefer to have them there!

aedans: Makes sense, I'll move them there.

pakoito: `forAll(intGen)` can become `forFew`, the rest you can move to the Generators file :D

aedans:
@raulraja Thanks, I'm happy to have it merged! And I think we should announce it once the docs are finished (which I'll be working on soon) so that Arrow users can look at tutorials in Kotlin rather than having to google tutorials in Haskell.

PR #832, opened 2018-05-04
Fix option monoid

pakoito:
https://github.com/typelevel/cats/blob/master/kernel/src/main/scala/cats/kernel/instances/option.scala#L73-L84


modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/option.kt
```diff
@@ -15,9 +15,9 @@ interface OptionSemigroupInstance<A> : Semigroup<Option<A>> {
     when (this) {
       is Some<A> -> when (b) {
         is Some<A> -> Some(SG().run { t.combine(b.t) })
-        None -> b
+        None -> this
       }
-      None -> this
+      None -> b
     }
```
pablisco: Two things, shouldn't it return just `None`?
Also, couldn't we flatten the when like:
```
when {
  this is Some<A> and b is Some<A> -> Some(SG().run { t.combine(b.t) })
  else -> empty()
}
```
Probably not for consistency, but thought worth a mention.

pakoito: It shouldn't return none because:

[1] + [] = [1]
1 + 0 = 1
1 * 1 = 1

So it's logical that

None + Some(1) = Some(1)

That's the way it's encoded in Scala too, so this was an oversight -.-

Issue #830, opened 2018-05-03
Constructor functions for Show to use receiver function


pakoito:
From gitter, current encoding

```
val catShow = Show<Cat> {
        val name = it.name
        val age = it.age
        val color = it.color
        "$name is a $age year old $color cat"
    }
```

with proposed encoding

```
val catShow = Show<Cat> {
        "$name is a $age year old $color cat"
    }
```

pakoito:
`Show` cannot have receivers for two parameters, silly me :P

ersin-ertan:
> <s>`Show`</s> `Eq` cannot ...

pakoito:
Closed by #835 . Thanks @ersin-ertan

PR #835, opened 2018-05-05
Show constructor to use receiver type to enable object property access

ersin-ertan:
Solves: https://github.com/arrow-kt/arrow/issues/830

### DIFF
--- a/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Show.kt
+++ b/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Show.kt
@@ -17,12 +17,12 @@ interface Show<in A> {
   companion object {
 
     /**
-     * Construct a [Show] instance from a function `(A) -> String`
+     * Construct a [Show] instance from a function `A.() -> String`
      *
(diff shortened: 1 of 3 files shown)


pakoito:
Could you please use this chance to add a reference to the `invoke` constructor in the docs? I'm surprised we don't use it in the codebase either, so good news hahaha

ersin-ertan:
Good tip - I'll create a snippet

pakoito:
This is a great chance to learn about Ank!! Ank is the tool we use to compile snippets. What is does is output the result of the last expression inside a block. With the changes you've introduced, as they're in the same block, they won't produce the snippet result. Compare the docs with the source: https://arrow-kt.io/docs/typeclasses/show/

If you'd like to test Ank, use `./gradlew runAnk`, and find the files in the arrow-docs output folder.

ersin-ertan:
Show and Eq examples moved to Create your own instance locations.
Previous commits squashed into a single commit.
Updated branch with other arrow commits. 
:+1:

pakoito:
Thanks @ersin-ertan for the contribution!

PR #836, opened 2018-05-06
DocsAdd: Eq invoke constructor example

ersin-ertan:


ersin-ertan:
Consolidated patches into a single push

Issue #838, opened 2018-05-06
Measure build time


pedrovgs:
As a developer, I'd like to know how long our contributors are building arrow or just waiting for the compiler or the tests execution 😃

raulraja:
Closing this old issue. @pedrovgs 💋

PR #839, opened 2018-05-06
Measure build times

pedrovgs:
### :pushpin: References
* **Issue:** Fixes #838

Measure build times as suggested by @raulraja.

### How is it being implemented?

To be able to measure build times we've added a new plugin to this repository enabled for every module. This plugin is named ``build-time-tracker`` and you can find the source code [here](https://github.com/passy/build-time-tracker-plugin). I tried to use [kuronometer](http://kuronometer.io/), but the plugin does not support multiple modules yet. The plugin configuration is the one provided by default in the project documentation but the summary configuration. I've disabled the summary feature because this project has so many modules that the output generated by this plugin could be annoying.

**The build times measured are saved into the ``build`` folder inside a ``times.csv`` file you can review when needed.**

@raulraja, I'm not really sure if this is the result you where expecting when you asked me for a tool to measure the build times, but if this is not the result you expected feel free to discard this PR. 

### How can it be tested?

Build the project executing any Gradle task and you'll see the time execution report and a ``times.csv`` file saved into the root ``build`` folder containing the build execution time information.


pedrovgs:
@JorgeCastilloPrz I'm not sure about this:

> Can we assume measuring the build times on this project is equivalent to what we look for? I guess it's directly related.

Once the arrow jar is built clients adding this jar to their projects should be able to build the whole project without executing all the different stages the compiler and Gradle as build tool performs when building the whole repository. If you'd like to measure the impact you'd need to create a sample project and build it with and without arrow added as a dependency.

About this one:

> Also it would be really handy if we could aggregate these results to get (at least) the average times for any projects using the Arrow modules

The execution time measured for every Gradle task is being appended to the ``csv`` file you can find inside the build folder. But as the information is saved just locally we can't aggregate this information by user.

About this comment:

> We've been very careful on it during implementation and we'd need a reliable (empirical) way to ensure it. Any ideas?

We could update [this Gralde plugin](https://github.com/pedrovgs/Kuronometer) implemented in Scala supporting multiple modules and send these reports to our servers. This could help us to measure how long our developers are waiting for Gradle, the compiler or the tests execution. Server-side code can be found [here](https://github.com/delr3ves/KuronometerServer). Just a simple Finatra project written in Scala using ElasticSearch as DB. Infrastructure code needed to get this project up and running in AWS using terraform can be found [here](https://github.com/davideme/KuronometerTerraform)

We could even show some cool dashboards using [Grafana](https://grafana.com/) and the ElasticSearch database as data source.

@JorgeCastilloPrz I don't know if this idea is what you are looking for, I tried to do this in the past with Kuronometer but I couldn't find users for my Gradle plugin 😢If you are interested maybe @davideme and @delr3ves could help us.

jereksel:
Why don't use native Build Scan: https://scans.gradle.com/ ?

JorgeCastilloPrz:
That's sounds quite interesting. I think we could get big benefits on aggregating those stats. Thats kind of the idea I had  (may be not such detailed) but still we would need it to be run implicitly by the arrow core dependency (no idea how to achieve that or if it's even possible, may be using the plugin @jerkesel suggested). Otherwise library users are not gonna install a plugin just for your help.

jereksel:
@JorgeCastilloPrz Nope

nomisRev:
@JorgeCastilloPr I use it at work. Anything you want me to find out?

JorgeCastilloPrz:
What does it provide exactly?

jereksel:
Damn, that's pretty cool (newest master, please don't DDOS me): https://scans.gradle.com/s/6ql3q7q3ep5tg/

raulraja:
I'm particularly interested in this functionality because 47 Degrees is developing a gradle plugin that includes many standard build features we plan on using on Arrow. Gathering build times among others so down the line we can measure the impact of new features in build times locally by any project not just arrow specially those dependent on meta programming facilities of arrow in particularly what we do with annotations and kapts. The meta modules and annotation processors we write in arrow impact directly build times of downstream users in the same way today they are impacted by dagger and other processors. Though bundling this plugin in our `org-policies` plugin would be a good feature to have and propose to arrow users. //cc @arturogutierrez

raulraja:
@pedrovgs I would like to try your solution since it what it was recommended initially and I think we can try both actually @jereksel  and perhaps compare both. At this point I'm don't have an strong opinion and as mention is ultimately for the org-policies-gradle plugin which ultimately will be used by Arrow and other libs

raulraja:
@pedrovgs once this is merged how can we see the reports and where? thanks!

JorgeCastilloPrz:
It's on the description. 

"The build times measured are saved into the build folder inside a times.csv file you can review when needed."

PR #847, opened 2018-05-17
Add ApplicativeError example to Validated docs 

raulraja:
This example shows how you can abstract away failure strategies


modules/docs/arrow-docs/docs/docs/datatypes/validated/README.md
```diff
@@ -214,6 +214,79 @@ val houseNumber = config.parse(Read.intRead, "house_number").withEither { either
...
+
+sealed class ValidationError(val msg: String)
+data class DoesNotContain(val value: String) : ValidationError("Did not contain $value")
+data class MaxLength(val value: Int) : ValidationError("Exceeded length of $value")
+data class NotAnEmail(val reasons: Nel<ValidationError>) : ValidationError("Not a valid email")
+
+data class FormField(val label: String, val value: String)
+data class Email(val value: String)
+
+abstract class Rules<F>(val A: ApplicativeError<F, Nel<ValidationError>>) : ApplicativeError<F, Nel<ValidationError>> by A {
```
pakoito: A doesn't have to be a field. Why the abstract class?

modules/docs/arrow-docs/docs/docs/datatypes/validated/README.md
```diff
@@ -214,6 +214,79 @@ val houseNumber = config.parse(Read.intRead, "house_number").withEither { either
 houseNumber
 ```
 
+## Abstracting aways validation strategies with `ApplicativeError`
```
pakoito: Alternative validation strategies to `Validated`: using `ApplicativeError`? It's a bit more specific than `abstracting away`. It'd be worthwhile to have a version of this example on the ApplicativeError docs!

modules/docs/arrow-docs/docs/docs/datatypes/validated/README.md
raulraja: There is nothing to implement, therefore no need for an interface and if it were an `interface` then it results in more boilerplate in the interpreters.
```kotlin
interface Rules<F> : ApplicativeError<F, Nel<ValidationError>> {

  val A: ApplicativeError<F, Nel<ValidationError>>

  private fun FormField.contains(needle: String): Kind<F, FormField> =
    if (value.contains(needle, false)) just(this)
    else raiseError(DoesNotContain(needle).nel())

  private fun FormField.maxLength(maxLength: Int): Kind<F, FormField> =
    if (value.length <= maxLength) just(this)
    else raiseError(MaxLength(maxLength).nel())

  fun FormField.validateEmail(): Kind<F, Email> =
    map(contains("@"), maxLength(250), {
      Email(value)
    }).handleErrorWith { raiseError(NotAnEmail(it).nel()) }

}

val errorAccumulatingApplicative : ApplicativeError<ValidatedPartialOf<Nel<ValidationError>>, Nel<ValidationError>> =
  Validated.applicativeError(NonEmptyList.semigroup())

object ErrorAccumulationStrategy : 
  Rules<ValidatedPartialOf<Nel<ValidationError>>>, 
  ApplicativeError<ValidatedPartialOf<Nel<ValidationError>>, Nel<ValidationError>> by errorAccumulatingApplicative {
    override val A: ApplicativeError<ValidatedPartialOf<Nel<ValidationError>>, Nel<ValidationError>>
      get() = Validated.applicativeError(NonEmptyList.semigroup())
}
```

Normally in tagless the practice I see is that if you can implement the algebra entirely with a type class and not leave any methods abstract to the interpreters then you don't need to use an interface or trait since the very definition of the algebra is abstract by relying on the type classes.

modules/docs/arrow-docs/docs/docs/datatypes/validated/README.md
```diff
@@ -214,6 +214,80 @@ val houseNumber = config.parse(Read.intRead, "house_number").withEither { either
...
+  
+  companion object {
+    infix fun <A> failFast(f: FailFastStrategy.() -> A): A = f(FailFastStrategy)
+    infix fun <A> accumulateErrors(f: ErrorAccumulationStrategy.() -> A): A = f(ErrorAccumulationStrategy)
+  }
+
+}
+```
+
+`Rules` defines abstract behaviors that can be composed and have access to the scoipe of `ApplicativeError` where we can invoke `just` to lift values in to the positive result and `raiseError` into the error context.
```
nomisRev: typo: scoipe

raulraja:
@nomisRev This example is considered advanced in the context that it requires a bit of knowledge on how Applicative and Applicative Error works though it contains a brief paragraph explaining how they are used in this context:
```kotlin
`Rules` defines abstract behaviors that can be composed and have access to the scope of `ApplicativeError` where we can invoke `just` to lift values in to the positive result and `raiseError` into the error context.
```
Are you implying we should not add it to the docs in Validated?

nomisRev:
No, I was not implying that at all. Just wondering if we could add something to mark it as advanced and/or point to Applicative/ApplicativeError.

Marking snippets or part of documentation as advanced might make them less intimidating/demotivating for new people.

raulraja:
@nomisRev Added an issue to discuss with @israelperezglez @calvellido . Feel free to contribute any ideas or thoughts there and we can address this in general for all docs since this same jump from beginner to advance happens in many other places.

JorgeCastilloPrz:
Thanks :-)

Issue #283, opened 2017-09-16
Invariant and Contravariant Functors


pakoito:
https://leanpub.com/fpmortals/read#leanpub-auto-variance

pakoito:
@ersin-ertan is working on Invariant Functors.

ersin-ertan:
@raulraja :eyes: 
- [ ] Ok, almost done - `KleisliFunctorInstance` has:  `fun FF(): Functor<F>` declared within them and the overridden method `fix().map(FF(), f)` uses `FF()`, because `Kleisli` has `map(FF: Functor<F>, f: (A) -> B)` does `Kleisli` and `EitherT, OptionT, Coproduct, StateT, WriterT` require two new function definitions? 

- [ ] Are there are any functions that take `Functor` as the argument, but now that both `Contravariant` and `Functor` extend `Invariant`, will change to take `Invariant` instead?

- [ ] Should `Contra/Invariant` instances be added to `core.dagger` `core.effects` `core.optics`?

- [ ] Some `ApplicativeInstances` extend `ABCFunctorInstance<E>` will these instances also need `ABCContra/InvariantInstance<E>` versions too?

- [ ] In `arrow.mlt` there exists `FunctorFilter`, is a `Contra/Invariant` version need too? If so would anything that extends the new `Contra/InvariantFilter` need specific versions too?

- [ ] Should the `arrow-infographic` update be included on each commit?

raulraja:
> KleisliFunctorInstance has: fun FF(): Functor<F> declared within them and the overridden method fix().map(FF(), f) uses FF(), because Kleisli has map(FF: Functor<F>, f: (A) -> B) does Kleisli and EitherT, OptionT, Coproduct, StateT, WriterT require two new function definitions?

> Are there are any functions that take Functor as the argument, but now that both Contravariant and Functor extend Invariant, will change to take Invariant instead?

I don't think that they require any changes. All those functions use the Covariant Functor. I could not find in cats either the use of Invariant inside those.Unless we have a use case by which to add them I think they are not required

> Should Contra/Invariant instances be added to core.dagger core.effects core.optics?

Yes please, add any new instances to all modules and docs possible where they make sense.

> Some ApplicativeInstances extend ABCFunctorInstance<E> will these instances also need ABCContra/InvariantInstance<E> versions too?

[...]

PR #849, opened 2018-05-18
Add typeclass invariant contravariant

ersin-ertan:
Solves: #283


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/InvariantLaws.kt
```diff
@@ -0,0 +1,44 @@
...
+import arrow.test.generators.genFunctionAToB
+import arrow.typeclasses.Applicative
+import arrow.typeclasses.Eq
+import arrow.typeclasses.Invariant
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+
+object InvariantLaws {
+
+  inline fun <F> laws(AP: Applicative<F>, EQ: Eq<Kind<F, Int>>): List<Law> =
```
pakoito: Inline isn't necessary. It's a leftover from before a refactor.

ersin-ertan: Should I remove `inline` from the others tests too, including `noinline` for those who use it?
But leave `inline` for those tests with `reified`

ersin-ertan:
Finished 2/3. For:
> add ContravariantLaws.test for the datatypes that implement Contravariant

Sometimes `ContravariantLaws.laws(ABC.contravariant(), ::ABC, lastParam)` the lastParam is `Eq.any()` or either `EQ()` or `EQ`. When is which required?

pakoito:
>Sometimes ContravariantLaws.laws(ABC.contravariant(), ::ABC, lastParam) the lastParam is Eq.any() or either EQ() or EQ. When is which required?

Simple data classes like `Either`, `Option` or `ListK` have `equals` written by the kotlin compiler so `Eq.any()` is enough. Other datatypes that need to be run like Function, Eval or IO require a specialized `Eq` instance that runs them for a value.

> SortedMapKTest
ContravariantLaws.laws(SortedMapK.contravariant<Int>(), { a: Int -> sortedMapOf("key" to a).k() }, EQ)

See that the error is telling you that your map has String as the key, but the instance your passing has Int as key! `SortedMapK.contravariant<String>()` should fix it.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/InvariantLaws.kt
pakoito: There should be no tests with reified AFAIK. Leave the fix for a different PR tho.

ersin-ertan:
Investigating why the tests are failing

modules/docs/arrow-docs/docs/docs/typeclasses/intro/README.md
```diff
@@ -86,9 +86,13 @@ We will list them by their hierarchy.
 
 - [`MonoidK`]({{ '/docs/typeclasses/monoidk/' | relative_url }}) - combinable datatypes have an empty value
 
-##### Functor
+##### Invariant, Contravariant, Functor
 
-- [`Functor`]({{ '/docs/typeclasses/functor/' | relative_url }}) - its contents can be mapped
+- [`Invariant`]({{ '/docs/typeclasses/Invariant/' | relative_url }}) - its contents can be bidirectionally mapped
```
pakoito: IIRC this is case sensitive. Please double check it!

modules/docs/arrow-docs/docs/docs/typeclasses/intro/README.md
```diff
@@ -86,9 +86,13 @@ We will list them by their hierarchy.
...
+
+- [`Contravariant`]({{ '/docs/typeclasses/contravariant/' | relative_url }}) - its contents can be mapped reverse sequentially
```
pakoito: To be honest I'd move these below `Bimonad`, as they're not as common as the others. Same in the menu.

ersin-ertan: The intro stated that "We will list them by their hierarchy" so I used that as the guide, however if commonality is more important, then I can switch them around, remove the quote, and change the section title back to Functor.

pakoito: Fair, up to you then :D

ersin-ertan:
I'm a bit confused at this point. Tests are failing because `@instance` annotated interfaces with `.contramap(f: (B) -> A): Data<B> = fix().contramap(f)` and `imap` point to themselves, thus stackoverflow.

Looking at cats/data, only a few implement `contramap` and almost none implement `imap`. I think I was over-eager my imap and contramap addition, as well as defined the function parameters incorrectly. 

My next step is to remove almost all `@instance` annotated imap declarations.
My next step is to remove most all `@instance` annotated contramap declarations.
Change contramap's within the remaining `@instance` annotated contramap declarations params to include the arg that are implicit in scala.
Thoughts?

pakoito:
You can implement `imap` in the instances in terms of `map` and `contramap` member methods, as far as I can see.

pakoito:
A new version of Invariant and Contravariant is already in the database. Let's close this :D Thanks @ersin-ertan for the effort!

PR #853, opened 2018-05-21
[arrow-core] Make param names more dev friendly

aballano:
### Current state

Currently in many DataTypes functions such as the common `fold`, have this format:

```
fun <B> fold(fa: (Throwable) -> B, fb: (A) -> B)
```

### The problem

Is common to find usages taking advantage of named parameters, in order to easily and quickly differentiate one function from the other. Such usages could look as follows with the current implementation:

```
Try { "Hello" }.fold(fa = {
    // is fa the error or success case?
}, fb = {
    // ditto
})
```

### Proposed solution
My proposal is to be more explicit with param names matching each possible case such as, for the Try example, `ifSuccess` and `ifFailure`. 

```
Try { "Hello" }.fold(ifFailure = {
    handleError(it)
}, ifSuccess = {
    print(it)
})
```

alternatively:

```
Try { "Hello" }.fold(
    ifFailure = ::handleError, 
    ifSuccess = ::print
)
```

#### Scope

I've reduced the scope to only change **non deprecated functions** and those which have **at least two parameters** referring to a specific state, if applied.

Current progress:
- [x] Try 
- [x] Either 

### Notes

Feel free to suggest better namings, I've mainly tried to match with Kotlin's similar functions, such as `fold` in Lists and so on. The rest are on my own subjective opinion.


modules/core/arrow-core/src/main/kotlin/arrow/core/Try.kt
```diff
@@ -123,16 +123,19 @@ sealed class Try<out A> : TryOf<A> {
     }
   }
 
-  fun toOption(): Option<A> = fold({ None }, { Some(it) })
+  fun toOption(): Option<A> = fold({ None }, {
+    arrayListOf<String>().fold
```
JorgeCastilloPrz: What's this about?

aballano: Euh, nothing :)
image

aballano:
@JorgeCastilloPrz Indeed, we could either release this as per "module-basis" or "whole lib-basis", I personally would prefer the former as I cannot assure being able to work on this for a much longer period :)

raulraja:
`fa` and `fb` are valid in polymorphic contexts like type classes but in concrete data types it's fine to use descriptive names like `ifLeft`, `ifRight`. The challenge is that instances and syntax inside type class blocks that affects those data types will still refer to those as `fa`, `fb`. For example:
```
ListK.applicative(). run {
  map(l1, l2, l3, l4, { ... }) // this comes from applicative so it won't have the friendly names
}
```

pakoito:
Renaming inherited parameters causes a compiler warning for a good reason, as it conflicts on call-by-name for parameters! It's okay to have different parameter names for datatype functions, but anything in instances has to retain the parent's name.

aballano:
@raulraja Good to know! Is there anything we could do there from here? Is that ok to go with or a blocker?

aballano:
I'll be merging for now. Hopefully we can continue with the rename in other modules too.

PR #857, opened 2018-05-24
Add SingleK wrapper for RxJava Single

### LINKED ISSUE #856: Add RxJava Single wrapper
The same way we're adding wrappers for `Flowable`, `Observable` and other types, we should have one for `Single`.

JorgeCastilloPrz:
Fixes #856

* Added `SingleK`
* Added instances
* Added some tests including law tests for the instances.


modules/effects/arrow-effects-rx2/src/test/kotlin/arrow/effects/SingleKTests.kt
```diff
@@ -0,0 +1,109 @@
...
+          subscription.dispose()
+        }
+      }.test()
+
+      test.awaitTerminalEvent(5, TimeUnit.SECONDS)
+      test.assertNotTerminated().assertNotComplete().assertNoErrors().assertNoValues()
+    }
+  }
+
+  // FIXME(paco): remove if this hasn't triggered in a while - 26 Jan 18
```
pakoito: lol is this in both Observable and Flowable? I wouldn't have it on by default.

JorgeCastilloPrz: Yeah It's copypasta

JorgeCastilloPrz:
Back to you @pakoito

PR #861, opened 2018-05-25
Add some clarifications of For- being reused in both Syntax and Witness

pakoito:


JorgeCastilloPrz:
Looks better now thx :)

PR #862, opened 2018-05-25
Add MaybeK wrapper for RxJava Maybe

### LINKED ISSUE #860: Add RxJava Maybe wrapper
The same way we're adding wrappers for `Flowable`, `Observable`, `Single`, and other types, we should have one for `Maybe`.

JorgeCastilloPrz:
Fixes #860 

* Added `MaybeK`
* Added instances
* Added some tests including law tests for the instances.

@pakoito I've added `fold()`, `foldLeft`, `foldRight` etc to it since I saw those where also defined in other "equivalent" data types like Option and maybe also represents code branching. Also a `Foldable` instance. But to be honest I'm not 100% sure the code on those methods / instance is correct, but `FoldableLaws` are passing, which seems suspicious. Could you take a look?

I've not added `Traverse` instance for it yet since traverse uses to rely on fold and I'm still not confident with my implementation.


JorgeCastilloPrz:
Yeees I'll add docs this weekend, but need your help on those folds :D

pakoito:
IIRC Maybe has the same semantics as Observable, except it's max 1 return.

https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-rx2/src/main/kotlin/arrow/effects/ObservableK.kt#L33-L42

JorgeCastilloPrz:
There you got the docs also, back to you @pakoito

modules/effects/arrow-effects-rx2/src/main/kotlin/arrow/effects/MaybeK.kt
```diff
@@ -0,0 +1,83 @@
...
+  fun <B> map(f: (A) -> B): MaybeK<B> =
+    maybe.map(f).k()
+
+  fun <B> ap(fa: MaybeKOf<(A) -> B>): MaybeK<B> =
+    flatMap { a -> fa.fix().map { ff -> ff(a) } }
+
+  fun <B> flatMap(f: (A) -> MaybeKOf<B>): MaybeK<B> =
+    maybe.flatMap { f(it).fix().maybe }.k()
+
+  fun <B> fold(ifEmpty: () -> B, ifSome: (A) -> B): B = when {
```
pakoito: According to RxJava's docs, `this.maybe.blockingGet()` returns `null` for no value and `T` for a value. Given that RxJava2 doesn't allow nulls in `Maybe`, you can rewrite this to be simpler:

maybe.blockingGet().let {
  if (it == null) ifEmpty() else ifSome(if)
}

pakoito: I would suggest adding a new test for it too.

JorgeCastilloPrz:
Done and done @pakoito 👍

JorgeCastilloPrz:
👍

PR #867, opened 2018-05-30
Fix wrong superclass

Takhion:
...right?


nomisRev:
Absolutely right!

nomisRev:
@maintainers seems I can't merge due to coverage dropping?? (Perhaps it's mobile site).

But on that note can we disable code coverage for docs?

aballano:
@nomisRev Is not required, I can merge from my PC, so maybe it's mobile site :)

PR #868, opened 2018-05-31
Fix types order in Either example

hamen:


pakoito:
Thank you much for finding this!

PR #876, opened 2018-06-06
Prevent Either::map and Either::flatMap creating a new Left

dmcg:
A number of Either operations could be optimized I think, this is the low-hanging fruit

### DIFF
--- a/modules/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
+++ b/modules/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
@@ -106,7 +106,7 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
   inline fun <C> map(crossinline f: (B) -> C): Either<A, C> =
       when (this) {
         is Right -> Right(f(b))
-        is Left -> this as Left<Nothing, C>
+        is Left -> this as Left<A, Nothing>
       }
 
   /**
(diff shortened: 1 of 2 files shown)


modules/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -102,8 +102,12 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
    * Left(12).map { "flower" }  // Result: Left(12)
    * ```
    */
+  @Suppress("UNCHECKED_CAST")
   inline fun <C> map(crossinline f: (B) -> C): Either<A, C> =
-    fold({ Left(it) }, { Right(f(it)) })
+      when (this) {
+        is Right -> Right(f(b))
+        is Left -> this as Left<Nothing, C>
```
pakoito: <A, C>?

dmcg: Actually I think that it should be Left<A, Nothing>. 

I'd like to remove the vestigial extra type param from Left and Right, so that Left<A> : Either<A, Nothing> and Right<B> : Either<Nothing, B>. That would allow the casts to go altogether, but is a breaking API change, so I thought I'd wait to see if this PR was accepted before chancing my arm.

dmcg: Github, <Nothing, B> is not a bold tag!

raulraja:
thanks @dmcg !

PR #877, opened 2018-06-06
Remove unnecessary type params on Either.Left and Either.Right - brea…

dmcg:
Remove the unused type parameters on Either.Left and Right, so instead of 

> Left<out A, out B> : Either<A, B> 

we have 

> Left<out A> : Either<A, Nothing>

and similarly for Right. This is a breaking API change, but will allow map and flatMap to be implemented without any additional object creation or casting, and just seems to me to be more the truth.

Most of the knock-on effects are removing type parameters in when statements that didn't really need them anyway.


pakoito:
This was for a while the way of encoding `Either` and it had a bunch of issues with type inference, and IDE support, that made us go back to the current encoding. Try to see if it passes the tests at least.

Right now it fails on detekt, our linter. To pass detekt, do `./gradlew clean detekt`

dmcg:
We've not seen issues with this approach in Result4K, although I do grant that Arrow may stretch type inference and the IDE further.

JorgeCastilloPrz:
I'd say we need them to be Right<L, R> and Left<L,R>. Both types explicit. I've tried using either on client projects and using the bottom type still breaks inference on very common use cases. I don't think we have tests in place covering those inferences atm, so this change can pass tests and still be potentially wrong.

dmcg:
I'd be interested to see cases where it is a problem, if only because as I say, we haven't seen them, and Result4K has been this way from the start

JorgeCastilloPrz:
If you get green CI I'd be okay move on with this approach. Even if I'm a bit sceptic, I don't have those use cases now and can't remember exactly how to reproduce them, so let's avoid blocking much longer.

dmcg:
Ready when you are

raulraja:
Thanks @dmcg !

PR #883, opened 2018-06-11
build.gradle configuration for fat jar

MaxNeutrino:
Paco, as we talk I created configuration for fatjar. This generate jar in build/libs. Sorry, I don't know how to integrate fatjar with bintray


pakoito:
Cool, I get the gist. We'll move it to a new module and build it as a separate dependency ;)

build.gradle
```diff
@@ -164,3 +166,61 @@ apply plugin: 'com.github.ben-manes.versions'
...
+    compile project(":arrow-docs")
+    compile project(":arrow-effects")
+    compile project(":arrow-effects-kotlinx-coroutines")
+    compile project(":arrow-effects-rx2")
+    compile project(":arrow-free")
+    compile project(":arrow-generic")
+    compile project(":arrow-instances-core")
+    compile project(":arrow-instances-data")
+    compile project(":arrow-kindedj")
+    // compile project(":arrow-mlt")
```
raulraja: Why wouldn't the `mtl` be included in the fat jar?

JorgeCastilloPrz:
In an effort to clean the PR list we've decided to close all the ones that are older than 2 months and consider them obsolete except for very specific circumstances that we will evaluate per PR. In this case we were preparing a fatjar that I assume is obsolete now given how the artifacts in the project have evolved so much (this is a PR from almost 2 years ago).

PR #885, opened 2018-06-12
Add extra tips for running IO

pakoito:
Requested on gitter, it explains why `runAsync` doesn't start immediately.

### DIFF
--- a/modules/docs/arrow-docs/docs/docs/effects/io/README.md
+++ b/modules/docs/arrow-docs/docs/docs/effects/io/README.md
@@ -23,6 +23,8 @@ Running in this context means evaluating the content of an `IO` object, and prop
 
 Note that `IO` objects can be run multiple times, and depending on how they are constructed they will evaluate its content again every time they're run.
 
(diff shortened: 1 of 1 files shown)


modules/docs/arrow-docs/docs/docs/effects/io/README.md
```diff
@@ -40,6 +42,8 @@ All exceptions that would happen on the function parameter are automatically cap
 
 It runs the current `IO` asynchronously, calling the callback parameter on completion and returning its result.
 
+The operation will not yield a result immediately; ultimately to start running the suspended computation you have to evaluate that new instance using an unsafe operator like `unsafeRunAsync` or `unsafeRunSync` for `IO`.
+
 ```kotlin
```
nomisRev: Any reason we are not using `ank` on these snippets?

Could we perhaps add an additional snippet to accompany the added docs?

```kotlin:ank
val io = IO<Int> { throw RuntimeException("Boom!") }
  .runAsync { result ->
    result.fold({ IO { println("Error") } }, { IO { println(it.toString()) } })
  }
```

The operation will ...

pakoito: Async operations fail to execute on Ank.

PR #888, opened 2018-06-13
Move Optics std to companion objects

nomisRev:
Like it was done for generated Optics in #825 this PR moves the Optics of the std to the companion objects of the respective data types.


pakoito:
I see you've created new Api objects fro List and such. Rather than having a new one just for this library, it'd be better if we added an object in one of the core or data module, and add them as extension functions. That way it can be extended by all modules, and we centralise all the APIs.

nomisRev:
@pakoito I made the changes 👍

modules/core/arrow-instances-data/src/main/kotlin/arrow/instances/list.kt
```diff
@@ -0,0 +1,3 @@
+package arrow.instances
+
+object ListInstances
```
pakoito: Could you please put this file in arrow-data instead? That way we don't need to depend on the instances package when adding other typeclasses to List.

modules/core/arrow-instances-data/src/main/kotlin/arrow/instances/set.kt
```diff
@@ -1,73 +1,3 @@
...
-  override fun <A> Kind<ForSetK, A>.combineK(y: Kind<ForSetK, A>): SetK<A> =
-    fix().setCombineK(y)
-}
-
-object SetKContext : SetKFoldableInstance, SetKMonoidKInstance
-
-infix fun <A> ForSetK.Companion.extensions(f: SetKContext.() -> A): A =
-  f(SetKContext)
\ No newline at end of file
+object SetInstances
```
pakoito: Same as ListInstances!

modules/core/arrow-instances-data/src/main/kotlin/arrow/instances/list.kt
nomisRev: Shouldn't these go to `arrow-core`?

pakoito: It's cheap so why not :D

Issue #884, opened 2018-06-11
Use awaitAll for DeferredK


pakoito:
Apply `awaitAll` to the implementation of DeferredK where it applies https://github.com/Kotlin/kotlinx.coroutines/pull/323/files

richard-gibson:
Hey, I can take this one

pakoito:
I believe @sanogueralorenzo is already looking into it. I really don't know where it could benefit us, maybe just for `ap`, but if it's better than what's there I won't say no :D

richard-gibson:
no problem, 
looks like awaitAll() could br implemented for any Traverse<DeferredK<A>> by  just using sequence followed by await on the result 
(as you said whether its worth having)

pakoito:
If you have a PR then first come first serve! :D

pakoito:
Completed and done, thanks @richard-gibson !!!

PR #899, opened 2018-06-21
AwaitAll for Traverse Deferred<T>

richard-gibson:
For #884, 
This adds awaitAll to any F<DeferredK<T>> that returns F<T> where F has a traverse instance

### DIFF
--- a/modules/effects/arrow-effects-kotlinx-coroutines/src/test/kotlin/arrow/effects/DeferredTest.kt
+++ b/modules/effects/arrow-effects-kotlinx-coroutines/src/test/kotlin/arrow/effects/DeferredTest.kt
@@ -1,13 +1,18 @@
 package arrow.effects
 
 import arrow.Kind
+import arrow.core.*
+import arrow.data.*
 import arrow.test.UnitSpec
 import arrow.test.generators.genIntSmall
 import arrow.test.laws.AsyncLaws
 import arrow.typeclasses.Eq
+import arrow.typeclasses.Functor
+import arrow.typeclasses.Traverse
 import io.kotlintest.KTestJUnitRunner
 import io.kotlintest.matchers.fail
 import io.kotlintest.matchers.shouldBe
+import io.kotlintest.properties.Gen
 import io.kotlintest.properties.forAll
 import kotlinx.coroutines.experimental.CoroutineStart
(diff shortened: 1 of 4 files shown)


modules/effects/arrow-effects-kotlinx-coroutines/src/test/kotlin/arrow/effects/DeferredTest.kt
```diff
@@ -127,5 +141,16 @@ class DeferredKTest : UnitSpec() {
         // Success
       }
     }
+
+    "awaitAll called on a Traverse instance of DeferredK<T> should return a Traverse instance of T" {
+      forAll(Gen.int(), {
+        runBlocking {
+          checkAwaitAll(ListK.applicative(), ListK.traverse(), it)
```
pakoito: If any of these fails, it just returns a boolean that's ignored. You have to && them!

If you were not using `forAll` you'd have to use `shouldBe` in checkAwaitAll, which throws an AssertionException instead.

modules/effects/arrow-effects-kotlinx-coroutines/src/test/kotlin/arrow/effects/DeferredTest.kt
```diff
@@ -20,6 +31,9 @@ class DeferredKTest : UnitSpec() {
     a.unsafeAttemptSync() == b.unsafeAttemptSync()
   }
 
+  suspend fun <F, A> checkAwaitAll(AP: Applicative<F>, T: Traverse<F>, v: A) =
+      AP.just(DeferredK { v }).awaitAll(T) == AP.just(v)
```
pakoito: To have more than 1 element per test I'd suggest doing something like:

```
suspend fun <F, A> checkAwaitAll(AP: Applicative<F>, T: Traverse<F>, col: HK<F, A>) = AP.run {
  col.map { v -> DeferredK { v }}.awaitAll(T) == col
}
```

modules/effects/arrow-effects-kotlinx-coroutines/src/test/kotlin/arrow/effects/DeferredTest.kt
richard-gibson: eek, school boy error. Thanks

PR #852, opened 2018-05-21
Improve optics processor

nomisRev:
Closes #848

PR #908, opened 2018-06-24
Update processors generated structures

nomisRev:
This PR updates the structure of the generated code as discussed [here](https://github.com/arrow-kt/arrow/pull/852#issuecomment-395221879)

To recap:
 * Code is generated in the same package as the annotated element
 * The folder structure follows the package.
 * Filename will be `TargetClass$$annotationName` for `TypeElement` (class, interface and object).

This way the generated code will be structured in the same way as the user's code. And will ease browsing / debugging if necessary. I think this is also more in line with how code gen is ussualy done in Java.


pakoito:
>Code is generated in the same package as the annotated element

won't this break Java10?

pakoito:
@nomisRev yeah, I meant this new thing where two different dependencies can't contribute to the same package.

nomisRev:
@Guardiola31337 you have any idea what is causing this?

```
:arrow-docs:runAnkException in thread "main" java.lang.NoClassDefFoundError: arrow/core/Instance_arrow_instances_EitherMonadErrorInstanceKt
	at arrow.ank.main.main(main.kt:16)
Caused by: java.lang.ClassNotFoundException: arrow.core.Instance_arrow_instances_EitherMonadErrorInstanceKt
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:335)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 1 more
 FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':arrow-docs:runAnk'.
> Process 'command '/usr/lib/jvm/java-8-oracle/bin/java'' finished with non-zero exit value 1
```

pakoito:
It's the second report of this error :/

nomisRev:
@pakoito actually that was me twice I think :D Both were about this.

nomisRev:
Closing this PR since it's completely outdated given the meta changes.

PR #913, opened 2018-06-28
Arrow-recursion docs

aedans:
Adds docs for arrow-recursion and makes the style more consistent with the rest of arrow.


modules/recursion-schemes/arrow-recursion/build.gradle
```diff
@@ -10,7 +10,6 @@ dependencies {
     testCompile project(':arrow-test')
 }
 
-// FIXME(paco): re-enable when it's prod-ready
-//apply from: rootProject.file('gradle/gradle-mvn-push.gradle')
+apply from: rootProject.file('gradle/gradle-mvn-push.gradle')
```
pakoito: My bad :(

PR #920, opened 2018-07-02
[Spitballing] More user friendly Optional constructor

jereksel:
I think most new user will use non-polymorphic Optional, but constructor still forces them to use Either which is not intuitive in that context. My idea is to introduce new constructor `(S) -> Option<A>`. Unfortunatelly due to type erasure, we can't have Function1 as second parameter, so I've added Function2, which also is frankly more user friendly.

note: after this change defaultHead was throwing type inference errors.
image

### DIFF
--- a/modules/optics/arrow-optics/src/main/kotlin/arrow/optics/Optional.kt
+++ b/modules/optics/arrow-optics/src/main/kotlin/arrow/optics/Optional.kt
@@ -67,6 +67,12 @@ interface POptional<S, T, A, B> : POptionalOf<S, T, A, B> {
       override fun set(s: S, b: B): T = set(b)(s)
     }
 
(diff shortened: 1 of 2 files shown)


nomisRev:
@jereksel awesome! Could you update docs? We should also do the same for `Prism` to make it more consistent. 

I was looking into this as well and there are a couple more improvements we could do.

Instead of the nested lambdas we have in places such as `Lens` we could change `((B) -> (S)) -> T` with `(S, B) -> T`. Which also resembles `Lens#set`. Thoughts?

jereksel:
@nomisRev Invoke change is breaking, so I've created separate issue for that #940

nomisRev:
Closed by #1061

PR #923, opened 2018-07-03
Move arrow version to variable

jereksel:


pakoito:
Can you please change in the README.md at the index of the docs? We sync them manually because they're almost the same but not quite.

PR #925, opened 2018-07-04
Add orNull() to Either, Validated, and Try

xenomachina:


modules/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -247,6 +247,18 @@ fun <A, B, C> EitherOf<A, B>.flatMap(f: (B) -> Either<A, C>): Either<A, C> =
...
+/**
+ * Returns the value from this [Either.Right] or null if this is a [Either.Left].
+ *
+ * Example:
+ * ```
+ * Right(12).orNull() // Result: 12
+ * Left(12).orNull()  // Result: null
+ * ```
+ */
+inline fun <B> EitherOf<*, B>.orNull(): B? =
```
pakoito: Why is it `*` here and `E` in the `Validated` one? If they're inconsistent across files we should fix it.

pakoito:
Good idea to fix this :) It's a bit wonky if the value contained was already nullable, but that's on the users I guess.

raulraja:
We may want to consider using this PR to normalize all these and also add them to `Foldable` seem they seem to be more generic than originally anticipated:
```kotlin
interface Foldable<F> {
  ...

  fun <A> Kind<F, A>.getOrElse(default: () -> A): A =
    foldRight(Eval.later(default), { i, _ -> Eval.now(i) }).value()

  fun <A> Kind<F, A>.orNull(): A? =
    getOrElse { null }

  ...
}
```

modules/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
xenomachina: I did notice that there were some inconsistencies between the different places I was changing. I tried to go with a "use the style of the code around you" approach, though, especially given that this is my first arrow PR, so I modeled each "orNull" after the "getOrElse" on the same class. Similarly, I modeled the tests after the corresponding "getOrElse" tests.

raulraja:
thanks @jereksel !

PR #926, opened 2018-07-04
Fix Android + Java 8 interaction

### LINKED ISSUE #807: Arrow 0.7.x causes builds to fail on Android projects
Just adding the Arrow dependencies on an Android project with Java 8 features enabled causes the builds to fail due to a missing traverse related class / lambda during the desugar task.

Sample repo here: https://github.com/jrgonzalezg/arrow-android-failing

pakoito:
Closes #807

Could @JorgeCastilloPrz @jrgonzalezg or @jereksel please check whether this fixes it? Or do we need to merge and use a snapshot to check it.

I'd say this is a blocker for 0.7.3, and we're ready for release afterwards.

### DIFF
--- a/modules/core/arrow-data/src/main/kotlin/arrow/data/ListK.kt
+++ b/modules/core/arrow-data/src/main/kotlin/arrow/data/ListK.kt
@@ -24,11 +24,10 @@ data class ListK<out A>(val list: List<A>) : ListKOf<A>, List<A> by list {
 
   fun <B> ap(ff: ListKOf<(A) -> B>): ListK<B> = ff.fix().flatMap { f -> map(f) }.fix()
 
-  fun <G, B> traverse(GA: Applicative<G>, f: (A) -> Kind<G, B>): Kind<G, ListK<B>> = GA.run {
-    foldRight(Eval.always { just(emptyList<B>().k()) }) { a, eval ->
-      f(a).map2Eval(eval) { (listOf(it.a) + it.b).k() }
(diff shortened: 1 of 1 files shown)


JorgeCastilloPrz:
The easiest way to test it would be to deploy a snapshot, if you can.

pakoito:
@JorgeCastilloPrz gimme a green and I'll merge then :D

jrgonzalezg:
I don't know how to find the snapshot or how to use it, but this will not fully fix the problem, cause what I posted was a tiny example of the set of related issues, so this will likely just fix one of the instances of the problem but the build will still fail.

jrgonzalezg:
Also, maybe #807 should not be closed until a working fix is on a release cause it is relevant for people looking to see if the issue is reported / handled.

jereksel:
If this doesn't fix everything, we should create simple android project with everything as dependency and build it during tests to make sure this won't happen again.

jrgonzalezg:
I already created a demonstrating minimal Android project that I linked on the original issue: https://github.com/jrgonzalezg/arrow-android-failing But I don't see 0.7.3-SNAPSHOT published anywhere to update that repo to it although I have updated it now to 0.7.2 and extracted arrow version to a constant to make it easier to change for tests. Also if you run the build as indicated on that repo's README you can see the many instances of the issue besides `ListK`.

jrgonzalezg:
Found the way to depend on 0.7.3-SNAPSHOT and updated the repo again: https://github.com/jrgonzalezg/arrow-android-failing

PR #930, opened 2018-07-05
Optics DSL processor

nomisRev:
This PR adds a processor to enhance the Optics DSL. Given a function that provides an optic the processor can generate the necessary code to write it up into the DSL.

See [docs changes](https://github.com/arrow-kt/arrow/compare/simon-add-dsl-processor?expand=1#diff-9ce8c30af0d4e43469e3efb8e63557f0) as example.

One small side note / question. Let's say we define the following.

```kotlin
@optics data class Employee(val name: String, val company: Company?)
@optics data class Employees(val employees: ListK<Employee>)

@optics fun everyEmployee(): Traversal<Employees, Employee> = Employees.employees.every(ListK.each())
```

A user might expect to be able to use it as follows `Employees.everyEmployee` which is not the case since the DSL is built using extension functions and in this case, there is no entry point. However `Iso.id<Employees>().everyEmployee` is valid.


nomisRev:
@jereksel I added tests 👍

raulraja:
@nomisRev this is good to merge once conflicts are resolved

nomisRev:
I am closing this. It's a little hacky with the processors and this PR is probably completely broken.
We can revisit later with a better solution than kapt.

PR #932, opened 2018-07-10
Remove kapt sources hax for non-Android projects

jereksel:
It's only not needed anymore, but in fact it breaks build (test `kaptKotlin` is not included - at least on Linux, on Windows it seems to be working fine, I don't get it :-( ).


jereksel:
Later I'll check Android Studio stable, beta and canary and see if it works. If not I'll create issues on Google's and our bugtracker for that.

pakoito:
Merge whenever you're happy then :D

raulraja:
thanks @jereksel !

PR #933, opened 2018-07-10
Replace let(::Left/Right) with .left()/.right()

jereksel:


jereksel:
@JorgeCastilloPrz I think I've missed import somewhere. I'll check it today/tomorrow.

pakoito:
You need to explicitly import the extfuns

PR #936, opened 2018-07-12
Add extfun for raiseError to match just

pakoito:


pakoito:
@raulraja it's derived from an existing function, it shouldn't need to.

Issue #841, opened 2018-05-09
Add remaining Monoid and Semigroup instances


pakoito:
At the moment several instances from Scala are not available in Arrow, like the one for `Try`. The task is to do a pass on the codebase and write them.

bassjacob:
I'd be happy to help with this

pakoito:
Neat, we got `Try` and `Either` now :D

What's left?

bassjacob:
Still to come:

* [ ] function0/ function1 -https://github.com/scalaz/scalaz/blob/series/7.3.x/core/src/main/scala/scalaz/std/Function.scala
* [ ] char
* [ ] id
* [ ] boolean
* [ ] const which can reuse some functions from the typeclasses file
* [ ] eval - I'm not sure what the semantics to choose for eval would be

I'll get started on id next

bassjacob:
Do you want arrow-data-instances as part of this issue or a separate one?

pakoito:
>Do you want arrow-data-instances as part of this issue or a separate one?

I'm not sure I follow.

pakoito:
>I'm not sure what the semantics to choose for eval would be

Let's do a live session when you reach it, I have an idea but would need to be in front of the code.

bassjacob:
The data types I listed above are from arrow-instances-core. There's also a few missing I think from arrow-instances-data that we might want to add?

nomisRev:
I'm closing this due to all the instances discussed here are available in Arrow Core. Re-open a new ticket if Monoid or Semigroup instances found missing please 👍 Thanks all for th work!

PR #942, opened 2018-07-17
#841 - Add Monoid instances

bassjacob:
Adding this as a WIP for now. Would there be more required for Try/Either?


modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/either.kt
```diff
@@ -3,12 +3,49 @@ package arrow.instances
...
+        is Either.Right -> Either.Right(SG().run { x.combine(b.b) })
+      }
+    }}
+}
+
+@instance(Either::class)
+interface EitherMonoidInstance<L, R> : EitherSemigroupInstance<L, R>, Monoid<Either<L, R>> {
+  fun MO(): Monoid<R>
+
+  override fun SG(): Semigroup<R> {
```
raulraja: style nit pick, favor simple expressions whenever possible:
```kotlin
override fun SG(): Semigroup<R> = MO()
```

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/try.kt
```diff
@@ -1,11 +1,56 @@
...
+      }
+      is Failure -> b
+    }
+}
+
+@instance(Try::class)
+interface TryMonoidInstance<A> : TrySemigroupInstance<A>, Monoid<Try<A>> {
+  fun MO(): Monoid<A>
+
+  override fun SG(): Semigroup<A> {
```
raulraja: style nit pick, favor simple expressions whenever possible:
```kotlin
override fun SG(): Semigroup<A> = MO()
```

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/either.kt
```diff
@@ -3,12 +3,49 @@ package arrow.instances
...
+interface EitherSemigroupInstance<L, R> : Semigroup<Either<L, R>> {
+
+  fun SG(): Semigroup<R>
+
+  override fun Either<L, R>.combine(b: Either<L, R>): Either<L, R> =
+    when (this) {
+      is Either.Left -> this
+      is Either.Right -> { val x = this.b; when (b) {
+        is Either.Left -> b
+        is Either.Right -> Either.Right(SG().run { x.combine(b.b) })
```
pakoito: Could this be done without the intermediate variable as `this.b.combine(b.b)`?

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/try.kt
```diff
@@ -1,11 +1,56 @@
...
+import arrow.typeclasses.Traverse
 import arrow.instances.traverse as tryTraverse
 
+@instance(Try::class)
+interface TrySemigroupInstance<A> : Semigroup<Try<A>> {
+
+  fun SG(): Semigroup<A>
+
+  override fun Try<A>.combine(b: Try<A>): Try<A> =
+    when (this) {
```
pakoito: I'd move these functions to the datatype, reference them here instead.

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/try.kt
```diff
@@ -1,11 +1,56 @@
 package arrow.instances
 
 import arrow.Kind
-import arrow.core.*
```
bassjacob: I'm not sure why my editor is making that change, but I can revert if you'd like?

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/try.kt
bassjacob: I couldn't find a way to do that given the dependency on SG and the semigroup interface. Is there a good example to follow given that core doesn't import typeclasses to avoid a circular dependency?

modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/try.kt
pakoito: Sure, although it isn't a blocker.

pakoito:
Tell us when it's ready to review!

PR #944, opened 2018-07-18
Remove non-null requirement in Either for Validation

jereksel:
There is no technical reason to have one


pakoito:
cc @MarioArias for this review

raulraja:
This validation library can be entirely based in `ApplicativeError<F>` and it would support a ton of target data types including and not limited to `Validated` and `Either`

PR #950, opened 2018-07-22
Add IO implementation of parMap

pakoito:
First pass at parallelising IO.

TODO:

- [X] Nice API
- [x] Tests
- [ ] Docs


pakoito:
@raulraja I'm happy to merge without docs for now, to validate our assumptions of how it should work and announce it later.

raulraja:
Looks great, awesome work!. Instead of creating the arities for mapN in the companion shouldn't these be in the type classes and exported with the work Eugenio is doing for type class extension generation?

pakoito:
@raulraja It's IO-exclusive for now, it isn't generalised to run for any effect because it uses `unsafeRunAsync()`.

pakoito:
@raulraja I've generalised for `Effect`, although it's a bit ugly because we don't have a nice `Future` box where to put the "starters" in.

PR #952, opened 2018-07-26
Implementing extension function optionFromNullable()

fabriciorissetto:
Not sure if there is a better way to do this:

```
val something: String? = null
something?.let { Some(it) } ?: None
```

If there is something better or already implemented in arrow this PR can be closed. If not, consider the function that I've implemted:

```
val something: String? = null
something.optionFromNullable()
```

if you want this PR to go forward, please let me know what is the next things need to be done (testing, etc). Do you have some docs with guidelines for contributions? Thanks


fabriciorissetto:
Closing the PR. This feature is already implemented in `toOption()` function.

PR #955, opened 2018-07-28
Attempt to fix 0.7.X crashing on Android

pakoito:
This is another attempt at palliating #807, which is blocking releases


pakoito:
Merged as admin to test the Android problem

Issue #966, opened 2018-08-01
Inline functions create additional anonymous class declaration


pakoito:
Each inline declaration of a function, i.e. `pipe`, it may force callsites to create an anonymous class to capture the scope. This is an issue in Android land, due to the dex format and its limitations around method count and size.

If we can validate this assumption in the bytecode viewer, it's worth consider doing a pass on the codebase and remove many unnecessary inline annotations.

What'cha think.

MarioAriasC:
I'll have a look

Laimiux:
Is there any update on this?

pakoito:
I did a pass on some of the core datatypes: https://github.com/arrow-kt/arrow/pull/971

For `arrow-syntax` we'll do it if necessary.

Laimiux:
I think it would be valuable to remove inline at least from `pipe` and `partially` functions (at least from our use cases). I could do a pass on `arrow-syntax`.

pakoito:
Go for it, thanks!

PR #967, opened 2018-08-01
Remove unnecessary inline in core

pakoito:
cc #966 && @trevorsummerssmith

When we started with the lib we liked inlines and crossinline because we thought it'd be efficient. It turned out to be a bytecode bloat instead, and it comes with bad interactions with coroutines.

This PR removes a bunch of gratuitous inline identifiers. I'm keeping the fold and constructor ones for now as they're mostly harmless.


pakoito:
This is not the fix we need. We need to remove crossinline instead.

PR #972, opened 2018-08-02
Added leftIfNull to Either.

atomcat1978:


modules/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -276,6 +276,23 @@ inline fun <A, B> EitherOf<A, B>.getOrHandle(crossinline default: (A) -> B): B =
...
+ * Example:
+ * ```
+ * Right(12).leftIfNull({ -1 }) // Result: Right(12)
+ * Right(null).leftIfNull({ -1 })  // Result: Left(-1)
+ *
+ * Left(12).leftIfNull({ -1 })      // Result: Left(12)
+ * ```
+ */
+inline fun <A, B> EitherOf<A, B?>.leftIfNull(crossinline default: () -> A): Either<A, B> =
+        fix().fold({ Left(it) }, { if ( it == null) Left(default()) else Right(it) })
```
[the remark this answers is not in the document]
atomcat1978: That's true.

modules/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -276,6 +276,23 @@ inline fun <A, B> EitherOf<A, B>.getOrHandle(crossinline default: (A) -> B): B =
...
+ * Example:
+ * ```
+ * Right(12).leftIfNull({ -1 }) // Result: Right(12)
+ * Right(null).leftIfNull({ -1 })  // Result: Left(-1)
+ *
+ * Left(12).leftIfNull({ -1 })      // Result: Left(12)
+ * ```
+ */
+inline fun <A, B> EitherOf<A, B?>.leftIfNull(crossinline default: () -> A): Either<A, B> =
+        fix().flatMap { if ( it == null) Left(default()) else Right(it) }
```
pakoito: I'm surprised our linter doesn't complain about the extra whitespace here

atomcat1978: U mean at the beginning. Yes, IntelliJ has added some extra here.

atomcat1978: Fixed. Before ```it == null``` as well.

pakoito:
Aaaaand merged. Thanks @atomcat1978 !!

PR #983, opened 2018-08-10
Add cancellation capabilities to IO

pakoito:
Make `IO` have a new cancellable runAsync operator that ties into the run loop for cancellation. It improves over the old way of doing it at coroutine (flatMap) boundaries because now we can cancel after every operator call instead.

I'm tempted to make the cancellable asyncRun the default, as it has the same overhead with and without it cc @raulraja for opinion.

With cancellation and parallelism we can build races, which means we can finally have a proper concurrency framework and libraries for Android!


pakoito:
I'll add the docs for parallelization and cancelation after the release. My plan is that because testing this is hard we'll do a beta with some maintainers and selected projects first, and promo it once it's proven to work correctly. Maybe add `@experimental` for now.

PR #984, opened 2018-08-11
Add CancellableEffect typeclass. Use it to give races to IO

pakoito:
This PR races two IOs, allowing cancellation between them.

It also adds the typeclass CancellableEffect, that was needed to achieve this.


pakoito:
@raulraja ready for review!

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IOParallel.kt
```diff
@@ -50,3 +92,98 @@ fun <A, B, C, D, E, F, G, H, I, J> IO.Companion.parallelMapN(ctx: CoroutineConte
...
+    ).unsafeRunAsync { it.fold({ complete(it.left()) }, { /* should never happen */ }) }
+  }.handleErrorWith {
+    if (it == OnCancel.CancellationException) {
+      complete(it.left())
+    }
+    raiseError(it)
+  }
+}
+
+fun <A, B, C, D> IO.Companion.raceN(ctx: CoroutineContext, a: IO<A>, b: IO<B>, c: IO<C>, d: IO<D>): IO<Either<Either<A, B>, Either<C, D>>> =
```
raulraja: Hoping these would make it to a `Concurrent` type class in the near future.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/CancellableEffect.kt
```diff
@@ -0,0 +1,8 @@
+package arrow.effects.typeclasses
+
+import arrow.Kind
+import arrow.core.Either
+
+interface CancellableEffect<F> : Effect<F> {
```
raulraja: I believe this should be just `Concurrent` and include the `race` style algebra combinators in this type class so they are available for all concurrent effects in the spirit that concurrent effects are cancelable. For what is worth in cats we have:
```
Concurrent[F] extends Async[F]
ConcurrentEffect[F] extends Concurrent[F] with Effect[F]
```

PR #991, opened 2018-08-19
EqLaws: Add law tests for an equivalence relation

paulcadman:
Ideally, instances of Eq would define an equivalence relation.

In an equivalence relation the reflexivity property says that an object
is equal to itself. I've added a `reflexivityEquality` test that now checks for this property.

I also remove the `identityEquality` check. `a.eqv(a) == a.eqv(a)` just seems to
be checking purity?


paulcadman:
If this looks ok I can add a transitivity law.

pakoito:
The end result is the same, so it'll work well. If you add transitivity I'll merge!

paulcadman:
Hi - I don’t think it’s the same? 

With the previous implementation If `a.eqv(a)` is `false` then the tests will still pass.

pakoito:
You're right! In that case, the current implementation is correct. The Laws should not prevent you from defining Eq as !=, in which case `a.eqv(a)` is `false`. They just ask for the result to be consistent, or as you said, pure.

paulcadman:
I think that `Eq` should codify equality which I think is a strict form of equivalence. So it seems appropriate to me for instances of Eq to be an equivalence relation.

It seems to also be the view of cats[1] and bow[2]? Would you reconsider its inclusion? I've also added transitivity.

Of course your library could take a different view :) in which case could I add separate laws to the test module for users who want to add tests for their instances of Eq that they intend to be equivalences?

[1] https://github.com/typelevel/cats/blob/master/kernel-laws/src/main/scala/cats/kernel/laws/EqLaws.scala
[2] https://github.com/arrow-kt/bow/blob/0d150b96a1d6fd7c7f59c46a14d8a4b7a66c9455/Tests/BowTests/Typeclasses/EqLaws.swift

pakoito:
@raulraja has convinced me that this is worthwhile, so I'll have to update some examples and unlawful instances I have around :D

pakoito:
Merging as soon as CI passes

paulcadman:
Thanks for merging 🥇

PR #992, opened 2018-08-20
Specify that MonadDefer focuses on synchronous effects

pakoito:
Remove the docs on Cancellation via coroutines too, as it's going away soon.

### DIFF
--- a/modules/docs/arrow-docs/docs/docs/effects/monaddefer/README.md
+++ b/modules/docs/arrow-docs/docs/docs/effects/monaddefer/README.md
@@ -9,11 +9,25 @@ permalink: /docs/effects/monaddefer/
 {:.intermediate}
 intermediate
 
-`MonadDefer` is a typeclass to abstract over lazy computations. This means that the computations are run *synchronously* at a different time than when they're created. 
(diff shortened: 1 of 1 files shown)


modules/docs/arrow-docs/docs/docs/effects/monaddefer/README.md
```diff
@@ -9,9 +9,9 @@ permalink: /docs/effects/monaddefer/
 {:.intermediate}
 intermediate
 
-`MonadDefer` is a typeclass representing suspension of execution via functions, allowing for asynchronous and lazy computations.
+`MonadDefer` is a typeclass to abstract over lazy computations. This means that the computations are run *synchronously* at a different time than when they're created.
```
raulraja: I'd say here instead:
```
`MonadDefer` is a typeclass to abstract over computations that cause side effects. This means that the computations are defered *synchronously* until the last possible time when effects are asked to be performed . Without effect suspension the effects would otherwise run immediately.

Applicative.pure(println("side effect")) // impure because `pure` is strict and the effects run uncontrolled, prints the value
MonadDefer.defer(println("side effect")) // pure because `println` is not invoked until the computation runs as part of the IO loop
```

PR #997, opened 2018-08-24
Improving the FpToTheMax example

enhan:
I just added TestIO based on the State datatype & included the version of @raulraja .


nfrankel:
As a newbie, I don't understand the need to wrap system calls in `IO` to "unwrap" them later with `bind()`?
```
IO { println("foo") }.bind()
```

It might be necessary in Scala, but it seems pretty purposeless in Kotlin.

raulraja:
Both `println` and `readLine` perform effects. `println` causes the external effect that when applied something gets printed to a log or system.out. `readLine` causes the effect that regardless of the input parameters you get always something different back therefore introduces the same kind of effect you would have if you had a random number generator in your computation. They both make functions impure since you can’t expect the function to produce the same output for a given input without either modifying the external world or making it non deterministic.
For that reason effects are encapsulated with either `IO` or if you are using type classes anything above `MonadDefer`.
Proof that these effects happen is:
```kotlin
fun Applicative<F>.printOpImpure: Kind<F, Unit> = pure(println("whatever")) 
// because pure is strict println gets evaluated each time `printOpImpure` is invoked and can't be part of the final IO loop that controls effects and exceptions. In your example it will be similar. If there were no previous bindings in the computation 
fun MonadDefer<F>.printOpPure: Kind<F, Unit> =  invoke { println("whatever") } 
// because `invoke` in MonadDefer is lazy invoking `printOpPure` will not cause printing to happen and the effect will be treated along the rest of computation when executing the `program.unsafeRunSync` at the edge```
(edited)
```

A different question is whether we can leave logging out of the effects we care to track.
To summarize, anything that causes a change to the external world or makes a function impure as a rule of thumb should be encapsulated in `IO#invoke` or suspended with a type class like `MonadDefer#invoke`.

In regards to Scala is the same as Kotlin, both langs allow devs to perform effect without being in IO.

pakoito:
Wrap, then merge!

enhan:
Hi @pakoito , what do you mean by wrap ? Should I squash the commits ?

pakoito:
@enhan Replace those `read/println().liftIO()` with `IO { read/println() }` to make them pure, and we'll merge afterwards!

enhan:
@pakoito I think there is no liftIO used in this PR ! @nfrankel used it in its own implementation of FP to the Max, if I'm not mistaking.

pakoito:
I see! I'll merge as soon as CI is ready, then

enhan:
nice :) !

PR #998, opened 2018-08-24
Add `arrow.generic.Coproduct` docs

abergfeld:
This adds docs for Coproducts. I largely based this off the Products docs that already existed. I'm not sure if I have to add anything else to get these setup with the web pages or how all that works. These are my first Arrow docs so feedback would be much appreciated :)


JorgeCastilloPrz:
If you wanna test the docs locally:

Run Ank, it'll deploy docs locally for you.
`./gradlew :arrow-docs:runAnk` in the root Arrow folder.

Once that's finished, go to modules/docs/arrow-docs/ and then

jekyll serve --source build/site/

That will launch the site on 127.0.0.1:4000. You'll need Jekyll installed.

abergfeld:
@JorgeCastilloPrz Thanks! I got the readme code compiling and hooked up the menu and it's all working locally 🎉

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -0,0 +1,167 @@
...
+compile 'io.arrow-kt:arrow-generic:$arrow_version'
+```
+
+### Features
+
+#### Coproduct
+
+Coproducts represent a container in which only one of the specified set of types exist. It's very similar in `Either` in that respect. `Either` supports one of two values, an `Either<A, B>` has to contain an instance of `A` or `B`. We can extrapolate that concept to `N` number of types. So a `Coproduct5<A, B, C, D, E>` has to contain an instance of `A`, `B`, `C`, `D`, or `E`. For example, perhaps there's a search function for a Car Dealer app that can show `Dealership`s, `Car`s and `SalesPerson`s, we could model that list of results as `List<Coproduct3<Dealership, Car, SalesPerson>`. The result would contain a list of heterogeneous elements but each element is one of `Dealership`, `Car` or `SalesPerson` and our UI can render the list elements based on those types.
+
+Let's say we have an api. Our api operates under the following conditions:
```
JorgeCastilloPrz: All this scenarios would work better using very simple code snippets.

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+```groovy
+compile 'io.arrow-kt:arrow-generic:$arrow_version'
+```
+
+### Features
+
+#### Coproduct
+
+Coproducts represent a container in which only one of the specified set of types exist. It's very similar in `Either` in that respect. `Either` supports one of two values, an `Either<A, B>` has to contain an instance of `A` or `B`. We can extrapolate that concept to `N` number of types. So a `Coproduct5<A, B, C, D, E>` has to contain an instance of `A`, `B`, `C`, `D`, or `E`. For example, perhaps there's a search function for a Car Dealer app that can show `Dealership`s, `Car`s and `SalesPerson`s, we could model that list of results as `List<Coproduct3<Dealership, Car, SalesPerson>`. The result would contain a list of heterogeneous elements but each element is one of `Dealership`, `Car` or `SalesPerson` and our UI can render the list elements based on those types.
```
JorgeCastilloPrz: Please add links to Either docs everytime you mention it.

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -0,0 +1,167 @@
...
+So now that we've got our api response modeled, we need to be able to create an instance of `Coproduct3`.
+
+```kotlin:ank
+import arrow.generic.*
+import arrow.generic.coproduct3.coproductOf
+
+val apiResult = coproductOf<CommonServerError, RegistrationError, Registration>(ServerError) //Returns Coproduct3<CommonServerError, RegistrationError, Registration>
+```
+
+There are `coproductOf` constructor functions for each Coproduct. All we have to do is pass in our value and have the correct type parameters on it, the value must be a type declared on the function call.
```
JorgeCastilloPrz: First line, I'd better say "...for each Coproduct regardless of the arity (number of types)."

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -0,0 +1,167 @@
...
+title: Generic
+permalink: /docs/generic/coproduct/
+---
+
+## Arrow Generic
+
+{:.beginner}
+beginner
+
+`arrow-generic` provides meta programming facilities over Product types like data classes, tuples, and heterogeneous lists; and Coproduct types like sealed classes.
```
JorgeCastilloPrz: What about "`arrow-generic` provides meta programming facilities over Product types (data classes, tuples, heterogeneous lists...). It also provides the Coproduct type, similar to sealed classes. 

(Even if by implementation we're providing N Coproduct types, semantically all of them are the same thing, the Coproduct, I don't think there's need here to state it's a bunch of them, at least not yet on this paragraph)

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
JorgeCastilloPrz: "represent a **sealed hierarchy of types** where only one of the specified set of types exist every time. Conceptually, it's very similar to the stdlib `sealed` class, or to `Either` if we move on to Arrow data types."

JorgeCastilloPrz: "...Car Dealer app that can show `Dealership`s, `Car`s and `SalesPerson`s **on a list**.."

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -0,0 +1,167 @@
...
+import arrow.generic.*
+import arrow.generic.coproduct3.cop
+
+val apiResult = ServerError.cop<CommonServerError, RegistrationError, Registration>() //Returns Coproduct3<CommonServerError, RegistrationError, Registration>
+```
+
+All we have to do is provide the type parameters and we can make a Coproduct using the `cop` extension method. Just like `coproductOf`, if the type of the value isn't in the type parameters of the method call, it won't compile:
+
+```kotlin:ank
+//val apiResult = ServerError.cop<String, RegistrationError, Registration>()
```
JorgeCastilloPrz: Can we write the exception thrown here?

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -0,0 +1,167 @@
...
+```
+
+##### fold
+
+Obviously, we're not just modeling errors for fun, we're going to handle them! All Coproducts have `fold` which allows us to condense the Coproduct down to a single type. For example, we could handle errors as such in a UI:
```
JorgeCastilloPrz: Maybe we can add something about exhaustive evaluation

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -0,0 +1,167 @@
...
+                    StolenCar -> "Car reported stolen!"
+                }
+            },
+            { registration ->
+                "Successfully Registered: $registration.car"
+            }
+    )
+```
+
+Here we're able to return the result of the `fold` and we're forced to handle all cases! Neat!
```
JorgeCastilloPrz: Ah! There it is. Maybe just explicitly mention exhaustive evaluation here. You're saying it but still mentioning it by its real name should help.

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+##### select
+
+Let's say we also want to store the `Registration` object into our database when we successfully register a car. We don't really want to have to `fold` over every single case just to handle something for the `Registration`, this is where `select<T>` comes to the rescue! We're able to take a Coproduct and `select` the type we care about from it. `select` returns an `Option`, if the value of the Coproduct was for the type you're trying to `select`, you'll get `Some`, if it was not the type used with `select`, you'll get `None`.
```
JorgeCastilloPrz: Maybe we could move all the first part until "...select comes to the rescue!" to the previous section. Then you present select as a new section, which would start on "We're able to take a...". That helps to start introducing the concept by the need for it before even giving it a name. It's just a more gradual way to put it.

abergfeld:
@JorgeCastilloPrz I updated the docs, I think I addressed everything. Let me know if there's anything else :)

abergfeld:
@ffgiraldez Would you be able to help me figure out what's happening with the code coverage testing here, it all looks like stuff that hasn't changed as part of this PR.

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -0,0 +1,193 @@
...
+
+`select` can only be called with a type that exists on the Coproduct, if the type doesn't exist, it won't compile:
+```kotlin:ank
+import arrow.generic.*
+import arrow.generic.coproduct3.Coproduct3
+import arrow.generic.coproduct3.select
+
+fun handleApiResult(apiResult: Coproduct3<CommonServerError, RegistrationError, Registration>): Unit {
+//    apiResult.select<String>()
+//error: type mismatch: inferred type is Coproduct3<CommonServerError, RegistrationError, Registration> but Coproduct3<String, *, *> was expected
```
raulraja: This reminds me that Ank should have a mode for expecting and printing errors similar to how we do now `ank:silent`. but `ank:error` so users are not forced to comment out code that fails to compile and can let the compiler provide the actual error that gets rendered along the snippet.

raulraja: /cc @Guardiola31337

ffgiraldez:
I will check It tomorrow

raulraja:
thanks @abergfeld !

PR #999, opened 2018-08-25
Add a Category typeclass, CategoryLaws and make Function1 a Category instance 

paulcadman:
In a category:
* Every object has an identity morphism. This is represented in the typeclass by `id()`.
* Every pair of morphisms can be composed. This is represented in the typeclass by `compose()`.
* Morphisms satisfy three laws: associativity, left identity and right identity.
  These are expressed as CategoryLaws.

`Function1` is a category where:
* The identity morphism is the identity function
* Morphism composition is function composition


paulcadman:
I'm not sure if I should commit all of the infographic file.?

Assuming you think this is OK and useful - what other things do I need to do to add a typeclass to arrow?

modules/core/arrow-core/src/main/kotlin/arrow/core/Function1.kt
```diff
@@ -32,5 +34,7 @@ class Function1<I, out O>(val f: (I) -> O) : Function1Of<I, O> {
     }
 
     fun <I, A, B> tailRecM(a: A, f: (A) -> Function1Of<I, Either<A, B>>): Function1<I, B> = { t: I -> step(a, t, f) }.k()
+
+    fun <I> id(): Function1<I, I> = { a : I -> a }.k()
```
paulcadman: Maybe `ask` should be defined in terms of `id` or vice-versa?

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Category.kt
```diff
@@ -0,0 +1,10 @@
+package arrow.typeclasses
+
+import arrow.Kind2
+
+interface Category<F> {
+
+  fun <A> id(): Kind2<F, A , A>
```
pakoito: Typo: whitespace

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Category.kt
```diff
@@ -0,0 +1,10 @@
...
+
+  fun <A, B, C> Kind2<F, B, C>.compose(arr: Kind2<F, A, B>): Kind2<F, A, C>
```
pakoito: Can we please have `andThen` as the dual of `compose` directly baked into the typeclass?

modules/core/arrow-core/src/main/kotlin/arrow/core/Function1.kt
pakoito: Yeah, maybe even use the ::identity function defined in `arrow-core`

modules/core/arrow-core/src/main/kotlin/arrow/core/Function1.kt
```diff
@@ -32,5 +34,7 @@ class Function1<I, out O>(val f: (I) -> O) : Function1Of<I, O> {
     }
 
     fun <I, A, B> tailRecM(a: A, f: (A) -> Function1Of<I, Either<A, B>>): Function1<I, B> = { t: I -> step(a, t, f) }.k()
+
+    fun <I> id(): Function1<I, I> = Function1(::identity)
```
paulcadman: I used the `Function1` constructor explicitly because the compiler
could not infer the type of `::identity` when I used `::identity.k()`

paulcadman:
I've made changes in the places you've suggested @pakoito - thanks 👍

pakoito:
Brilliant, thanks for the contribution! If you'd like a follow-up PR, you could add the docs, and a reference to the typeclass in the typeclass/intro doc. Else we'll take care of it before the next release :D

PR #1003, opened 2018-09-01
Add effectM to Monad, to ignore the result of a second monadic operation

pakoito:
Simple helper and simple use case

### DIFF
--- a/modules/docs/arrow-docs/docs/docs/typeclasses/monad/README.md
+++ b/modules/docs/arrow-docs/docs/docs/typeclasses/monad/README.md
@@ -95,7 +95,7 @@ ForOption extensions {
 
 #### effectM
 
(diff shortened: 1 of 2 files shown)


modules/docs/arrow-docs/docs/docs/typeclasses/monad/README.md
```diff
@@ -92,6 +93,18 @@ ForOption extensions {
 }
 ```
 
+#### effectM
+
+Executes sequentially two elements and ignores the result of the second. This is useful for effects like logging.
```
JorgeCastilloPrz: "Executes two elements sequentially"

pakoito: Done!

PR #1004, opened 2018-09-01
Add small snippet about datatype representation on the glossary

pakoito:
Title


modules/docs/arrow-docs/docs/docs/patterns/glossary/README.md
```diff
@@ -24,6 +24,21 @@ branching in code with [`Either`]({{ '/docs/datatypes/either' | relative_url }})
 catching exceptions with [`Try`]({{ '/docs/datatypes/try' | relative_url }}),
 or interacting with the platform the program runs in using [`IO`]({{ '/docs/effects/io' | relative_url }}).
 
+The name datatype comes from the representation of these patterns using plain data, with `data` and `sealed` classes.
```
JorgeCastilloPrz: Is it like that? I thought it was more related to the data types being "the data" of a functional program (as in they're used to hold the data to operate on with typeclasses combinators which are the connectors)

modules/docs/arrow-docs/docs/docs/patterns/glossary/README.md
```diff
@@ -24,6 +24,21 @@ branching in code with [`Either`]({{ '/docs/datatypes/either' | relative_url }})
...
+For example, the internal representation of an `Option` is a sealed class with two data classes `Some<A>(val a: A)` and `None`,
+and `Ior` is a sealed class with three data class inheritors, `Left(val a: A)`, `Right(val b: B)`, and `Both(val a: A, val b: B)`.
+
+Lazy datatypes can be recursive, nesting instances of themselves until execution. One example is `IO`.
```
JorgeCastilloPrz: For some reason I feel like there's something missing before this paragraph. You talk about lazy data types and I'm not sure if reader would not what you're talking about exactly. I could imagine what is it, but I wouldn't expect devs not familiarized with FP to know. I would at least introduce laziness around data types before or link to it.

modules/docs/arrow-docs/docs/docs/patterns/glossary/README.md
raulraja: We use the term `Data type` to refer to any data structure for which a type class instance can be targeted so it covers much more encodings than the traditional Algebraic data type encoding,in kotlin based in ADTs. @pakoito in your change you are referring exclusively to Algebraic data types but other data types for example church encoded data types we call them also `data types`. Ex `List` in the Kotlin std lib

pakoito:
@JorgeCastilloPrz there's an ank snippet below where you can see them nested. This is how it renders:

```kotlin
import arrow.effects.*

IO { 0 }
 .continueOn(UI)
 .flatMap { IO { it * 2 } }
 .map { it + 1 }
// Map(source=Bind(cont=ContinueOn(cont=Suspend(thunk=() -> arrow.effects.IO.Pure<A>), cc=Unconfined), g=(A) -> arrow.effects.IO<B>), g=(kotlin.Int) -> kotlin.Int, index=0)
```

pakoito:
I agree I should improve the wording. On it!

JorgeCastilloPrz:
That one looks better 👍. Thanks for the changes 🤗

PR #1006, opened 2018-09-02
Adds docs for Category typeclass

paulcadman:
I'm not sure what to write as a brief description of Category. Should I add the definition into the docs?

So far I've opted to link to Bartosz Milewski's work.


pakoito:
What would you use a category for? Better than focusing on the what it is, go for the what it can do!

paulcadman:
@pakoito maybe I need some more instances in the library before I can show what it could be used for. I will try to add some.

paulcadman:
I'm on holiday until the end of the month, so I won't be working on this until then.

pakoito:
No hurries, I'm just happy you're contributing :D

JorgeCastilloPrz:
In an effort to clean the PR list we've decided to close all the ones that are older than 2 months and consider them obsolete except for very specific circumstances that we will evaluate per PR. This PR has been in standby for too long so we're closing it. Note it can be brought back to life or browsed if at some point we need these docs.

PR #1011, opened 2018-09-04
Try's success/failure extfuns 

### LINKED ISSUE #1009: Add .success() and .failure()
https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-core/src/main/kotlin/arrow/core/Try.kt#L240

Similar to `left`, `right`, `some`, or `liftIO`, we need better constructor extfuns for `Try`.

alissonfpmorais:
Added `success` and `failure` extfuns to `Try` datatype.
Added tests for `success` and `failure` to `TryTest`

It fixes #1009 

Let me know if there's something to change

### DIFF
--- a/modules/core/arrow-data/src/test/kotlin/arrow/data/TryTest.kt
+++ b/modules/core/arrow-data/src/test/kotlin/arrow/data/TryTest.kt
@@ -331,11 +331,12 @@ class TryTest : UnitSpec() {
     }
 
     "success" {
-      "1".toInt().success() shouldBe success
+      10.success() shouldBe success
     }
 
     "failure" {
-      NumberFormatException().failure<Int>() shouldBe failure
+      val ex = NumberFormatException()
+      ex.failure<Int>() shouldBe Failure(ex)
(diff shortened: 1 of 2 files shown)


modules/core/arrow-data/src/test/kotlin/arrow/data/TryTest.kt
```diff
@@ -330,6 +330,14 @@ class TryTest : UnitSpec() {
       failure.transform({ Try { it.toString() } }) { Try { "NaN" } }.get() shouldBe "NaN"
     }
 
+    "success" {
+      "1".toInt().success() shouldBe success
```
pakoito: There's no need to do a transformation here, as if it fails it'll throw instead of create a `Try` :(

alissonfpmorais: Okey! Fixed 😃

alissonfpmorais:
Cool!! 🎉

Issue #1000, opened 2018-08-29
Flipped infix Applicative::ap


toxxmeister:
Current signature of `Applicative`'s `ap` is:
```kotlin
Kind<F, A>.(Kind<F, (A) -> B>): Kind<F, B>
```

By flipping the receiver and parameter signature to:
```kotlin
Kind<F, (A) -> B>.ap(Kind<F, A>): Kind<F, B>
```
and making the function infix, we could achieve pretty nice syntax for working with curried functions:
```kotlin
pure(f.curried()) ap fa ap fb ap fc

// or multiline (parentheses are mandatory):
(pure(f.curried())
        ap complexExpression1
        ap complexExpression2
        ap complexExpression3)
```

Do you think it'd make sense to change the signature of `ap`, or even just include an infix version alongside the existing one?

pakoito:
I'd be happy with a new infix `ap` alongside the existing one, yeah.

toxxmeister:
@pakoito Great!

What would you recommend as a name for the infix version?

There is a platform declaration clash between the original and infix version, and since it's an infix function, we cannot add a dummy parameter to resolve the issue, nor can we add a `@JvmName` annotation since it's an interface method.
This means that we'd either have to drop the original version, which I suppose is not an option, or think of a different name.
Something like `app` or `apply` might work, in my opinion.

pakoito:
`apply` is already a language construct, so let's go with something else. `apPipe` maybe?

toxxmeister:
@pakoito 
I wouldn't consider `apply` a language construct (it's merely a scoping function), and there are no cases I can foresee where there'd be confusion between the two.
I'd definitely be open to `apply` as I think it makes the most sense semantically.

I'm not entirely satisfied with `apPipe` as, in my opinion, piping implies some kind of left-to-right functional composition, while this is functional application in context.
It doesn't entirely make its intent clear.

Another possibility is overloading the `*` operator due to similarity to Haskell's `<*>` operator:

```kotlin
operator fun <A, B> Kind<F, (A) -> B>.times(fa: Kind<F, A>): Kind<F, B>
```

This'd in turn allow for even more elegant syntax:

```kotlin
pure(f.curried()) * fa * fb * fc

// or multiline (parentheses are mandatory):
(pure(f.curried())
        * complexExpression1
        * complexExpression2
        * complexExpression3)
```

[...]

PR #1012, opened 2018-09-04
Flipped infix Applicative::ap (#1000)

toxxmeister:
Implementation of https://github.com/arrow-kt/arrow/issues/1000

Renames old Applicative::ap to apPipe and introduces an infix version
with flipped receiver and parameter under the old name

### DIFF
--- a/modules/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
+++ b/modules/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
@@ -306,8 +306,12 @@ inline fun <A, B> EitherOf<A, B?>.leftIfNull(crossinline default: () -> A): Eith
 fun <A, B> EitherOf<A, B>.contains(elem: B): Boolean =
   fix().fold({ false }, { it == elem })
 
-fun <A, B, C> EitherOf<A, B>.ap(ff: EitherOf<A, (B) -> C>): Either<A, C> =
+fun <A, B, C> EitherOf<A, B>.apPipe(ff: EitherOf<A, (B) -> C>): Either<A, C> =
   ff.fix().flatMap { f -> fix().map(f) }.fix()
+
+@Suppress("NOTHING_TO_INLINE")
+inline infix fun <A, B, C> EitherOf<A, (B) -> C>.ap(fb: EitherOf<A, B>): Either<A, C> =
+  fb.apPipe(this)
 
 fun <A, B> EitherOf<A, B>.combineK(y: EitherOf<A, B>): Either<A, B> =
   when (this) {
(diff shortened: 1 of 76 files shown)


modules/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -306,9 +306,13 @@ inline fun <A, B> EitherOf<A, B?>.leftIfNull(crossinline default: () -> A): Eith
 fun <A, B> EitherOf<A, B>.contains(elem: B): Boolean =
   fix().fold({ false }, { it == elem })
 
-fun <A, B, C> EitherOf<A, B>.ap(ff: EitherOf<A, (B) -> C>): Either<A, C> =
+fun <A, B, C> EitherOf<A, B>.apPipe(ff: EitherOf<A, (B) -> C>): Either<A, C> =
   ff.fix().flatMap { f -> fix().map(f) }.fix()
 
+@Suppress("NOTHING_TO_INLINE")
+inline infix fun <A, B, C> EitherOf<A, (B) -> C>.ap(fb: EitherOf<A, B>): Either<A, C> =
```
pakoito: Don't inline it for now, we can do it later if necessary.

modules/core/arrow-core/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -211,3 +211,6 @@ fun <A> Boolean.maybe(f: () -> A): Option<A> =
 fun <A> A.some(): Option<A> = Some(this)
 
 fun <A> none(): Option<A> = None
+
+@Suppress("NOTHING_TO_INLINE")
+inline infix fun <A, B> OptionOf<(A) -> B>.ap(fa: OptionOf<A>): Option<B> = fa.fix().apPipe(this)
```
raulraja: In 0.8.0 all these functions will be automatically exported because `ap` appears in `Applicative`, so all instances of Applicative will get auto exports of package level extension functions that delegate to the instance. Nothing wrong in having them here for people that include the `data` module but not the `typeclass` one.

toxxmeister: If that is the case, is there really a need for them to be here?
I doubt someone not using typeclasses would find them useful.

raulraja: no need, they can be removed from here since the projection will add this exact same function in the type class syntax packages

JorgeCastilloPrz:
In an effort to clean the PR list we've decided to close all the ones that are older than 2 months and consider them obsolete except for very specific circumstances that we will evaluate per PR. In this case the changes were interesting but have been in standby for too long. Thanks for the contribution and note we can get back to it at some point if it makes sense (PR can be restored). Our syntax has changed so much since 1 year ago also so this might not apply anymore at this point.

1Jajen1:
I'll continue this in a new pr. It is still relevant and we should do this sooner rather than later.

Issue #1015, opened 2018-09-07
Array-backed generic tuples


toxxmeister:
This is an attempt at implementing performant array-backed type-safe generic tuples.
It's very similar to `HList` on type-level, which due to its nature is not very performant.
Furthermore, convenience typealiases are introduced to make them more flexible than current `HList`s.

Do you find this useful? Could the API be improved/expanded?

## Implementation and API

Let's start by introducing `TupN` which serves as our size and type marker:

```kotlin
sealed class TupN

class TupCons<out H, out T : TupN> private constructor() : TupN()

@Suppress("CanSealedSubClassBeObject")
class TupNil private constructor() : TupN()
```

The actual tuple is just a simple array tagged with the specified `N : TupN`:

```kotlin
class Tup<out N : TupN>
  internal constructor(internal vararg val data: Any?)
```

### Structure preserving types

`Tup{N}Cons<A1, ..., AN, TS>` allows us to only focus on the relevant part of the tuple (`A1, ..., AN`), while capturing the rest of the tuple through `TS`.

```kotlin
typealias Tup0Cons<TS> = Tup<TS>
typealias Tup1Cons<A, TS> = Tup0Cons<TupCons<A, TS>>
typealias Tup2Cons<A, B, TS> = Tup1Cons<A, TupCons<B, TS>>
```

### Exact size types

Exact size tuples are defined through `Tup{N}` types:

```kotlin
typealias Tup0 = Tup0Cons<TupNil>
typealias Tup1<A> = Tup1Cons<A, TupNil>
typealias Tup2<A, B> = Tup2Cons<A, B, TupNil>
```

### Lower bound size types

We can define lower bound on tuple size without necessarily caring about the rest of the tuple outside of that size:

```kotlin
typealias Tup0Plus = Tup0Cons<*>
typealias Tup1Plus<A> = Tup1Cons<A, *>
typealias Tup2Plus<A, B> = Tup2Cons<A, B, *>
```

### Structure preserving copy

Copy functions which preserve the rest of the structure are defined:

```kotlin
fun <A, B, TS : TupN> Tup2Cons<A, B, TS>.copy(
  a: A = data[0] as A,
  b: B = data[1] as B
): Tup2Cons<A, B, TS> =
  Tup(a, b, *data.copyOfRange(2, data.size))
```

### Focused copy

If we don't want to carry potential additional data, we can `focusN` instead of `copy`:

```kotlin
[...]

PR #1017, opened 2018-09-07
Array-backed generic tuple (#1015)

toxxmeister:
Implementation of https://github.com/arrow-kt/arrow/issues/1015

### DIFF
--- a/modules/core/arrow-generic/src/main/kotlin/arrow/generic/tup/TupFocus.kt
+++ b/modules/core/arrow-generic/src/main/kotlin/arrow/generic/tup/TupFocus.kt
@@ -1,0 +1,170 @@
+@file:Suppress("UNCHECKED_CAST")
+
+package arrow.generic.tup
+
+fun @receiver:Suppress("unused") Tup0Plus.focus0(): Tup0 = Tup()
+
+fun <A> Tup1Plus<A>.focus1(a: A = data[0] as A): Tup1<A> = Tup(a)
+
+fun <A, B> Tup2Plus<A, B>.focus2(a: A = data[0] as A, b: B = data[1] as B): Tup2<A, B> = Tup(a, b)
+
+fun <A, B, C> Tup3Plus<A, B, C>.focus3(
+  a: A = data[0] as A, b: B = data[1] as B, c: C = data[2] as C
+): Tup3<A, B, C> =
+  Tup(a, b, c)
+
+fun <A, B, C, D> Tup4Plus<A, B, C, D>.focus4(
+  a: A = data[0] as A, b: B = data[1] as B, c: C = data[2] as C, d: D = data[3] as D
+): Tup4<A, B, C, D> =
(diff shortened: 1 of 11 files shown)


pakoito:
I sincerely hope you used codegen for this. I've done it manually before and it was a PITA with just one file, I don't want to imagine 10 of them :D

modules/core/arrow-generic/src/main/kotlin/arrow/generic/tup/TupFocus.kt
```diff
@@ -0,0 +1,170 @@
+@file:Suppress("UNCHECKED_CAST")
+
+package arrow.generic.tup
+
+fun @receiver:Suppress("unused") Tup0Plus.focus0(): Tup0 = Tup()
```
pakoito: TIL

toxxmeister:
@pakoito Yes, I only had to do some reformatting in the end, and even that was painful due to its sheer size.
By the way, do you include your codegen files anywhere in the repository? It might be useful to have it somewhere for future refactoring.

toxxmeister:
@pakoito 
I'll add the script tomorrow, it needs some cleaning up first.

You suggest using `toTup` for `A.with(b: B)` and `to` for the rest, if I understood you right?
Ideally, It'd be great if we could find a word which makes sense semantically, is short, and can be used for all arities (unlike `toTup`/`to`).
Of course, if we don't find anything (which I suspect we won't, since naming things in FP is usually harder than writing the code itself), `toTup` will work.

Another question is whether we should shy away from using `with` here since it's completely unambiguous from Kotlin's `with(T, T.() -> R)`.

pakoito:
I'd rather stay unambiguous, yeah. We get enough pushback for coming up with new things, overloading the existing ones would get even more hahaha

Can do gitter for a naming session :D

raulraja:
I'm concerned about having multiple TupleN representations in Arrow and the confusion that may create for people getting started or even just realizing they just wrote some code for an incompatible type. We need to understand how much bytecode this increases the artifact because if it's similar to the TupleN impl in core and this is more performant and does not expose am unsafe API to end users this should probably live in `core` instead of `generic` replacing the existing TupleN implementation. Tuples are in core becuse we need them in certain type classes like `Applicative`,

JorgeCastilloPrz:
In an effort to clean the PR list we've decided to close all the ones that are older than 2 months and consider them obsolete except for very specific circumstances that we will evaluate per PR. In this case all this should be generated and will be highly influenced by Arrow meta. That, on top of the time this has been on standby requires us to declare it obsolete for the time being. Note this can always be brought back to life if we have it as a priority in our roadmap at any point in time. Thank you for all the effort since it was a really good job! 🙏

raulraja:
@toxxmeister id you are still interested in exploring this with arrow meta type proofs it will be possible to encode a lot of the proposed here in a few weeks when the design is finished and published

PR #1020, opened 2018-09-12
Comonadic UIs datatypes

Cotel:
The purpose of this PR is adding the datatypes shown on [this paper](https://functorial.com/the-future-is-comonadic/main.pdf) and this [simplified version](https://docs.google.com/presentation/d/1UkRASOzbN67KinFq7SqJgsH_Q27iE4jxwgeveznWowM/edit#slide=id.g3d9244ddc6_0_330).

Once we have them implemented I would like to make an example of use and maybe write some kind of tutorial if I get something useful.

For now I have implemented:

- [x] **Store**: Which can be seen as a React Component if you are familiar with the library. It is a datatype which holds a state and a render function which given a state it returns some representation of it.

- [x] **Moore**: A Moore machine acts similarly to Redux or the Elm architecture. It holds a state and in order to change it you need to dispatch some event of a determined type.

- [x] **Cofree**: Cofree is a special case of Moore which defines transitions in terms of functors. It was already implemented.

- [x] **Sum**: The sum of two comonads is also a comonad. We can use this datatype to represent two tabs of an user interface and display one at a time.


pakoito:
Busy afternoon I see.

modules/core/arrow-data/src/main/kotlin/arrow/data/Moore.kt
```diff
@@ -0,0 +1,17 @@
+package arrow.data
+
+import arrow.higherkind
+
+@higherkind
+data class Moore<E, V>(val view: V, val handle: (E) -> Moore<E, V>) : MooreOf<E, V> {
```
pakoito: Can you please check if MooreOfJ is available?

modules/core/arrow-data/src/main/kotlin/arrow/data/Sum.kt
```diff
@@ -0,0 +1,52 @@
...
+    val side: Side = Side.Left
+) : SumOf<F, G, V> {
+
+  sealed class Side {
+    object Left : Side()
+    object Right : Side()
+  }
+
+  fun <A> extend(CF: Comonad<F>, CG: Comonad<G>, f: (Sum<F, G, V>) -> A): Sum<F, G, A> = CF.run {
+    CG.run {
```
pakoito: Android D8 compiler won't like this nesting, sadly. You need to put them just in the function where they're applied.

modules/core/arrow-data/src/main/kotlin/arrow/data/Sum.kt
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  fun <A> map(FF: Functor<F>, FG: Functor<G>, f: (V) -> A): Sum<F, G, A> = FF.run {
+    FG.run {
+      Sum(left.map(f), right.map(f), side)
+    }
+  }
+
+  fun extract(CF: Comonad<F>, CG: Comonad<G>): V = CF.run {
+    CG.run {
```
pakoito: Same comment about D8.

modules/core/arrow-data/src/test/kotlin/arrow/data/MooreTest.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    val EQ = object : Eq<MooreOf<Int, Int>> {
+      override fun MooreOf<Int, Int>.eqv(b: MooreOf<Int, Int>): Boolean =
+          this.fix().extract() == b.fix().extract()
+    }
+
+    ForMoore<Int>() extensions {
+      testLaws(
+          FunctorLaws.laws(Moore.functor(), intMoore, EQ),
+          ComonadLaws.laws(Moore.comonad(), intMoore, EQ)
```
pakoito: Doesn't Comonad imply Functor? You may not need to test both.

modules/core/arrow-data/src/main/kotlin/arrow/data/Moore.kt
```diff
@@ -0,0 +1,17 @@
...
+  fun <A> extend(f: (Moore<E, V>) -> A): Moore<E, A> =
+      Moore(f(Moore(view, handle))) { update -> handle(update).extend(f) }
+
+  fun <A> map(f: (V) -> A): Moore<E, A> =
+      Moore(f(view)) { update -> handle(update).map(f) }
+
+  fun extract(): V = view
+
+  companion object
+}
```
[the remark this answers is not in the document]
pakoito: The logo doesn't show in the pull page, just on the diff one.

modules/core/arrow-data/src/main/kotlin/arrow/data/Store.kt
```diff
@@ -0,0 +1,16 @@
+package arrow.data
+
+import arrow.higherkind
+
+@higherkind
+data class Store<S, V>(val state: S, private val render: (S) -> V) : StoreOf<S, V> {
```
pakoito: Why make render private here?

modules/core/arrow-data/src/test/kotlin/arrow/data/MooreTest.kt
```diff
@@ -0,0 +1,90 @@
...
+
+@RunWith(KTestJUnitRunner::class)
+class MooreTest : UnitSpec() {
+
+  init {
+
+    fun handle(x: Int): Moore<Int, Int> = Moore(x, ::handle)
+    val intMoore: (Int) -> MooreOf<Int, Int> = { x: Int -> Moore(x, ::handle) }
+
+    val EQ = object : Eq<MooreOf<Int, Int>> {
```
pakoito: EQ has a lambda constructor for which you only need to annotate the parameters, it may make this easier.

modules/core/arrow-data/src/test/kotlin/arrow/data/MooreTest.kt
Cotel: You are right 👌

modules/core/arrow-data/src/main/kotlin/arrow/data/Store.kt
Cotel: I made it private because I see no sense in allowing the user to use that function with any other state than the one of that Store and we have `extract` for that case.

modules/core/arrow-data/src/main/kotlin/arrow/data/Sum.kt
Cotel: I'm not sure if I understood this. How can I tell the compiler that F and G are Comonads then?

Cotel: Is this correct?

```kotlin
fun <A> extend(CF: Comonad<F>, CG: Comonad<G>, f: (Sum<F, G, V>) -> A): Sum<F, G, A> =
  Sum(
    CF.run { left.coflatMap { f(Sum(left, right, Side.Left)) } },
    CG.run { right.coflatMap { f(Sum(left, right, Side.Right)) } },
    side
  )  
```

pakoito: yeah, like that

modules/core/arrow-data/src/main/kotlin/arrow/data/Store.kt
pakoito: I don't think we're making properties in Kleisli or Reader/Writer/State, so it's okay to keep them public.

modules/core/arrow-data/src/main/kotlin/arrow/data/Moore.kt
Cotel: There is a `MooreKindedJ` if that's what you mean. `StoreKindedJ` and `SumKindedJ` exist too.

Cotel:
I'm having some troubles to implement `Day#extend` so let me explain myself so you can see where is the error.

Day datatype has the following aspect:

```kotlin
data class Day<F, G, X, Y, A>(val left: Kind<F, X>, val right: Kind<G, Y>, val get: (X, Y) -> A)
```

And it's extend function should be:

```kotlin
fun <B> extend(f: (Day<F, G, X, Y, A) -> B): Day<F, G, X, Y, B>
```

I have implemented it this way:

```kotlin
fun <B> extend(CF: Comonad<F>, CG: Comonad<G>, f: (Day<F, G, X, Y, A>) -> B): Day<F, G, X, Y, B> = Day(
    CF.run { left.coflatMap { f(Day(left, right, get)) } },
    CG.run { right.coflatMap { f(Day(left, right, get)) } },
    { left, right -> f(Day(left, right, get)) }
)
````

But this returns `Day<F, G, B, B ,B>`. The only way I see this can work is to ignore left and right and just change the `get` function but I am pretty sure this is not the solution as the sides of the datatype won't be evolving.

```kotlin
fun <B> extend(CF: Comonad<F>, CG: Comonad<G>, f: (Day<F, G, X, Y, A>) -> B): Day<F, G, X, Y, B> = Day(
    left,
    right,
    { _, _ -> f(Day(left, right, get)) }
)
```

Any ideas? 🤔

modules/core/arrow-data/src/main/kotlin/arrow/data/Moore.kt
pakoito: Yeah those I meant. Make the datatypes implement them!

pakoito:
Your signature is for map, if I understand correctly. Shouldn't extend be `fun <B> extend(f: (Day<F, G, X, Y, A) -> B): Day<F, G, X, Y, Day<F, G, X, Y, A>>`?

pakoito:
One change: `extend` is called `duplicate` in Arrow. I'll fix it myself.

pakoito:
It passes the laws, so I'll take it :shrug:

pakoito:
Purescript is also betraying me: https://github.com/paf31/purescript-day/blob/master/src/Data/Functor/Day.purs#L108-L109

Cotel:
I've managed to pass the laws again but I'm not sure if this will do the trick. The thing that is itching me the most is that for `coflatMap` you need `Applicative` instances and for `ap` you need `Comonad`s 😵. The `invoke` operator I added as a constructor needs `Applicative` instances too.

I want to make some more tests besides the laws to see if this is correct.

pakoito:
After another pass at the encoding I found something that meets all requirements, passes the tests, and looks like the purescript implementation.

Missing: tests in context to learn how to use it.

Cotel:
I'm writing some usage tests for Day and I just realized we cannot access left or right sides, so when using `coflatMap` we cannot transform them and the result will remain the same. The only thing we can use to alter these sides is `extract` but it does not allow to treat them separately except if the result type is a Tuple. Is this correct?

For example, If I have a Day with two Id holding Ints, when extending it (coflatMap) I would like the result to be a String indicating which side is greater. I cannot make this happen because I don't have access to left nor right.

Cotel:
I'm writing some docs for these datatypes before merging so we can have something to show.

pakoito:
@Cotel was the fix just type hinting then?

Cotel:
Yep 👍

pakoito:
Can you please do me a favor. Check whether using the lazy version of the typeclass methods 
i.e. `mapLazy` or `coflatMapLazy` makes them fail the tests, or have any meaningfully different behavior.

Cotel:
There is a test for `coflatMapLazy` if that's what you mean and it works with no difference from the eager version. I can write a test for `mapLazy` too if you want.

Cotel:
I've replaced them locally and both lazy and eager versions pass the laws. I would leave the lazy versions as this is taken from Haskell and for example Store is a `lazy unfolding of all possible states`. What do you think? 🤔

pakoito:
Yeah, let's keep the lazy versions then. To have consistency let's add an alias for `extract` called `runDay` and keep both, then rename the current private `runDay` to something else like `stepDay`.

modules/core/arrow-data/src/main/kotlin/arrow/data/Sum.kt
```diff
@@ -17,7 +17,7 @@ data class Sum<F, G, V>(
     object Right : Side()
   }
 
-  fun <A> coflatMap(CF: Comonad<F>, CG: Comonad<G>, f: (Sum<F, G, V>) -> A): Sum<F, G, A> = Sum(
+  fun <A> coflatmap(CF: Comonad<F>, CG: Comonad<G>, f: (Sum<F, G, V>) -> A): Sum<F, G, A> = Sum(
```
pakoito: It's coflatMap AFAIK.

Cotel: You changed it recently from `extend` to `coflatmap` and this was the only left to change. I'll change them all to `coflatMap`.

Cotel:
So this is all I could write as docs for these datatypes. I know it is too little (specially Day) but I don't know how to give some guidance until I test this on a project or something.

Talking about `Day`, I still have no idea on how this datatype can evolve its states. I've been trying to write some example for `coflatMap` but since I cannot access its left or right side it's not useful to use it and `map` can only transform the `extract` result taking as a parameter the last `extract` result so we cannot use it to evolve states.

Cotel:
This is ready for merging. I've been working on some examples on Android with `Store` and `Moore` but there is work to do before showing anything. I'll keep you updated 😁

pakoito:
Party time then! Awesome job Miguel, thanks 💃💃💃💃💃

Ping me on Slack whenever you'd like some advice.

PR #1024, opened 2018-09-13
[pipe] remove inline modifier

Laimiux:
## What:
I removed all inline modifiers from all `pipe` functions that return a lambda. This will avoid implicitly allocating anonymous classes at the call-site and instead will allocate them once for each `pipe` method declaration. 

Part of #966

### DIFF
--- a/modules/core/arrow-syntax/src/main/kotlin/arrow/syntax/function/pipe.kt
+++ b/modules/core/arrow-syntax/src/main/kotlin/arrow/syntax/function/pipe.kt
@@ -1,47 +1,47 @@
 package arrow.syntax.function
 
+infix fun <P1, R> P1.pipeLazy(t: (P1) -> R): () -> R = { t(this) }
+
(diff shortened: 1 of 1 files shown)


Laimiux:
I left inline modifier for 
```infix inline fun <P1, R> P1.pipe(t: (P1) -> R): R = t(this)```
as it doesn't allocate any new lambdas. I can remove it here as well for consistency sake.

Laimiux:
Let me know if you have questions or want some changes @pakoito

Laimiux:
It seems there are some weird inconsistencies with codecov analysis. map.kt file hasn't changed, curious why it's adding it to impacted files. 

Also, I don't have the rights to merge this. What's the usual process around contributing?

pakoito:
Once CI passes and the branch is in line with master and any contributor has approved the diff, a maintainer can merge it.

Over time we add frequent contributors to the list of reviewers.

PR #1028, opened 2018-09-15
[WIP] Add Bracket

### LINKED ISSUE #1023: Bracket
https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/Bracket.scala

https://github.com/scalaz/scalaz-zio/blob/master/microsite/src/main/tut/usage/bracket.md

Bracket models the standard imperative `try catch finally` block but as a Typeclass working over Higher Kinds. This ticket will unlock other typeclasses we need to work on, like `Concurrent`, since it's kind of in the top of the hierarchy, right below `MonadError`. 

typeclass_diagram

`MonadError` and `Async` and `MonadDefer` (Sync) are already there. Bracket instance must always extend MonadError instance for the same data type, and MonadDefer instance extend from the Bracket one. Instances must follow the same hierarchy than the actual Typeclasses.

It also requires `IOBracket` implementation which is a bit more complex.

pakoito:
Closes #1023


JorgeCastilloPrz:
Ah! I'll need this for Kollect

pakoito:
@JorgeCastilloPrz Most of the things on the diff that you saw are merged now. Check the latest version.

I'm going to work on Concurrent before working on Bracket. AFAIK it shouldn't be hard.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Bracket.kt
```diff
@@ -0,0 +1,21 @@
...
+
+  fun <F, A, B> Kind<F, A>.bracketCase(release: (A, ExitCase<E>) -> Kind<F, Unit>, use: (A) -> Kind<F, B>): Kind<F, B>
+
+  fun <A, B> Kind<F, A>.bracket(release: (A) -> Kind<F, Unit>, use: (A) -> Kind<F, B>): Kind<F, B> =
+    bracketCase({ a, _ -> release(a) }, use)
+
+  fun <A> Kind<F, A>.uncancelable(): Kind<F, A> =
+    bracket({ just(Unit) }, { just(it) })
+
+  fun <A> Kind<F, A>.guarantee(finalizer: Kind<F, Unit>): Kind<F, A> =
```
alissonfpmorais: I was reading this code to learn more about typeclass implementation and I was left with a question. In [Cats docs](https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/Bracket.scala) `guarantee` has this signature and implementation:

`def guarantee[A](fa: F[A])(finalizer: F[Unit]): F[A] = bracket(unit)(_ => fa)(_ => finalizer)`

In this case, shouldn't signature use the receiver type `Kind<F, Unit>` instead of `Kind<F, A>`? Something like: 

`fun <A> Kind<F, Unit>.guarantee(finalizer: Kind<F, Unit>, fa: Kind<F, A>): Kind<F, A> = bracket({ _ -> finalizer }, { _ -> fa })`

Sorry about newbie level question and thanks in advance 😄

JorgeCastilloPrz: Note that the Scala one gets an additional argument which is Fine<A>. That's what we move to be the receiver in Arrow.

alissonfpmorais: Yes, `guarantee` has `fa: F<A>`  as a parameter, but it call `bracket` passing `unit` as `acquire: F<A>` and `fa` as `use: A => F<B>`.

In Kotlin, aren't we using the receiver type as `acquire` param? If so, it looks like the receiver type `Kind<F, A>` are used in both `acquire` and `use` params of the `bracket` method.

pakoito: The method guarantee reads as follows: for the resource `A` provided, run `finalizer` when it completes. So your use would be as so:

```
val resource: Res = newRes()
just(resource).guarantee(invoke { resource.close() }) // Kind<F, A>
```

Which is a simpler version of:
```
val resource: Res = newRes()
bracket({ invoke { resource.close() } }) { just(it) } // Kind<F, A>
```

alissonfpmorais: Oh, now I got the idea. Reading that way makes totally sense the signature used!

Thank you both! :smile:

JorgeCastilloPrz:
Not needed anymore.

PR #1029, opened 2018-09-18
Add Collection<Either<*, A>>.collect: Collection<A>

nxtstep:
A nice little convenience function for Collections with Either<> types to filter out the `Rights`.

### DIFF
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -361,3 +361,13 @@ fun <A, B> B?.rightIfNotNull(default: () -> A): Either<A, B> = when (this) {
   null -> Either.Left(default())
   else -> Either.Right(this)
 }
+
+/**
+ * Convenience extension for [Collection<Either<*, A>>] to get a [Collection<A>] with only the Either.rights.
+ *
+ * Example:
+ * ```
+ * listOf("value".right(), "other".left()).collect() // ["value"]
+ * ```
+ */
+fun <A> Collection<Either<*, A>>.collect(): Collection<A> where A : Any = mapNotNull { it.orNull() }
--- a/modules/core/arrow-core-data/src/test/kotlin/arrow/core/EitherTest.kt
+++ b/modules/core/arrow-core-data/src/test/kotlin/arrow/core/EitherTest.kt
@@ -129,6 +129,12 @@ class EitherTest : UnitSpec() {
       }
     }
 
+    "collect should return only the Right values" {
+      forAll { a: Int, b: Int ->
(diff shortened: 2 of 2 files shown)


nxtstep:
@pakoito Thanks for your feedback. I've renamed the function to `collect` as the signature already indicates `A` is non-optional I deemed it not necessary to include that into the function name. (Of course A could be `.rights` of Option<*>)

pakoito:
Nice, that's a good first step. My point about `Option` is that what I'd like is something that looks like

`fun <F, A> Collection<Kind<F, A>>.collect(XXX): Collection<A> where A : Any =`

where I could call as

listOf(1,2,3).collect(Option.XXX())
listOf(1,2,3).collect(Some.XXX())
listOf(1,2,3).collect(Try.XXX())
listOf(1,2,3).collect(Observable.XXX())

etc...

Now for what XXX can be, it's probably `Foldable` or `Traverse`, but that's for you to discover :D

modules/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -340,3 +340,13 @@ fun <A, B> B?.rightIfNotNull(default: () -> A): Either<A, B> = when (this) {
...
+
+/**
+ * Convenience extension for [Collection<Either<*, A>>] to get a [Collection<A>] with only the Either.rights.
+ *
+ * Example:
+ * ```
+ * listOf("value".right(), "other".left()).collect() // ["value"]
+ * ```
+ */
+fun <A> Collection<Either<*, A>>.collect(): Collection<A> where A : Any = mapNotNull { it.orNull() }
```
raulraja: This can be generalized for all `Foldable` data types not just either. In Arrow 0.8.0 all extensions declared in Foldable will be projected as extension functions over the datatype. We already have also a version of `Collect` elsewhere which uses `PartialFunction` that could potentially be moved to Foldable as well. In general, we prefer in Arrow to not provide concrete extension functions for a particular datatype if the same functionality can be expressed polymorphically as part of a type class that many data types provide instances for.

I would encourage you to explore moving collect with `PartialFunction` inside the `Foldable` interface

JorgeCastilloPrz:
In an effort to clean the PR list we've decided to close all the ones that are older than 2 months and consider them obsolete except for very specific circumstances that we will evaluate per PR. In this case, this has been on standby after requesting changes for too long (more than a year), and the extension was kind of [already there](https://github.com/arrow-kt/arrow/blob/93091c50314b22b2589455bfa127b735ce2da7d2/modules/core/arrow-syntax/src/main/kotlin/arrow/syntax/collections/iterable.kt) even if defined for `Iterable` not specific to `Iterables` with `Either` elements. It can also be written using `filter` or `filterMap` from `FunctorFilter`, so we're closing this for the time being.

PR #1031, opened 2018-09-26
Publish a shadowed uber jar of Arrow

raulraja:
The uber jar contains relocation packages so it may be consumed by `arrow-meta`.
Uses Arrow previous version to build Arrow and other `arrow-meta` dependents.


modules/shadow/arrow-shadow/build.gradle
```diff
@@ -0,0 +1,46 @@
...
+    dependencies {
+        classpath 'com.github.jengelman.gradle.plugins:shadow:4.0.0'
+    }
+}
+
+apply plugin: 'com.github.johnrengelman.shadow'
+apply plugin: 'java'
+
+ext {
+    metaArrowVersion = '0.7.3'
```
pakoito: Should this be on the root build.gradle instead?

raulraja: yeah, I'll move it there as part of the `arrow-meta` uber PR

raulraja:
@pakoito what do we do in cases where codecov BS shows up decreasing like this even if there are no code changes?

pakoito:
@raulraja we always merge regardless of codecov, AFAIK. It's a metric, but not one we follow hahah

raulraja:
LOL, I like that 🔥

PR #1035, opened 2018-10-06
Add `compareTo` overload for `Order`

### LINKED ISSUE #729: Add operator overloads to Syntax classes
Syntax modules enable functions that weren't possible before. In this case we could define Kotlin's operators:
* plus() in SemiGroup
* equals() in Eq
* compareTo() in Order

And any others in here: https://kotlinlang.org/docs/reference/operator-overloading.html

Mishkun:
This pull-request provides `compareTo` operators for all `Order` instances as described in #729.
There are two more operators mentioned in #729:

* Equality operator (`==`, `!=`) for `Eq` instances  - _it is impossible to implement that as extension function, because member functions always shadow extensions and equals is defined for every object in Kotlin (because of Java legacy)_
* Plus operator (`+`) for `Semigroup` instances - _already added in #768 _

It seems that all three items of issue would resolved after this PR is merged, so it will close #729

### DIFF
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/OrderLaws.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/OrderLaws.kt
@@ -22,7 +22,8 @@ object OrderLaws {
       Law("Order law: totality order") { O.totalityOrder(fGen) },
       Law("Order law: compare order") { O.compareOrder(fGen) },
       Law("Order law: min order") { O.minOrder(fGen) },
-      Law("Order law: max order") { O.maxOrder(fGen) }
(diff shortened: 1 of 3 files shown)


nomisRev:
Should we replace `compare` by `compareTo` since they're the same function?

This would also make a nice example of the usefulness of typeclasses. Since many people are familiar with the `compareTo` operator magic, we can explain how you could do it his easily with typeclasses yourself.

pakoito:
For consistency with the other typeclasses, I'd keep them separated for now. We can still use it as an example, and deprecate later.

@Mishkun Thanks for the PR, well done adding the laws and docs too :)

PR #1037, opened 2018-10-09
Adding integration for Retrofit

leandroBorgesFerreira:
This PR is a adapter for Arrow's Async classes. 

Example of use:

```
interface ApiClientTest {
  @GET("test")
  fun test(): CallK<ResponseMock>

  @GET("testIO")
  fun testIO(): IO<ResponseMock>
}
```

Using this plugin, it will be possible to set IO as a return type for a endpoint. It will be also possible the set CallK to create polymorphic return types that can be concretised in a latter moment.

### DIFF
--- a/modules/integrations/arrow-integrations-retrofit-adapter/build.gradle
+++ b/modules/integrations/arrow-integrations-retrofit-adapter/build.gradle
@@ -1,0 +1,21 @@
+dependencies {
+    def retrofitVersion = "2.4.0"
+
+    compile project(':arrow-effects')
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
+    compile project(':arrow-annotations')
+    compile "com.squareup.retrofit2:retrofit:2.4.0"
+    kapt project(':arrow-annotations-processor')
+    kaptTest project(':arrow-annotations-processor')
+    compileOnly project(':arrow-annotations-processor')
+    testCompileOnly project(':arrow-annotations-processor')
+    testCompile "io.kotlintest:kotlintest:$kotlinTestVersion"
+    testCompile project(':arrow-test')
+    testCompile "com.squareup.retrofit2:converter-gson:${retrofitVersion}"
+    testCompile 'io.kotlintest:kotlintest-runner-junit5:3.1.10'
+    testCompile 'com.squareup.okhttp3:mockwebserver:3.11.0'
+    testCompile project(':arrow-effects-rx2')
+}
+
+apply from: rootProject.file('gradle/gradle-mvn-push.gradle')
+apply plugin: 'kotlin-kapt'
--- a/modules/integrations/arrow-integrations-retrofit-adapter/build.gradle
+++ b/modules/integrations/arrow-integrations-retrofit-adapter/build.gradle
@@ -9,10 +9,8 @@ dependencies {
     kaptTest project(':arrow-annotations-processor')
     compileOnly project(':arrow-annotations-processor')
     testCompileOnly project(':arrow-annotations-processor')
-    testCompile "io.kotlintest:kotlintest:$kotlinTestVersion"
     testCompile project(':arrow-test')
     testCompile "com.squareup.retrofit2:converter-gson:${retrofitVersion}"
-    testCompile 'io.kotlintest:kotlintest-runner-junit5:3.1.10'
     testCompile 'com.squareup.okhttp3:mockwebserver:3.11.0'
     testCompile project(':arrow-effects-rx2')
 }
--- a/modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/Async2CallAdapterFactory.kt
+++ b/modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/Async2CallAdapterFactory.kt
@@ -1,0 +1,38 @@
+package arrow.integrations.retrofit.adapter
+
+import arrow.effects.IO
+import retrofit2.CallAdapter
+import retrofit2.Retrofit
+import java.lang.reflect.ParameterizedType
+import java.lang.reflect.Type
+
+class Async2CallAdapterFactory : CallAdapter.Factory() {
+
+  companion object {
+    fun create() = Async2CallAdapterFactory()
+  }
+
+  override fun get(returnType: Type, annotations: Array<Annotation>, retrofit: Retrofit): CallAdapter<*, *>? {
+    val rawType = CallAdapter.Factory.getRawType(returnType)
+
+    if (returnType !is ParameterizedType) {
+      val name = parseTypeName(returnType)
+      throw IllegalArgumentException("Return type must be parameterized as " +
+        "$name<Foo> or $name<out Foo>")
+    }
+
+    val effectType = CallAdapter.Factory.getParameterUpperBound(0, returnType)
+
+    return when (rawType) {
+      IO::class.java -> IO2CallAdapter<Type>(effectType)
+      CallK::class.java -> CallKind2CallAdapter<Type>(effectType)
+      else -> null
+    }
+  }
+}
+
+private fun parseTypeName(type: Type) =
+  type.typeName
+    .split(".")
+    .last()
+
--- a/modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/CallK.kt
+++ b/modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/CallK.kt
@@ -1,0 +1,33 @@
+package arrow.integrations.retrofit.adapter
+
+import arrow.Kind
+import arrow.effects.typeclasses.Async
+import arrow.effects.typeclasses.MonadDefer
+import arrow.typeclasses.MonadError
+import retrofit2.Call
+import retrofit2.HttpException
+import retrofit2.Response
+
+data class CallK<R>(val call: Call<R>) {
+  fun <F> async(async: Async<F>): Kind<F, R> =
+    async.async { proc -> call.enqueue(ResponseCallback(proc)) }
+
+  fun <F> defer(defer: MonadDefer<F>): Kind<F, R> =
+    defer {
+      handleResponse(call.execute())
+    }
+
+  fun <F> catch(defer: MonadError<F, Throwable>): Kind<F, R> =
+    defer.run {
+      catch {
+        handleResponse(call.execute())
+      }
+    }
+}
+
+private fun <R> handleResponse(response: Response<R>) =
+  if (response.isSuccessful) {
+    response.body() ?: throw IllegalStateException("The request returned a null body")
+  } else {
+    throw HttpException(response)
+  }
--- a/modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/IO2CallAdapter.kt
+++ b/modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/IO2CallAdapter.kt
@@ -1,0 +1,12 @@
+package arrow.integrations.retrofit.adapter
+
+import arrow.effects.IO
+import retrofit2.Call
+import retrofit2.CallAdapter
+import java.lang.reflect.Type
+
+class IO2CallAdapter<R>(private val type: Type) : CallAdapter<R, IO<R>> {
+  override fun adapt(call: Call<R>): IO<R> = IO.async { proc -> call.enqueue(ResponseCallback(proc)) }
+
+  override fun responseType(): Type = type
+}
--- a/modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/ResponseCallback.kt
+++ b/modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/ResponseCallback.kt
@@ -1,0 +1,28 @@
+package arrow.integrations.retrofit.adapter
+
+import arrow.core.Either
+import arrow.core.left
+import arrow.core.right
+import retrofit2.Call
+import retrofit2.Callback
+import retrofit2.HttpException
+import retrofit2.Response
+
+class ResponseCallback<R>(private val proc: (Either<Throwable, R>) -> Unit) : Callback<R> {
+  override fun onResponse(call: Call<R>, response: Response<R>) {
+    if (response.isSuccessful) {
+      val body = response.body()
+      if (body != null) {
+        proc(body.right())
+      } else {
+        proc(IllegalStateException("The request returned a null body").left())
+      }
+    } else {
+      proc(HttpException(response).left())
+    }
+  }
+
+  override fun onFailure(call: Call<R>, throwable: Throwable) {
+    proc(throwable.left())
+  }
+}
--- a/modules/integrations/arrow-integrations-retrofit-adapter/src/test/kotlin/arrow/integrations/retrofit/adapter/Async2CallAdapterFactoryTest.kt
+++ b/modules/integrations/arrow-integrations-retrofit-adapter/src/test/kotlin/arrow/integrations/retrofit/adapter/Async2CallAdapterFactoryTest.kt
@@ -1,0 +1,50 @@
+package arrow.integrations.retrofit.adapter
+
+import arrow.effects.IO
+import arrow.test.UnitSpec
+import com.google.common.reflect.TypeToken
+import io.kotlintest.KTestJUnitRunner
+import io.kotlintest.shouldBe
+import io.kotlintest.shouldThrow
+import io.kotlintest.specs.StringSpec
+import org.junit.runner.RunWith
+import retrofit2.Retrofit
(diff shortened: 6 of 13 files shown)


modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/CallK.kt
```diff
@@ -0,0 +1,33 @@
...
+import arrow.Kind
+import arrow.effects.typeclasses.Async
+import arrow.effects.typeclasses.MonadDefer
+import arrow.typeclasses.MonadError
+import retrofit2.Call
+import retrofit2.HttpException
+import retrofit2.Response
+
+data class CallK<R>(val call: Call<R>) {
+  fun <F> async(async: Async<F>): Kind<F, R> =
```
pakoito: I'd replace this version with the one that uses `Kind<F, Response<R>>` as the return type. I wouldn't impose the semantics of `handleResponse` because you don't give users the chance to inspect other fields of the response.

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/CallK.kt
```diff
@@ -0,0 +1,33 @@
...
+    defer.run {
+      catch {
+        handleResponse(call.execute())
+      }
+    }
+}
+
+private fun <R> handleResponse(response: Response<R>) =
+  if (response.isSuccessful) {
+    response.body() ?: throw IllegalStateException("The request returned a null body")
```
pakoito: Don't throw, use `raiseException` instead.

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/ResponseCallback.kt
```diff
@@ -0,0 +1,28 @@
...
+import arrow.core.Either
+import arrow.core.left
+import arrow.core.right
+import retrofit2.Call
+import retrofit2.Callback
+import retrofit2.HttpException
+import retrofit2.Response
+
+class ResponseCallback<R>(private val proc: (Either<Throwable, R>) -> Unit) : Callback<R> {
+  override fun onResponse(call: Call<R>, response: Response<R>) {
```
pakoito: Same as mentioned above, I wouldn't unpack the `Response<R>` and would let users decide instead.

modules/integrations/arrow-integrations-retrofit-adapter/src/test/kotlin/arrow/integrations/retrofit/adapter/Async2CallAdapterFactoryTest.kt
```diff
@@ -0,0 +1,47 @@
...
+import okhttp3.HttpUrl
+import org.junit.runner.RunWith
+import retrofit2.Retrofit
+import retrofit2.converter.gson.GsonConverterFactory
+
+private val NO_ANNOTATIONS = emptyArray<Annotation>()
+
+@RunWith(KTestJUnitRunner::class)
+class Async2CallAdapterFactoryTest : UnitSpec() {
+  private val retrofit = retrofit(HttpUrl.parse("http://localhost:1")!!)
```
pakoito: This can be cached outside of the test, same as NO_ANNOTATION.

modules/integrations/arrow-integrations-retrofit-adapter/build.gradle
```diff
@@ -0,0 +1,19 @@
+dependencies {
+    def retrofitVersion = "2.4.0"
```
pakoito: I'd move this to the main `build.gradle`.

modules/integrations/arrow-integrations-retrofit-adapter/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+
+    compile project(':arrow-effects')
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
+    compile project(':arrow-annotations')
+    compile "com.squareup.retrofit2:retrofit:2.4.0"
```
pakoito: Use `retrofitVersion` here.

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/CallK.kt
leandroBorgesFerreira: Do you mean `raiseError`? I don't see a raiseException in the code. 

But, anyway, then I would need to wrap an then unwrap the result to conform with the signature:
```
invoke(f: () -> A)
```

defer is already catching any kind of Throwable. Wouldn't throw make the code simpler?

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/ResponseCallback.kt
leandroBorgesFerreira: You're right, I am imposing an semantics on the client and we should let then decide between `Response<R>` and just `R`. But there's no need to create an extension function for this, or to change the API. The API could be the same, we just need the to read the ReturnType at the Async2CallAdapterFactory and then create the right adapter for each response. If one uses CallK<R> him/she gets the current solution, but if he/she uuses CallK<Response<R>> him/she gets yours. WDYT?

leandroBorgesFerreira: That's pretty much how the RxJava2 plugin works

pakoito: My concern is with discoverability, as it won't be immediately apparent that `CallK<Response<X>>` has different semantics than any other value. I'd rather it was a different function of the `CallK` API, like `asyncResponse` for example.

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/CallK.kt
pakoito: Then, in this case `defer` should behave differently than `catch`. Defer returns a value, catch will throw.

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/ResponseCallback.kt
```diff
@@ -0,0 +1,28 @@
...
+    if (response.isSuccessful) {
+      val body = response.body()
+      if (body != null) {
+        proc(body.right())
+      } else {
+        proc(IllegalStateException("The request returned a null body").left())
```
pakoito: I've been thinking about it. Doesn't POST and HEAD request imply the possibility of an empty body? Or does retrofit return `null` for failures to deserialize? Maybe it should be `R?` or `Option<R>` instead.

leandroBorgesFerreira: Post would be an IO<Unit> or CallK<Unit>. It wouldn't return `null`, it would return Unit. Error to deserialize go straight to `onFailure`. But I should include tests for those cases

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/Async2CallAdapterFactory.kt
```diff
@@ -0,0 +1,38 @@
+package arrow.integrations.retrofit.adapter
```
raulraja: For what I can see in the file name this is in a folder with `.` but each one of the sub packages should be on its own folder

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/Async2CallAdapterFactory.kt
```diff
@@ -0,0 +1,38 @@
...
+import arrow.effects.IO
+import retrofit2.CallAdapter
+import retrofit2.Retrofit
+import java.lang.reflect.ParameterizedType
+import java.lang.reflect.Type
+
+class Async2CallAdapterFactory : CallAdapter.Factory() {
+
+  companion object {
+    fun create() = Async2CallAdapterFactory()
```
raulraja: All return types in Arrow are always fully typed and we don't rely on inference.

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/CallK.kt
```diff
@@ -0,0 +1,33 @@
...
+
+  fun <F> catch(defer: MonadError<F, Throwable>): Kind<F, R> =
+    defer.run {
+      catch {
+        handleResponse(call.execute())
+      }
+    }
+}
+
+private fun <R> handleResponse(response: Response<R>) =
```
raulraja: type return types of all functions

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/CallK.kt
raulraja: `catch` does not return a kinded value but `deffer` does. In whatever case, we should not throw exceptions since these integrations are used by others as an example and Arrow provides mechanisms to avoid throwing exceptions and instead of favors representing them as value. You don't need `catch` wrapping those if you just call `raiseError` instead of throwing exceptions.

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/IO2CallAdapter.kt
```diff
@@ -0,0 +1,12 @@
+package arrow.integrations.retrofit.adapter
+
+import arrow.effects.IO
+import retrofit2.Call
+import retrofit2.CallAdapter
+import java.lang.reflect.Type
+
+class IO2CallAdapter<R>(private val type: Type) : CallAdapter<R, IO<R>> {
```
raulraja: I believe you can define these adapters polymorphically so they can support a lot more data types than just Call and IO. Something along the lines of:
```kotlin
class GenericCallAdapter<F, A>(private val type: Type, private val delay: Async<F>) : CallAdapter<A, Kind<F, A>>, Async<F> by delay {
  override fun adapt(call: Call<A>): Kind<F, A> = async { proc -> call.enqueue(ResponseCallback(proc)) }
   override fun responseType(): Type = type
}
```

raulraja: You may then make them concrete where needed by passing in the async instance of each of the data types you wish to support. With a generic adapter you may already be supporting Observable, Flux, Deferred and many others

pakoito: That's what the other adapter, `CallK`, does. We cannot create and register an adapter because retrofit is implemented by doing runtime checks on a Class return using reflection, and it doesn't get all the generics we need for us to create our own internally. `CallK` fixes that, check the code.

raulraja: If `CallK` is already polymorphic why do we need then an `IO` call adapter? Couldn;t the user use the `CallK` adapter to get back responses on `IO`?

pakoito: To avoid the extra step if users want to. I'm on the fence about it because we have to decide which semantics we'll need to enforce for it.

raulraja: The way I see it this needs `arrow-retrofit-coroutines`, `arrow-retrofit-rx2`, etc. for each one of the currently supported effect types and with those modules you get the specialization for the runtime you are gonna support

leandroBorgesFerreira:
Shouldn't have pushed code... Still working on the solution.

raulraja:
@leandroBorgesFerreira no worries, we will squash and merge when ready, intermediate broken commits are fine.

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow.integrations.retrofit.adapter/IO2CallAdapter.kt
leandroBorgesFerreira: My idea for the IO adapter is to create a way for a developer be able to use the plugin without understanding how to create polymorphic code, which is not a very simple concept. 
I don't mind removing the adapter, but we should keep in mind that documentation must teach how to concretise the type for someone that is not very used advanced concepts.

leandroBorgesFerreira:
I believe I took care of all the observations. Let me know if there's still more to do

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow/integrations/retrofit/adapter/CallK.kt
```diff
@@ -0,0 +1,30 @@
...
+
+data class CallK<R>(val call: Call<R>) {
+  fun <F> async(AC: Async<F>): Kind<F, Response<R>> = call.runInAsyncContext(AC)
+
+  fun <F> defer(defer: MonadDefer<F>): Kind<F, Response<R>> = call.runSyncDeferred(defer)
+
+  fun <F> catch(monadError: MonadError<F, Throwable>): Kind<F, Response<R>> = call.runSyncCatch(monadError)
+}
+
+fun <F, A> Call<A>.runInAsyncContext(AC: Async<F>): Kind<F, Response<A>> =
```
pakoito: Move these to `Ext.kt` probably.

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow/integrations/retrofit/adapter/Ext.kt
```diff
@@ -0,0 +1,18 @@
+package arrow.integrations.retrofit.adapter
```
pakoito: Worth renaming to `syntax.kt` to be consistent

pakoito:
Approved from me, with one suggestion to move some methods. @raulraja needs to approve it too.

We can do the docs later :)

raulraja:
@leandroBorgesFerreira This is ready to go once you rebase from master

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow/integrations/retrofit/adapter/syntax.kt
```diff
@@ -0,0 +1,36 @@
...
+    if (body != null) {
+      apError.just(body)
+    } else {
+      apError.raiseError(IllegalStateException("The request returned a null body"))
+    }
+  } else {
+    apError.raiseError(HttpException(this))
+  }
+
+fun <F, A> Call<A>.runInAsyncContext(AC: Async<F>): Kind<F, Response<A>> =
```
pakoito: `runAsync`? I believe this name is a holdover from when `Async` was called `AsyncContext`. Have you seen it somewhere else?

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow/integrations/retrofit/adapter/Async2CallAdapterFactory.kt
```diff
@@ -0,0 +1,36 @@
+package arrow.integrations.retrofit.adapter
+
+import retrofit2.CallAdapter
+import retrofit2.Retrofit
+import java.lang.reflect.ParameterizedType
+import java.lang.reflect.Type
+
+class Async2CallAdapterFactory : CallAdapter.Factory() {
```
pakoito: Now that we don't do `IO`, what about just `CallKindAdapterFactory`?

modules/integrations/arrow-integrations-retrofit-adapter/src/main/kotlin/arrow/integrations/retrofit/adapter/syntax.kt
leandroBorgesFerreira: I just didn't know any good name haha. `runAsync` looks good

PR #1038, opened 2018-10-10
MonadDefer implementation for Kleisli

### LINKED ISSUE #910: Missing ReaderT extensions including the ones in Effect hierarchy
ReaderT's can provide instances for any type class for which their inner monad provides instances for. Currently we lack any extensions for `ReaderT` in Arrow. We are missing instances such as `ReaderT.monadDefer` that are possible provided the Monad that ReaderT is wrapping also provides a `MonadDefer` instance. Same for `Concurrent` or any other type classes.

enhan:
I'm adding the implementation for closes #910. Some open points: 

- Is the new module OK for this instance
- I am not sure the tests are correctly written


modules/effects/arrow-effects-data/build.gradle
```diff
@@ -0,0 +1,16 @@
+dependencies {
+    compile project(':arrow-typeclasses')
```
pakoito: Data implies typeclasses, you shouldn't need this here.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/KleisliInstances.kt
```diff
@@ -0,0 +1,30 @@
...
+import arrow.effects.typeclasses.MonadDefer
+import arrow.instance
+
+
+@instance(Kleisli::class)
+interface KleisliMonadDeferInstance<F, D>: MonadDefer<KleisliPartialOf<F, D>> {
+
+    fun FF(): MonadDefer<F>
+
+    override fun <A> defer(fa: () -> Kind<KleisliPartialOf<F, D>, A>): Kind<KleisliPartialOf<F, D>, A> = Kleisli { d: D ->
```
pakoito: Can you please move this constructor to the Kleisli file instead? That way everyone can use it.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/KleisliInstances.kt
```diff
@@ -0,0 +1,30 @@
...
+        FF().defer { fa().fix().run(d) }
+    }
+
+    override fun <A> just(a: A): Kind<KleisliPartialOf<F, D>, A> = Kleisli { FF().just(a) }
```
pakoito: Same as above, I'm sure this'll be useful.

modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/KleisliTest.kt
```diff
@@ -0,0 +1,42 @@
...
+class KleisliTest: UnitSpec() {
+
+    val IO_DEFER = IO.monadDefer()
+
+    val EQ_OPTION = Option.eq(Eq.any())
+
+    fun <A> EQ(): Eq<Kind<KleisliPartialOf<ForIO, Unit>, A>> {
+        return Eq {a, b ->
+            EQ_OPTION.run {
+                a.fix().run(Unit).fix().attempt().unsafeRunTimed(60.seconds).eqv(b.fix().run(Unit).fix().attempt().unsafeRunTimed(60.seconds))
```
pakoito: In this case `==` for comparison should suffice.

pakoito:
>Is the new module OK for this instance

Yeah, let's not add a dependency on data just for this instance.

>I am not sure the tests are correctly written

They look good to me.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/KleisliInstances.kt
```diff
@@ -0,0 +1,30 @@
...
+import arrow.data.KleisliPartialOf
+import arrow.data.fix
+import arrow.effects.typeclasses.MonadDefer
+import arrow.instance
+
+
+@instance(Kleisli::class)
+interface KleisliMonadDeferInstance<F, D>: MonadDefer<KleisliPartialOf<F, D>> {
+
+    fun FF(): MonadDefer<F>
```
pakoito: `DF` as a name, `FF` implies it should be a functor.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/KleisliInstances.kt
enhan: I think it is not possible to move it to Kleisli because of the dependency on MonadDefer

enhan:
@pakoito is it ok to merge it, now that CI completed with success ?

raulraja:
@enhan we have to merge in the changes for `@extension` in 0.8.0 first this coming week and then this PR which will also be in 0.8.0

raulraja:
@enhan  This is ready to go once you rebase from master

pakoito:
```
e: /home/travis/build/arrow-kt/arrow/modules/effects/arrow-effects-data/build/tmp/kapt3/stubs/main/arrow/effects/KleisliMonadDeferInstance.java:6: error: incompatible types: NonExistentClass cannot be converted to Annotation
@error.NonExistentClass()
      ^
> Task :arrow-effects-data:kaptKotlin FAILED
```

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/KleisliInstances.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import arrow.Kind
+import arrow.core.Either
+import arrow.data.Kleisli
+import arrow.data.KleisliPartialOf
+import arrow.data.fix
+import arrow.effects.typeclasses.MonadDefer
+import arrow.instance
+
+@instance(Kleisli::class)
```
pakoito: This is deprecated, use `@extends` instead.

settings.gradle
```diff
@@ -62,6 +62,7 @@ include {
             _ 'effects-kotlinx-coroutines'
             _ 'effects-kotlinx-coroutines-instances'
             _ 'effects-reactor'
+            _ 'effects-data'
```
pakoito: As per Raul's changes this module should be called `effects-data-instances`

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/KleisliInstances.kt
```diff
@@ -0,0 +1,29 @@
+package arrow.effects
```
pakoito: @raulraja has told me a specific package is required for instances. Check with him before merging, please.

pakoito:
@enhan after the latest changes to the intances modules, this PR need to move to this file: https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-data-instances-effects/src/main/kotlin/arrow/instances/kleisli.kt

It should be simpler than all the new changes to create the new module :D

JorgeCastilloPrz:
In an effort to clean the PR list we've decided to close all the ones that are older than 2 months and consider them obsolete except for very specific circumstances that we will evaluate per PR. In this case, we already got effect extensions for `Kleisli` ready [in the mtl library](https://github.com/arrow-kt/arrow/blob/master/modules/fx/arrow-fx-mtl/src/main/kotlin/arrow/fx/mtl/kleisli.kt#L52) including `MonadDefer` and `Concurrent`.

PR #1040, opened 2018-10-10
@extension Toward KEEP-87

raulraja:
`@instance(Target::class) -> @extension` + tons of goodies.

1.  This PR brings first-class extension functions for all data types exported directly from the type class instances they provide instances for. This includes wrapped platform types such as `kotlin.collections.List` which in Arrow is wrapped by `ListK` in order to provide higher kinds emulation. Extensions are projected as functions in the same package that the instance is part of. For users, this would allow them to use the platform types directly:

```kotlin
import arrow.instances.list.traverse.*

listOf(Some(1), Some(2)).sequence(Option.applicative()) 
// Some(listOf(1, 2))
```

which if [KEEP-87](https://github.com/Kotlin/KEEP/pull/87) is approved it would look just like:

```kotlin
listOf(Some(1), Some(2)).sequence()
// Some(listOf(1, 2))
```

This eliminates the need to use the `For[Type] extensions { }` DSL which has been deprecated and removed from references in the documentation.

2. Granular modules for instances where all instances are now in a different module including the `arrow-effects` integrations.

3. An Early DRAFT of Arrow-Meta utilized in the [`ExtensionProcessor`](https://github.com/arrow-kt/arrow/pull/1040/files#diff-8ea7f23c5424c5d88b3402dc9f61f7b6) including a full AST of Kotlin sources reified in the JVM with TypeElement, Kotlin metadata lib by @Takhion and Kotlin Poet. It's an early still undocumented draft because it's being evolved as the rest of the arrow processors are migrated to use arrow-meta. At that time it will be split into its own module with a shadowed version of Arrow that is always the last previous release. This will allow us to expose the Arrow API for anyone out there.

4. Relocation of Ank back as a module that will be published with each release of Arrow until we get to 1.0 to ensure it requires no side maintenance on each Arrow release.


modules/dagger/arrow-dagger/src/main/kotlin/arrow/dagger/instances/optiont.kt
```diff
@@ -51,12 +53,13 @@ class DaggerOptionTFoldableInstance<F> @Inject constructor(val FFF: Foldable<F>)
 
 class DaggerOptionTTraverseInstance<F> @Inject constructor(val FFF: Traverse<F>) : OptionTTraverseInstance<F> {
   override fun FFF(): Traverse<F> = FFF
+  override fun FFT(): Traverse<F> = FFF
```
pakoito: ???

modules/dagger/arrow-dagger/src/main/kotlin/arrow/dagger/instances/optiont.kt
```diff
@@ -39,10 +39,12 @@ class DaggerOptionTFunctorInstance<F> @Inject constructor(val FF: Functor<F>) :
 
 class DaggerOptionTApplicativeInstance<F> @Inject constructor(val FF: Monad<F>) : OptionTApplicativeInstance<F> {
   override fun FF(): Monad<F> = FF
+  override fun MF(): Monad<F> = FF
```
pakoito: Duplicated?

settings.gradle
```diff
@@ -55,13 +56,18 @@ include {
...
+            _ 'effects-rx2-instances'
             _ 'effects-kotlinx-coroutines'
+            _ 'effects-kotlinx-coroutines-instances'
             _ 'effects-reactor'
+            _ 'effects-reactor-instances'
         }
 
         folder('recursion-schemes') {
             _ 'recursion'
+            _ 'instances-recursion'
```
pakoito: recursion-instances for consistency

modules/core/arrow-annotations-processor/models/src/main/java/arrow/ap/objects/deriving/Deriving.kt
```diff
@@ -0,0 +1,12 @@
...
+import arrow.deriving
+import arrow.typeclasses.Functor
+
+class ForBox private constructor() { companion object }
+typealias BoxOf<A> = arrow.Kind<ForBox, A>
+
+@deriving(Functor::class)
+data class Box<A>(val value: A): BoxOf<A> {
+  fun <B> map(f: (A) -> B): Box<B> = Box(f(value))
+}
```
pakoito: Isn't this just `Id`?

modules/core/arrow-data/src/test/kotlin/arrow/data/WriterTTest.kt
```diff
@@ -29,46 +30,44 @@ import org.junit.runner.RunWith
...
-      testLaws(
-        MonadLaws.laws(this, Eq.any()),
-        MonoidKLaws.laws(
-          WriterT.monoidK<ForListK, Int>(ListK.monoidK()),
-          WriterT.applicative(ListK.monad(), Int.monoid()),
-          Eq { a, b ->
-            a.fix().value == b.fix().value
-          }),
+    testLaws(
+      MonadLaws.laws(WriterT.monad(Option.monad(), Int.monoid()), Eq.any()),
```
pakoito: Quick fix: this shouldn't be needed as it's part of the MonadWriterLaws.

modules/dagger/arrow-dagger/src/main/kotlin/arrow/dagger/instances/optiont.kt
raulraja: Different interdependent instances in the hierarchy declared a dependency on the nested monad multiple times with different names. It wasn't in the scope of this massive refactoring to adjust all names in type class hierarchies because they are all wrong and it would have been another massive change

PR #1039, opened 2018-10-10
Interactive code snippets

calvellido:
This PR make docs Kotlin code snippets to be rendered through a Kotlin playground / Codemirror instance, but maintaining current look and feel. By this, any code snippet is now available to be runnable, though code snippets are read-only by default.

You can make a snippet editable, or add a coding box into a doc page, by appending a `{: data-executable='true'}` line on top of a code fence. Like this:

````md
{: data-executable='true'}
```kotlin
import arrow.*
import arrow.core.*

fun main(args: Array<String>) {
//sampleStart
val left: Either<String, Int> = Either.Left("Something went wrong")
//sampleEnd
println(left)
left
}
```
````

* Add proper <!DOCTYPE> declaration to all layouts.
* Load a minified JS version of playground and make it render Kotlin code snippets.
* Add CodeMirror specific styles to make all code snippets look the same.
* Various small docs fixes found along the way, like some bad names or URLs.

This closes #1013

PR #1042, opened 2018-10-11
Revert "Interactive code snippets"

calvellido:
Reverts arrow-kt/arrow#1039

We need to add a https URL to send the snippets code to, otherwise as we are serving our page through https the communication will fail.


modules/docs/arrow-docs/docs/docs/datatypes/store/README.md
```diff
@@ -1,5 +1,5 @@
 ---
-layout: docs
```
pakoito: please keep this change. And also, if you could add `/` at the end of the permalink to fix this, day, moore, and sum, it'd be awesome.

`permalink: /docs/datatypes/store/`

calvellido: Done :+1:

PR #1043, opened 2018-10-11
Bracket

### LINKED ISSUE #1023: Bracket

JorgeCastilloPrz:
fixes #1023

* [x] `Bracket` 
* [x] `BracketLaws`
* [x] Instances for `Kleisli`, `DeferredK`, `FluxK`, `MonoK`, `FlowableK`, `ObservableK`, `MaybeK`, `SingleK`
* [x] `IOBracket` instance (this one is quite more complex)
* [x] Law tests passing for all the data types providing a Bracket instance.
* [x] IOConnection
* [x] IOConnectionTests


modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/IOPlatform.kt
```diff
@@ -0,0 +1,16 @@
...
+internal object IOPlatform {
+
+  /**
+   * Composes multiple errors together, meant for those cases in which error suppression, due to a second error being
+   * triggered, is not acceptable.
+   *
+   * On top of the JVM this function uses Throwable#addSuppressed, available since Java 7. On top of JavaScript the
+   * function would return a CompositeException.
+   */
+  fun composeErrors(first: Throwable, rest: List<Throwable>): Throwable {
```
JorgeCastilloPrz: Used from `CancelUtils` which is used from `IOConnection`. Not sure if we want to get into this rabbit hole though.

JorgeCastilloPrz: Used from `CancelUtils` which is used from `IOConnection`. Not sure if we want to get into this rabbit hole though.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
```diff
@@ -0,0 +1,92 @@
...
+import arrow.test.generators.genConstructor
+import arrow.test.generators.genFunctionAToB
+import arrow.test.generators.genThrowable
+import arrow.typeclasses.Eq
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+
+object BracketLaws {
+  inline fun <F> laws(BF: Bracket<F, Throwable>,
+                      noinline cf: (Int) -> Kind<F, Int>,
```
pakoito: You don't need this parameter, as it's `just` from `Applicative`.

modules/effects/arrow-effects-reactor/src/main/kotlin/arrow/effects/FluxK.kt
```diff
@@ -27,6 +32,14 @@ data class FluxK<A>(val flux: Flux<A>) : FluxKOf<A>, FluxKKindedJ<A> {
   fun <B> flatMap(f: (A) -> FluxKOf<B>): FluxK<B> =
     flux.flatMap { f(it).fix().flux }.k()
 
+  fun <B> bracketCase(use: (A) -> FluxKOf<B>, release: (A, ExitCase<Throwable>) -> FluxKOf<Unit>): FluxK<B> =
+    flatMap { a ->
+      use(a).fix().flux
+        .doOnNext { release(a, ExitCase.Completed) }
```
pakoito: It shouldn't be `doOnNext` because `a` can be used multiple times. It should be on `doOnComplete`.

In any case, check if `doOnTerminate` exists for both cases, and whether they're called when the flux is `cancelled`.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Bracket.kt
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * Meant for specifying tasks with safe resource acquisition and release in the face of errors and interruption.
+   * It would be the the equivalent of `try/catch/finally` statements in mainstream imperative languages for resource
+   * acquisition and release.
+   *
+   * @param release is the action that's supposed to release the allocated resource after `use` is done, irregardless
+   * of its exit condition.
+   */
+  fun <A, B> Kind<F, A>.bracket(use: (A) -> Kind<F, B>, release: (A) -> Kind<F, Unit>): Kind<F, B> =
```
pakoito: For readability purposes, I'd move the use to the last parameter. That way we have the execution always right-aligned.

`bla.bracket({ a -> callNetwork()}) { a.cancel() }`

vs 

`bla.bracket({ a.cancel() }) { a -> callNetwork()}`

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/CancelUtils.kt
```diff
@@ -0,0 +1,59 @@
+package arrow.effects.internal
+
+import arrow.effects.ForIO
+import arrow.effects.IO
+import arrow.effects.IOFrame
+import arrow.effects.fix
+import com.sun.tools.javac.util.ListBuffer
```
pakoito: We shouldn't depend on sun tools.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/CancelUtils.kt
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * utilities for dealing with cancelable thunks.
+ */
+internal object CancelUtils {
```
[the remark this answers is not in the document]
pakoito: Yes, check this branch for them: https://github.com/arrow-kt/arrow/pull/984

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Bracket.kt
JorgeCastilloPrz: Turns out I kept the same ordering found in cats mostly because of the bracket semantics: You take the resource, use it and then release it. But your point is also good though.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
JorgeCastilloPrz: Alright, removed now!

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/CancelUtils.kt
JorgeCastilloPrz: Removed on later pushes, sorry, I had some code pending to push

JorgeCastilloPrz:
Hey @pakoito I fixed / answered some of your comments, also pushed new changes needed to fix all previously broken stuff from `IOBracket`. Tests are passing now. I think I've lost some of your comments after rebasing or something. Could you review again this new version? It should be more solid than the one you reviewed.

Still I have some concerns about the cancellation implementation. Here I'm relying on the new `IOConnection` which is ported over from cats and is used from a couple of places to achieve cancelation, recovering and a couple more things. Looks like it works good, but it probably doesn't fit well with what you shared on that cancellation related PR. I'm not sure if there's some overlapping or what. Any suggestions on how to move on and homogeneize in case it's needed?

I'd also love to get your thoughts on this @raulraja.

One love ❤️

modules/core/arrow-instances-data/build.gradle
```diff
@@ -3,6 +3,7 @@ dependencies {
     compile project(':arrow-annotations')
     compile project(':arrow-typeclasses')
     compile project(':arrow-data')
+    compile project(':arrow-effects')
```
pakoito: This should be the other way around. Effects should have the MonadDefer instance of Kleisli.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
```diff
@@ -14,18 +14,18 @@ import kotlinx.coroutines.experimental.CommonPool
...
       Law("Sync bind: unsafe binding") { SC.asyncBindUnsafe(EQ) },
-      Law("Sync bind: unsafe binding failure") { SC.asyncBindUnsafeError(EQERR) },
+      Law("Sync bind: unsafe binding failure") { SC.asyncBindUnsafeError(EQ) },
       Law("Sync bind: binding in parallel") { SC.asyncParallelBind(EQ) },
       Law("Sync bind: binding cancellation before flatMap") { SC.asyncCancellationBefore(EQ) },
       Law("Sync bind: binding cancellation after flatMap") { SC.asyncCancellationAfter(EQ) },
       Law("Sync bind: bindingInContext cancellation before flatMap") { SC.inContextCancellationBefore(EQ) },
       Law("Sync bind: bindingInContext cancellation after flatMap") { SC.inContextCancellationAfter(EQ) },
-      Law("Sync bind: bindingInContext throw equivalent to raiseError") { SC.inContextErrorThrow(EQERR) },
+      Law("Sync bind: bindingInContext throw equivalent to raiseError") { SC.inContextErrorThrow(EQ) },
```
pakoito: Are you sure this needs to be changed?

modules/effects/arrow-effects-reactor/src/main/kotlin/arrow/effects/FluxK.kt
```diff
@@ -80,7 +94,7 @@ data class FluxK<A>(val flux: Flux<A>) : FluxKOf<A>, FluxKKindedJ<A> {
 
     fun <A> runAsync(fa: Proc<A>): FluxK<A> =
       Flux.create { emitter: FluxSink<A> ->
-        fa { either: Either<Throwable, A> ->
+        fa(IOConnection()) { either: Either<Throwable, A> ->
```
pakoito: Can't we register the `IOConnection` to work in sync with `emitter`? AFAIK they represent the same semantics.

modules/effects/arrow-effects-reactor/src/main/kotlin/arrow/effects/MonoK.kt
```diff
@@ -58,7 +68,7 @@ data class MonoK<A>(val mono: Mono<A>) : MonoKOf<A>, MonoKKindedJ<A> {
 
     fun <A> async(fa: Proc<A>): MonoK<A> =
       Mono.create { emitter: MonoSink<A> ->
-        fa { either: Either<Throwable, A> ->
+        fa(IOConnection()) { either: Either<Throwable, A> ->
```
pakoito: Same comment about `emitter` and `IOConnection`

modules/effects/arrow-effects-rx2/src/main/kotlin/arrow/effects/FlowableK.kt
```diff
@@ -28,6 +34,14 @@ data class FlowableK<A>(val flowable: Flowable<A>) : FlowableKOf<A>, FlowableKKi
   fun <B> flatMap(f: (A) -> FlowableKOf<B>): FlowableK<B> =
     flowable.flatMap { f(it).fix().flowable }.k()
 
+  fun <B> bracketCase(use: (A) -> FlowableKOf<B>, release: (A, ExitCase<Throwable>) -> FlowableKOf<Unit>): FlowableK<B> =
+    flatMap { a ->
+      use(a).fix().flowable
+        .doOnNext { release(a, ExitCase.Completed) }
```
pakoito: doOnComplete

modules/effects/arrow-effects-reactor/src/main/kotlin/arrow/effects/MonoK.kt
```diff
@@ -27,6 +29,14 @@ data class MonoK<A>(val mono: Mono<A>) : MonoKOf<A>, MonoKKindedJ<A> {
   fun <B> flatMap(f: (A) -> MonoKOf<B>): MonoK<B> =
     mono.flatMap { f(it).fix().mono }.k()
 
+  fun <B> bracketCase(use: (A) -> MonoKOf<B>, release: (A, ExitCase<Throwable>) -> MonoKOf<Unit>): MonoK<B> =
+    flatMap { a ->
+      use(a).fix().mono
+        .doOnNext { release(a, ExitCase.Completed) }
```
pakoito: Check if onNext == onComplete for Mono

modules/effects/arrow-effects-rx2/src/main/kotlin/arrow/effects/ObservableK.kt
```diff
@@ -27,6 +33,14 @@ data class ObservableK<A>(val observable: Observable<A>) : ObservableKOf<A>, Obs
   fun <B> flatMap(f: (A) -> ObservableKOf<B>): ObservableK<B> =
     observable.flatMap { f(it).fix().observable }.k()
 
+  fun <B> bracketCase(use: (A) -> ObservableKOf<B>, release: (A, ExitCase<Throwable>) -> ObservableKOf<Unit>): ObservableK<B> =
+    flatMap { a ->
+      use(a).fix().observable
+        .doOnNext { release(a, ExitCase.Completed) }
```
pakoito: doOnComplete

modules/effects/arrow-effects-rx2/src/main/kotlin/arrow/effects/SingleK.kt
```diff
@@ -58,7 +68,7 @@ data class SingleK<A>(val single: Single<A>) : SingleKOf<A>, SingleKKindedJ<A> {
 
     fun <A> async(fa: Proc<A>): SingleK<A> =
       Single.create { emitter: SingleEmitter<A> ->
-        fa { either: Either<Throwable, A> ->
+        fa(IOConnection()) { either: Either<Throwable, A> ->
```
pakoito: Same about emitter

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -90,7 +100,7 @@ sealed class IO<out A> : IOOf<A> {
     IORunLoop.start(this, cb, null)
 
   fun runAsyncCancellable(onCancel: OnCancel = Silent, cb: (Either<Throwable, A>) -> IOOf<Unit>): IO<Disposable> =
-    IO.async { ccb ->
+    IO.async { conn, ccb ->
```
pakoito: It seems like this is the point where my cancellation should be replaced with yours.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IOInstances.kt
```diff
@@ -67,7 +81,22 @@ interface IOMonadErrorInstance : IOMonadInstance, MonadError<ForIO, Throwable> {
 }
 
 @instance(IO::class)
-interface IOMonadDeferInstance : IOMonadErrorInstance, MonadDefer<ForIO> {
+interface IOBracketInstance : IOMonadErrorInstance, Bracket<ForIO, Throwable> {
+  override fun <A, B> Kind<ForIO, A>.bracketCase(use: (A) -> Kind<ForIO, B>, release: (A, ExitCase<Throwable>) -> Kind<ForIO, Unit>): IO<B> =
+    this@bracketCase.fix().bracketCase({ a -> use(a).fix() }, { a, e -> release(a, e).fix() })
```
pakoito: This shouldn't be needed here

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IOParallel.kt
```diff
@@ -18,7 +18,7 @@ fun <A, B, C> IO.Companion.parallelMapN(ctx: CoroutineContext, ioA: IO<A>, ioB:
     cbf.unsafeGet().forEach { it(result) }
   }
 
-  return IO.async { cb: (Either<Throwable, C>) -> Unit ->
+  return IO.async { _, cb: (Either<Throwable, C>) -> Unit ->
```
pakoito: Cancellation for Parallels needs to be setup here.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/IOConnection.kt
```diff
@@ -0,0 +1,112 @@
...
+    override fun push(token: CancelToken<ForIO>) = Unit
+    override fun pop(): CancelToken<ForIO> = IO.unit
+    override fun tryReactivate(): Boolean = true
+    override fun pushPair(lh: IOConnection, rh: IOConnection): Unit = Unit
+  }
+
+  /**
+   * Default [[IOConnection]] implementation.
+   */
+  private class Impl : IOConnection() {
```
pakoito: I'd call it a different name, like `DefaultIOConnection()`.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/IOConnection.kt
```diff
@@ -0,0 +1,112 @@
...
+ */
+sealed class IOConnection {
+
+  /**
+   * Cancels the unit of work represented by this reference.
+   *
+   * Guaranteed idempotency - calling it multiple times should have the same side-effect as calling it only
+   * once. Implementations of this method should also be thread-safe.
+   */
+  abstract fun cancel(): CancelToken<ForIO>
```
pakoito: All the `CancelToken` in the PR are fixed to `ForIO`. Doesn't this imply that there should be other fixes for other datatypes?

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/IOPlatform.kt
```diff
@@ -0,0 +1,21 @@
+package arrow.effects.internal
+
+internal object IOPlatform {
```
pakoito: We don't really need this object because IO will never work outside the JVM. Merge it in another file if you want.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/ParallelUtils.kt
```diff
@@ -14,7 +14,7 @@ import kotlin.coroutines.experimental.suspendCoroutine
 /* See parMap3 */
 internal fun <F, A, B, C> Effect<F>.parMap2(ctx: CoroutineContext, ioA: Kind<F, A>, ioB: Kind<F, B>, f: (A, B) -> C,
   /* start is used because this has to start inside the coroutine. Using Future won't work */
-                                            start: (Kind<F, Unit>) -> Unit): Proc<C> = { cc ->
+                                            start: (Kind<F, Unit>) -> Unit): Proc<C> = { _, cc ->
```
pakoito: Cancellation needs to be setup here.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Bracket.kt
pakoito: I prefer the release/use style, for the reasons above. We can ask in the channel.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/IOConnection.kt
```diff
@@ -0,0 +1,112 @@
...
+
+  abstract fun isCanceled(): Boolean
+
+  /**
+   * Pushes a cancelable reference on the stack, to be popped or canceled later in FIFO order.
+   */
+  abstract fun push(token: CancelToken<ForIO>): Unit
```
pakoito: I understand the semantics behind this function name. But, for the user's sake, I'd replace it with something more like `push` -> `register` and `pop` -> `unregisterLatest`

pakoito: More inline with `setDisposable` and `setCancellable` here http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/ObservableEmitter.html#setDisposable-io.reactivex.disposables.Disposable-

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Bracket.kt
```diff
@@ -0,0 +1,83 @@
+package arrow.effects.typeclasses
+
+import arrow.Kind
+import arrow.core.Either
+import arrow.effects.typeclasses.ExitCase.Error
+import arrow.typeclasses.MonadError
+
+sealed class ExitCase<out E> {
```
nomisRev: Should this be moved to `arrow.effects` like cats?

raulraja:
This is ready to for a rebase from master. The Extensions PR is in already

modules/effects/arrow-effects-reactor/src/main/kotlin/arrow/effects/FluxK.kt
JorgeCastilloPrz: I was mistaken about how the callbacks work, sorry. I'm refactored it to use the 3 different callbacks depending on the case now:

```
.doOnError { release(a, ExitCase.Error(it)) }
.doOnCancel { release(a, ExitCase.Cancelled) }
.doOnComplete { release(a, ExitCase.Completed) }
```

I've added a test to ensure the expected behaviour in cancellation.

JorgeCastilloPrz: I was mistaken about how the callbacks work, sorry. I'm refactored it to use the 3 different callbacks depending on the case now:

```
.doOnError { release(a, ExitCase.Error(it)) }
.doOnCancel { release(a, ExitCase.Cancelled) }
.doOnComplete { release(a, ExitCase.Completed) }
```

I've added a test to ensure the expected behaviour in cancellation.

JorgeCastilloPrz: I was mistaken about how the callbacks work, sorry. I'm refactored it to use the 3 different callbacks depending on the case now:

```
.doOnError { release(a, ExitCase.Error(it)) }
.doOnCancel { release(a, ExitCase.Cancelled) }
.doOnComplete { release(a, ExitCase.Completed) }
```

I've added a test to ensure the expected behaviour in cancellation.

JorgeCastilloPrz: I was mistaken about how the callbacks work, sorry. I'm refactored it to use the 3 different callbacks depending on the case now:

```
.doOnError { release(a, ExitCase.Error(it)) }
.doOnCancel { release(a, ExitCase.Cancelled) }
.doOnComplete { release(a, ExitCase.Completed) }
```

I've added a test to ensure the expected behaviour in cancellation.

JorgeCastilloPrz:
Ok GitHub is trolling me and can't comment this into an outdated comment. @pakoito:

I was mistaken about how the callbacks work for `FluxK`, sorry. I've refactored it now to use the 3 different callbacks depending on the case:

```
.doOnError { release(a, ExitCase.Error(it)) }
.doOnCancel { release(a, ExitCase.Cancelled) }
.doOnComplete { release(a, ExitCase.Completed) }
```

Also added a test to ensure the expected behaviour in cancellation.

modules/effects/arrow-effects-reactor/src/main/kotlin/arrow/effects/FluxK.kt
```diff
@@ -35,9 +35,10 @@ data class FluxK<A>(val flux: Flux<A>) : FluxKOf<A>, FluxKKindedJ<A> {
 
   fun <B> bracketCase(use: (A) -> FluxKOf<B>, release: (A, ExitCase<Throwable>) -> FluxKOf<Unit>): FluxK<B> =
     flatMap { a ->
-      use(a).fix().flux
-        .doOnComplete { release(a, ExitCase.Completed) }
+      use(a).value()
         .doOnError { release(a, ExitCase.Error(it)) }
+        .doOnCancel { release(a, ExitCase.Cancelled) }
+        .doOnComplete { release(a, ExitCase.Completed) }
```
pakoito: 👍

modules/effects/arrow-effects-reactor/src/test/kotlin/arrow/effects/FluxKTests.kt
```diff
@@ -1,58 +1,52 @@
...
+          err
+        }
+
+        errA == errB
+      }
   }
 
   init {
-
+/*
```
pakoito: Removed for testing?

JorgeCastilloPrz: Oops sorry, commented this for reducing noise amount when testing something else. Uncommented now 👍

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Bracket.kt
JorgeCastilloPrz: [It comes along with `Bracket` into the same file in `cats-effect`](https://github.com/typelevel/cats-effect/blob/dbbbf3c30bd8b52d31bd95b3084aad1054cb483a/core/shared/src/main/scala/cats/effect/Bracket.scala#L146), so did the same here.

modules/core/arrow-instances-data/build.gradle
JorgeCastilloPrz: I'm probably missing something but, are you sure about this? The only instances I could find on `arrow-effects-instances` are the ones from `IO`, but just because it's an effects data type. Kleisli is not. That seems to be the approach until today.

The module does not contain the `MonadDefer` / `Bracket` / `Async` / `Effect` (or any other effects Type classes) instances for all the data types in Arrow.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
JorgeCastilloPrz: EQERR is an EQ instance for `Kind<F, Int>`, exactly the same than other param: `EQ`. If you look at callers, feels like we're passing the same instance twice everywhere:

captura de pantalla 2018-11-10 a las 8 53 27

So I removed it since I didn't understand why we need two if they're just structural comparisons over the same `Kind<F, Int>`, so can't find a reason for them to be different. Please let me know if I'm wrong here. Actually I think I forgot to remove `EQERR` also in `AsyncLaws`, since it has the same issue.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/IOConnection.kt
JorgeCastilloPrz: Yep. We need to support Cancellation over any `F` but that's part of [ConcurrentEffect](https://github.com/typelevel/cats-effect/blob/76d354153926b30217cc1aed9d1b181c7ea20d32/site/src/main/tut/typeclasses/concurrent-effect.md) and [Concurrent](https://github.com/typelevel/cats-effect/blob/77931fd2fc6d7d2891e321cd59689a47b2e13a56/site/src/main/tut/typeclasses/concurrent.md). That will be done when we work on those types completely (we just have `ConcurrentEffect` on very early steps for now).

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/IOPlatform.kt
nomisRev: > never work outside the JVM

We never plan to support IO on js?

modules/effects/arrow-effects-reactor/src/main/kotlin/arrow/effects/MonoK.kt
JorgeCastilloPrz: Actually it's not, the correct one was `onSuccess()`, which checks the `Mono` has completed successfully. Fixed it and added a test to cover cancellation.

modules/core/arrow-instances-data/build.gradle
nomisRev: @JorgeCastilloPr This breaks the module system. If I include instances for `data` then I don't want to implicitly import `effects`.

What did you need `effects` for anyway?

modules/effects/arrow-effects-rx2/src/main/kotlin/arrow/effects/ObservableK.kt
JorgeCastilloPrz: Fixed. Following [this](https://github.com/ReactiveX/RxJava/issues/4456) and [this](https://github.com/ReactiveX/RxJava/pull/4458) looks like `Observable.onDispose()` would be the equivalent to `Flowable.onCancel()`, so I'm notifying the `Cancelled` ExitCase when the observer is disposed.

Added a test to cover the cancellation.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/IOPlatform.kt
JorgeCastilloPrz: These functions are used from a couple of places so not sure where to place them. I've renamed the file to `ErrorUtils` for now, following the same style than `CancelUtils`.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/IOConnection.kt
JorgeCastilloPrz: Actually this is not a stack, but a queue. It's **FIFO**, so `unregisterLatest` wouldn't be accurate either. Ideas?

modules/core/arrow-instances-data/build.gradle
pakoito: Nevermind that, do a new module `arrow-data-instances-effects` instead.

JorgeCastilloPrz: Alright. Did that as a compromise solution for now, mostly to unlock this PR. We definitely need to have that maintainers call asap to decide what to do with these instances, it doesn't look great though. Here's [the commit](https://github.com/arrow-kt/arrow/pull/1043/commits/9609ea0ce4945d4916d0ee8a107c5372b776bfdb).

I like the new approach with everything super segregated, but there are still some rough edges.

modules/effects/arrow-data-instances-effects/gradle.properties
```diff
@@ -0,0 +1,4 @@
+# Maven publishing configuration
+POM_NAME=Arrow Type Classes Effects Instances for data types in Data
```
JorgeCastilloPrz: :D the fact that I can't describe this properly with human words could highlight the "smell" on the approach. Anyways don't wanna get into the rabbit hole more than what I already did, so keeping it for now.

modules/effects/arrow-effects-reactor/src/main/kotlin/arrow/effects/FluxK.kt
JorgeCastilloPrz: `IOConnection` is not exposed through Async anymore, mimicking cats.

modules/effects/arrow-effects-reactor/src/main/kotlin/arrow/effects/MonoK.kt
JorgeCastilloPrz: `IOConnection` is not exposed through Async anymore, mimicking cats.

modules/effects/arrow-effects-rx2/src/main/kotlin/arrow/effects/SingleK.kt
JorgeCastilloPrz: `IOConnection` is not exposed through Async anymore, mimicking cats. Further explanations in a comment in the end.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/ParallelUtils.kt
JorgeCastilloPrz: `IOConnection` is not exposed through Async anymore, mimicking cats. Further explanations in a comment in the end.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IOParallel.kt
JorgeCastilloPrz: `IOConnection` is not exposed through Async anymore, mimicking cats. Further explanations in a comment in the end.

JorgeCastilloPrz:
After some cats digging, I made my mind around how `IOConnection` is used.

Turns out I initially added `IOConnection` as a requirement to `Proc` (`Async` callback), and that was incorrect. Following cats implementation, `IOConnection` is not part of the processing callback but a separate requirement by `IO.Async`. So it’s an ad hoc cancellation solution for `IO`. I’ve fixed it to follow the same constraints it has in typelevel/cats now.

That removes the need for passing in the `IOConnection` for each async() implementation on each data type. That was wrong since data types should not be tied to IO in any way.

About achieving a generalized way to control cancellation **across any data types supporting `Async`**, sounds totally legit, but I couldn’t find any relation between that concept and `IOConnection` in cats. I think that’s part of `Concurrent` and `ConcurrentEffect` work since those are indeed using the concept of `CancelToken<F>`, but they use it differently (not through `IOConnection`). Therefore that’d be a bit too out of scope of this PR and more in the line of #1090.

(IOConnection was required on this one because it’s part of the `IOBracket` implementation).

Said that, could I get a third review pass on everything @pakoito @nomisRev ? Thanks guys ❤️ .

modules/core/arrow-data/build.gradle
```diff
@@ -12,6 +12,7 @@ dependencies {
     testCompile project(':arrow-test')
     testCompile project(':arrow-free')
     testCompile project(':arrow-instances-free')
+    testCompile project(':arrow-data-instances-effects')
```
JorgeCastilloPrz: This is required since tests are all on this module but this particular instance is in the newly added module.

modules/core/arrow-instances-data/build.gradle
JorgeCastilloPrz: Yep! removed now, sorry. I noticed some mins ago. The commit links got broken cause of rebases, you'd need to look at complete diff though.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/ForwardCancelable.kt
```diff
@@ -0,0 +1,89 @@
...
+import arrow.effects.IORunLoop
+import arrow.effects.fix
+import arrow.effects.internal.ForwardCancelable.Companion.State.Active
+import arrow.effects.internal.ForwardCancelable.Companion.State.Empty
+import java.util.concurrent.atomic.AtomicReference
+
+/**
+ * A placeholder for a [[CancelToken]] that will be set at a later time, the equivalent of a
+ * `Deferred[IO, CancelToken]`. Used in the implementation of `bracket`, see [[IOBracket]].
+ */
```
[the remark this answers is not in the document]
JorgeCastilloPrz: fixed

modules/effects/arrow-effects-instances/build.gradle
```diff
@@ -1,5 +1,7 @@
 dependencies {
     compile project(':arrow-effects')
+    compile project(':arrow-data')
+    compile project(':arrow-instances-data')
```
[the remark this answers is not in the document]
JorgeCastilloPrz: That's removed already, sorry, missed it!

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/ErrorUtils.kt
```diff
@@ -0,0 +1,21 @@
+package arrow.effects.internal
+
+internal object ErrorUtils {
+
+  /**
+   * Composes multiple errors together, meant for those cases in which error suppression, due to a second error being
+   * triggered, is not acceptable.
+   *
+   * On top of the JVM this function uses Throwable#addSuppressed, available since Java 7. On top of JavaScript the
+   * function would return a CompositeException.
```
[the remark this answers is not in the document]
JorgeCastilloPrz: Looks like it does. This is docs just copied over from cats because seemed to be important to highlight.

modules/dagger/arrow-dagger/src/test/kotlin/arrow/dagger/instances/tests/instances.kt
```diff
@@ -24,7 +63,7 @@ class LocalOptionTTInstances : OptionTInstances<ForOption>()
 class LocalFunction1Instances : Function1Instances<ForOption>()
 
 @Module
-class LocalKleisliInstances : KleisliInstances<ForOption, Unit>()
+class LocalKleisliInstances : KleisliInstances<ForOption, Unit, Unit>()
```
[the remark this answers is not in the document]
JorgeCastilloPrz: This is not needed anymore. KleisliInstances contained the bracket instance before and therefore it required the error type, and that one does not belong to this dagger instances module.

modules/core/arrow-data/src/test/kotlin/arrow/data/KleisliTest.kt
```diff
@@ -17,19 +28,34 @@ import io.kotlintest.KTestJUnitRunner
 import io.kotlintest.matchers.shouldBe
 import org.junit.runner.RunWith
 
+@Suppress("TestFunctionName")
```
pakoito: What is this annotation for?

modules/core/arrow-instances-data/src/main/kotlin/arrow/instances/kleisli.kt
```diff
@@ -33,69 +51,54 @@ interface KleisliContravariant<F, D> : Contravariant<Conested<Kind<ForKleisli, F
 @extension
 interface KleisliApplicativeInstance<F, D> : Applicative<KleisliPartialOf<F, D>>, KleisliFunctorInstance<F, D> {
 
-  fun AF(): Applicative<F>
+  override fun FFF(): Applicative<F>
```
pakoito: Can you do this? We've a few of these through the codebase, and we reasigned them instead, like:

override fun FFF() = AFF()

fun AFF(): Applicative<F>

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
pakoito: Thinking about this, there's a potential issue where one of the tests returns an unexpected error, and our Eq accepts it because it's made to compare errors. Same for error cases, they may succeed and the Eq treats them as correct.

This is probably an issue with the callers, more than it is with the laws.

modules/core/arrow-data/src/test/kotlin/arrow/data/KleisliTest.kt
JorgeCastilloPrz: Otherwise you get the lint warning about the test names

captura de pantalla 2018-11-11 a las 18 02 21

modules/effects/arrow-data-instances-effects/gradle.properties
pakoito: "Instances of arrow-effects for the datatypes in arrow-data"

modules/effects/arrow-effects-instances/src/main/kotlin/arrow/effects/instances/io.kt
```diff
@@ -4,9 +4,22 @@ import arrow.Kind
...
+import arrow.effects.typeclasses.Proc
 import arrow.extension
-import arrow.typeclasses.*
+import arrow.typeclasses.Applicative
+import arrow.typeclasses.ApplicativeError
+import arrow.typeclasses.Functor
+import arrow.typeclasses.Monad
+import arrow.typeclasses.MonadError
+import arrow.typeclasses.Monoid
+import arrow.typeclasses.Semigroup
```
pakoito: I suspect your IDE is configured not to collapse these after seeing 5+ of them.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IORunLoop.kt
```diff
@@ -133,31 +133,27 @@ internal object IORunLoop {
...
   // Hitting an async boundary means we have to stop, however
   // if we had previous `flatMap` operations then we need to resume
   // the loop with the collected stack
     when {
       bFirst != null || (bRest != null && bRest.isNotEmpty()) ->
         IO.Async { conn, cb ->
-          val rcb = RestartCallback(cb as Callback, null)
-          rcb.prepare(bFirst, bRest)
-          register(conn, rcb)
+          loop(currentIO, conn, cb as Callback, null, bFirst, bRest)
```
JorgeCastilloPrz: why was this so different than cats? I'm not 100% sure of this change, but this is what you can find in cats for this function

modules/core/arrow-instances-data/src/main/kotlin/arrow/instances/kleisli.kt
JorgeCastilloPrz: Added #1104 to work on that separately. Ideally we'd have the complete instance hierarchy for all the data types following this approach which reduces the amount of boilerplate quite a bit.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/IOConnection.kt
```diff
@@ -3,8 +3,15 @@ package arrow.effects.internal
 import arrow.effects.ForIO
 import arrow.effects.IO
 import arrow.effects.fix
+import arrow.effects.typeclasses.Disposable
 import java.util.concurrent.atomic.AtomicReference
 
+fun IOConnection.toDisposable(): Disposable = object : Disposable {
```
pakoito: Disposable is an alias for `() -> Unit`, so `= { cancel() }` should be enough

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IORunLoop.kt
```diff
@@ -239,6 +247,20 @@ internal object IORunLoop {
           bFirst = currentIO as BindF
           currentIO = currentIO.source
         }
+        is IO.ContextSwitch -> {
```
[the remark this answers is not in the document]
pakoito: EDIT: I don't really know what this does..

modules/dagger/arrow-dagger/src/test/kotlin/arrow/dagger/instances/tests/instances.kt
```diff
@@ -1,9 +1,48 @@
...
+import arrow.typeclasses.Foldable
+import arrow.typeclasses.Functor
+import arrow.typeclasses.Monad
+import arrow.typeclasses.MonadError
+import arrow.typeclasses.Monoid
+import arrow.typeclasses.MonoidK
+import arrow.typeclasses.Order
+import arrow.typeclasses.Semigroup
+import arrow.typeclasses.SemigroupK
+import arrow.typeclasses.Traverse
```
pakoito: Configure your imports to fold after a 5+ of them. This becomes unmaintainable after a while.

modules/effects/arrow-effects-instances/src/test/kotlin/arrow/effects/IOTest.kt
```diff
@@ -243,7 +243,7 @@ class IOTest : UnitSpec() {
     }
 
     "unsafeRunTimed times out with None result" {
-      val never = IO.async<Int> { }
+      val never = IO.async().never<Unit>().fix()
```
pakoito: *squint* I wrote it, but it's still ugly. Maybe it's being generated now and it's just an import?

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Bracket.kt
pakoito: Reaffirming myself on the function order here :D

modules/effects/arrow-effects-instances/src/main/kotlin/arrow/effects/instances/io.kt
JorgeCastilloPrz: True story 👍

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IORunLoop.kt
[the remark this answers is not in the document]
JorgeCastilloPrz: It's 2 different reifications of IO.

* `IO.ContinueOn` switches coroutine context.
* `IO.ContextSwitch` represents an `IOConnection` switch. It tries to modify the current connection using it's `modify` function with type `(IOConnection) -> IOConnection`. If it's udpated successfully, it updates the connection in the `RestartCallback` so next time it's invoked it'll be updated. `ContextSwitch` is also able to restore a failed connection using it's `restore` function which relies on the `RestoreContext`.

modules/aql/arrow-query-language/src/test/kotlin/arrow/aql/tests/IOExample.kt
```diff
@@ -39,7 +39,7 @@ class RemoteDataSource : DataSource {
     mapOf(User(UserId("user2")) to listOf(Task("Remote Task assigned to user2")))
 
   override fun allTasksByUser(user: User): IO<List<Task>> =
-    IO.async { cb ->
+    IO.async { _, cb ->
```
JorgeCastilloPrz: I think this one should be ignored. As in: no matter what connection the new async gets passed, it'll reuse the one passed in for the current `loop()` call.

modules/effects/arrow-effects-instances/src/test/kotlin/arrow/effects/IOTest.kt
JorgeCastilloPrz: No idea

modules/dagger/arrow-dagger/src/test/kotlin/arrow/dagger/instances/tests/instances.kt
JorgeCastilloPrz: Done! Fixed all over the place now.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IORunLoop.kt
```diff
@@ -233,8 +233,8 @@ internal object IORunLoop {
 
           bFirst = { c: Any? -> IO.just(c) }
 
-          currentIO = IO.async { TODO, cc ->
-            loop(localCont, conn, cc.asyncCallback(currentCC), null, null, null)
+          currentIO = IO.async { _, callback ->
```
JorgeCastilloPrz: I think this one should be ignored. As in: no matter what connection the new async gets passed, it'll reuse the one passed in for the current `loop()` call. Can't see a reason to use a modified one or anything here, since there's no new cancellation reasons. Does it make sense to you @pakoito ?

pakoito: Add a comment there, just in case, to explain your reasoning.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Bracket.kt
JorgeCastilloPrz: Yeye, don't think I forgot, that'll be done before merging.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
JorgeCastilloPrz: Alright, I can see the problem now. I've reverted it to be as it was til today, since it doesn't look like a boyscout we wanna do here, given the highlighted caveats of the approach. Still, it might be something to address in the maintainer's meeting.

raulraja:
@JorgeCastilloPrz Is this good to go?, feel free to merge when conflicts are resolved if you and @nomisRev are happy with it

JorgeCastilloPrz:
It still requires some connection related work over IOParallel

modules/effects/arrow-data-instances-effects/gradle.properties
JorgeCastilloPrz: Done

JorgeCastilloPrz:
Hey @pakoito, I've been trying to understand how to wire the new IOConnection for cancellation on IOParallel utilities, but I've not been able to. I fear I don't have a deep scope on how that works and I'd say there's no IOConnection cancellation wired in cats for this. Since you probably were the one working on that it would be faster if you give some help with that. Could you plssss? 🙏

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/AsyncLaws.kt
```diff
@@ -13,8 +13,14 @@ import io.kotlintest.properties.forAll
 import kotlinx.coroutines.newSingleThreadContext
 
 object AsyncLaws {
-  fun <F> laws(AC: Async<F>, EQ: Eq<Kind<F, Int>>, EQ_EITHER: Eq<Kind<F, Either<Throwable, Int>>>, EQERR: Eq<Kind<F, Int>> = EQ): List<Law> =
-    MonadDeferLaws.laws(AC, EQERR, EQ_EITHER, EQ) + listOf(
+
```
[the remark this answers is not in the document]
JorgeCastilloPrz: Looks like around 50% of the Laws have the blank line at the beginning and the other 50% don't. Added it to all of them now for coherence.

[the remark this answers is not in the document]
JorgeCastilloPrz: Fixed

modules/core/arrow-instances-data/src/main/kotlin/arrow/instances/kleisli.kt
```diff
@@ -33,69 +33,54 @@ interface KleisliContravariant<F, D> : Contravariant<Conested<Kind<ForKleisli, F
...
+    Kleisli { FFF().run { fix().run(it).product(fb.fix().run(it)) } }
 }
 
 @extension
 interface KleisliMonadInstance<F, D> : Monad<KleisliPartialOf<F, D>>, KleisliApplicativeInstance<F, D> {
 
-  fun MF(): Monad<F>
-
-  override fun AF(): Applicative<F> = MF()
+  override fun FFF(): Monad<F>
```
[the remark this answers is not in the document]
JorgeCastilloPrz: Nope. I mean, it was working good but @raulraja informed that it was not okay to do that. Reverting it now.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
```diff
@@ -13,8 +13,14 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.newSingleThreadContext
 
 object MonadDeferLaws {
-  fun <F> laws(SC: MonadDefer<F>, EQ: Eq<Kind<F, Int>>, EQ_EITHER: Eq<Kind<F, Either<Throwable, Int>>>, EQERR: Eq<Kind<F, Int>> = EQ): List<Law> =
-    MonadErrorLaws.laws(SC, EQERR, EQ_EITHER, EQ) + listOf(
+
```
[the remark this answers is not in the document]
JorgeCastilloPrz: fixed

JorgeCastilloPrz:
Added a comment on #1091 to work on `IOParallel` cancellation through `IOConnection` whenever `Parallel` is coded.

PR #1046, opened 2018-10-12
adding Tuple2 extension functions

### LINKED ISSUE #1044: Support for Pair extension functions to Tuple2
It would be nice to have the same extension functions if Pair in the Tuple2, like `toMap`, for example.

What do you think? I would be happy to add then.

leandroBorgesFerreira:
Fixes #1044 

Adding the extension functions of Pair for Tuple2. 

This is basically the same code in Kotlin std, with some small alterations.

### DIFF
--- a/modules/core/arrow-core/src/main/kotlin/arrow/core/TupleN.kt
+++ b/modules/core/arrow-core/src/main/kotlin/arrow/core/TupleN.kt
@@ -1,6 +1,7 @@
 package arrow.core
 
 import arrow.*
+import java.util.*
 
 @higherkind
 data class Tuple2<out A, out B>(val a: A, val b: B) : Tuple2Of<A, B> {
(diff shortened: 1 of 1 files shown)


raulraja:
The extensions look good but in the spirit of keeping Arrow immutable and pure in its public interface at least, I don't believe we should include any extensions or helpers for mutable collections.

leandroBorgesFerreira:
Now it only provides public extensions and helps for immutable structures

modules/core/arrow-core/src/main/kotlin/arrow/core/TupleN.kt
```diff
@@ -135,4 +136,96 @@ data class Tuple22<out A, out B, out C, out D, out E, out F, out G, out H, out I
...
+  0 -> emptyMap()
+  1 -> mapOf(this[0])
+  else -> toMap(LinkedHashMap<K, V>(mapCapacity(size)))
+}
+
+fun <K, V> Sequence<Tuple2<K, V>>.toMap(): Map<K, V> = toMap(LinkedHashMap()).optimizeReadOnlyMap()
+
+fun <K, V> mapOf(pair: Tuple2<K, V>): Map<K, V> = Collections.singletonMap(pair.a, pair.b)
+
+internal fun <K, V, M : MutableMap<in K, in V>> Iterable<Tuple2<K, V>>.toMap(destination: M): M =
```
raulraja: what are these `internal` ones needed for?

leandroBorgesFerreira: They are handling the MutableMaps. The concrete implementation being used in this PR is LinkedHashMap (this is the same as Kotlin standard library), so it need a function to deal with this MutableMap. 

Basically is creates the MutableMap and adds all the object in the Iterable/Array/Sequence in it, but the PR only exposes the Map (unmutable).

PR #1047, opened 2018-10-13
Optics state API

nomisRev:


modules/optics/arrow-optics/src/main/kotlin/arrow/optics/Optional.kt
```diff
@@ -263,4 +265,57 @@ interface POptional<S, T, A, B> : POptionalOf<S, T, A, B> {
...
+/**
+ * Modify the focus [A] viewed through the [Optional] and returns its *new* value.
+ */
+fun <S, A> Optional<S, A>.mod(f: (A) -> A): State<S, Option<A>> =
+  modo(f).map { it.map(f) }
+
+/**
+ * Modify the focus [A] viewed through the [Optional] and returns its *old* value.
+ */
+fun <S, A> Optional<S, A>.modo(f: (A) -> A): State<S, Option<A>> =
```
pakoito: We can have `modify,` `modifyOld`, `assign`, `assignOld`. In this case longer names will be helpful.

modules/optics/arrow-optics/src/main/kotlin/arrow/optics/Optional.kt
```diff
@@ -263,4 +265,57 @@ interface POptional<S, T, A, B> : POptionalOf<S, T, A, B> {
...
+  /**
+   * Transforms a [POptional] into a [State].
+   * Alias for [extract].
+   */
+  fun toState(): State<S, Option<A>> = extract()
+
+  /**
+   * Extracts the focus [A] viewed through the [POptional] and applies [f] to it.
+   */
+  fun <C> extracts(f: (A) -> C): State<S, Option<C>> = extract().map { it.map(f) }
```
pakoito: This should be called `extractMap`.

nomisRev:
@pakoito & @raulraja also the `mod` & `assign` variants are now forced to monomorphic optics since we lack https://github.com/nomisRev/kategory/blob/simon-dontfeartheoptics-wip/kategory-core/src/main/kotlin/kategory/data/IndexedStateT.kt

[polymoprhic version](https://github.com/nomisRev/kategory/blob/simon-dontfeartheoptics-wip/kategory-optics/src/main/kotlin/kategory/optics/data/LensState.kt)

But we lose type safety in `binding` with `IndexedStateT`. Not sure if we have a solution for that now. And/or if we're interested in having `IndexedStateT` despite this short coming?

modules/optics/arrow-optics/src/main/kotlin/arrow/optics/Optional.kt
pakoito: `update` then

raulraja:
We can bring `IndexedStateT` back if we need it to bring polymorphism back to `mod` and `assign`. I don't recall what the problem with `binding` was. We can potentially make it a private data type in optics that does not explicitly adhere to Monad and therefore does not inherit binding.

raulraja:
Let's wait until #1040 is merged first please

raulraja:
@nomisRev This is ready to go once you rebase from master

PR #1048, opened 2018-10-14
FreeC implementation

nomisRev:
This PR is the starting point of a FS2 port or FS2 inspired stream lib.

In discussion with @raulraja we had the idea to move this to `arrow-free` as a replacement for the current `Free` implementation since we can then provide instance up to `MonadDefer` (and later also `Sync`).

If we are to move `FreeC` to `arrow-free` I'd like to improve the `Result` and `ViewL` code. Does any1 know if we could do something with `inline classes` to create `sealed class` wrappers without runtime overhead? I am assuming since they're going to be nested that they'll auto-box?


reports/baseline.xml
```diff
@@ -6,5 +6,6 @@
     <ID>ComplexMethod:tuple.kt$Tuple10EqInstance$override fun Tuple10&lt;A, B, C, D, E, F, G, H, I, J&gt;.eqv(b: Tuple10&lt;A, B, C, D, E, F, G, H, I, J&gt;): Boolean</ID>
     <ID>ComplexMethod:IORunLoop.kt$IORunLoop$fun &lt;A&gt; step(source: IO&lt;A&gt;): IO&lt;A&gt;</ID>
     <ID>ComplexMethod:IORunLoop.kt$IORunLoop$private fun loop( source: Current, cb: (Either&lt;Throwable, Any?&gt;) -&gt; Unit, rcbRef: RestartCallback?, bFirstRef: BindF?, bRestRef: CallStack?, isCancelled: (() -> Boolean)?): Unit</ID>
+    <ID>ThrowRuntimeException:FreeC.kt$internal inline fun &lt;F, R, A&gt; ViewL&lt;F, R&gt;.fold(pure: (FreeC.Pure&lt;F, R&gt;) -> A, fail: (FreeC.Fail&lt;F, R&gt;) -> A, interrupted: (FreeC.Interrupted&lt;F, R, Any?&gt;) -> A, view: (ViewL.Companion.View&lt;F, Any?, R&gt;) -> A ): A</ID>
```
nomisRev: Since like this is not picked up correctly. Any1 know how to fix this?

raulraja:
@nomisRev AFAIK inline classes properties of erasing the wrapper are not applied to any kind of generic container and they only work with primitives but @pakoito may know more.

modules/core/arrow-core/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -191,7 +191,7 @@ fun <T> Option<T>.getOrElse(default: () -> T): T = fold({ default() }, ::identit
  *
  * @param alternative the default option if this is empty.
  */
-inline fun <A, B : A> OptionOf<B>.orElse(alternative: () -> Option<B>): Option<B> = if (fix().isEmpty()) alternative() else fix()
+inline fun <A> OptionOf<A>.orElse(alternative: () -> Option<A>): Option<A> = if (fix().isEmpty()) alternative() else fix()
```
raulraja: nice one

modules/streams/arrow-streams/src/test/kotlin/arrow/streams/internal/FreeCTest.kt
```diff
@@ -0,0 +1,180 @@
...
+          .translate(EitherToTry)
+          .run(Try.monadError()) == Success(Some(s))
+      }
+    }
+
+  }
+}
+
+internal object EitherToTry : FunctionK<EitherPartialOf<Throwable>, ForTry> {
+  override fun <A> invoke(fa: Kind<EitherPartialOf<Throwable>, A>): Kind<ForTry, A> =
```
raulraja: We have `Try#toEither`. Perhaps this implementation can be exposed in the core as `Either<Throwable, A>.toTry()`

nomisRev: Yeah sure. Btw should we add `FunctionK` instances like this one to `Arrow`? Or is there a ticket for this that I missed?

pakoito: @nomisRev they won't hurt, but it isn't a priority. Only the ones between Try/Either/Option and maybe ListK may be more interesting for now.

modules/streams/arrow-streams/build.gradle
```diff
@@ -0,0 +1,14 @@
+dependencies {
```
raulraja: 🎉 it's alive!

modules/streams/arrow-streams/src/main/kotlin/arrow/streams/internal/FreeC.kt
```diff
@@ -0,0 +1,345 @@
...
+
+import arrow.Kind
+import arrow.core.*
+import arrow.effects.ExitCase
+import arrow.streams.CompositeFailure
+import arrow.streams.internal.FreeC.Result
+import arrow.typeclasses.MonadError
+
+/**
+ * Free Monad with Catch (and Interruption).
```
raulraja: We should mention and include fs2 here https://github.com/arrow-kt/arrow/blob/master/LICENSE#L20

modules/streams/arrow-streams/src/main/kotlin/arrow/streams/internal/FreeC.kt
```diff
@@ -0,0 +1,345 @@
...
+      view.step.attempt().flatMap { e ->
+        val next = view.next
+        val result: Result<Any?> = Result.fromEither(e)
+        next(result).run(ME)
+      }
+    }
+  }
+)
+
+/**  @higherkind doesn't respect access modifier **/
```
raulraja: once it is migrated to the meta processor you may be able to get that info

modules/streams/arrow-streams/src/main/kotlin/arrow/streams/internal/instances.kt
```diff
@@ -0,0 +1,98 @@
...
+import arrow.streams.internal.ap as apply
+
+internal fun <F> FreeC.Companion.functor(): Functor<FreeCPartialOf<F>> = object : FreeCFunctor<F> { }
+
+internal interface FreeCFunctor<F> : Functor<FreeCPartialOf<F>> {
+  override fun <A, B> Kind<FreeCPartialOf<F>, A>.map(f: (A) -> B): Kind<FreeCPartialOf<F>, B> =
+    this.fix().map(f)
+}
+
+internal fun <F> FreeC.Companion.applicative(): Applicative<FreeCPartialOf<F>> = object : FreeCApplicative<F> { }
```
raulraja: Are these handwritten because they are internal? After 0.8.0 we should be able to encode internal instances in the `ExtensionProcessor` so all the syntax they export it's also internal.

modules/streams/arrow-streams/src/test/kotlin/arrow/streams/internal/FreeCTest.kt
```diff
@@ -0,0 +1,180 @@
...
+    ))
+
+    testLaws(ApplicativeErrorLaws.laws(
+      AE = FreeC.applicativeError<ForTry>(),
+      EQ = Eq { a, b -> a.run(Try.monadError()) == b.run(Try.monadError()) },
+      EQERR = Eq { a, b -> a.run(Try.monadError()) == b.run(Try.monadError()) },
+      EQ_EITHER = Eq { a, b -> a.run(Try.monadError()) == b.run(Try.monadError()) }
+    ))
+
+    testLaws(MonadDeferLaws.laws(
```
raulraja: I thought the laws where encoded such as if you just test the `MonadDeferLaws` it will include all others in its hierarchy.

modules/streams/arrow-streams/src/test/kotlin/arrow/streams/internal/FreeCTest.kt
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    "Running a Bind value" {
+      forAll(Gen.string()) { s ->
+        FreeC.Bind(FreeC.pure("")) { _ -> FreeC.pure<EitherPartialOf<Throwable>, String>(s) }
+          .run(Either.monadError()) == Right(Some(s))
+      }
+    }
+
+    "Running an suspended value"{
```
raulraja: s/an/a

modules/streams/arrow-streams/src/test/kotlin/arrow/streams/internal/FreeCTest.kt
raulraja: Just a friendly reminder that in 0.8.0 the use of `ListK` and wrappers of one type arguments are rarely seen in user code unless they have polymorphic functions that expect values in the kind hierarchy.

raulraja:
@nomisRev This is ready to go once you rebase from master

modules/streams/arrow-streams/src/test/kotlin/arrow/streams/internal/FreeCTest.kt
nomisRev: Yes, they're. This is left over from when I was debugging combinators in smaller sets at a time :D

nomisRev:
@raulraja could you review again? Also API / encoding of `FreeC`. I think this is about as good as it gets and if everyone agrees I'll move it to `arrow-free`.

modules/streams/arrow-streams/src/main/kotlin/arrow/streams/internal/FreeC.kt
```diff
@@ -0,0 +1,432 @@
...
+      is Pure<*, R> -> {
+        try {
+          FreeC.Pure<F, R2>(f(free.r))
+        } catch (t: Throwable) {
+          FreeC.Fail<F, R2>(t)
+        }
+      }
+      is Fail<*, R> -> free
+      is Interrupted<*, R, *> -> free
+      else -> throw AssertionError("Unreachable")
```
raulraja: In the event that this is ever thrown we should indicate in the exception message that this is potentially caused by a bug and provide a link to report it in the arrow issue tracker.

PR #1049, opened 2018-10-15
Ref

nomisRev:


nomisRev:
Should we consider adding [kotlinx.atomicfu](https://github.com/Kotlin/kotlinx.atomicfu) for multiplatform in the future?

raulraja:
sure thing but we are still not close to multiplatform because of the coroutines hacks, annotation processors etc.

raulraja:
@nomisRev This is ready to go once it's rebased with master

PR #1054, opened 2018-10-20
Fix Foldable foldMapM restrictions

AdrianRaFo:
We were applying the Monad and Monoid restrictions to the same type `TC`.

This PR and the monoid parameter separately solving that.

### DIFF
--- a/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Foldable.kt
+++ b/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Foldable.kt
@@ -47,7 +47,10 @@ interface Foldable<F> {
       }
     }
 
-  fun <A, B> Kind<F, A>.reduceRightToOption(f: (A) -> B, g: (A, Eval<B>) -> Eval<B>): Eval<Option<B>> =
+  fun <A, B> Kind<F, A>.reduceRightToOption(
+    f: (A) -> B,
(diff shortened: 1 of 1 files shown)


modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Foldable.kt
```diff
@@ -157,9 +161,11 @@ interface Foldable<F> {
    *
    * Similar to foldM, but using a Monoid<B>.
    */
-  fun <G, A, B, TC> Kind<F, A>.foldMapM(tc: TC, f: (A) -> Kind<G, B>): Kind<G, B>
-    where TC : Monad<G>, TC : Monoid<B> = tc.run {
-    foldM(tc, tc.empty()) { b, a -> f(a).map { b.combine(it) } }
+  fun <G, A, B, MA, MO> Kind<F, A>.foldMapM(ma: MA, mo: MO, f: (A) -> Kind<G, B>): Kind<G, B>
+      where MA : Monad<G>, MO : Monoid<B> = ma.run {
```
pakoito: This will crash Android builds with the new compiler. You cannot nest two `run` directly.

AdrianRaFo: Fixed here https://github.com/arrow-kt/arrow/pull/1058

PR #1056, opened 2018-10-20
Remove @Deprecated functions + General warning cleanup

raulraja:
This PR removes all the previous deprecated functions that have been lingering around since several versions ago and cleans up most of the compiler warnings


detekt.yml
```diff
@@ -36,7 +36,7 @@ complexity:
     threshold: 300
     active: false
   ComplexMethod:
```
JorgeCastilloPrz: Does it make sense to remove this check? Not sure if we're getting any value from it. Looks like one of those you end up raising over and over when they blame.

JorgeCastilloPrz:
Thanks for this. One doubt. Why are we removing all those inline modifiers?

raulraja:
@JorgeCastilloPrz all the removed `inline` modifiers were being flagged as warnings by the Kotlin compiler since what they were inlining was potentially being insignificant and creating more code than necessary.

detekt.yml
raulraja: I raised the threshold of this check because the `FreeApplicative.foldMap` triggers it but it's necessary that it remains all in one function because it's inside a big mutable loop for performance reasons.

modules/core/arrow-data/src/test/kotlin/arrow/data/OptionTest.kt
```diff
@@ -33,6 +30,7 @@ class OptionTest : UnitSpec() {
 
   init {
 
+    @Suppress("UNUSED_VARIABLE")
```
pakoito: Why supress and not delete?

modules/docs/arrow-examples/src/test/kotlin/arrow/FpToTheMax.kt
```diff
@@ -40,7 +40,7 @@ class MonadAndConsoleRandom<F>(M: Monad<F>, C: Console<F>, R: FRandom<F>) : Mona
 data class TestData(val input: List<String>, val output: List<String>, val nums: List<Int>) {
     fun putStrLn(s: String): Tuple2<TestData, Unit> = copy(output = output.plus(s)) toT Unit
     fun getStrLn(): Tuple2<TestData, String> = copy(input = input.drop(1)) toT input[0]
-    fun nextInt(upper: Int): Tuple2<TestData, Int> = copy(nums = nums.drop(1)) toT nums[0]
+    fun nextInt(@Suppress("UNUSED_PARAMETER") upper: Int): Tuple2<TestData, Int> = copy(nums = nums.drop(1)) toT nums[0]
```
pakoito: Rename to `_` instead

PR #1057, opened 2018-10-20
Bump Kotlin version to 1.3-RC and updated Coroutines

raulraja:
Includes all changes from previous PR from @pakoito that was still unmerged


modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadContinuations.kt
```diff
@@ -38,9 +39,10 @@ open class MonadContinuation<F, A>(M: Monad<F>, override val context: CoroutineC
       }
 
       override fun resumeWithException(exception: Throwable) {
-        error = exception
+        error = null
```
pakoito: Are we sure this change was supposed to happen?

Issue #940, opened 2018-07-15
Change A -> B -> C functions in Optics to (A, B) -> C


jereksel:
Breaking change so former must be deprecated (Fortunatelly there won't be any issues with generic erasure)

jereksel:
I think we don't have to remove old ones: https://kotlinlang.org/docs/reference/java-to-kotlin-interop.html#handling-signature-clashes-with-jvmname

pakoito:
cc @nomisRev !!

nomisRev:
Closed by #1061

PR #1061, opened 2018-10-22
Improve optics constructors

nomisRev:
A small improvement to remove nested lambdas in the constructors of Optics.


nomisRev:
This will close #920 and #940

PR #1064, opened 2018-10-23
Adding Iterable Extensions that returns Option

guilherme-pohlmann:
## What

Adding `firstOrNone`, `singleOrNone`, `lastOrNone`, `elementAtOrNone` as Iterable extensions methods.

## Why

This methods allows us to maintain a friendly API syntax when operating on lists while avoiding null handling.

The motivation behind this appeared when we had to iterate over entries of a Map using `firstOrNull`. We are beginning with FP and trying do build a 100% pure Domain. Our domain is very permissive, and almost all of the inputs are Option, so the `firstOrNull` like methods goes against our principles, and `firstOrNone` like methods appears to be a solution to this problem.

Example:

```
fun main() {
    val foxMap = mapOf(1 to "The", 2 to "Quick", 3 to "Brown", 4 to "Fox")

    val ugly = foxMap.entries.firstOrNull { it.key == 5 }?.value.let { it?.toCharArray() }.toOption()
    val pretty = foxMap.entries.firstOrNone { it.key == 5 }.map { it.value.toCharArray() }
    
    //Do something with pretty Option
}
```

### DIFF
--- a/modules/core/arrow-core/src/main/kotlin/arrow/core/Option.kt
+++ b/modules/core/arrow-core/src/main/kotlin/arrow/core/Option.kt
@@ -202,14 +202,14 @@ fun <A> none(): Option<A> = None
 
 fun <T> Iterable<T>.firstOrNone(): Option<T> = this.firstOrNull().toOption()
 
-inline fun <T> Iterable<T>.firstOrNone(predicate: (T) -> Boolean): Option<T> = this.firstOrNull(predicate).toOption()
+fun <T> Iterable<T>.firstOrNone(predicate: (T) -> Boolean): Option<T> = this.firstOrNull(predicate).toOption()
 
 fun <T> Iterable<T>.singleOrNone(): Option<T> = this.singleOrNull().toOption()
 
-inline fun <T> Iterable<T>.singleOrNone(predicate: (T) -> Boolean): Option<T> = this.singleOrNull(predicate).toOption()
+fun <T> Iterable<T>.singleOrNone(predicate: (T) -> Boolean): Option<T> = this.singleOrNull(predicate).toOption()
 
 fun <T> Iterable<T>.lastOrNone(): Option<T> = this.lastOrNull().toOption()
 
-inline fun <T> Iterable<T>.lastOrNone(predicate: (T) -> Boolean): Option<T> = this.lastOrNull(predicate).toOption()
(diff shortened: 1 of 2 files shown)


modules/core/arrow-core/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -199,3 +199,17 @@ fun <A> Boolean.maybe(f: () -> A): Option<A> =
...
+
+inline fun <T> Iterable<T>.firstOrNone(predicate: (T) -> Boolean): Option<T> = this.firstOrNull(predicate).toOption()
+
+fun <T> Iterable<T>.singleOrNone(): Option<T> = this.singleOrNull().toOption()
+
+inline fun <T> Iterable<T>.singleOrNone(predicate: (T) -> Boolean): Option<T> = this.singleOrNull(predicate).toOption()
+
+fun <T> Iterable<T>.lastOrNone(): Option<T> = this.lastOrNull().toOption()
+
+inline fun <T> Iterable<T>.lastOrNone(predicate: (T) -> Boolean): Option<T> = this.lastOrNull(predicate).toOption()
```
pakoito: We're slowly doing away with the inlines. They don't give much microperf, increase the bytecode size on the clients, and they allow non-local returns which are scary :D

modules/docs/arrow-docs/docs/docs/datatypes/option/README.md
```diff
@@ -138,6 +138,29 @@ val nullableValue: String? = "Hello"
 nullableValue.toOption()
 ```
 
+Some Iterable extensions were added, so you can maintain a friendly API syntax while avoiding null handling (`firstOrNull()`)
+
+```
```
pakoito: We use `ank` for snippets. Check other code blocks in this file to understand them, but tl;dr each snippet block gets executed on builds and the output is then inlined.

The scripting engine we use doesn't quite work for async calls and coroutines, but these would fit just right!

modules/docs/arrow-docs/docs/docs/datatypes/option/README.md
```diff
@@ -138,6 +138,29 @@ val nullableValue: String? = "Hello"
 nullableValue.toOption()
 ```
 
+Some Iterable extensions were added, so you can maintain a friendly API syntax while avoiding null handling (`firstOrNull()`)
```
pakoito: `were added,` -> `are available,`

guilherme-pohlmann:
Thank you for your help @pakoito.

PR #1067, opened 2018-10-25
Listk clean

leandroBorgesFerreira:
I was reading the code of ListK and notice that it has some unnecessary `fix()` calls. This PR removes then for simplicity.


leandroBorgesFerreira:
The commit list of this PR is really weird o.O

PR #1068, opened 2018-10-25
removing unnecessary `fix()`call

leandroBorgesFerreira:
I was reading the code of ListK and notice that it has some unnecessary fix() calls. This PR removes then.

### DIFF
--- a/modules/core/arrow-data/src/main/kotlin/arrow/data/ListK.kt
+++ b/modules/core/arrow-data/src/main/kotlin/arrow/data/ListK.kt
@@ -4,6 +4,7 @@ import arrow.Kind
 import arrow.core.*
 import arrow.higherkind
 import arrow.typeclasses.Applicative
+import arrow.typeclasses.fix
 
 /**
  * A wrapper data type also considered by the @extension mechanisms to forward type class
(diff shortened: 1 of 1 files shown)


modules/core/arrow-data/src/main/kotlin/arrow/data/ListK.kt
```diff
@@ -19,33 +19,33 @@ import arrow.typeclasses.Applicative
 @higherkind
 data class ListK<out A>(val list: List<A>) : ListKOf<A>, List<A> by list {
 
-  fun <B> flatMap(f: (A) -> ListKOf<B>): ListK<B> = this.fix().list.flatMap { f(it).fix().list }.k()
+  fun <B> flatMap(f: (A) -> ListKOf<B>): ListK<B> = this.list.flatMap { f(it).fix().list }.k()
```
pakoito: I don't even think you need `this` here.

modules/core/arrow-data/src/main/kotlin/arrow/data/ListK.kt
```diff
@@ -19,33 +19,33 @@ import arrow.typeclasses.Applicative
...
 
-  fun <B> map(f: (A) -> B): ListK<B> = this.fix().list.map(f).k()
+  fun <B> map(f: (A) -> B): ListK<B> = this.list.map(f).k()
```
pakoito: or here

modules/core/arrow-data/src/main/kotlin/arrow/data/ListK.kt
```diff
@@ -19,33 +19,33 @@ import arrow.typeclasses.Applicative
...
 
-  fun <B> foldLeft(b: B, f: (B, A) -> B): B = this.fix().fold(b, f)
+  fun <B> foldLeft(b: B, f: (B, A) -> B): B = this.fold(b, f)
```
pakoito: or here xD

PR #1073, opened 2018-10-28
[WIP] map and flatMap as extension functions 

leandroBorgesFerreira:
*Proposal*

The `map` and `flatMap` should be extension function for all containers of iterables. This avoids implementations in the concrete classes and make the code easier to read as the `map` and `flatMap`  live in just one place. 

With this PR a SetK.map is going to return a ListK instead of List and it holds true for future implementations of iterable holders. 

*Problem*: Right now it is not possible to create the version of flatMap, because the ForListK doesn't inherits from a `ForIterableHolder` interface (which doesn't even exists right now). 

The code for ForListK is generated and I am not sure how it works, so I don't know how to change it. Some guidance would be great =]. (Maybe some KindedJ docs?)

If it is possible to make something like: 
```
class ForListK private constructor() : ForIterableK { companion object } // ForIterableK Should be a interface
```

Then it would be possible to do this:
```
fun <B, F: ForIterableHolder> flatMap(f: (A) -> Kind<F, B>): ListK<B> = list.flatMap { f(it).fix().list }.k()
```

I believe this would be a great advantage to the library, because we would make the use of Higher Kinded Types more powerful in Arrow. Right now those types doesn't accept inheritance. 

Obs: I just changed the name of `map` to `mapK` to make de change more clean, but it can use `map` as the final name

So what do you guys think? It is a polemic PR I know...


leandroBorgesFerreira:
You're right, I don't need the wrapper as those guys are already Iterables (duh). Also I just added the map for SequenceK. 

The functor doesn't solve this problem, as the implementations of functor still uses the methods that I am trying to remove. I mean, `ListKFunctorInstance` uses the `map` from ListK, so you still have to deal with this method, even using Functors or Monads. 

For each new implementation of IterableK or SequenceK (I don't know... SuperListK, SuperSequenceK...) you're gonna need to implement the `map`, `ap` and `flatMap`, if you use extension functions from Iterable (like `mapK`, `flatMapK`), you get then "for free". 

But I still need to figure out how to have some kind of inheritance in the `ForListK` to be able to implement a `flatMap`.

JorgeCastilloPrz:
Hi @leandroBorgesFerreira! In an effort to clean the PR list we've decided to close all the ones that are older than 2 months and consider them obsolete except for very specific circumstances that we will evaluate per PR. Let me close this one since it remains untouched since more than 1 year ago and probably doesn't apply anymore given how many changes Arrow has suffered. Note we can bring the proposal back to life if at some point we find it's a priority and the library is not providing it already. Thanks for the proposal 🙏

PR #1079, opened 2018-11-01
The Arrow Query Language.

raulraja:
The Arrow query language (name TBD) offers SQL, Monoid comprehensions and LINQ
style queries generalized to all data types that can provide instances for the type classes that it uses internally.

This works is inspired in random ideas drawn from:
- [Comprehending Ringads](http://www.cs.ox.ac.uk/jeremy.gibbons/publications/ringads.pdf)
- [LINQ and F# Query expressions](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/query-expressions)
- [Comprehending Monoids with class](https://lptk.github.io/programming/2018/10/04/comprehending-monoids-with-class.html)

AQL gives you a unified single language that is familiar to users and that does not require users to understand the type class combinators.

AQL covers any use cases in which you are extracting and transforming data and your data is contained within a data type for which a kinded representation exists.

Ex:
- query a remote data source `Deferred, IO`
- query a remote spark distributed data set: `Dataset`
- query an in-memory collection `List, Set, ..`
- query nested type constructors. Ex: `Deferred<List<A>>`
- query JSON trees and in general any kinded sum type since we can derive `Foldable` instances for those

In general, you can use DSL combinators over any data type value as long as the data type provides instances for the combinator in use.

The current set of type operators and type class dependencies are:

| TypeClass | Dependency         |
|-----------|--------------------|
| Select    | Functor            |
| From      | (Data type source) |
| Join      | Applicative        |
| Where     | FunctorFilter      |
| Count     | Foldable           |
| Sum       | Foldable           |
| GroupBy   | Foldable           |
| OrderBy   | Foldable           |

Instances for some of the common data types will be included as well as submodules for the different arrow integrations.

Before this is merged I will fix the `@extension` processor to support `infix` type class extension projections in order to achieve true SQL like syntax beside the fluent builder.

Additionally in AQL we have freedom for naming combinators like `bind` or `flatMap` to other names like `chain. This is in the same line all other SQL expressions try to ressemble plain english, and these would be just a functional superset over sql similar in spirit to `LINQ` but supporting other type classes like error handling, traversing, effects, etc.

A few examples of usage of AQL in action over the`List` below:

```kotlin
package arrow.aql.tests

import arrow.aql.Ord
import arrow.aql.instances.id.select.value
import arrow.aql.instances.list.count.count
import arrow.aql.instances.list.count.value
import arrow.aql.instances.list.from.join
import arrow.aql.instances.list.groupBy.groupBy
import arrow.aql.instances.list.orderBy.order
import arrow.aql.instances.list.orderBy.orderMap
import arrow.aql.instances.list.select.query
import arrow.aql.instances.list.select.select
import arrow.aql.instances.list.select.value
import arrow.aql.instances.list.sum.sum
import arrow.aql.instances.list.union.union
import arrow.aql.instances.list.where.where
import arrow.aql.instances.list.where.whereSelection
import arrow.aql.instances.listk.select.select
import arrow.aql.instances.listk.select.selectAll
import arrow.core.Id
import arrow.instances.order
import arrow.test.UnitSpec
import io.kotlintest.matchers.shouldBe

class AQLTests : UnitSpec() {

  init {

    "AQL is able to `select`" {
      listOf(1, 2, 3).query {
        select { this * 10 }
      }.value() shouldBe listOf(10, 20, 30)
    }

    "AQL is able to `select count`" {
      listOf(1, 2, 3).query { select { this } }.count()
        .value() shouldBe 3L
    }

    "AQL is able to `select`, transform and filter data with `where`" {
      listOf(1, 2, 3).query {
        selectAll() where { this > 2 }
      }.value() shouldBe listOf(3)
    }

    "AQL is able to `select`, transform and filter data with `in`" {
      listOf(1, 2, 3).query {
        selectAll() where { this in listOf(3) }
      }.value() shouldBe listOf(3)
    }

    "AQL is able to `join` and transform data for List" {
      (listOf(1) join listOf("a")).query {
        select { "$a$b" } where { a > 0 } whereSelection { startsWith("1") }
      }.value() shouldBe listOf("1a")
    }

    "AQL is able to `groupBy`" {
      data class Student(val name: String, val age: Int)

      val john = Student("John", 30)
      val jane = Student("Jane", 32)
      val jack = Student("Jack", 32)
      listOf(john, jane, jack).query {
        selectAll() groupBy { age }
      }.value() shouldBe Id(mapOf(30 to listOf(john), 32 to listOf(jane, jack)))
    }

    data class Student(val name: String, val age: Int)

    val john = Student("John", 30)
    val jane = Student("Jane", 32)
    val jack = Student("Jack", 32)
    val chris = Student("Chris", 40)

    "AQL is able to `groupBy`" {
      listOf(john, jane, jack).query {
        selectAll() where { age > 30 } groupBy { age }
      }.value() shouldBe mapOf(32 to listOf(jane, jack))
    }

    "AQL is able to `sum`" {
      listOf(john, jane, jack).query {
        selectAll() where { age > 30 } sum { age.toLong() }
      }.value() shouldBe 64L
    }

    "AQL is able to `order by Asc` simple selects" {
      listOf(1, 2, 3).query {
        select { this * 10 } order Ord.Asc(Int.order())
      }.value() shouldBe listOf(10, 20, 30)
    }

    "AQL is able to `order by Desc` simple selects" {
      listOf(1, 2, 3).query {
        select { this * 10 } order Ord.Desc(Int.order())
      }.value() shouldBe listOf(30, 20, 10)
    }

    "AQL is able to `order by Desc` simple selects with explicit instance" {
      listOf(1, 2, 3).query {
        select { this * 10 } order Ord.Desc(Int.order())
      }.value() shouldBe listOf(30, 20, 10)
    }


    "AQL is able to `groupBy` and then order `keys`"{
      listOf(john, jane, jack).query {
        selectAll() where { age > 30 } groupBy { age } orderMap Ord.Desc(Int.order())
      }.value() shouldBe mapOf(32 to listOf(jane, jack))
    }

    "AQL is able to `union`" {
      val queryA = listOf(
        "customer" to john,
        "customer" to jane
      ).select { this }
      val queryB = listOf(
        "sales" to jack,
        "sales" to chris
      ).select { this }
      queryA.union(queryB).value() shouldBe listOf(
        "customer" to john,
        "customer" to jane,
        "sales" to jack,
        "sales" to chris
      )
    }
  }
}
```

The initial DRAFT of AQL with just support for the core type classes is in. I have not added yet support for the rest of the modules to give enough time for the design to evolve before we are dependent on a ton of instances and code to get updated.


gradle.properties
```diff
@@ -4,7 +4,6 @@ VERSION_NAME=0.8.0-SNAPSHOT
 # Gradle options
 org.gradle.jvmargs=-Xmx4g
 # Kotlin configuration
-kotlin.coroutines=enable
```
pakoito: This is where this was

modules/aql/arrow-query-language/build.gradle
```diff
@@ -0,0 +1,14 @@
...
+    compile project(':arrow-mtl')
+    kapt project(':arrow-annotations-processor')
+    kaptTest project(':arrow-annotations-processor')
+    compileOnly project(':arrow-annotations-processor')
+    testCompileOnly project(':arrow-annotations-processor')
+    testCompile "io.kotlintest:kotlintest:$kotlinTestVersion"
+    testCompile project(':arrow-test')
+}
+
+apply from: rootProject.file('gradle/gradle-mvn-push.gradle')
```
pakoito: Do we plan to have this out for 0.8.0?

modules/aql/arrow-query-language/src/main/kotlin/arrow/aql/Prelude.kt
```diff
@@ -0,0 +1,6 @@
+package arrow.aql
```
pakoito: we're calling these `syntax` for now

modules/aql/arrow-query-language/src/main/kotlin/arrow/aql/OrderBy.kt
```diff
@@ -0,0 +1,71 @@
...
+      select = ::identity,
+      from = sortedMap.toMap().just()
+    )
+  }
+
+  fun <Z> Query<ForId, List<Z>, List<Z>>.value(): List<Z> =
+    this@value.from.value()
+
+  fun <Z, X> Query<ForId, Map<X, List<Z>>, Map<X, List<Z>>>.value(dummy: Unit = Unit): Map<X, List<Z>> =
+    this@value.from.value()
```
pakoito: This `this` shouldn't be necessary.

modules/aql/arrow-query-language/build.gradle
raulraja: 0.9.0, It's in early design phase

modules/aql/arrow-query-language/src/main/kotlin/arrow/aql/Sum.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import arrow.core.*
+import arrow.data.ForListK
+import arrow.data.fix
+import arrow.data.k
+import arrow.instances.monoid
+import arrow.typeclasses.Foldable
+import com.sun.org.apache.xpath.internal.operations.Bool
+
+interface Sum<F> {
```
pakoito: We already have a `Sum` in the codata package. Could it be renamed to `Add`?

modules/aql/arrow-query-language/src/main/kotlin/arrow/aql/instances/either.kt
```diff
@@ -0,0 +1,20 @@
...
+import arrow.aql.Select
+import arrow.core.Either
+import arrow.core.EitherPartialOf
+import arrow.extension
+import arrow.instances.either.applicative.applicative
+import arrow.instances.either.functor.functor
+import arrow.typeclasses.Applicative
+import arrow.typeclasses.Functor
+
+@extension interface EitherSelect<L> : Select<EitherPartialOf<L>> {
```
pakoito: Consistent formatting of @extension in the instances in this file

modules/aql/arrow-query-language/src/main/kotlin/arrow/aql/instances/eval.kt
```diff
@@ -0,0 +1,20 @@
...
+import arrow.instances.eval.functor.functor
+import arrow.typeclasses.Applicative
+import arrow.typeclasses.Functor
+
+@extension
+interface EvalFrom : From<ForEval> {
+  override fun applicative(): Applicative<ForEval> = Eval.applicative()
+}
+
+@extension interface EvalSelect : Select<ForEval> {
```
pakoito: Same :D

modules/aql/arrow-query-language/src/test/kotlin/arrow/aql/tests/AQLTests.kt
```diff
@@ -0,0 +1,141 @@
...
+    "AQL is able to `groupBy` and then order `keys`" {
+      listOf(john, jane, jack)
+        .select { it }
+        .where { it.age > 30 }
+        .groupBy { it.age }
+        .order(Ord.Desc, Int.order())
+        .value() shouldBe mapOf(32 to listOf(jane, jack))
+    }
+
+//    "AQL is able to `select count` and `groupBy`" {
```
pakoito: ???

raulraja: WIP

modules/core/arrow-core/src/main/kotlin/arrow/core/Id.kt
```diff
@@ -33,4 +33,10 @@ data class Id<out A>(val value: A) : IdOf<A> {
 
     fun <A> just(a: A): Id<A> = Id(a)
   }
+
+  override fun equals(other: Any?): Boolean =
```
pakoito: What was the issue with the default implementation?

pakoito: Oh, right, you're comparing wrapped and unwrapped values.

modules/core/arrow-mtl/src/main/kotlin/arrow/mtl/typeclasses/MonadCombine.kt
```diff
@@ -3,6 +3,8 @@ package arrow.mtl.typeclasses
 import arrow.Kind
 import arrow.Kind2
 import arrow.core.Tuple2
+import arrow.core.curry
```
pakoito: Does it need curry?

modules/core/arrow-data/src/main/kotlin/arrow/data/ListK.kt
```diff
@@ -33,6 +33,13 @@ data class ListK<out A>(val list: List<A>) : ListKOf<A>, List<A> by list {
     return Eval.defer { loop(this) }
   }
 
+  override fun equals(other: Any?): Boolean =
+    when (other) {
```
pakoito: I don't disagree, but we should also do it for the other wrappers, and to be honest if we're using 1.3 we should just use inline classes instead.

modules/core/arrow-core/src/main/kotlin/arrow/core/Id.kt
raulraja: `Id` and the `K` Wrappers should all express their equality to the value they wrap or things can go wrong when comparing wrappers and wrapped values with `==`. This surfaced as I was testing the DSL

rafaparadela:
It would be beautiful to provide the ability for using `FROM` to define the source in the same order that an SQL query does. Something like:

```kotlin
query
  .select { it }
  .from(listOf(1, 2, 3))
  .where { it > 2 } shouldBe listOf(3)
```

I mean, the syntax might operate over queries instead over data.

modules/core/arrow-core/src/main/kotlin/arrow/core/Id.kt
pakoito: We can probably fix this with inline classes. I expect them to be the same.

pakoito: Well, they aren't -.- Still, we should re-encode them as inlined at some point.

pakoito:
We also need a `join` or `selectMany` that uses `flatMap` behind the scenes.

JorgeCastilloPrz:
+1 to supporting From also as a Typeclass, since we'll need that for achieving human like language whenever we get infix notation ready. "select whatever from wherever"

modules/aql/arrow-query-language/src/main/kotlin/arrow/aql/Count.kt
```diff
@@ -0,0 +1,27 @@
...
+
+  fun <A, Z> Query<F, A, Z>.count(): Query<ForListK, Long, Long> =
+    foldable().run {
+        Query(
+          select = ::identity,
+          from = listOf(from.size(Long.monoid())).k()
+        )
+    }
+
+  fun Query<ForListK, Long, Long>.value(): Long =
```
JorgeCastilloPrz: I assume the plan with value is to get a terminal operation enforcing the complete chain to resolve. Would make more sense to call it "run" or something similar?

JorgeCastilloPrz:
IMO this is potentially a powerful language for teaching FP concepts using "human" words (tutorials, courses ..etc). Good job on implementing those papers.

modules/aql/arrow-query-language/src/main/kotlin/arrow/aql/Count.kt
raulraja: Yes, they should be renamed to run but in some cases the resulting value may be a pure expression for example when working with IO

raulraja:
@jorgecastillopr we can't place select first without typing the Lambda or type arguments explicitly. This a limitation of inference in Kotlin which happens in order left to right.

pakoito:
>@pakoito join is already there and it delegates to Applicative#product.

I mean another, new one, that delegates to `flatMap`.

modules/aql/arrow-query-language/src/main/kotlin/arrow/aql/GroupBy.kt
```diff
@@ -10,7 +10,7 @@ interface GroupBy<F> {
 
   fun foldable(): Foldable<F>
 
-  infix fun <A, Z, X> Query<F, A, Z>.groupBy(group: (Z) -> X): Query<ForId, Map<X, List<Z>>, Map<X, List<Z>>> =
+  infix fun <A, Z, X> Query<F, A, Z>.groupBy(group: Z.() -> X): Query<ForId, Map<X, List<Z>>, Map<X, List<Z>>> =
```
pakoito: This is something I thought at some point could be more widespread. I'm not sure about the consequences...

modules/aql/arrow-query-language/src/main/kotlin/arrow/aql/Select.kt
```diff
@@ -21,6 +21,9 @@ interface Select<F> {
   infix fun <A, Z> Source<F, A>.select(f: Selection<A, Z>): Query<F, A, Z> =
     Query(f, this)
 
+  fun <A> Source<F, A>.selectAll(): Query<F, A, A> =
```
pakoito: It'll probably feel better as a field

raulraja:
I'll address some of your comments in other PRs. We need to primarily do an overhaul on names and make sure it all makes sense before 1.0.0 and many of the concerns here are related to that. As for query extensions on combinators like `.value()` we need to provide at a global level one for each terminal type which is `Kind`, `List`, `Map`, `Id` and `Long` so far.

modules/core/arrow-annotations-processor/src/main/java/arrow/meta/encoder/jvm/JvmMetaApi.kt
```diff
@@ -93,9 +94,9 @@ interface JvmMetaApi : MetaApi, TypeElementEncoder, ProcessorUtils, TypeDecoder
   /**
    * @see [MetaApi.removeConstrains]
    */
-  override fun Func.removeConstrains(): Func =
+  override fun Func.removeConstrains(keepModifiers: Set<Modifier>): Func =
     copy(
-      modifiers = emptyList(),
+      modifiers = modifiers.filterNot { it !in keepModifiers },
```
leandroBorgesFerreira: `modifiers.filter { it !in keepModifiers }`?

modules/aql/arrow-query-language/src/main/kotlin/arrow/aql/Count.kt
```diff
@@ -0,0 +1,27 @@
...
+    foldable().run {
```
leandroBorgesFerreira: Is there a need to change scope to Foldable? This run is just changing scope right?

modules/aql/arrow-query-language/src/main/kotlin/arrow/aql/Union.kt
```diff
@@ -0,0 +1,23 @@
...
+
+interface Union<F> {
+
+  fun foldable(): Foldable<F>
+
+  infix fun <A, B, Z> Query<F, A, Z>.union(query: Query<F, B, Z>): Query<ForListK, Z, Z> =
+    foldable().run {
+      val la: ListK<Z> = from.foldMap(ListK.monoid()) { listOf(select(it)).k() }
+      val lb: ListK<Z> = query.from.foldMap(ListK.monoid()) { listOf(query.select(it)).k() }
+      val result: ListK<Z> = la.combineK(lb).k()
```
leandroBorgesFerreira: No need for `k()`. Could be just `val result: ListK<Z> = la.combineK(lb)`

raulraja: `result` has to be a Kinded value and `List` on its own isn't.

modules/aql/arrow-query-language/src/main/kotlin/arrow/aql/GroupBy.kt
```diff
@@ -0,0 +1,29 @@
...
+      Id.just(from.foldLeft(emptyMap()) { map, a ->
+        val z: Z = select.invoke(a)
+        val key: X = group(z)
+        val grouped: List<Z> = map[key]?.let { it + z } ?: listOf(z)
+        map + mapOf(key toT grouped)
+      }))
+    }
+
+  fun <Z, X> Query<ForListK, Map<X, List<Z>>, Map<X, List<Z>>>.value(): Map<X, List<Z>> =
+    foldable().run {
```
leandroBorgesFerreira: No need for the `foldable().run`. 

Could be just: 
```
fun <Z, X> Query<ForListK, Map<X, List<Z>>, Map<X, List<Z>>>.value(): Map<X, List<Z>> =
    from.fix().firstOrNone().getOrElse { emptyMap() }
```

modules/aql/arrow-query-language/src/main/kotlin/arrow/aql/Sum.kt
```diff
@@ -0,0 +1,25 @@
...
+    foldable().run {
+        Query(
+          select = ::identity,
+          from = Id(from.foldLeft(0L) { acc, a ->
+            acc + f(a)
+          })
+        )
+    }
+
+  fun Query<ForId, Long, Long>.value(): Long =
```
leandroBorgesFerreira: No need to the `foldable().run {` could be
```
fun Query<ForId, Long, Long>.value(): Long = from.value()
```

PR #1087, opened 2018-11-06
changing visibility for Call inside CallK

leandroBorgesFerreira:
Call shouldn't be exposed so CallK can have a simpler API. 

Call and CallK were not meant to be used together. One can choose between Call's extension functions or use CallK directly, but using both only allows confusion.


raulraja:
I think it's fine if it's encapsulated in this case because in 0.8.1 the usage of the wrapper is transparent and users have no need to use neither `.k()` nor `callK.call`. All the syntax and constructors are projected directly over the wrapped object. We just need to fix the annotation processor to handle cases like this one where there is no inheritance and therefore we can't just cast but we have to instead wrap and unwrap the values with property access instead of safe downcasts. Safe downcasting is the only strategy currently considered in `ExtensionProcessor` for wrapped types and it relies on the wrapper to be a subtype of the wrapped type like in the case of `ListK<A>` which extends `List<A>`.

raulraja:
I think we need to go always in the spirit that wrappers do not exist and we are truly complementing other third-party data types extending their syntax and making them available as runtime of polymorphic programs. Other than that in my opinion we should hide as much as we can our impl details because that means an easier API for most use cases.

pakoito:
The point is that `Call` has methods other than the enqueuing ones we use that are important and should still be accessible to whomever is using the wrapper.

leandroBorgesFerreira:
I was taking a look at the code... And it seams that we don't need the Wrapper at all to be true. I'll take a look to see if it makes sense

pakoito:
Yeah the wrapper was more to signify our own API if we'd like to extend it, but we could just work directly on Calls.

leandroBorgesFerreira:
Folks, I used the CallK on a small project and I am not sure about just having the extension functions anymore. CallK can be a much easier way to use it, since you only have 3 method that you can choose from. With only the extension function it is not very clear with are the new method and with are the old ones. So I still think this PR makes sense. WDYT?

raulraja:
@leandroBorgesFerreira sounds good to me

pakoito:
I prefer if it was public but won't block on it.

raulraja:
@leandroBorgesFerreira @pakoito let's keep this open for now and think of a generalized strategy for wrappers. In my opinion, wrappers need to disappear from user code because the main use case is not to use a Class with fewer combinators or restrict access to the wrapped object. 

The goal is to enhance and be a `functional companion` that is to enrich 3rd party data types with functional behaviors. I don't think we are guardians or gatekeepers of their API and we are also not supposed to expose our implementations details. We should strive to use wrappers for the sole purpose of providing Kind support for third party types because if HKs become something in the future in the Kotlin compiler we need to remove all these wrappers and provide proper instances for the 3rd party types directly.

raulraja:
I'm gonna close this for now but we need to address the entire Wrapper issue as a whole. I have some ideas as to how we can accomplish this with the `@extension` processor

PR #1088, opened 2018-11-07
[WIP] Move to Kotlin 1.3.0 and coroutines 1.0.0

kitttn:
Need help with tests here 😞 
Currently, tests fail in DeferredTest.kt, it looks like `.invokeOnCompletion` which throws Exception is now forbidden according to [docs](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-completion-handler.html) - it gets wrapped with `CompletionHandlerException` and propagated. 

What has been done: 
* Tried to use Dispatchers.IO - throws `AssertionError`, which means re-thrown exceptions are caught in run block. 
* Tried to catch `CompletionHandlerException` - still getting that exception in wrong `catch { .. }` block though.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -21,9 +21,9 @@ buildscript {
         gradleVersionsPluginVersion = '0.20.0'
         javaVersion = JavaVersion.VERSION_1_7
         kotlinTestVersion = '2.0.7'
-        kotlinVersion = '1.3.0-rc-190'
+        kotlinVersion = '1.3.0'
         daggerVersion = '2.17'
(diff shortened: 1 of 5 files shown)


pakoito:
I didn't have access to your branch, so I had to create my own in #1088. I used the opportunity to update some of the API changes in kotlinx.coroutines too.

pakoito:
Thank you for the contribution, I didn't even know how to make KotlinPoet work with 1.3! Closing it now that it's been merged.

PR #1103, opened 2018-11-11
Fiber 

### LINKED ISSUE #1089: Fiber
Fiber is an abstraction representing the **pure** result of an asynchronous computation by an `Async` data type (i.e: `IO`), where the computation started concurrently and can be either joined or canceled.

[Here is the cats implementation](https://github.com/typelevel/cats-effect/blob/84696fb35ec8464d277f221991a64882c79363a0/core/shared/src/main/scala/cats/effect/Fiber.scala#L55) for reference.

`Concurrent` requires `Fiber`, so this should be done first.

nomisRev:
I left the docs unpublished because this doesn't make a lot of sense without `Concurrent`.

For that reason, I have also not published the code for `Deferred` and left it in docs. This should be enabled with the `Concurrent` type class.

Closes #1089

### DIFF
--- a/modules/docs/arrow-docs/docs/docs/effects/fiber/README.MD
+++ b/modules/docs/arrow-docs/docs/docs/effects/fiber/README.MD
@@ -1,0 +1,50 @@
+---
+layout: docs
+title: Fiber
+permalink: /docs/effects/fiber/
+---
+
+## Fiber
+
+{:.advanced}
+advanced
+
+A `Fiber` is a concurrency primitive for describing parallel operations or multi-tasking.
+Concurrently started tasks can either be joined or canceled and this are the only two operators available on `Fiber`.
+
+Using `Fiber` we can verily easily describe parallel operations such as `parallelMap`.
+
+```kotlin:ank
+import arrow.effects.*
+import arrow.effects.deferredk.monad.monad
+import arrow.effects.typeclasses.Fiber
+import arrow.typeclasses.binding
+
+fun <A, B, C> parallelMap(first: DeferredK<A>,
+                     second: DeferredK<B>,
+                     f: (A, B) -> C): DeferredK<C> =
+  DeferredK.monad().binding {
+    val fiberOne: Fiber<ForDeferredK, A> = first.startF().bind()
+    val fiberTwo: Fiber<ForDeferredK, B> = second.startF().bind()
+    f(fiberOne.join.bind(), fiberTwo.join.bind())
+  }.fix()
+
+val first = DeferredK.sleep(5000).map {
+  println("Hi, I am first")
+  1
+}
+
+val second = DeferredK.sleep(3000).map {
+  println("Hi, I am second")
+  2
+}
+```
+
+```kotlin
+parMap(first, second, Int::plus).await()
+
+//Hi, I am second
+//Hi, I am first
+//3
+```
+
--- a/modules/docs/arrow-docs/docs/docs/effects/fiber/README.MD
+++ b/modules/docs/arrow-docs/docs/docs/effects/fiber/README.MD
@@ -18,7 +18,6 @@ Using `Fiber` we can verily easily describe parallel operations such as `paralle
 import arrow.effects.*
 import arrow.effects.deferredk.monad.monad
 import arrow.effects.typeclasses.Fiber
-import arrow.typeclasses.binding
 
 fun <A, B, C> parallelMap(first: DeferredK<A>,
                      second: DeferredK<B>,
--- a/modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Fiber.kt
+++ b/modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Fiber.kt
@@ -1,6 +1,7 @@
 package arrow.effects.typeclasses
 
 import arrow.Kind
+import arrow.effects.internal.CancelToken
 
 /**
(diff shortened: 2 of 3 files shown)


modules/docs/arrow-docs/docs/docs/effects/fiber/README.MD
```diff
@@ -0,0 +1,50 @@
+---
```
JorgeCastilloPrz: You said you're not publishing the docs yet but this is docs. This is gonna be automatically deployed as soon as Ank runs. Is it intended?

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Fiber.kt
```diff
@@ -0,0 +1,47 @@
...
+   * until that result is available.
+   */
+  val join: Kind<F, A>
+
+  /**
+   * Triggers the cancellation of the fiber.
+   *
+   * @returns a task that trigger the cancellation upon evaluation.
+   */
+  val cancel: Kind<F, Unit>
```
JorgeCastilloPrz: We already got `CancelToken<F>` ongoing in `Bracket` PR. Should we wait to merge that one and update this to use the typealias ? (this is ` def cancel: CancelToken[F]` in cats).

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Fiber.kt
```diff
@@ -0,0 +1,47 @@
...
+
+  companion object {
+
+    /**
+     * [Fiber] constructor.
+     *
+     * @param join task that will trigger the cancellation.
+     * @param cancel task that will await for the completion of the underlying fiber.
+     */
+    operator fun <F, A> invoke(join: Kind<F, A>, cancel: Kind<F, Unit>): Fiber<F, A> = object : Fiber<F, A> {
```
JorgeCastilloPrz: same here, you'd have `cancel: CancelToken<F>` as an argument instead of `Kind<F, Unit>`.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Fiber.kt
```diff
@@ -0,0 +1,47 @@
...
+      override val cancel: Kind<F, Unit> = cancel
+      override val join: Kind<F, A> = join
+      override fun toString(): String = "Fiber(join= $join, cancel= $cancel)"
```
JorgeCastilloPrz: just to satisfy my curiosity, are we overriding toString for testsing purposes or something? are we doing the same in other places in arrow?

JorgeCastilloPrz:
Good job dude :) you're fast! Shouldn't we add a couple of tests for both methods?

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Fiber.kt
```diff
@@ -0,0 +1,47 @@
...
+  val join: Kind<F, A>
+
+  /**
+   * Triggers the cancellation of the fiber.
+   *
+   * @returns a task that trigger the cancellation upon evaluation.
+   */
+  val cancel: Kind<F, Unit>
+
+  operator fun component1(): Kind<F, A> = join
```
JorgeCastilloPrz: 👍

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Fiber.kt
nomisRev: `type CancelToken[F[_]] = F[Unit]` in cats. Anyhow, I am fine with waiting until `Bracket` is merged and do it in 1 shot.

modules/docs/arrow-docs/docs/docs/effects/fiber/README.MD
nomisRev: With `not publishing` I meant not add a menu item to navigate to `Fiber`. I should've been more specific. So the doc is published but it can not be found by navigating documentation.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Fiber.kt
nomisRev: > just to satisfy my curiosity, are we overriding toString for testsing purposes or something?

I guess so I can't remember what made me do this 😅

> are we doing the same in other places in arrow?

I think not actually.

nomisRev:
@JorgeCastilloPrz there is not much to write tests for until `Concurrent` is ready? There is no logic implemented whatsoever.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Fiber.kt
JorgeCastilloPrz: Yep that's what I meant about the typealias. Let's wait if there's no hurry for a merge.

JorgeCastilloPrz:
True story, it's just an interface xD. Sry for the confusion.

JorgeCastilloPrz:
#1043 is ready, let's rebase master and reuse the CancelToken added with Bracket here. Then ready to merge this one too.

PR #1110, opened 2018-11-12
Reasonable equality for wrapper types

raulraja:
For wrappers that can extend an interface the wrapped value is now private so that it does not leak into user cases.

Something like `ListK<A>` is now equal to `List<A>`.

Other wrappers like `ObservableK` that can't inherit from `Observable` keep their wrapped value public but also implement equality based on delegation to that property.


modules/core/arrow-instances-data/src/main/kotlin/arrow/instances/setk.kt
```diff
@@ -26,7 +26,7 @@ interface SetKEqInstance<A> : Eq<SetK<A>> {
   fun EQ(): Eq<A>
 
   override fun SetK<A>.eqv(b: SetK<A>): Boolean =
-    if (size == b.size) set.map { aa ->
+    if (size == b.size) map { aa ->
```
pakoito: The formatting here ;____;

nomisRev:
@raulraja  @pakoito 
> Other wrappers like ObservableK that can't inherit from Observable

Is that so? Why?

I just tried this and all tests pass.

```kotlin
@higherkind
data class ObservableK<A>(val observable: Observable<A>) : ObservableKOf<A>, ObservableKKindedJ<A>, Observable<A>() {
  override fun subscribeActual(observer: Observer<in A>) =
    observable.subscribe(observer)
```

raulraja:
We can change those wrapper if they can inherit from those base clssdes and providing a concrete impl makes sense. Thoughts ?

raulraja:
@nomisRev in future PRs I'd like to remove all occurrences of the wrappers and work exclusively with their kind `ForListK` where needed in polymorphic context but make all concrete usage based on the wrapper types. Additionally, I would not export the wrappers syntax that we are exporting today so the `listk` package version of exported extensions functions would disappear. This is kind of out of scope for this PR though.

PR #1116, opened 2018-11-12
Optics Snoc

nomisRev:


modules/docs/arrow-docs/docs/docs/optics/snoc/README.MD
```diff
@@ -0,0 +1,34 @@
...
+title: Snoc
+permalink: /docs/optics/snoc/
+---
+
+## Snoc
+
+{:.beginner}
+beginner
+
+`Snoc` provides a [Prism]({{ '/docs/optics/prism' | relative_url }}) between `S` and its init `A` and last element `S`.
```
pakoito: `last element S. It provides a way to attach or detach elements on the end side of a structure.`

pakoito:
Clojure calls it `conj`

nomisRev:
@pakoito updated docs :)

nomisRev:
> Clojure calls it `conj`

Not sure how this is better than `Snoc`. I've come across `Snoc` in Haskell, Scala and even Java but granted it's uncommon.

One argument of `Snoc` over `Conj` would probably be that `Snoc` is `"Cons".reversed()` :D

Something like `Append` and `Prepend` wouldn't really make sense because it's more powerful than that, it can also destructure. I am open for suggestions / opinions.

modules/docs/arrow-docs/docs/docs/optics/snoc/README.MD
```diff
@@ -0,0 +1,53 @@
...
+```kotlin:ank
+import arrow.optics.instances.list.snoc.unsnoc
+
+listOf(1, 2, 3).unsnoc()
+```
+```kotlin:ank
+emptyList<Int>().unsnoc()
+```
+
+### Instances
```
raulraja: This can be automated once https://github.com/arrow-kt/arrow/pull/1119 is in

nomisRev:
> Any papers or original literature around snoc?

@raulraja not that I know off.

So what about the name? Shall we merge this as is yay or nay?

pakoito:
F#Collections going for Conj: http://jackfoxy.com/a-unified-collection-of-purely-functional-list-like-data-structures/

Haskell uses Snoc.

Okay, let's go for Snoc, with a typealias to Conj on the same file :D

raulraja:
I'm fine just fine with snoc but I'm sure it's gonna trip some people but we also have Mu, Nu, Yoneda and friends 😆

PR #1118, opened 2018-11-13
Refined types

### LINKED ISSUE #1099: Arrow-refinement module
We should define what things we want to include in this API and choose the best encoding from there.

The main purpose of this module is offering a set of refined types over primitives like `PositiveInt`, `RegexString`, `NonEmptyString`, etc. And a DSL for easily define new refined types.

We could also include a compiler plugin for checking literals and prevent situations like `-1.positive()` at compile time. 

Some of the possible encodings are:

**Class with private constructor and `invoke` operator in its companion object to check the refinement**

```kotlin
object RefinedPredicateException : IllegalArgumentException("Value does not fulfill predicate")

class NonZeroInt private constructor(val value: Int) {
  companion object {

    private val predicate = { x: Int -> x != 0 }

    operator fun invoke(x: Int): Validated<RefinedPredicateException, NonZeroInt> = 
      if (predicate(x)) NonZeroInt(x).valid() else RefinedPredicateException.invalid()

    operator fun <F> invoke(x: Int, AE: ApplicativeError<F, RefinedPredicateException>): Kind<F, NonZeroInt> = AE.run {
      if (predicate(x)) just(NonZeroInt(x)) else raiseError(RefinedPredicateException)
    }
  }
}

// Usage
val validResult = NonZeroInt(12) // Valid(NonZeroInt(12))
val invalidResult = NonZeroInt(0) // Invalid(e=RefinedPredicateException)

val rightResult = NonZeroInt(12, Either.applicativeError()) // Right(NonZeroInt(12))
```

**Creating a Refined typeclass and then extending the typeclass for every new refined type**

This approach deviates less from the Arrow conventions.

```kotlin
interface Refinement<F, A> {

  fun applicativeError(): ApplicativeError<F, Nel<RefinedPredicateException>>

  fun A.refinement(): Boolean

  fun refine(value: A): Kind<F, A> = applicativeError().run {
    value.refinement()
      .maybe { just(value) }
      .getOrElse { raiseError(Nel.of(RefinedPredicateException)) }
}

interface PositiveInt<F> : Refinement<F, Int> {
  override fun Int.refinement(): Boolean = this >= 0

  fun Int.positive(): Kind<F, Int> = refine(this)
} 

// Usage

val validResult = 0.positive() // Valid(0)
val invalidResult = (-1).positive() //Invalid(e=NonEmptyList(all=RefinedPredicateException))
```

Cotel:
Close #1099 

The goal of this PR is to provide a `Refinement<F, A>` typeclass for creating refined types and also offering a basic set of them.


modules/core/arrow-validation/src/main/kotlin/arrow/validation/refinedTypes/NonZero.kt
```diff
@@ -0,0 +1,47 @@
...
+import arrow.instances.validated.applicativeError.applicativeError
+import arrow.typeclasses.ApplicativeError
+import arrow.validation.RefinedPredicateException
+import arrow.validation.Refinement
+
+interface NonZero<F, A : Number> : Refinement<F, A> {
+
+  override fun A.refinement(): Boolean = this != 0
+
+  fun A.nonZero(): Kind<F, A> = refine(this)
```
Cotel: If I add a method `fun nonZero(a: A): Kind<F, A> = a.nonZero()` the IDE tells me its signature clashes with this method 🤔

Cotel:
I'm trying to create the same refining mechanism as in https://github.com/fthomas/refined but I don't think I'm doing it right or that if that will work for Kotlin 🤔 

I've defined `interface GreaterEqual<F, A> : Not<F, A, Less<F, A>>` and it needs an overriden `Refinement` which `Not` will use to negate the refinement. The problem is that, as you can see in `GreaterEqualTest` you need to pass a `Less` instance when doing `x.greaterEqual()` and I there is no place where I can get that instance.

Besides, `GreaterEqual` in the Scala's refined library is just a typealias and I was expecting to do the same but it seems that is not possible. Maybe it is even better to define this refined type from 0 like `Less` instead of extending `Not`.

modules/core/arrow-validation/src/test/kotlin/arrow/validation/refinedTypes/numeric/GreaterEqualTest.kt
```diff
@@ -0,0 +1,55 @@
...
+          Nel<RefinedPredicateException>> =
+          Validated.applicativeError(Nel.semigroup())
+
+        override fun invalidValueMsg(a: Int): String = Less.errorMsg(a, max)
+      }
+    }
+
+    "Can create GreaterEqual for every number greater or equal than min defined by instance" {
+      forAll(GreaterEqualGen(min)) { x: Int ->
+        x.greaterEqual(LESS(min)).isValid
```
Cotel: I would like to have a function with the same signature as `less(Ord<Int>, max: Int)` instead of passing a `Less` instance 🤷‍♂️

pakoito:
We wanted to do away with arrow-validation soon. Do you mind keeping this open for a few more days while we decide?

Cotel:
Sure there is no hurry 👌

raulraja:
Arrow validation has to change from the current use of Either to be a validation library similar in scope to refined. I think this code can be merged in as is as we evolve and deprecate other functions in Arrow Validation which for now can just be ignored for this development.

Cotel:
I've been trying to implement [Modulo](https://github.com/fthomas/refined/blob/85305c55bedb41edd4a15ca701f4f0a88a581ed5/modules/core/shared/src/main/scala/eu/timepit/refined/numeric.scala#L42) and [Positive](https://github.com/fthomas/refined/blob/85305c55bedb41edd4a15ca701f4f0a88a581ed5/modules/core/shared/src/main/scala/eu/timepit/refined/numeric.scala#L51) but I'm having some troubles.

For `Modulo` I need to have access to the `Int.rem(x: Int): Int` operation but as I'm working with `Modulo<F, A: Number>` Number has not access to this operation. I would need A to be a concrete type but I would have to write lots of instances (ValidatedModuloInt, EitherModuloInt, ValidatedModuloFloat, EitherModuloFloat...)

For `Positive` I can define it like `interface Positive<F, A : Number> : Greater<F, A>` and then override the `fun min(): A` function from `Greater` to always return 0. But it depends on the concrete type again which 0 to return. In Scala's refined library they make use of `_0` from shapeless but I cannot do something similar in Kotlin or at least I don't know how to do it.

For the moment I think I will stick to the refined types defined in https://github.com/gcanti/newtype-ts which I think I can port them all even though they are very few and silly. Besides, I don't like the `101.greaterEqual(Less(100))` which I told in my previous comments so I will remove the `Not` refined type and remake `GreaterEqual` and `LessEqual` to be more like their non-`Equal` types. Once I have them implemented we can work from there.

Tell me your opinions pls.

raulraja:
Sounds like a good path forward, let me know when ready for review.

Cotel:
All the refined types (besides their `Integer` specific version) from `fp-ts` are implemented now. This is ready for review.

Which should be our next step? Should I write some documentation on the usage and the definition of custom refined types or should I try to implement more types from the scala library? I would like to stay in this branch if there is no hurry in finishing this 😄

modules/core/arrow-validation/src/main/kotlin/arrow/validation/refinedTypes/bool/Not.kt
```diff
@@ -0,0 +1,16 @@
...
+import arrow.Kind
+import arrow.validation.Refinement
+
+interface Not<F, B> : Refinement<F, B> {
+
+  fun REF(): Refinement<F, B>
+
+  override fun B.refinement(): Boolean = REF().run { !refinement() }
+
+  fun B.not(): Kind<F, B> = refine(this)
```
raulraja: Should we make these `operator`?

modules/core/arrow-validation/src/main/kotlin/arrow/validation/refinedTypes/numeric/Greater.kt
```diff
@@ -0,0 +1,53 @@
...
+  ORD.run { a.gt(min) }
+
+interface Greater<F, A : Number> : Refinement<F, A> {
+
+  fun ORD(): Order<A>
+  fun min(): A
+
+  override fun A.refinement(): Boolean = isGreaterThan(ORD(), this, min())
+
+  fun A.greater(): Kind<F, A> = refine(this)
```
raulraja: should we make these `operator`?

modules/core/arrow-validation/src/main/kotlin/arrow/validation/refinedTypes/numeric/NonNegative.kt
```diff
@@ -0,0 +1,46 @@
...
+import arrow.instances.validated.applicativeError.applicativeError
+import arrow.typeclasses.ApplicativeError
+import arrow.typeclasses.Order
+import arrow.validation.RefinedPredicateException
+import arrow.validation.Refinement
+
+interface NonNegative<F, A : Number> : Refinement<F, A> {
+  fun ORD(): Order<A>
+
+  override fun A.refinement(): Boolean = !isNegative(ORD(), this)
```
raulraja: We need Kdocs with runnable Ank snippets in all the public defined APIs

modules/core/arrow-validation/src/main/kotlin/arrow/validation/refinedTypes/string/NonEmptyString.kt
```diff
@@ -0,0 +1,39 @@
...
+import arrow.data.ValidatedPartialOf
+import arrow.extension
+import arrow.instances.either.applicativeError.applicativeError
+import arrow.instances.nonemptylist.semigroup.semigroup
+import arrow.instances.validated.applicativeError.applicativeError
+import arrow.typeclasses.ApplicativeError
+import arrow.validation.RefinedPredicateException
+import arrow.validation.Refinement
+
+interface NonEmptyString<F> : Refinement<F, String> {
```
raulraja: Can `NonEmpty<A>` be defined instead of specializing this constraint per type?

modules/core/arrow-validation/src/main/kotlin/arrow/validation/refinedTypes/numeric/NonZero.kt
```diff
@@ -0,0 +1,51 @@
...
+import arrow.instances.validated.applicativeError.applicativeError
+import arrow.typeclasses.ApplicativeError
+import arrow.validation.RefinedPredicateException
+import arrow.validation.Refinement
+
+internal fun <A : Number> isNonZero(a: A): Boolean = a != 0
+
+interface NonZero<F, A : Number> : Refinement<F, A> {
+
+  override fun A.refinement(): Boolean = isNonZero(this)
```
raulraja: All of these are lawful type classes. We should consider including the laws of each one of them or just for `Refinement` for now.
That will give users a framework to test their new refined rules that are composed with multiple refinements. No need to be in this PR but would be great to have.

modules/core/arrow-validation/src/main/kotlin/arrow/validation/refinedTypes/bool/Not.kt
Cotel: Do you mean infix functions? I can overload the not operator which I suppose it would be `!` but I cannot overload `greater` as an operator 🤔

modules/core/arrow-validation/src/main/kotlin/arrow/validation/refinedTypes/string/NonEmptyString.kt
Cotel: Sure, we just have to define an empty value for type `A`

Cotel:
I'm trying to run dokka but it tells me that `import arrow.validation.xxx <- unresolved reference: validation` for every refined type 🤔

Cotel:
I still don't know whats going on with dokka/ank. I have tried to compile the arrow-validation project in the arrow-docs project but then ank won't parse a single file. Could anyone look into this? The branch is almost finished except for the docs.

1Jajen1:
@Cotel 
Ank fails because the import is not correct. It should be `import arrow.validation.refinedTypes.numeric.validated.nonZero.nonZero` (or `import arrow.validation.refinedTypes.numeric.validated.nonZero.*`) instead. A good way to test snippets like this is to create a small class (top level or something to not implicitly import anything) and then actually try to compile and run the snippets.

Cotel:
Soo I've been trying to fix this for some days now but I have no idea of what to do. The only way I can fix the imports is by including `arrow-validation` as a dependency for `arrow-docs` but then Ank won't run on a single file. It fails on the first document it needs to generate. Any ideas ?

raulraja:
@Cotel Can we see the output and command you are using?

Cotel:
I'm doing the same the CI is doing (`clean build dokka :arrow-docs:runAnk)`

And its failing with this

```
> Task :arrow-docs:runAnk
[35m      :::     ::::    ::: :::    :::
    :+: :+:   :+:+:   :+: :+:   :+:
   +:+   +:+  :+:+:+  +:+ +:+  +:+
  +#+     ++: +#+ +:+ +#+ +#++:++
  +#+     +#+ +#+  +#+#+# +#+  +#+
  #+#     #+# #+#   #+#+# #+#   #+#
  ###     ### ###    #### ###    ###[0m
[31m[50%] âœ— /home/travis/build/arrow-kt/arrow/modules/docs/arrow-docs/build/site/docs/effects/effect/README.md [1 of 2][0m
Exception in thread "main" 

"""
import arrow.reflect.*
import arrow.effects.typeclasses.*

TypeClass(Effect::class).dtMarkdownList()
"""

[31mjava.lang.ExceptionInInitializerError
	at arrow.reflect.TypeClassKt.extensions(TypeClass.kt:17)
	at arrow.reflect.TypeClassKt.supportedDataTypes(TypeClass.kt:41)
	at arrow.reflect.ExtensionsHelperKt.dtMarkdownList(ExtensionsHelper.kt:31)
Caused by: java.lang.ClassNotFoundException: arrow.core.Eitherrrow.data.NonEmptyList
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:335)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:264)
	at arrow.reflect.TypeClassExtensionKt.unKind(TypeClassExtension.kt:57)
	at arrow.reflect.TypeClassExtensionKt.asClassTypeExtension(TypeClassExtension.kt:25)
	at arrow.reflect.TypeClassExtensionKt.<clinit>(TypeClassExtension.kt:51)
	at arrow.reflect.TypeClassKt.extensions(TypeClass.kt:17)
	at arrow.reflect.TypeClassKt.supportedDataTypes(TypeClass.kt:41)
	at arrow.reflect.ExtensionsHelperKt.dtMarkdownList(ExtensionsHelper.kt:31)
	at Line_1.<init>(Unknown Source)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at org.jetbrains.kotlin.cli.common.repl.GenericReplEvaluator.eval(GenericReplEvaluator.kt:95)
	at org.jetbrains.kotlin.cli.common.repl.GenericReplCompilingEvaluator.eval(GenericReplCompilingEvaluator.kt:78)
	at org.jetbrains.kotlin.cli.common.repl.GenericReplCompilingEvaluator.compileAndEval(GenericReplCompilingEvaluator.kt:53)
	at org.jetbrains.kotlin.cli.common.repl.ReplAtomicEvalAction$DefaultImpls.compileAndEval$default(ReplApi.kt:168)
	at org.jetbrains.kotlin.cli.common.repl.KotlinJsr223JvmScriptEngineBase.compileAndEval(KotlinJsr223JvmScriptEngineBase.kt:61)
	at org.jetbrains.kotlin.cli.common.repl.KotlinJsr223JvmScriptEngineBase.eval(KotlinJsr223JvmScriptEngineBase.kt:31)
	at javax.script.AbstractScriptEngine.eval(AbstractScriptEngine.java:264)
	at arrow.ank.InterpreterKt$monadDeferInterpreter$1$compileCode$$inlined$run$lambda$1$1.invokeSuspend(interpreter.kt:142)
	at arrow.ank.InterpreterKt$monadDeferInterpreter$1$compileCode$$inlined$run$lambda$1$1.invoke(interpreter.kt)
	at arrow.typeclasses.Monad$binding$wrapReturn$1.invokeSuspend(Monad.kt:56)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:32)
	at kotlin.coroutines.ContinuationKt.startCoroutine(Continuation.kt:127)
	at arrow.typeclasses.Monad$DefaultImpls.binding(Monad.kt:57)
	at arrow.typeclasses.MonadError$DefaultImpls.binding(MonadError.kt)
	at arrow.typeclasses.MonadThrow$DefaultImpls.binding(MonadError.kt)
	at arrow.effects.typeclasses.MonadDefer$DefaultImpls.binding(MonadDefer.kt)
	at arrow.effects.instances.IOMonadDeferInstance$DefaultImpls.binding(io.kt)
	at arrow.effects.instances.io.monadDefer.IOMonadDeferInstanceKt$monadDefer$1.binding(IOMonadDeferInstance.kt:83)
	at arrow.typeclasses.MonadContinuation.binding(MonadContinuations.kt)
	at arrow.ank.InterpreterKt$monadDeferInterpreter$1$compileCode$$inlined$run$lambda$1.invokeSuspend(interpreter.kt:133)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:32)
	at arrow.typeclasses.MonadContinuation$bind$$inlined$suspendCoroutineUninterceptedOrReturn$lambda$1.invoke(MonadContinuations.kt:61)
	at arrow.typeclasses.MonadContinuation$bind$$inlined$suspendCoroutineUninterceptedOrReturn$lambda$1.invoke(MonadContinuations.kt:15)
	at arrow.effects.IO$flatMap$1.invoke(IO.kt:89)
	at arrow.effects.IO$flatMap$1.invoke(IO.kt:34)
	at arrow.effects.IORunLoop.step(IORunLoop.kt:120)
	at arrow.effects.IO.unsafeRunTimed(IO.kt:124)
	at arrow.effects.IO.unsafeRunSync(IO.kt:121)
	at arrow.ank.main.main(main.kt:14)[0m

> Task :arrow-docs:runAnk FAILED
```

raulraja:
@Cotel let's get this merged in once CI passes and continue with more granular features. More action and contributions are coming to `arrow-validated` and we should make sure all this great work is considered. Thoughts?

PR #1125, opened 2018-11-15
Docs improvements: Auto Type class hierarchy graphs

raulraja:
This PR adds a hierarchy graph to each of the type class documents which is driven by `arrow-reflect`, `ank:outFile` new directive and the [nomnoml](http://nomnoml.com/) js library.

Ex. on `MonadCombine`:

monadcombine

1. We use `arrow-reflect` as the source of data where we extract the type class hierarchy. 
2. The doc site provides functions that turn these reflect calls into a nomnoml diagram that looks like the example below.
3. These functions to construct the diagram are invoked by `ank` through `kotlin:ank:outFile(fileName)` directive which replaces the code fence entirely with an empty string and writes the result of the evaluated code into the specified file.
4. The diagram file is stored alongside the page and loaded by the `nomnoml` js library to render the actual diagram on an html canvas.

All processing except the chart rendering happens at build time since it runs in the `ank` process.

Ex: Nomnoml generated code:
```
#font: monoidregular
#arrowSize: 1
#bendSize: 0.3
#direction: down
#gutter: 5
#edgeMargin: 0
#edges: rounded
#fillArrows: false
#fontSize: 10
#leading: 1.25
#lineWidth: 1
#padding: 8
#spacing: 40
#stroke: #485C8A
#title: MonadCombine
#zoom: 1
#.typeclass: fill=#FFFFFF visual=class bold
[<typeclass>MonadCombine|separate|unite]
[<typeclass>MonadFilter]<-[<typeclass>MonadCombine]
[<typeclass>Alternative]<-[<typeclass>MonadCombine]
[<typeclass>Monad]<-[<typeclass>MonadFilter]
[<typeclass>FunctorFilter]<-[<typeclass>MonadFilter]
[<typeclass>Applicative]<-[<typeclass>Monad]
[<typeclass>Functor]<-[<typeclass>Applicative]
[<typeclass>Invariant]<-[<typeclass>Functor]
[<typeclass>Functor]<-[<typeclass>FunctorFilter]
[<typeclass>Applicative]<-[<typeclass>Alternative]
[<typeclass>MonoidK]<-[<typeclass>Alternative]
[<typeclass>SemigroupK]<-[<typeclass>MonoidK]
```

Produces

monadcombine


modules/docs/arrow-docs/src/main/kotlin/ExtensionsHelper.kt
```diff
@@ -1,8 +1,10 @@
 package arrow.reflect
 
+import arrow.aql.Box
 import kotlin.reflect.KClass
+import kotlin.reflect.full.declaredFunctions
 
-private val lineSeparator: String = System.getProperty("line.separator")
+private val lineSeparator: String = "\n" //System.getProperty("line.separator")
```
pakoito: Cleanup!

Issue #333, opened 2017-09-24
`Traverse` docs


raulraja:

vejeta:
I can take this task.

pakoito:
@vejeta PLEASE YES

calvellido:
Please consider it yours though you are not assigned :+1:

PR #1127, opened 2018-11-16
[Issue-333] Documentation for Traverse

vejeta:
This is some documentation for the Traverse Type Class,
as requested in: https://github.com/arrow-kt/arrow/issues/333

It is inspired on: 
https://www.scala-exercises.org/cats/traverse that was also inspired by https://www.enhan.eu/how-to-in-fp/


modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+sealed class SecurityError {
+    data class RuntimeSecurityError(val cause: String) : SecurityError()
+}
+
+interface Credential
+data class Profile(val id: String)
+data class User(val id: String, val name: String)
+```
+
+`parseInt`: is a function that will try to convert an `String` parameter `s` to an `Int`, it if succeeds it will return the number inside the `Some`. If it fails, it will return `None`.
```
JorgeCastilloPrz:
```suggestion
`parseInt`: is a function that will try to convert an `String` parameter `s` to an `Int`, it if succeeds it will return the number inside `Some`. If it fails, it will return `None`.
```

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+```
+
+Next, we will define some data classes and functions to show case different examples of traverse below.
+
+```kotlin:ank:silent
+sealed class SecurityError {
+    data class RuntimeSecurityError(val cause: String) : SecurityError()
+}
+
+interface Credential
```
JorgeCastilloPrz: maybe make it be a `sealed` class and add these two for context?
```kotlin
UserName : Credential()
Password : Credential()
```

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+```
+
+`parseInt`: is a function that will try to convert an `String` parameter `s` to an `Int`, it if succeeds it will return the number inside the `Some`. If it fails, it will return `None`.
+
+ `validateLogin`: This function can be something that could either fail or be a sauccessful operation when validating login credentials */
+
+`userInfo`: This function could be something that could return a Profile asynchronously.
+
+```kotlin:ank:silent
+interface OurSideEffectsFunctions {
```
JorgeCastilloPrz: Maybe?
```suggestion
interface SideEffectingFunctions {
```

Just a suggestion, feel free to ignore

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+
+    fun validateLogin(cred: Credential): Either<SecurityError, Unit>
+
+    fun userInfo(user: User): DeferredK<Profile>
+}
+```
+
+As we can see, every function above only takes as parameter the argument to perform its operation on.
+
+The next objects are just for mocking and simulate results of a side effect performed by an external system so we can see the output of our examples.
```
JorgeCastilloPrz:
```suggestion
This is just a mock implementation of the previous interface, where we simulate results of a side effect performed by an external system. We'll make good use of it later.
```

Sometimes it helps when you're super explicit. I use to think that the person reading this does not have any knowledge about FP yet.

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+
+```kotlin:ank:silent
+object ValidEffects : OurSideEffectsFunctions {
+
+    override fun validateLogin(cred: Credential): Either<SecurityError, Unit> {
+        return Either.right(Unit)
```
JorgeCastilloPrz: I'd even think about adding clarifying comments for dummies. i.e:
```suggestion
return Either.right(Unit) // assuming successful login
```

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+    }
+
+    override fun userInfo(user: User): DeferredK<Profile> {
+        return DeferredK.async { Profile(id = user.id) }
```
JorgeCastilloPrz:
```suggestion
return DeferredK.async { Profile(id = user.id) } // assuming profile details successfully fetched
```

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+    }
+
+    fun savingProfiles(): DeferredK<Unit>  = GlobalScope.async(Dispatchers.Default, CoroutineStart.LAZY){Unit}.k()
```
JorgeCastilloPrz:
```suggestion
fun savingProfiles(): DeferredK<Unit>  = GlobalScope.async(Dispatchers.Default, CoroutineStart.LAZY){Unit}.k() // assuming saving profiles successfully
```

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+
+}
+```
+
+We just defined results for our functions when everything went "right", the next object is also just for simulating the result when something goes "wrong".
```
JorgeCastilloPrz:
```suggestion
We just defined results for our functions for the happy case, or in other words, when everything went "right". The next mock for our `SideEffectingFunctions` simulates the result when something goes "wrong".
```

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+        return Either.left(SecurityError.RuntimeSecurityError("Invalid credentials"))
+    }
+
+    override fun userInfo(user: User): DeferredK<Profile> {
+        return DeferredK.failed(Throwable("Error retrieving profile"))
+    }
+}
+```
+
+If you want to run these examples you can run them inside a main body.
```
JorgeCastilloPrz: I'm not sure what's the intention here, but if you wanna have a runnable version why don't you provide the code? (you're leaving it empty). Not sure whether we should just remove this instead. What's the playgrounds support status @raulraja ?

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+
+```kotlin:ank:silent
+fun main(args: Array<String>) {
+
+}
+```
+
+If we need to extract the information for a List of Users we can create a function that composes with the function defined previously `userInfo`.
```
JorgeCastilloPrz:
```suggestion
If we need to extract the profile information for a List of Users we can create a function that reuses the `userInfo` function defined previously.
```

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+
+```kotlin:ank
+fun profilesFor(users: List<User>): List<DeferredK<Profile>> = users.map { u -> ValidEffects.userInfo(u) }
+```
+
+Notice how we are returning a List of deferred computations. It would be nice for the ones using this function if we could aggregate the results and return the List of Profile under a single DeferredK, something like `DeferredK<List<Profile>>`
+
+To be able to do this transformation we have the `Traverse` type class.
```
JorgeCastilloPrz: This overall thing was a simple and direct example of the need for it. I like it.

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+
+
+`Traverse` is defined with the following signature:
+
+interface Traverse<F> : Functor<F>, Foldable<F> {
```
JorgeCastilloPrz: I can see why you're giving details about how it's implemented here, so you can explain how it should be called. 

I usually don't like to add details about how things are working under the hood, since the user should just be interested on **how to call it** and **what's the role on the library** for the function / feature. But you're basically trying to explain how to call it here.

Could we link to the corresponding higher kinds explanation on arrow docs from here? If the reader is not really aware of what's a Kind yet, he's gonna suffer a bit. He should have an easy way to go there and learn about that before stepping into this.

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+  fun <G, A, B> Kind<F, A>.traverse(AP: Applicative<G>, f: (A) -> Kind<G, B>): Kind<G, Kind<F, B>>
+}
+
+For our example, F would be `List` (the initial container), G would be the data type representing the side effect: `Option`, `Either` or `DeferredK`.
+
+So, if we have a `List<User>` (the ones we want to obtain their profiles) and a function `User -> DeferredK<Profile>`, with traverse we can transform and instead of obtaining a `List<DeferredK, Profile>` it can aggregate all the results to obtain a single `DeferredK<List,Profile>`.
+
+In this case, `traverse` can go over the collection, apply the function and aggregate the resulting values (with side effects) in a `List`.
+
+Basically, `F` is some sort of context which may contain a value. We are using List in the example, but there are `Traverse` implementations for `Option`, `Either` or `Validated`.
```
JorgeCastilloPrz: Can we link the data type docs here also? thanks

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+
+Let's see another example further clarifying this.
```
JorgeCastilloPrz:
```suggest
Let's see another example for further clarification.
```

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+
+```kotlin:ank
+fun parseIntEither(s: String): Either<NumberFormatException, Int> =
+  Either.applicativeError<NumberFormatException>().catch(
```
JorgeCastilloPrz: There's is a leap of faith here. The reader could not know about what's `ApplicativeError` yet and how it works. We should probably clarify that a bit before stepping in. I know many people already knows FP but many people comes to this docs to learn about all this things for the first time. We must always give them means to reach the information they require to avoid getting lost in a point like this one, therefore lower the learning curve as much as possible.

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+    { NumberFormatException("Error converting $s to Int") },
+    { s.toInt() }
+  ).fix()
+
+fun parseIntValidated(s: String): ValidatedNel<NumberFormatException, Int> =
```
JorgeCastilloPrz: should we explain with a one liner what's `ValidatedNel` for and link to its docs?

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+  Validated.fromEither(parseIntEither(s)).toValidatedNel()
+```
+
+We can use these two functions to traverse a collection containing strings, converting them to integers and accumulating the errors with Either or Validated.
```
JorgeCastilloPrz: Maybe I'd move this down, after the following snippet and before "Examples of going through..."

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+listOfValidNumbers.traverse(
+  ValidatedNel.applicative(Nel.semigroup<NumberFormatException>())) { element -> parseIntValidated(element)}
+
+listOfInvalidNumbers.traverse(
+  ValidatedNel.applicative(Nel.semigroup<NumberFormatException>())) { element -> parseIntValidated(element)}
+```
+
+Let's explain what's going on here when traversing a `list` with `Validated`.
+We are using an `Applicative` instance of `ValidatedNel`, and for that we need to provide a "proof" that the non-empty-list (Nel) is a `SemiGroup`.
+The `Applicative` typeclass instance for `ValidatedNel` allows to run independent computations. And the `SemiGroup` typeclass instance allows us to combine elements of the same type, in this case, it helps `ValidatedNel` with the task of accumulating the errors.
```
JorgeCastilloPrz: pls link to `ValidatedNel` and `Semigroup` docs every time you mention them.

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+
+If you want to see other example, you could visit: https://www.enhan.eu/how-to-in-fp/
+
+
+## Sequence
+
+When we want to traverse a collection that each of its elements already contains an effect, for example, List<Option<A>>, we may want to convert it to Option<List<A>> to work easily with the elements. To do that we could traverse the list applying the identity function.
```
JorgeCastilloPrz:
```suggestion
When we want to traverse a collection that each of its elements already contains an effect, for example, List<Option<A>>, we may want to convert it to Option<List<A>> to work easily with the elements. To do that we could traverse the list applying the `::identity` transformation function to each one of the elements.
```

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+
+```kotlin:ank
+val listofOptionalNumbers: List<Option<Int>> = listOf(Option(1), Option(2), Option(3))
+
+listofOptionalNumbers.traverse(Option.applicative(), ::identity)
+```
+
+The equivalent to do a traverse applying identity is using sequence.
```
JorgeCastilloPrz: I think this should be **bold** or highlighted in some way since it's quite important. When I learned that sequence was just syntax for traverse applying identity, my mind clicked.

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+
+```kotlin:ank
+val sequenceOptions = listofOptionalNumbers.sequence(Option.applicative())
+```
+
+We could also use sequence on a list of Either.
```
JorgeCastilloPrz: when you mention functions or types pls always us the highlighted syntax (`sequence`, `Either`...) so you can easily differentiate when we're talking about code.

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+
+```kotlin:ank
+val listOfEither: List<Either<NumberFormatException, Int>> = listOfValidNumbers.map { s -> parseIntEither(s) }
+
+listOfEither.sequence(Either.applicative<NumberFormatException>())
+```
+
+## traverse_ and sequence_
+
+Another usage for sequence is when we are traversing a list of data to which we apply some effectful function and we do not care about the returned values.
```
JorgeCastilloPrz:
```suggestion
Another usage for `sequence` is when we are traversing a list of data to which we apply some effectful function and we do not care about the returned values.
```

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+
+Continuing with our first example, imagine a function saveProfile that perform a side effect, saving a profile in a database, and returning Unit asynchronously.
```
JorgeCastilloPrz:
```suggestion
Continuing with our first example, imagine a function `saveProfile` that perform a side effect, saving a profile in a database, and returning `Unit` asynchronously.
```

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+`Traverse`, also known as `Traversable` is a Typeclass used to perform traversals over an structure with an effect.
 
+We will see some of the use cases for Traverse. Let's start by looking for an example where the side effects are modeled as data types. Side effects in functional programming are changes outside of the scope of the function, for example, performing some IO operation, modifying global variables, etc...
+
+In Kotlin with Arrow, these aforementioned data types can be modeled as `Option` for missing values, `Either` and `Validated` for things that could either provide a right result or give an error, and `IO`, `Async` for asynchronous computations.
+
+
+Let's show some examples that will need the following imports.
+
+```kotlin:ank:silent
```
raulraja: We don't need to ever talk about imports. We should model each of the code snippets as executable with the playgrounds and each one of them will contain the import that it needs in order to operate. See how the AQL docs are done and ask @calvellido @dominv or myself if you have any questions.

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+}
+
+interface Credential
+data class Profile(val id: String)
+data class User(val id: String, val name: String)
+```
+
+`parseInt`: is a function that will try to convert an `String` parameter `s` to an `Int`, it if succeeds it will return the number inside the `Some`. If it fails, it will return `None`.
+
+ `validateLogin`: This function can be something that could either fail or be a sauccessful operation when validating login credentials */
```
raulraja:
```suggestion
 `validateLogin`: This function can be something that could either fail or be a successful operation when validating login credentials */
```

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
raulraja: no need to tell users how to run things, we can make them runnable

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+If we need to extract the information for a List of Users we can create a function that composes with the function defined previously `userInfo`.
+
+```kotlin:ank
+fun profilesFor(users: List<User>): List<DeferredK<Profile>> = users.map { u -> ValidEffects.userInfo(u) }
+```
+
+Notice how we are returning a List of deferred computations. It would be nice for the ones using this function if we could aggregate the results and return the List of Profile under a single DeferredK, something like `DeferredK<List<Profile>>`
+
+To be able to do this transformation we have the `Traverse` type class.
+
```
raulraja: Start new ## section here

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
raulraja: There is no need to prefix `catch`. It is exported for Either as a extension function since 0.8.0. There is rarely the need to ever call `Whatever.typeClass()` any more because all the type class extension functions are exported for all data types that provide instances.

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+Example of what these functions do:
+
+```kotlin:ank
+parseIntEither("1")
+parseIntEither("jimmy")
+```
+
+Examples of going through a list with `map` and with `traverse` for `Either`:
+
+```kotlin:ank
```
raulraja: All these snippets should runnable with playgrounds. See the AQL docs

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+```
+
+Let's explain what's going on here when traversing a `list` with `Validated`.
+We are using an `Applicative` instance of `ValidatedNel`, and for that we need to provide a "proof" that the non-empty-list (Nel) is a `SemiGroup`.
+The `Applicative` typeclass instance for `ValidatedNel` allows to run independent computations. And the `SemiGroup` typeclass instance allows us to combine elements of the same type, in this case, it helps `ValidatedNel` with the task of accumulating the errors.
+
+If you want to see other example, you could visit: https://www.enhan.eu/how-to-in-fp/
+
+
+## Sequence
```
raulraja: `Sequence` is something else. When referring to combinators always use their actual name. In this case `sequence`

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
...
+Let's see what we would get with `traverse_`:
+
+```kotlin:ank
+val l = listOfUsers.traverse_(DeferredK.applicative(), {user -> saveProfile(user)})
+l.fix().unsafeRunSync()
+```
+
+That should return kotlin.Unit
+
+Now, if we have a list already containing effects with results we do not care about: listOfAsyncResults, we can apply `sequence_` to aggregate the result.
```
raulraja: Missing data type table and hierarchy diagram. See docs in other type classes as an example.

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
```diff
@@ -9,8 +9,259 @@ permalink: /docs/typeclasses/traverse/
 {:.intermediate}
 intermediate
 
-TODO. Meanwhile you can find a short description in the [intro to typeclasses]({{ '/docs/typeclasses/intro/' | relative_url }}).
+`Traverse`, also known as `Traversable` is a Typeclass used to perform traversals over an structure with an effect.
 
+We will see some of the use cases for Traverse. Let's start by looking for an example where the side effects are modeled as data types. Side effects in functional programming are changes outside of the scope of the function, for example, performing some IO operation, modifying global variables, etc...
+
+In Kotlin with Arrow, these aforementioned data types can be modeled as `Option` for missing values, `Either` and `Validated` for things that could either provide a right result or give an error, and `IO`, `Async` for asynchronous computations.
```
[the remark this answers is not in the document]
vejeta: Added

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: changed

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: I realize that this part of the examples, Credential and the functions validateLogin are not providing any support to explaining the traverse.

This was used in the original Scala examples and I think `Credential` and `validateLogin` can be safely removed. 

What are your thoughts, @JorgeCastilloPrz ?

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: changed

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: changed

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: changed

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: If we remove the validateLogin function, as I would like, we may not need this.

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: added

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: added

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: changed

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: I will remove this.

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: changed

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: Thanks :)

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: I am still not sure about how to rework this, but following Raul comments below, my intention is reworking the document, and provide first the explanation of the definition briefly and then adding the examples. Otherwise, what are the higher kind explanation you are referring to, @JorgeCastilloPrz ?

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: done

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: I imported the instance (extension function) for catch, so the example should be simpler now.

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: I linked the documentation for Validated. What would you suggest as one-liner to explain it?

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: Links to Validated and Semigroup added

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: Moved

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: changed

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: changed

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: changed

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: changed

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: changed

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: I think we can delete the validateLogin example

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: Deleted this part

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: changed

modules/docs/arrow-docs/docs/docs/typeclasses/traverse/README.md
vejeta: I have seen that they have been added automatically after running the ank task

vejeta:
Since this way of documenting is obsolete, closing this PR.

Issue #1124, opened 2018-11-15
Async.asyncF


JorgeCastilloPrz:
We need to include `asyncF` in `Async` to be able to encode `Concurrent#defaultCancelable`. #1090 depends on this one.

nomisRev:
`Async` should also define `cancelable` since it's derived from `asyncF` and `bracketCase` which are available in `Async`.

This way we can also reduce the restrictions on `Promise`, `MVar`, `Semaphore` to only have 1 implementation based on `Async` instead of 2 based on `Concurrent` & `Async`.

nomisRev:
Closed by #1198

PR #1132, opened 2018-11-16
Concurrent

nomisRev:
- [x] Add `cancelable` method. ~Blocked by #1124~
- [x] Finish laws, requires ~Blocked by #1171 & #1124~
- [x] Update Semaphore to work with Concurrent. ~Blocked by #1124~ 
- [X] Update Promsie to work with Concurrent. ~Blocked by #1124~ 
- [X] Update MVar to work with Concurrent. ~Blocked by #1124~
- [x] Write IO instance

Excluded from scope of this PR.
~~Nice to have~~
~~[ ] Write RxJava instance~~
~~[ ] Write Reactor instance~~
~~[ ] Write Deferred instance~~


JorgeCastilloPrz:
Looks like Concurrent depends on Semaphore and Promise, and Semaphore and Promise depend on Concurrent at the same time. 

Sounds good to me to integrate Semaphore and Promise with uncancellable support using Async, then iterate those to use Concurrent (cancellable) in Concurrent PR.

modules/effects/arrow-effects-instances/src/main/kotlin/arrow/effects/IOStart.kt
```diff
@@ -0,0 +1,49 @@
...
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.startCoroutine
+import kotlin.coroutines.suspendCoroutine
+
+fun <A> IOOf<A>.startF(ctx: CoroutineContext): Kind<ForIO, Fiber<ForIO, A>> = Promise.uncancelable<ForIO, A>(IO.async()).flatMap { promise ->
+  IO.async<Fiber<ForIO, A>> { ioConnection, cb ->
+    val a: suspend () -> A = {
+      suspendCoroutine { ca: Continuation<A> ->
+        this.attempt().unsafeRunSync().fold({ error ->
```
pakoito: As it is now, this blocks the thread this coroutine runs on. Use unsafeRunAsync instead and make the callback call the paths of the continuation.

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt
```diff
@@ -112,9 +113,32 @@ data class DeferredK<out A>(private val deferred: Deferred<A>, val scope: Corout
...
         }.k()
       }
+
+    fun <A, B> racePair(ctx: CoroutineContext,
+                        lh: DeferredKOf<A>,
+                        rh: DeferredKOf<B>): DeferredK<Either<Tuple2<A, Fiber<ForDeferredK, B>>, Tuple2<Fiber<ForDeferredK, A>, B>>> =
+      lh.startF(ctx).flatMap { fiberA ->
+        rh.startF(ctx).flatMap { fiberB ->
+          DeferredK.async<Either<Tuple2<A, Fiber<ForDeferredK, B>>, Tuple2<Fiber<ForDeferredK, A>, B>>> { cb ->
+            fiberA.join.fix().invokeOnCompletion { error ->
```
pakoito: Use use normal operators like `unsafeRunAsync` for this case.

If that's not available I made a new function that's sometimes required to trigger `invokeOnCompletion` for Deferred. Check `forcePropagateExceptions` or something like that.

modules/effects/arrow-effects-instances/src/main/kotlin/arrow/effects/IOStart.kt
nomisRev: This should be fixed now.

nomisRev:
@pakoito @JorgeCastilloPrz I am seeing something really weird and I am unsure what the issue is. I was investigating why the [race laws](https://github.com/arrow-kt/arrow/pull/1132/files#diff-3d061ce877a7468475fa61a0b9fe6b15R62) and when using `IO.never` or `IO.async().never()` these law fail so I made this example but became none the wiser. Any thoughts or directions? 

```kotlin
package problem

import arrow.effects.IO
import arrow.effects.instances.io.concurrent.race
import kotlinx.coroutines.Dispatchers

fun bug(fb: IO<Int>): IO<Unit> = IO.defer {
  val end = 1000

  fun IO<Unit>.loop(i: Int): IO<Unit> =
    if (i == end) this
    else flatMap { _ ->
      race(Dispatchers.IO, IO.just(i), fb)
        .flatMap { either ->
          either.fold({ IO.just(it) }, { IO.raiseError(IllegalStateException("never() finished race")) })
        }.map(::println)
        .loop(i + 1)
    }

  IO.unit.loop(0)
}

fun main(args: Array<String>) =
//  bug(IO.sleep(100).map { -1 }).unsafeRunSync() //runs fine
  bug(IO.never).unsafeRunSync() //hangs around 62 iterations.
```

modules/effects/arrow-effects/src/test/kotlin/arrow/effects/RefTest.kt
```diff
@@ -15,7 +15,7 @@ class RefTest : UnitSpec() {
     with(IO.monadDefer()) {
       "set get" {
         forAll(Gen.int(), Gen.int()) { a, b ->
-          Ref.of(a, IO.monadDefer()).flatMap { ref ->
+          Ref.of(a, arrow.effects.IO.monadDefer()).flatMap { ref ->
```
pakoito: Nit: is this necessary?

modules/streams/arrow-streams/src/main/kotlin/arrow/streams/internal/FreeC.kt
```diff
@@ -69,10 +68,10 @@ sealed class FreeC<F, out R> : FreeCOf<F, R> {
   @Suppress("UNCHECKED_CAST", "ThrowRuntimeException")
   fun asHandler(e: Throwable): FreeC<F, R> = when (val view = ViewL(this.fix())) {
     is FreeC.Pure -> FreeC.Fail(e)
-    is FreeC.Fail -> FreeC.Fail(CompositeFailure(view.error, e))
+    is FreeC.Fail -> FreeC.Fail(arrow.streams.CompositeFailure(view.error, e))
```
pakoito: NIT: this seems like a mistake in a refactor?

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Fiber.kt
```diff
@@ -7,20 +7,20 @@ import arrow.effects.internal.CancelToken
...
    * Returns a new task that will await for the completion of the
-   * underlying fiber, (asynchronously) blocking the current run-loop
+   * underlying Fiber, (asynchronously) blocking the current run-loop
    * until that result is available.
    */
   val join: Kind<F, A>
 
   /**
-   * Triggers the cancellation of the fiber.
+   * Triggers the cancellation of the Fiber.
```
pakoito: Nit: `[Fiber]`

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Fiber.kt
```diff
@@ -7,20 +7,20 @@ import arrow.effects.internal.CancelToken
...
    *
    * @returns a task that trigger the cancellation upon evaluation.
    */
```
pakoito: NIT: Make these functions instead of fields even if they're cached. We don't have field-based APIs and it's harder to generate code for them.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/Semaphore.kt
```diff
@@ -60,7 +60,7 @@ interface Semaphore<F> {
 
   /**
    * Acquires [n] resources
-   * Suspending the fiber running the action until the resources are available.
+   * Suspending the Fiber running the action until the resources are available.
```
pakoito: Nit: `[Fiber]`

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/IOBracket.kt
```diff
@@ -15,7 +15,7 @@ internal object IOBracket {
    * Implementation for `IO.bracketCase`.
    */
   operator fun <A, B> invoke(acquire: IO<A>, release: (A, ExitCase<Throwable>) -> IO<Unit>, use: (A) -> IO<B>): IO<B> =
-    IO.Async { conn, cb ->
+    IO.async { conn, cb ->
```
pakoito: 👍

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
```diff
@@ -0,0 +1,82 @@
...
+  fun <A, B> raceN(ctx: CoroutineContext, fa: Kind<F, A>, fb: Kind<F, B>): Kind<F, Either<A, B>> =
+    racePair(ctx, fa, fb).flatMap {
+      it.fold({ (a, b) ->
+        b.cancel.map { a.left() }
+      }, { (a, b) ->
+        a.cancel.map { b.right() }
+      })
+    }
+
+  fun <A, B, C> raceN(ctx: CoroutineContext, fa: Kind<F, A>, fb: Kind<F, B>, fc: Kind<F, C>): Kind<F, Either<A, Either<B, C>>> =
```
pakoito: _eye twitch_ is this nested Either how it's presented in Scala?

pakoito: Probably not. They have proper `Product`s _sigh_

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
```diff
@@ -0,0 +1,82 @@
...
+
+  fun <A, B, C, D, E, G, H, I, J> raceN(ctx: CoroutineContext, a: Kind<F, A>, b: Kind<F, B>, c: Kind<F, C>, d: Kind<F, D>, e: Kind<F, E>, g: Kind<F, G>, h: Kind<F, H>, i: Kind<F, I>, j: Kind<F, J>): Kind<F, Either<Either<Either<A, Either<B, C>>, Either<D, E>>, Either<Either<G, H>, Either<I, J>>>> =
+    raceN(ctx,
+      raceN(ctx, a, b, c),
+      raceN(ctx, d, e),
+      raceN(ctx, g, h),
+      raceN(ctx, i, j)
+    )
+
+}
```
pakoito: Extra endline! :D

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IOStart.kt
```diff
@@ -0,0 +1,44 @@
+package arrow.effects
+
+import arrow.core.*
+import arrow.effects.internal.*
+import arrow.effects.typeclasses.Fiber
+import kotlin.coroutines.*
+import kotlin.coroutines.Continuation
+
+/**
+ * Create a new [IO] that executes the receiver [IO].
```
pakoito: `Create a new [IO] that eagerly executes the receiver [IO] in the current thread.`

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IOStart.kt
```diff
@@ -0,0 +1,44 @@
...
+      }
+    }
+  }
+
+  a.startCoroutine(asyncContinuation(ctx) { either ->
+    either.fold(
+      { promise.complete(it.left()) },
+      { promise.complete(it.right()) }
+    )
+  })
```
pakoito: 👍 very clean

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IOStart.kt
```diff
@@ -0,0 +1,44 @@
...
+
+  IO.just(Fiber(promise, conn))
```
pakoito: So clean.

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt
```diff
@@ -548,9 +549,32 @@ sealed class DeferredK<A>(
...
         }
       }
+
+    fun <A, B> racePair(ctx: CoroutineContext,
+                        lh: DeferredKOf<A>,
+                        rh: DeferredKOf<B>): DeferredK<Either<Tuple2<A, Fiber<ForDeferredK, B>>, Tuple2<Fiber<ForDeferredK, A>, B>>> =
+      lh.startF(ctx).flatMap { fiberA ->
+        rh.startF(ctx).flatMap { fiberB ->
+          DeferredK.async<Either<Tuple2<A, Fiber<ForDeferredK, B>>, Tuple2<Fiber<ForDeferredK, A>, B>>> { cb ->
+            fiberA.join.unsafeRunAsync { eith ->
```
pakoito: What happens with this pair on cancellation of the parent `DeferredK`? Do we have a test for that? They don't seem connected by sharing a parent `Job()` for example.

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt
```diff
@@ -548,9 +549,32 @@ sealed class DeferredK<A>(
...
+              cb(eith.map { (fiberA toT it).right() })
+            }
+          }
+        }
+      }
+
   }
 }
 
+fun <A> DeferredKOf<A>.startF(ctx: CoroutineContext): DeferredK<Fiber<ForDeferredK, A>> {
```
pakoito: If `startF` works like the `IO` one, this should be 

```
fun <A> DeferredKOf<A>.startF(ctx: CoroutineContext) =
  DeferredK.defer {
    val join = scope().asyncK(ctx = ctx, start = CoroutineStart.DEFAULT) { await() }
    val cancel = DeferredK(start = CoroutineStart.LAZY) { join.cancel() }
    Fiber(join, cancel)
  }
```

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/Promise.kt
```diff
@@ -0,0 +1,104 @@
...
+import arrow.effects.Promise
+import java.util.concurrent.atomic.AtomicReference
+
+fun <A> Promise.Companion.unsafe(): Promise<ForId, A> = UnsafePromise()
+
+internal class UnsafePromise<A> : Promise<ForId, A> {
+
+  private val state: AtomicReference<State<A>> = AtomicReference(State.Pending(emptyList()))
+
+  override val get: Id<A>
```
pakoito: If we're overriding the get, might as well make it if a function. I know `def` acts to define both values and functions in Scala, but here it's better to be explicit.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/Promise.kt
```diff
@@ -0,0 +1,104 @@
...
+        }
+        return if (state.compareAndSet(oldState, newState)) Unit else calculateNewState()
+      }
+
+      calculateNewState()
+      return loop()
+    }
+
+  override val tryGet: Kind<ForId, Option<A>>
+    get() = when (val oldState = state.get()) {
```
pakoito: Same as `get`

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/Promise.kt
```diff
@@ -0,0 +1,104 @@
...
+  override fun tryError(throwable: Throwable): Id<Boolean> = state.get().let { oldState ->
+    when (oldState) {
+      is State.Pending -> throw throwable
+      is State.Full -> Id(false)
+      is State.Error -> Id(false)
+    }
+  }
+
+  override fun complete(a: A): Id<Unit> {
+    tailrec fun calculateNewState(): Unit {
```
pakoito: This creates a new lambda on every execution. Better move it outside and mark it private.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/Promise.kt
```diff
@@ -0,0 +1,104 @@
...
+
+  override val tryGet: Kind<ForId, Option<A>>
+    get() = when (val oldState = state.get()) {
+      is State.Pending<A> -> Id(None)
+      is State.Full -> Id(Some(oldState.value))
+      is State.Error -> Id(None)
+    }
+
+  override fun tryComplete(a: A): Id<Boolean> {
+    tailrec fun calculateNewState(): Unit {
```
pakoito: This creates a new lambda on every execution. Better move it outside and mark it private.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/KindConnection.kt
```diff
@@ -0,0 +1,139 @@
...
+          else current.first()
+      }
+    }
+
+    override fun tryReactivate(): Boolean =
+      state.compareAndSet(null, emptyList())
+  }
+}
+
+internal object CancelUtils2 {
```
pakoito: This shouldn't be necessary. Make the functions internal instead.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/KindConnection.kt
```diff
@@ -0,0 +1,139 @@
...
+    override fun push(token: CancelToken<F>) = Unit
+    override fun pop(): CancelToken<F> = FA.just(Unit)
+    override fun tryReactivate(): Boolean = true
+    override fun pushPair(lh: KindConnection<F>, rh: KindConnection<F>): Unit = Unit
+  }
+
+  /**
+   * Default [IOConnection] implementation.
+   */
+  private class DefaultKindConnection<F>(val AS: Async<F>) : KindConnection<F>() {
```
pakoito: I'm checking and I haven't seen any use of `AS.async`, just `defer`, so why not make it `MonadDefer` instead?

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/KindConnection.kt
```diff
@@ -0,0 +1,139 @@
...
+   * Tries to reset an IOConnection, from a cancelled state, back to a pristine state, but only if possible.
+   *
+   * @return true on success, false if there was a race condition (i.e. the connection wasn't cancelled) or if
+   * the type of the connection cannot be reactivated.
+   */
+  abstract fun tryReactivate(): Boolean
+
+  companion object {
+
+    operator fun <F> invoke(AS: Async<F>): KindConnection<F> = DefaultKindConnection(AS)
```
pakoito: See note about making this MonadDefer instead.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/KindConnection.kt
```diff
@@ -0,0 +1,139 @@
...
+          else -> CancelUtils2.cancelAll(list.iterator(), AS)
+        }
+      }
+    }
+
+    override fun isCanceled(): Boolean = state.get() == null
+
+    override fun push(token: CancelToken<F>): Unit = state.get().let { list ->
+      when (list) {
+//        null -> token.fix().unsafeRunAsync { } TODO("Can this ever occur??? & why do we need to run on push???")
```
pakoito: ????? where does this come from?

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/KindConnection.kt
```diff
@@ -0,0 +1,139 @@
...
+        }
+      }
+    }
+
+    override fun pushPair(lh: KindConnection<F>, rh: KindConnection<F>): Unit =
+      push(CancelUtils2.cancelAll(listOf(lh.cancel(), rh.cancel()).iterator(), AS))
+
+    override fun pop(): CancelToken<F> = state.get().let { current ->
+      when (current) {
+        null, listOf<CancelToken<ForIO>>() -> AS.just(Unit)
```
pakoito: In cancel you use this instead:

```
          list == null || list.isEmpty() -> AS.just(Unit)
```

Worth conciliating them?

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/KindConnection.kt
```diff
@@ -0,0 +1,139 @@
...
+  }
+}
+
+internal object CancelUtils2 {
+
+  fun <F> cancelAll(cursor: Iterator<CancelToken<F>>, AS: Async<F>): CancelToken<F> =
+    if (!cursor.hasNext()) AS.just(Unit)
+    else AS.defer {
+      val frame = CancelAllFrame2(cursor, AS)
+      frame.loop()
```
pakoito: Does this restart the iterator or?

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IORace.kt
```diff
@@ -0,0 +1,114 @@
...
+import arrow.effects.internal.asyncContinuation
+import arrow.effects.internal.unsafe
+import arrow.effects.typeclasses.Fiber
+import java.util.concurrent.atomic.AtomicBoolean
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.startCoroutine
+import kotlin.coroutines.suspendCoroutine
+
+typealias Pair<A, B> = Either<Tuple2<A, Fiber<ForIO, B>>, Tuple2<Fiber<ForIO, A>, B>>
```
pakoito: FiberPair maybe? Something that doesn't collide.

pakoito: I've seen this somewhere else in the review. Maybe in DeferredK. Maybe we need `FiberPair<F, A, B>`?

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IORace.kt
```diff
@@ -0,0 +1,114 @@
...
+
+    // Cancelable connection for the right value
+    val connB = IOConnection()
+    val promiseB = Promise.unsafe<Either<Throwable, B>>()
+
+    conn.pushPair(connA, connB)
+
+    val a: suspend () -> A = {
+      suspendCoroutine { ca: Continuation<A> ->
+        IORunLoop.startCancelable(ioA.fix(), connA) { either: Either<Throwable, A> ->
```
pakoito: If we need to fix it, maybe make IORunLoop take `IOOf`

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IORace.kt
```diff
@@ -0,0 +1,114 @@
...
+            ca.resumeWith(Result.success(b))
+          })
+        }
+      }
+    }
+
+    a.startCoroutine(asyncContinuation(ctx) { either ->
+      either.fold({ error ->
+        if (active.getAndSet(false)) { //if an error finishes first, stop the race.
+          connB.cancel().fix().unsafeRunAsync { r2 ->
```
pakoito: Nit: in the past I've been bitten by `cancel()` not immediately triggering a cancellacion. Is it worth renaming it to something that suggest it's returning an object you have to act one?

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
nomisRev: @pakoito this is how it's currently presented on [IO.Companion](arrow-effects-instances) and is why I choose it. I am not a huge fan of this tho.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IOStart.kt
nomisRev: I would rather make it `Create a new [IO] that upon execution starts the receiver [IO] within a [Fiber].`

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IORace.kt
nomisRev: I'm unsure, to be honest. That's why I liked the property based API because it was giving me a false feeling that things still needed to be run 😅

We do this all over tho. Same for `Fiber` with suggested review changes. To be fair I don't I don't have a strong opinion.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/KindConnection.kt
nomisRev: That snippet comes from [IOConnection](https://github.com/arrow-kt/arrow/blob/a8e198eb346a6aeb86d48df7623b1207676554a5/modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/IOConnection.kt#L92).

I think the new change should be the correct behavior.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
```diff
@@ -24,7 +24,7 @@ object BracketLaws {
       Law("Bracket: bracketCase with just Unit is uncancelable") { BF.bracketCaseWithJustUnitIsUncancelable(EQ) },
       Law("Bracket: bracketCase failure in acquisition remains failure") { BF.bracketCaseFailureInAcquisitionRemainsFailure(EQ) },
       Law("Bracket: bracket is derived from bracketCase") { BF.bracketIsDerivedFromBracketCase(EQ) },
-      Law("Bracket: uncancelable prevents Cancelled case") { BF.uncancelablePreventsCanceledCase(BF.just(Unit), BF.just(Unit), EQ) },
+      Law("Bracket: uncancelable prevents Canceled case") { BF.uncancelablePreventsCanceledCase(BF.just(Unit), BF.just(Unit), EQ) },
```
pakoito: 😠

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Bracket.kt
```diff
@@ -8,9 +8,13 @@ import arrow.typeclasses.MonadError
 
 sealed class ExitCase<out E> {
 
-  object Completed : ExitCase<Nothing>()
+  object Completed : ExitCase<Nothing>() {
+    override fun toString() = "ExitCase.Completed"
+  }
 
-  object Cancelled : ExitCase<Nothing>()
+  object Canceled : ExitCase<Nothing>() {
```
pakoito: 😠

raulraja:
@pakoito `arrow-uk` would make a nice module with a bunch of type-aliases

modules/core/arrow-extras-extensions/src/main/kotlin/arrow/data/extensions/eithert.kt
```diff
@@ -8,8 +8,10 @@ import arrow.core.extensions.either.traverse.traverse
 import arrow.data.*
 import arrow.extension
 import arrow.typeclasses.*
+import arrow.undocumented
 
 @extension
+@undocumented
```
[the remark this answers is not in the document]
nomisRev: You can use this annotation to tell the documentation compiler not to inline the documentation in this because.

This is done because we're currently missing some features to figure out how to currently to inline and adjust code snippets to be valid for monad transformers.

modules/core/arrow-extras-extensions/src/main/kotlin/arrow/data/extensions/eithert.kt
```diff
@@ -142,3 +152,36 @@ fun <F, A, B, G, C> EitherTOf<F, A, B>.traverse(FF: Traverse<F>, GA: Applicative
 
 fun <F, G, A, B> EitherTOf<F, A, Kind<G, B>>.sequence(FF: Traverse<F>, GA: Applicative<G>): Kind<G, EitherT<F, A, B>> =
   traverse(FF, GA, ::identity)
+
+fun <F, L> EitherT.Companion.applicativeError(MF: Monad<F>): ApplicativeError<EitherTPartialOf<F, L>, L> =
```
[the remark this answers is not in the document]
nomisRev: I am using the `@extension` system but an `ApplicativeError` instance for `EitherT` can be derived with both an `ApplicativeError` for `F` and a `Monad`.

So I opted for a `ApplicativeError` constraint in the actual typeclass and to offer an utility function to obtain an `ApplicativeError` instance by providing a monad.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -131,10 +136,9 @@ sealed class IO<out A> : IOOf<A> {
 
   internal abstract fun unsafeRunTimedTotal(limit: Duration): Option<A>
 
-  /**
-   * Makes the source `IO` uninterruptible such that a [Fiber.cancel] signal has no effect.
-   */
-  fun uncancelable(): IO<A> = IOCancel.uncancelable(this)
+  /** Makes the source [IO] uncancelable such that a [Fiber.cancel] signal has no effect. */
+  fun uncancelable(): IO<A> =
+    IO.ContextSwitch(this, ContextSwitch.makeUncancelable, ContextSwitch.disableUncancelable())
```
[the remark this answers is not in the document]
nomisRev: It was just a small util file with only code related to `IO.ContextSwitch` so I moved that code to `IO.ContextSwitch` directly.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/internal/CancelablePromise.kt
```diff
@@ -0,0 +1,126 @@
...
+
+  internal sealed class State<out A> {
+    data class Pending<A>(val joiners: Map<Token, (Either<Throwable, A>) -> Unit>) : State<A>()
+    data class Full<A>(val value: A) : State<A>()
+    data class Error<A>(val throwable: Throwable) : State<A>()
+  }
+
+  private val state: AtomicReference<State<A>> = AtomicReference(State.Pending(emptyMap()))
+
+  override fun get(): Kind<F, A> = defer {
```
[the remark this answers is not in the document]
nomisRev: It is and it does. However you can cancel that operation, i.e. when it never completes, and when it does cancel we need to be able to unregister the callback. Hence a `Cancelable` version.

However I think I will convert these naming back to `Concurrent` & `Async`. And we need documentation that explains the cancellation is not wired correctly when using `Async` instances of Promise.

I think that'll be much more clear than `Cancelable` and `Uncancelable`.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/internal/CancelablePromise.kt
```diff
@@ -0,0 +1,126 @@
...
+      is State.Full -> just(current.value)
+      is State.Pending -> cancelable { cb ->
+        val id = unsafeRegister(cb)
+        delay { unregister(id) }
+      }
+      is State.Error -> raiseError(current.throwable)
+    }
+  }
+
+  override fun tryGet(): Kind<F, Option<A>> = delay {
```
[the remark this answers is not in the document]
nomisRev: It's a safe operation since it returns `Option`. Documentation is not provided in the internal implementations, it's explained in the `Promise` interface.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/internal/ImmediateContext.kt
```diff
@@ -0,0 +1,13 @@
+package arrow.effects.internal
+
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.ContinuationInterceptor
+import kotlin.coroutines.CoroutineContext
+
+internal object ImmediateContext : ContinuationInterceptor {
```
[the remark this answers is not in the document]
nomisRev: We can't use that immediate dispatcher since it lives in `coroutines-core`. However, I am removing this as we speak in favor of `EmptyCoroutineContext`. That's the only adjustment I am still making in this PR because `ImmediateContext` shouldn't get merged :D 

This is currently only used to start some internal code that should be blocking. See [CancelableMVar](https://github.com/arrow-kt/arrow/pull/1132/files#diff-46b3d40c91597898c01b1d177e803b25R221) or [CancelablePromise](https://github.com/arrow-kt/arrow/pull/1132/files#diff-cf09624ff9e2cce5f9f4ee032a495cf7R92).

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/internal/UncancelablePromise.kt
```diff
@@ -0,0 +1,95 @@
...
+import arrow.core.*
+import arrow.effects.Promise
+import arrow.effects.typeclasses.Async
+import java.util.concurrent.atomic.AtomicReference
+
+internal class UncancelablePromise<F, A>(AS: Async<F>) : Promise<F, A>, Async<F> by AS {
+
+  internal sealed class State<out A> {
+    data class Pending<A>(val joiners: List<(Either<Throwable, A>) -> Unit>) : State<A>()
+    data class Full<A>(val value: A) : State<A>()
```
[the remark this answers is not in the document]
nomisRev: I was trying to go for a different AST naming altogether. Something that lies more in line with `Promise`, something like `Fulfilled` and `Unfulfilled` but I reverted back during the PR somewhere I think :D

Naming here is for our own benefit mostly since the state is internal to the internal implementations. I'll rename both to `Complete` to be more in line with the APIs.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
nomisRev: @pakoito I made an attempt to make this nicer. Check the aliases at the bottom of the file. Thoughts?

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt
nomisRev: This is removed and postponed to become a separate ticket.

modules/core/arrow-extras/src/test/kotlin/arrow/data/EitherTTest.kt
```diff
@@ -25,24 +28,20 @@ import org.junit.runner.RunWith
...
 class EitherTTest : UnitSpec() {
 
-  private fun IOEQ(): Eq<Kind<EitherTPartialOf<ForIO, Throwable>, Int>> = Eq { a, b ->
-    a.value().attempt().unsafeRunSync() == b.value().attempt().unsafeRunSync()
-  }
-
-  private fun IOEitherEQ(): Eq<Kind<EitherTPartialOf<ForIO, Throwable>, Either<Throwable, Int>>> = Eq { a, b ->
-    a.value().attempt().unsafeRunSync() == b.value().attempt().unsafeRunSync()
+  fun <A> EQ(): Eq<Kind<EitherTPartialOf<ForIO, Throwable>, A>> = Eq { a, b ->
+    a.value().attempt().unsafeRunTimed(60.seconds) == b.value().attempt().unsafeRunTimed(60.seconds)
```
pakoito: 👍

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/Semaphore.kt
```diff
@@ -195,11 +184,30 @@ interface Semaphore<F> {
...
+     * import arrow.effects.extensions.io.concurrent.concurrent
+     *
+     * fun main(args: Array<String>) {
+     *   //sampleStart
+     *   val semaphore = Semaphore<ForIO>(5, IO.concurrent())
+     *   //sampleEnd
+     * }
+     */
+    operator fun <F> invoke(n: Long, CF: Concurrent<F>): Kind<F, Semaphore<F>> = CF.run {
+      assertNonNegative(n).flatMap {
```
pakoito: Could we use ULong here? Is that a thing?

modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/predef.kt
```diff
@@ -0,0 +1,14 @@
+package arrow.effects
+
+import arrow.Kind
+import arrow.core.extensions.either.eq.eq
+import arrow.core.extensions.option.eq.eq
+import arrow.effects.typeclasses.Duration
+import arrow.effects.typeclasses.seconds
+import arrow.typeclasses.Eq
+
+fun <A> EQ(EQA: Eq<A> = Eq.any(), timeout: Duration = 60.seconds): Eq<Kind<ForIO, A>> = Eq { a, b ->
```
pakoito: We have at least 2 uses of this that I've seen during the review. We should replace those too in another diff.

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/observablek.kt
```diff
@@ -10,6 +10,14 @@ import arrow.effects.rx2.extensions.observablek.monad.monad
 import arrow.effects.rx2.extensions.observablek.monadError.monadError
 import arrow.effects.rx2.fix
 import arrow.effects.typeclasses.*
+import arrow.effects.typeclasses.Async
```
pakoito: Are these necessary when we already have `import arrow.effects.typeclasses.*` in line 12?

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
```diff
@@ -0,0 +1,660 @@
...
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+import io.kotlintest.properties.map
+import kotlinx.coroutines.Dispatchers
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicReference
+import kotlin.coroutines.CoroutineContext
+
+@Suppress("LargeClass")
```
pakoito: lol

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -56,7 +55,13 @@ sealed class IO<out A> : IOOf<A> {
             IO { callback(Left(t)) }
           }
 
-          IORunLoop.startCancelable(fa, conn2, mapUnit)
+          //DEV: If fa cancels conn2 like so `conn.cancel().map { cb(Right(Unit)) }`
```
pakoito: nit: space after //

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -56,7 +55,13 @@ sealed class IO<out A> : IOOf<A> {
...
+          //It doesn't run the stack of conn2, instead the result is seen it cb of startCancelable.
+          IORunLoop.startCancelable(fa, conn2) {
+
+            if (it.fold({ e -> e == OnCancel.CancellationException }, { false }) && conn.isNotCanceled()) IORunLoop.start(conn.cancel(), mapUnit)
```
pakoito: nit: take the fold into boolean in a new variable. Put the comment on that variable instead

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
JorgeCastilloPrz: That sort of things should be reported by detekt / ktlint if we care enough

PR #1136, opened 2018-11-18
Deprecate `invoke`, add a `delay` function that does the same thing

### LINKED ISSUE #1135: Replace `invoke` for `delay` in `MonadDefer`
The current `invoke` operator is confusing because it requires to use the type class as if it was creating new instances of it when really this combinator is known elsewhere as `delay`.

LordRaydenMK:
Fixes #1135

the `delay` has the same implementation as the `invoke` operator.
I added `@Deprecated` with `@ReplaceWith` to the `invoke` operator. 
I replaced all usages of `invoke` with `delay` (that I could find).

Is that all? Or am I missing something?

Note:
When I run the tests with `./gradlew test` (both before and after this change) a test inside `IOTests` fails. If I run trough IntelliJ only for that class, everything works fine. Is this a know issue? Should I report a bug?

### DIFF
--- a/modules/docs/arrow-examples/src/test/kotlin/arrow/FpToTheMax.kt
+++ b/modules/docs/arrow-examples/src/test/kotlin/arrow/FpToTheMax.kt
@@ -22,8 +22,8 @@ interface Console<F> {
 }
 
 class ConsoleInstance<F>(val delay: MonadDefer<F>) : Console<F> {
-    override fun putStrLn(s: String): Kind<F, Unit> = delay { println(s) }
-    override fun getStrLn(): Kind<F, String> = delay { readLine().orEmpty() }
+    override fun putStrLn(s: String): Kind<F, Unit> = delay.delay { println(s) }
+    override fun getStrLn(): Kind<F, String> = delay.delay { readLine().orEmpty() }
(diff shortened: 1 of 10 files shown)


modules/docs/arrow-examples/src/test/kotlin/arrow/FpToTheMax.kt
```diff
@@ -22,16 +22,16 @@ interface Console<F> {
 }
 
 class ConsoleInstance<F>(val delay: MonadDefer<F>) : Console<F> {
-    override fun putStrLn(s: String): Kind<F, Unit> = delay { println(s) }
-    override fun getStrLn(): Kind<F, String> = delay { readLine().orEmpty() }
+    override fun putStrLn(s: String): Kind<F, Unit> = delay.delay { println(s) }
+    override fun getStrLn(): Kind<F, String> = delay.delay { readLine().orEmpty() }
```
raulraja: No need to call the instance name delay anymore

LordRaydenMK: yeah... but how would you name it?

PR #1138, opened 2018-11-19
Implement hash typeclass and its interfaces

1Jajen1:
WIP because its missing docs and I have some open questions.

- Some datatypes (specifically ListK/SetK etc.) override hashCode to use kotlins Any.hashCode. Should they instead point to their instance of hash() and if so should the parameter if needed be substituted with `Hash.any()`?
- Are there any datatypes I have missed? I skipped some where I was unsure (Coproduct, Sum etc) so please take a look there :)
- As for laws: Hashes only really need to be consistent (`hash(a) == hash(a)`) and hold the law if `a == b` => `hash(a) == hash(b)`. Is that enough?

@pakoito This should be ready some docs I hope :see_no_evil: 

Solves #687

### DIFF
--- a/modules/core/arrow-instances-data/src/main/kotlin/arrow/instances/nonemptylist.kt
+++ b/modules/core/arrow-instances-data/src/main/kotlin/arrow/instances/nonemptylist.kt
@@ -139,6 +139,17 @@ interface NonEmptyListSemigroupKInstance : SemigroupK<ForNonEmptyList> {
     fix().nelCombineK(y)
 }
 
+@extension
+interface NonEmptyListHashInstance<A> : Hash<NonEmptyList<A>>, NonEmptyListEqInstance<A> {
+  fun HA(): Hash<A>
+
+  override fun EQ(): Eq<A> = HA()
+
+  override fun NonEmptyList<A>.hash(): Int = foldLeft(1) { hash, a ->
+    31 * hash + HA().run { a.hash() }
+  }
+}
+
 fun <F, A> Reducible<F>.toNonEmptyList(fa: Kind<F, A>): NonEmptyList<A> =
   fa.reduceRightTo({ a -> NonEmptyList.of(a) }, { a, lnel ->
     lnel.map { nonEmptyList -> NonEmptyList(a, listOf(nonEmptyList.head) + nonEmptyList.tail) }
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/HashLaws.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/HashLaws.kt
@@ -8,7 +8,8 @@ import io.kotlintest.properties.forAll
 object HashLaws {
   fun <F> laws(HF: Hash<F>, EQ: Eq<F>, cf: (Int) -> F): List<Law> =
     listOf(
-      Law("Hash Laws: Equality implies equal hash") { equalHash(HF, EQ, cf) }
+      Law("Hash Laws: Equality implies equal hash") { equalHash(HF, EQ, cf) },
+      Law("Hash Laws: Multiple calls to hash should result in the same hash") { equalHashM(HF, cf) }
     )
 
   fun <F> equalHash(HF: Hash<F>, EQ: Eq<F>, cf: (Int) -> F): Unit {
@@ -16,6 +17,13 @@ object HashLaws {
       val a = cf(f)
       val b = cf(f)
       EQ.run { a.eqv(b) } && HF.run { a.hash() == b.hash() }
+    }
+  }
+  
+  fun <F> equalHashM(HF: Hash<F>, cf: (Int) -> F): Unit {
+    forAll(Gen.int()) { f ->
+      val a = cf(f)
+      HF.run { a.hash() == a.hash() }
     }
   }
 }
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/HashLaws.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/HashLaws.kt
@@ -1,0 +1,21 @@
+package arrow.test.laws
+
+import arrow.typeclasses.Eq
+import arrow.typeclasses.Hash
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+
+object HashLaws {
+  fun <F> laws(HF: Hash<F>, EQ: Eq<F>, cf: (Int) -> F): List<Law> =
+    listOf(
+      Law("Hash Laws: Equality implies equal hash") { equalHash(HF, EQ, cf) }
+    )
+
+  fun <F> equalHash(HF: Hash<F>, EQ: Eq<F>, cf: (Int) -> F): Unit {
+    forAll(Gen.int()) { f ->
+      val a = cf(f)
+      val b = cf(f)
+      EQ.run { a.eqv(b) } && HF.run { a.hash() == b.hash() }
+    }
+  }
+}
--- a/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Hash.kt
+++ b/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Hash.kt
@@ -1,0 +1,24 @@
+package arrow.typeclasses
+
+interface Hash<in F> : Eq<F> {
+
+  fun F.hash(): Int
+
+  companion object {
+
+    inline operator fun <F> invoke(crossinline hashF: (F) -> Int): Hash<F> = object : Hash<F> {
+      override fun F.hash(): Int = hashF(this)
+
+      // either this.hash == hash, or this == b or supply an eq function, dunno what is best
+      override fun F.eqv(b: F): Boolean = hashF(b) == hashF(this)
+    }
+
+    fun any(): Hash<Any> = HashAny
+
+    private object HashAny : Hash<Any> {
+      override fun Any.hash(): Int = hashCode()
+
+      override fun Any.eqv(b: Any): Boolean = Eq.any().run { eqv(b) }
+    }
+  }
+}
--- a/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Hash.kt
+++ b/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Hash.kt
@@ -1,24 +1,45 @@
 package arrow.typeclasses
 
+/**
+ * A type class used to represent hashing for objects of type [F]
+ *
+ * @see // Fancy link to docs
+ */
 interface Hash<in F> : Eq<F> {
 
+  /**
+   * Produces a hash for an object of type [F].
+   *
+   * @receiver The object to hash
+   * @returns an int representing the object hash
+   */
   fun F.hash(): Int
 
   companion object {
 
+    /**
+     * Construct an instance of [Hash] from a function `(F) -> Int`.
+     *
+     * @param hashF function that computes a hash for any object of type [F]
+     * @returns an instance of [Hash] that is defined by the hashF function
+     */
     inline operator fun <F> invoke(crossinline hashF: (F) -> Int): Hash<F> = object : Hash<F> {
       override fun F.hash(): Int = hashF(this)
 
-      // either this.hash == hash, or this == b or supply an eq function, dunno what is best
-      override fun F.eqv(b: F): Boolean = hashF(b) == hashF(this)
+      override fun F.eqv(b: F): Boolean = this == b
     }
 
-    fun any(): Hash<Any> = HashAny
+    /**
+     * Retrieve an instance of [Hash] where the hash function delegates to kotlin's `Any?.hashCode()` function
+     *
+     * @returns an instance of [Hash] that always delegates to kotlin's native hashCode functionality
+     */
+    fun any(): Hash<Any?> = HashAny
 
-    private object HashAny : Hash<Any> {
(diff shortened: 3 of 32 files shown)


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/HashLaws.kt
```diff
@@ -0,0 +1,29 @@
+package arrow.test.laws
+
+import arrow.typeclasses.Eq
+import arrow.typeclasses.Hash
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+
+object HashLaws {
+  fun <F> laws(HF: Hash<F>, EQ: Eq<F>, cf: (Int) -> F): List<Law> =
+    listOf(
```
pakoito: Because Hash inherits from Eq then this implies that:

EqLaws.laws(...) +  listOf(...

This way you can test both Eq and Hash instances in a single call, replacing the current EqLaws call for HashLaws in the existing tests.

pakoito:
>Some datatypes (specifically ListK/SetK etc.) override hashCode to use kotlins Any.hashCode. Should they instead point to their instance of hash() and if so should the parameter if needed be substituted with Hash.any()?

The idea is that the instance delegates to the original implementation, which in this case would be the one of the types they're wrapping: List, Set, Sequence, etc...

> Are there any datatypes I have missed?

Per Java's indication, anything with `equals` qualifies for `hashCode` too. You don't have to do all of them in one go, but do list the missing ones in a new Issue.

> Laws
The hashLaws in Scala are here: https://github.com/typelevel/cats/blob/master/kernel-laws/src/main/scala/cats/kernel/laws/HashLaws.scala. 

IMO given that we don't have universal hash calculation we should be good with the two that you added :)

See the comment I've added, to make sure that Hash instances also imply correct Eq instances.

modules/docs/arrow-docs/docs/_data/menu.yml
```diff
@@ -234,6 +234,9 @@ options:
       - title: Alternative
         url: /docs/typeclasses/alternative/
 
+      - title: Hash
```
pakoito: I'd put it below `Eq` instead, to simplify that's simple, it inherits from it, and this may be somewhat common anyway :D

modules/docs/arrow-docs/docs/docs/typeclasses/hash/README.md
```diff
@@ -0,0 +1,100 @@
...
+It also might make more sense to provide your own instance in for some cases, for example when there is already a unique property on an object.
+Hash has a constructor to create a `Hash` instance from any function `(F) -> Int`.
+
+```kotlin:ank
+import arrow.typeclasses.Hash
+data class User(val id: String, val name: String)
+val user = User("MyId", "MyName")
+
+// This is fine
+Hash.any().run { user.hash() }
```
pakoito: Split this snippet into 2 ank snippets at this line, so there's an output from this line too.

modules/docs/arrow-docs/docs/docs/typeclasses/hash/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+// This might be better because id usually is a unique value in itself
+val userHash = Hash<User> { u -> u.id.hashCode() }
+userHash.run { user.hash() }
+```
+
+```kotlin:ank
+import arrow.core.*
+import arrow.typeclasses.Hash
```
pakoito: Imports are per file, not per snippet, so this one is unnecessary.

modules/docs/arrow-docs/docs/docs/typeclasses/hash/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+// This will fail because it's not evaluated for hashing
+Hash.any().run { Eval.later { 1 }.hash() == Eval.later { 1 }.hash() }
+```
+
+```kotlin:ank
+import arrow.typeclasses.Hash
```
pakoito: Same duped import here.

modules/docs/arrow-docs/docs/docs/typeclasses/hash/README.md
```diff
@@ -0,0 +1,100 @@
...
+val stringHash = Hash<String> { a -> a.hashCode() }
+```
+
+See [Deriving and creating custom typeclass]({{ '/docs/patterns/glossary' | relative_url }}) to provide your own `Hash` instances for custom datatypes.
+
+### Data types
+
+```kotlin:ank:replace
+import arrow.reflect.*
+import arrow.typeclasses.Hash
```
pakoito: This one may not be necessary either.

modules/core/arrow-instances-data/src/main/kotlin/arrow/instances/nonemptylist.kt
```diff
@@ -139,6 +139,17 @@ interface NonEmptyListSemigroupKInstance : SemigroupK<ForNonEmptyList> {
...
 }
 
+@extension
+interface NonEmptyListHashInstance<A> : Hash<NonEmptyList<A>>, NonEmptyListEqInstance<A> {
+  fun HA(): Hash<A>
+
+  override fun EQ(): Eq<A> = HA()
+
+  override fun NonEmptyList<A>.hash(): Int = foldLeft(1) { hash, a ->
+    31 * hash + HA().run { a.hash() }
```
pakoito: @raulraja any opinions about this one?

1Jajen1:
I know :) You are too fast :D Was about to push the fix for that. I also added instances for Tuple, Coproduct (derived from either) and sum. And added the tests for them as well. If they are not correct or should not exist I can easily revert them, so please let me know :)

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/HashLaws.kt
```diff
@@ -0,0 +1,30 @@
+package arrow.test.laws
+
+import arrow.typeclasses.Eq
+import arrow.typeclasses.Hash
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+
+object HashLaws {
+  fun <F> laws(HF: Hash<F>, EQ: Eq<F>, cf: (Int) -> F): List<Law> =
```
pakoito: Nit: empty line here.

modules/docs/arrow-docs/docs/docs/typeclasses/hash/README.md
```diff
@@ -51,6 +51,11 @@ val user = User("MyId", "MyName")
 
 // This is fine
 Hash.any().run { user.hash() }
+```
+
+```kotlin:ank
+data class User(val id: String, val name: String)
```
pakoito: Same as with the imports, there's no need to redefine the class either!

pakoito:
If you'd like to run detekt to find errors like the one in CI, you have to do it on a clean build:

`./gradlew clean detekt`

```
UnusedImports - [option.kt] at modules/core/arrow-instances-core/src/main/kotlin/arrow/instances/option.kt:7:1
```

1Jajen1:
Yeah, will do from now on. Takes a while to get used to a different workflow

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Hash.kt
```diff
@@ -0,0 +1,45 @@
+package arrow.typeclasses
+
+/**
+ * A type class used to represent hashing for objects of type [F]
```
raulraja: We should probably explain what a Hash is here

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Hash.kt
```diff
@@ -0,0 +1,45 @@
...
+ */
+interface Hash<in F> : Eq<F> {
+
+  /**
+   * Produces a hash for an object of type [F].
+   *
+   * @receiver The object to hash
+   * @returns an int representing the object hash
+   */
+  fun F.hash(): Int
```
raulraja: Use kdoc embedded ank snippets to show each one of the public api methods with examples. In the next version of Arrow these will show up in the browser as runnable:
```
/**
   * Commented method or class
   * 
   * {: data-executable='true'}
   *
   * ```kotlin:ank
   * import arrow.aql.instances.list.select.*
   * import arrow.aql.instances.listk.select.select
   *
   * fun main(args: Array<String>) {
   *   //sampleStart
   *   val result: List<Int> =
   *     listOf(1, 2, 3).query {
   *       select { this + 1 }
   *     }.value()
   *  //sampleEnd
   *  println(result)
   *  }
   *  ```
   */
```

modules/docs/arrow-docs/docs/docs/typeclasses/hash/README.md
```diff
@@ -0,0 +1,95 @@
...
+```kotlin:ank
+// This might be better because id usually is a unique value in itself
+val userHash = Hash<User> { u -> u.id.hashCode() }
+userHash.run { user.hash() }
+```
+
+```kotlin:ank
+import arrow.core.*
+
+// This will fail because it's not evaluated for hashing
```
raulraja:
```suggestion
// This will return `false` because it's not evaluated for hashing
```
Because if this really fails blowing up the stack we have something wrong in our code.

modules/docs/arrow-docs/docs/docs/typeclasses/hash/README.md
raulraja: The ones that are playable which if possible should be most of the ones used to document API require all the context in which they run in them

1Jajen1: Ok I'll change that then. But that only goes for context as in classes and variables, not imports right?

raulraja: See the `arrow-query-language` docs for examples. You have to include everything in the snippet that it needs to run. All imports, a main method and a couple of comments that delimit the code you are not collapsing in the snippet when presented on the site.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Hash.kt
raulraja: What a hash value is and what they are useful for. A simple introductory paragraph should be enough IMO.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Hash.kt
raulraja: I ment actually in the sources too. The example will then show up in the API docs beside the docs page.

1Jajen1: Yeah, that confused me a bit :) 
Added it now for `F.hash()`, `Hash.invoke()` and `Hash.any()`

modules/docs/arrow-docs/docs/docs/typeclasses/hash/README.md
1Jajen1: Done (I hope) :)

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Hash.kt
1Jajen1: Added something. Not sure if enough, or good. :see_no_evil:

pakoito:
@1Jajen1 is this good to go?

1Jajen1:
If the docs are good then yes. They compile and look alright, but I feel like with the runnable docs you never know (or is there a good way to test them locally with new features)

pakoito:
Good for me :D Thanks for your first contribution, I hope it was interesting! If you'd like to take other tickets just let us know.

1Jajen1:
It definitly was. It was something new, I learned a lot and it'll definitly change how I set up my projects in the future. Lots of new tools that feel so much better than what I usually use at work (which is no wonder considering how old codebases tend to work :/). I'll definitly continue, work and university are not that busy atm. The more to learn the better :)

raulraja:
There seems to be something wrong with this function. After it has been introduced it always bails with the same error https://api.travis-ci.org/v3/job/457554572/log.txt

1Jajen1:
An exact copy of this runs fine locally (without ank). Thats the example from the source right? Have you ever encountered errors like this before? To me it seems that the way ank runs this somehow breaks it. I can't imagine that snippet breaking memory limits, as its literally just creating an object of Hash<String> and running its extension function. The created object is almost the same as the String.hash() instance (except for the eq part) so I'd suspect ank, but then again I have no idea how it works internally so I might be wrong :confused:

1Jajen1:
So correct me if I am wrong: The way ank works top level is =
- create target dir
- get list of all relevant files
- read all those files
- preprocess all files
- magic
- output

Is there any reason to do those sequential, having all files and their content loaded in memory, as opposed to doing this whole workflow for each file one by one?

pakoito:
It may be because it's importing all instances, instead of just the necessary ones? There's a lot of code there.

1Jajen1:
While it might make the problem worse I don't think its the cause. There is a lot of code in the docs that uses wildcard imports, that has previously worked fine. (The docs for equal use exactly the same import)

raulraja:
@1Jajen1 It might be very well a problem with Ank. I wasn't built to process that many api docs files. I'm pretty sure that is the issue. Forget about this and I'll try to fix it in that branch before pushing to master

PR #1139, opened 2018-11-19
Add Λnk README

Guardiola31337:
- Adds Λnk `README`

### DIFF
--- a/modules/ank/README.md
+++ b/modules/ank/README.md
@@ -1,0 +1,73 @@
+# Λnk
+
+_Compile time docs verification and evaluation for Kotlin and Java._
+
+_Λnk_ short for [Ankhesenpaaten](https://en.wikipedia.org/wiki/Ankhesenamun) wife and sister of [Tutankhamun](https://en.wikipedia.org/wiki/Tutankhamun), is a [Gradle](https://gradle.org/) plugin to verify code snippets in library docs for the Kotlin and Java programming languages.
+
+Λnk is inspired by the awesome docs and tutorial generator for Scala [`tut`](http://tpolecat.github.io/tut/).
+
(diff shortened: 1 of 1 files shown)


pakoito:
Amazing job :D

modules/ank/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+```
+./gradlew :module-name:runAnk
+```
+
+This will process all `*.md` files in `source`, write them to `target`, while providing all runtime dependencies (`sourceSets.main.runtimeClasspath`) in the `classpath`.
+
+## Errors output
+
+When something goes wrong Λnk shows what snippet has failed and where, including the compiler errors.
```
raulraja: This should probably show an example with a snippet that fails and how ank reports it.

modules/ank/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Modifiers
+
+By default Λnk compiles and evaluates the snippets of code included in `ank` sheds. However sometimes you might want a definition without printing the result out, or might want to replace an entire snippet with the output of the resulting evaluation. For these occasions Λnk provides a number of modifiers that you can add to the shed declaration.
```
raulraja: I would use ````kotlin:ank:*` fences instead of just `ank`

modules/ank/README.md
```diff
@@ -0,0 +1,73 @@
...
+The language used (Kotlin or Java) should be prepended to `:ank` e.g. `kotlin:ank:replace`. The following modifiers are supported.
+
+| Modifier | Explanation |
+|---|---|
+| `:silent` | Suppresses output; under this modifier the input and output text are identical. |
```
raulraja: We should provide an example for each one of these.

PR #1140, opened 2018-11-19
Fixes StackOverFlow issue in MonoidK typeclass

### LINKED ISSUE #1126: fold option monoidk algebra stackoverflow
Running: `listOf(Option(1)).fold(Option.monoidK().algebra())`

Produces: `Exception in thread "main" java.lang.StackOverflowError
	at arrow.typeclasses.MonoidK$algebra$1.empty(MonoidK.kt:20)`

franciscodr:
This pull request fixes the `StackOverflowError` exception that is thrown when running the following code:
```
listOf(Option(1)).fold(Option.monoidK().algebra())
```

It closes #1126

### DIFF
--- a/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonoidK.kt
+++ b/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonoidK.kt
@@ -17,7 +17,7 @@ interface MonoidK<F> : SemigroupK<F> {
 
   override fun <A> algebra(): Monoid<Kind<F, A>> = object : Monoid<Kind<F, A>> {
 
-    override fun empty(): Kind<F, A> = empty()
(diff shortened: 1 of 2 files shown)


pakoito:
Nice catch! Accept2Unblock.

I'd rather this was part of the laws instead: https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonoidKLaws.kt

franciscodr:
> Nice catch! Accept2Unblock.
> 
> I'd rather this was part of the laws instead: https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonoidKLaws.kt

Sounds good, I'm gonna add the new test to the laws then

pakoito:
@franciscodr @raulraja This should go in the laws! Follow up in #1143

PR #1141, opened 2018-11-19
Integrate API Docs into Arrow doc site + Ank Macros

raulraja:
The following PR introduces an integration based on [Dokka]() as part of the Arrow doc verification flow. It additionally introduces the concept of ank macros that can be used in KDoc and Markdown to automatically generate multiple fences or simplifications of generated boilerplate.

- [x] New `API Docs` section in the top level menu
- [x] New `ank_macro_hierarchy(name)` macro that generates a class hierarchy diagram for any `name` that is a fully qualified class name.
- [x] Added `ank_macro_hierarchy` to all the current type classes in Arrow and their modules.
- [x] Adapted macro to understand when a page has multiple type classes mentioned in their documentation and it automatically adds a single diagram with all the type classes aggregated and show as they relate in the hierarchy.
- [x] Integrated Dokka docs as part of the current design and Jekyll config to be displayed alongside the rest.
See the example below for an auto-calculated diagram that shows the type classes in the `arrow.typeclasses` package

screen shot 2018-11-19 at 6 08 44 pm

- [x] Enabled `ank` postprocessing of Kdocs that allows us to have runnable example in the actual apidocs and not just in the markdown docs that are more tutorial style. See example below for a runnable example of a playground that is applied as a kdoc over a method and how it renders in the arrow docs site:
The documented method below `Bracket#bracketCase` has a kdoc like this one:
```kotlin
/**
   * A generalized version of [bracket] which uses [ExitCase] to distinguish between different exit cases when
   * releasing the acquired resource.
   *
   * @param release is the action supposed to release the allocated resource after `use` is done, by observing and
   * acting on its exit condition.
   *
   * {: data-executable='true'}
   *
   * ```kotlin:ank
   * import arrow.effects.IO
   * import arrow.effects.typeclasses.ExitCase
   *
   * class File(url: String) {
   *   fun open(): File = this
   *   fun close(): Unit {}
   *   override fun toString(): String = "This file contains some interesting content!"
   * }
   *
   * fun openFile(uri: String): IO<File> = IO { File(uri).open() }
   * fun closeFile(file: File): IO<Unit> = IO { file.close() }
   * fun fileToString(file: File): IO<String> = IO { file.toString() }
   *
   * fun main(args: Array<String>) {
   *   //sampleStart
   *   val safeComputation = openFile("data.json").bracketCase(
   *     release = { file, exitCase ->
   *       when (exitCase) {
   *         is ExitCase.Completed -> { /* do something */ }
   *         is ExitCase.Cancelled -> { /* do something */ }
   *         is ExitCase.Error -> { /* do something */ }
   *       }
   *       closeFile(file)
   *    },
   *    use = { file -> fileToString(file) }
   *  )
   *  //sampleEnd
   *  println(safeComputation)
   *  }
   *  ```
   *
   */
  fun <A, B> Kind<F, A>.bracketCase(release: (A, ExitCase<E>) -> Kind<F, Unit>, use: (A) -> Kind<F, B>): Kind<F, B>
```

And it renders in the Arrow doc site like

screen shot 2018-11-19 at 6 33 16 pm


modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Bracket.kt
```diff
@@ -44,6 +81,33 @@ interface Bracket<F, E> : MonadError<F, E> {
    *
    * @param release is the action that's supposed to release the allocated resource after `use` is done, irregardless
    * of its exit condition.
+   *
+   * {: data-executable='true'}
+   *
+   * ```kotlin:ank
+   * import arrow.effects.IO
+   *
+   * class File(url: String) {
```
nomisRev: What happens with this? In the screenshot, it's all gone.

raulraja: It shows everything collapsed except for what is in between `//sampleStart` and `//sampleEnd`

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -77,6 +77,8 @@ sealed class IO<out A> : IOOf<A> {
         }
       }
 
+    val never: IO<Nothing> = async { _, _ -> Unit }
```
pakoito: This should come from the Async instance.

raulraja: What is the advantage? It ends up calling this same code after three indirections and `async` is also public and in scope.

raulraja: I think it'd be better if the IOAsyncInstance called this directly instead of `async` in the type class so that there is less indirection

pakoito: I don't have a strong preference, just pointing out that codegen should bring that constructor in anyway. IIRC it was there at some point.

PR #1142, opened 2018-11-19
Address PR Λnk README comments

Guardiola31337:
- Follow up from https://github.com/arrow-kt/arrow/pull/1139


raulraja:
thanks @Guardiola31337 !

PR #1147, opened 2018-11-20
creating lift for bifunctor

leandroBorgesFerreira:
It adds lift for the Bifunctor. that's it.


leandroBorgesFerreira:
Ok, no problem

leandroBorgesFerreira:
Added the Dokka docs and some ANK samples. But I am not sure if it is correct, is it?

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Bifunctor.kt
```diff
@@ -29,4 +96,4 @@ private interface RightFunctor<F, X> : Functor<Kind<F, X>> {
 
   override fun <A, B> Kind2<F, X, A>.map(f: (A) -> B): Kind2<F, X, B> =
     F.run { bimap(::identity, f) }
-}
\ No newline at end of file
+}
```
pakoito: <3

Issue #590, opened 2018-01-20
MonadDeferLaws


pakoito:
A port of these laws:

https://github.com/typelevel/cats-effect/blob/master/laws/shared/src/main/scala/cats/effect/laws/SyncLaws.scala

in here:

https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt#L23

(Sync is called MonadDefer in Arrow)

sanogueralorenzo:
I'll take this one 😄

1Jajen1:
So if I understand this correctly the current MonadDeferLaws only include laws for binding over monad defer and should be extended by those present in cats-effect. The cats-effect typeclass seems to have different functions on top of what MonadDefer in arrow has (or maybe I just can't map the names correctly).
I'll try and get as many ported as possible and then see how it works :)

pakoito:
You're correct, the ones in the current laws are ad-hoc for monadic comprehensions. We also need the ones from cats which are the original test suite for the typeclass :D

1Jajen1:
Closed by #1150

PR #1150, opened 2018-11-21
Monad defer laws

1Jajen1:
Monad defer laws portet from [cats-sync-laws](https://github.com/typelevel/cats-effect/blob/master/laws/shared/src/main/scala/cats/effect/laws/SyncLaws.scala).

Solves #590 

Excluded laws are: (by function name from cats)
- suspendConstantIsPureJoin (what is the suspend/flatten equivalent in arrow)
- suspendThrowIsRaiseError (^)
- unsequencedDelayIsNoop (what is this for? is that not covered by delaySuspendsEvaluation, is it the same? it is a confusing name either way...)
- repeatedSyncEvaluationNotMemoized (is this a thing in arrow)
- stack safety laws (all of them) (not sure what the case is here, I had them initially, and they passed for IO and deferred, but failed for all otheres, so I left them out)

Also there seems to be something wrong with DeferredK as it fails the test for delay suspends evaluation. I believe it is because it is using an async instance of DeferredK which overrides delay to invoke `DeferredK.invoke` which does not start a coroutine with `CoroutineStart.LAZY` but with `CoroutineStart.DEFAULT`.

### DIFF
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
@@ -23,7 +23,8 @@ object MonadDeferLaws {
     SC: MonadDefer<F>,
     EQ: Eq<Kind<F, Int>>,
     EQ_EITHER: Eq<Kind<F, Either<Throwable, Int>>>,
-    EQERR: Eq<Kind<F, Int>> = EQ
+    EQERR: Eq<Kind<F, Int>> = EQ,
+    testStackSafety: Boolean = true
   ): List<Law> =
     BracketLaws.laws(SC, EQ, EQ_EITHER, EQERR) + listOf(
       Law("Sync bind: binding blocks") { SC.asyncBind(EQ) },
@@ -47,12 +48,17 @@ object MonadDeferLaws {
       Law("Sync laws: delay suspends evaluation") { SC.delaySuspendsEvaluation(EQ) },
       Law("Sync laws: flatMap suspends evaluation") { SC.flatMapSuspendsEvaluation(EQ) },
       Law("Sync laws: map suspends evaluation") { SC.mapSuspendsEvaluation(EQ) },
-      Law("Sync laws: Repeated evaluation not memoized") { SC.repeatedSyncEvaluationNotMemoized(EQ) },
-      Law("Sync laws: stack safety over repeated left binds") { SC.stackSafetyOverRepeatedLeftBinds(5000, EQ) },
-      Law("Sync laws: stack safety over repeated right binds") { SC.stackSafetyOverRepeatedRightBinds(5000, EQ) },
-      Law("Sync laws: stack safety over repeated attempts") { SC.stackSafetyOverRepeatedAttempts(5000, EQ) },
-      Law("Sync laws: stack safety over repeated maps") { SC.stackSafetyOnRepeatedMaps(5000, EQ) }
-    )
+      Law("Sync laws: Repeated evaluation not memoized") { SC.repeatedSyncEvaluationNotMemoized(EQ) }
+    ) + if (testStackSafety) {
+      listOf(
+        Law("Sync laws: stack safety over repeated left binds") { SC.stackSafetyOverRepeatedLeftBinds(5000, EQ) },
+        Law("Sync laws: stack safety over repeated right binds") { SC.stackSafetyOverRepeatedRightBinds(5000, EQ) },
+        Law("Sync laws: stack safety over repeated attempts") { SC.stackSafetyOverRepeatedAttempts(5000, EQ) },
+        Law("Sync laws: stack safety over repeated maps") { SC.stackSafetyOnRepeatedMaps(5000, EQ) }
+      )
+    } else {
+      emptyList()
+    }
 
   fun <F> MonadDefer<F>.delayConstantEqualsPure(EQ: Eq<Kind<F, Int>>): Unit {
     forAll(genIntSmall()) { x ->
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
@@ -2,12 +2,17 @@ package arrow.test.laws
 
 import arrow.Kind
 import arrow.core.*
+import arrow.data.k
 import arrow.effects.data.internal.BindingCancellationException
 import arrow.effects.typeclasses.MonadDefer
+import arrow.instances.list.foldable.foldLeft
 import arrow.test.concurrency.SideEffect
 import arrow.test.generators.genIntSmall
 import arrow.test.generators.genThrowable
 import arrow.typeclasses.Eq
+import io.kotlintest.matchers.Matcher
+import io.kotlintest.matchers.should
+import io.kotlintest.matchers.shouldBe
 import io.kotlintest.properties.forAll
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.newSingleThreadContext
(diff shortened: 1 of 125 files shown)


1Jajen1:
Tests will fail btw because of the one DeferredK test (or they better should :D)

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
```diff
@@ -31,9 +36,95 @@ object MonadDeferLaws {
...
+    val sideEffect = SideEffect(counter = 0)
+    val df = delay { sideEffect.increment(); sideEffect.counter }
+
+    Thread.sleep(10)
+
+    sideEffect.counter shouldBe 0
+    df.equalUnderTheLaw(just(1), EQ) shouldBe true
+  }
+
+  fun <F> MonadDefer<F>.bindSuspendsEvaluation(EQ: Eq<Kind<F, Int>>): Unit {
```
pakoito: Rename to `flatMapSuspendsEvaluation`

pakoito:
Flatten is in Monad, which means it's available in MonadDefer: https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Monad.kt#L22

Defer constant is a new function you can add in MonadDefer:
`fun <A> defer(Kind<F, A>): Kind<F, A>`

RaiseError is in ApplicativeError, of which MonadDefer inherits too: https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/ApplicativeError.kt#L8

> unsequencedDelayIsNoop

Yeah, it's likely a duplicated, but keep it just in case :D

> repeatedSyncEvaluationNotMemoized

*> is a shorter version of flatMap, just use flatMap instead

>stack safety laws

If the instances or the implementations are broken, we need to check the cause 🔍 

>DeferredK.invoke which does not start a coroutine with CoroutineStart.LAZY but with CoroutineStart.DEFAULT.

Seems like a bug to fix to to me then hahaha

pakoito:
Please fix DeferredK.async too! https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt#L87-L89

1Jajen1:
So is a function `deferConstant` really needed? What would it look like? Simply returning its argument or would it invoke `defer { arg }`. I don't really get the use so I don't really know how to implement it :confused: 

>> unsequencedDelayIsNoop

> Yeah, it's likely a duplicated, but keep it just in case :D

Well they don't have a delay/defer suspends eval law so that might it. So add it regardless?

>> DeferredK.invoke which does not start a coroutine with CoroutineStart.LAZY but with
>> CoroutineStart.DEFAULT.

> Seems like a bug to fix to to me caught by the tests then hahaha

Well other instances (IO) override delay as well, to delegate it to invoke which in turn does exactly the same as delay in MonadDefer except it does not catch errors.
So how is delay in async supposed to work? Or does it need to be overridden at all?

And for async and invoke I am guessing it should just start lazy instead of default.

I am just asking all of these questions because there might have been a reason to implement it like that :see_no_evil:

pakoito:
>would it invoke defer { arg }

Yeah, that's it, ez :D

>So add it regardless?

yeah, go for it

>And for async and invoke I am guessing it should just start lazy instead of default.

Yep!

>Well other instances (IO) override delay as well, to delegate it to invoke which in turn does exactly the same as delay in MonadDefer except it does not catch errors. So how is delay in async supposed to work? Or does it need to be overridden at all?

So if I understand correctly, the cause of the test failing is that one of invoke/suspend/defer is not catching errors in the block when it's supposed to. Easy fix, try/catch the implementations in IO.kt then :D

pakoito:
Yes, it seems incorrect. Just change it to the original implementation in the datatype then.

1Jajen1:
So I added repeatedSyncEvaluationNotMemoized not sure how useful it is tho. I skipped unsequencedDelayIsNoop because it ends up exactly like delaySuspendsEvaluation.

Regarding stack safety. Are all MonadDefer instances supposed to be stack safe?

1Jajen1:
Then there is a problem with reactor and rx2 instances of MonadDefer because they all fail the tests (I'll push them in a sec) for stacksafety while IO and DeferredK work just fine

raulraja:
@1Jajen1 any way to see where the stack is recursing?

1Jajen1:
For reactor (MonoK) the stacktrace points to methods from reactor. 

StackTrace excerpt: 
`Exception in thread "main" java.lang.StackOverflowError
	at reactor.core.publisher.Operators$MonoSubscriber.actual(Operators.java:1103)
	at reactor.core.publisher.InnerOperator.currentContext(InnerOperator.java:32)
	at reactor.core.publisher.InnerOperator.currentContext(InnerOperator.java:32)
	at reactor.core.publisher.InnerOperator.currentContext(InnerOperator.java:32)
	at reactor.core.publisher.InnerOperator.currentContext(InnerOperator.java:32)
	at reactor.core.publisher.InnerOperator.currentContext(InnerOperator.java:32)
	at reactor.core.publisher.InnerOperator.currentContext(InnerOperator.java:32)
	at reactor.core.publisher.MonoFlatMap$FlatMapMain.onNext(MonoFlatMap.java:123)
	at reactor.core.publisher.Operators$MonoSubscriber.complete(Operators.java:1083)`

The last two lines get cycled over and over again.

FluxK:
`Exception in thread "main" java.lang.StackOverflowError
	at reactor.core.publisher.FluxFlatMap$FlatMapMain.onSubscribe(FluxFlatMap.java:331)
	at reactor.core.publisher.FluxFlatMap$FlatMapMain.onSubscribe(FluxFlatMap.java:331)
	at reactor.core.publisher.FluxFlatMap$FlatMapMain.onSubscribe(FluxFlatMap.java:331)
	at reactor.core.publisher.FluxFlatMap$FlatMapMain.onSubscribe(FluxFlatMap.java:331)`

And this as well cycles on and on...

I think rx2 will be similar. I don't really know what to do with this, or if something from inside arrow can/should be done.

raulraja:
I'm not familiar with those data types. Maybe @pakoito @Cotel can help, but it's bad. If they don't pass the laws we are gonna have to disable the integrations.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
```diff
@@ -31,9 +36,132 @@ object MonadDeferLaws {
...
+
+    sideEffect.counter shouldBe 0
+    df.equalUnderTheLaw(just(1), EQ) shouldBe true
+  }
+
+  fun <F> MonadDefer<F>.repeatedSyncEvaluationNotMemoized(EQ: Eq<Kind<F, Int>>): Unit {
+    val sideEffect = SideEffect()
+    val df = delay { sideEffect.increment(); sideEffect.counter }
+
+    df.flatMap { df }.flatMap { df }.equalUnderTheLaw(just(3), EQ)
```
pakoito: shouldBe true, else this is not checked

1Jajen1: Can't believe I forgot that :sweat_smile:. Added it and it works for io, reactor and rx2 but fails for DeferredK.

raulraja:
@pakoito can you link them here?

1Jajen1:
Welp yeah will do. Should that be added to the docs somewhere? (better yet create a ticket for it since that is not really the scope of this pr...) 
I mean for most it'll hopefully be a non-issue anyway...

pakoito:
Nah, leave it as the commented out tests with that FIXME. We have a couple of those in the codebase already.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
1Jajen1: That is a good question and might be an issue or feature of coroutines. Will test this more thoroughly later on.
A possible scenario would be that coroutines simply after finishing return their result instead of rerunning when awaited again

pakoito:
Not mandatory for this ticket, but if you could add the FIXME to the tailrecM implementations of Observable, Flowable, Mono, and Flux here we'd save ourselves a new PR :D

https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-rx2/src/main/kotlin/arrow/effects/ObservableK.kt#L113

https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-reactor/src/main/kotlin/arrow/effects/MonoK.kt#L90

1Jajen1:
Instead of commenting them out can't we use two lists of laws? Then at least io and deferred are well tested and any future changes that might break will be caught

pakoito:
In those cases what we normally do is add the tests directly over their files, i.e. IOTest, DeferredTest, etc. even if they're almost literal duplicates. I'm not sure if this is the best approach, but I'm now sure we want to have `MonadDeferLaws` split in two files/functions :/ 

@raulraja for second opinion

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
1Jajen1: Well it makes sense now... Thats just how coroutines work:
```kotlin
  var sideEffect = 0
  val a = GlobalScope.async(start = CoroutineStart.LAZY) {
    ++sideEffect
  }

  println("Started a: ${a.start()}")

  runBlocking {
    println("Result a: ${a.await()}")
  }

  val b = GlobalScope.async(start = CoroutineStart.LAZY) {
    a.await()
  }

  println("Started b: ${b.start()}")

  runBlocking {
    println("Result b: ${b.await()}")
  }

  println(sideEffect)
```

Which prints: 
```
Started a: true
Result a: 1
Started b: true
Result b: 1
1
```

So what should be done about this? This is not like the stack-safety issue, if someone were to try and build a retry/repeat scheduler that won't work with DeferredK. Also other code which reuses MonadDefer instances (and then gets a deferredK) simply won't work. And the more annoying problem here is that it would be very hard to track down to Coroutines in that case :/

I don't really know how to solve this, wrapping deferred in DeferredK into a function has other challenges (stack-safety etc) that might not work... Maybe someone at jb knows what to do here...

pakoito: Right, I forgot they are memoized -.- This cannot be solved by JBs as it'd require moving the stack labels, which is the same issue we have with monad comprehensions, and it's just a design choice instead.

cc @raulraja for ideas

1Jajen1: Ok so how about I create a ticket for this. Comment the test out for now with a fixme linked to the ticket, add a copy to io's tests (also with a comment and link) and let that ticket deal with it then.

raulraja:
I think the docs on each one of the wrappers that are not stack-safe should state that, so users are aware of the reasons why it's not. Including proper links to the issues that @pakoito created in their repos. We can also display examples how we can make them stack-safe with `bindingStackSafe` examples that show how we have overcome the limitation with a classical example with over 5k iterations.

raulraja:
I'd rather have a way to tell the monad-suspend laws to skip the problematic tests so we can override in those data types and potentially emit a warning that they have been skipped because of [links here]

1Jajen1:
Added an option to skip and warnings with the link to the tickets. Only test that fails now is the one that ensures MonadDefer instances don't memoize things for DeferredK. Don't really know what to do with it yet... (Might make sense to skip it as well for now and add a ticket/pr trying to fix it, because that may be some work...)

Using bindingStackSafe would only be stacksafe inside that binding right? I mean it'd be fine if well documented that without it its not supported, but a general stack safe integration would be better if it does not mean lots of added complexety or overhead.

I'll do docs later on or tomorrow morning. If its not too much I'll keep it in this pr.

1Jajen1:
Added some documentation about rx2 and reactor not being stacksafe, with links to both tickets.
Last thing to do (hopefully :see_no_evil:) in this pr is dealing with DeferredK memoizing results. :/

modules/docs/arrow-docs/docs/docs/integrations/rx2/README.md
```diff
@@ -181,6 +181,49 @@ disposable()
 // Boom! caused by BindingCancellationException
 ```
 
+### Stack safety
```
pakoito: <3

modules/docs/arrow-docs/docs/docs/integrations/reactor/README.md
```diff
@@ -148,6 +148,49 @@ disposable()
 // Boom! caused by BindingCancellationException
 ```
 
+### Stack safety
```
pakoito: <3

pakoito:
>is dealing with DeferredK memoizing results. :/

I'll take a look, I had some "shower" ideas last morning

pakoito:
We'll need to refactor Deferred for that test to pass, I've fixed it bit needs some cleanup. I say disable it for now, merge this PR and we'll make a ticket to work on deferred ASAP.

1Jajen1:
Commented out and linket to #1156 (which will be the next ticket created if no one does anything the next few minutes :see_no_evil:)

I'll create the ticket and link back to this as well.

modules/effects/arrow-effects-reactor-instances/src/test/kotlin/arrow/effects/FluxKTests.kt
```diff
@@ -53,10 +55,15 @@ class FluxKTest : UnitSpec() {
       }
   }
 
+  override fun interceptSpec(context: Spec, spec: () -> Unit) {
```
pakoito: Neat!

Issue #1153, opened 2018-11-22
Store docs pages in S3 bucket


dominv:

dominv:
Related PR https://github.com/arrow-kt/arrow/pull/1152 -> Merged

PR #1155, opened 2018-11-23
Add video for Simple Dependency Management in Kotlin

pakoito:


raulraja:
This won't show up live until the Arrow docs are deployed in S3 which @calvellido and @dominv are working on. See #1153

pakoito:
Yeah I was a bit suspicious that travis kept on failing

Issue #1156, opened 2018-11-23
DeferredK memoizes its result


1Jajen1:
DeferredK implements the MonadDefer which has a law stating its results should not be memoized. Kotlinx coroutines Deferred (which most of DeferredK delegates to) however does not follow this.

See #1150 for some discussion and I think @pakoito had some ideas as well on how to solve this.

Please also re-enable the law once done :)

pakoito:
I've given Jorge and Simon the context and the prototype for the new Deferred architecture :D

JorgeCastilloPrz:
Thanks for the issue. We'll take a look at this, not sure if it can happen before Q1 though, we really wanna finish the other effects blockers first.

pakoito:
Note that we can fix this #1131 in one go too :D

1Jajen1:
Closed by #1157

Issue #1131, opened 2018-11-16
Fix Deferred<A>.k() and CoroutineScope.asyncK(...) to set the right scope


pakoito:
https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt#L11-L15

Currently they set the scope to Global, due to the default value in the constructor of `DeferredK`. It should inherit it from `scope` in `asyncK`, and get it from the context in `k()`.

pakoito:
The architecture proposed in #1156 should fix this.

1Jajen1:
Not sure if this is closed, `.k()` still sets globalscope because afaik there is no way to get the scope from a created Deferred. `asyncK` works correctly now tho.

nomisRev:
@pakoito this can be closed with #1220, right? It also sets all the correct scope.

pakoito:
Yep! Close now :D

PR #1157, opened 2018-11-25
Attempt to make DeferredK rerunnable

1Jajen1:
Solves #1156 and maybe #1131 
Re-enable no-memoization law and fix deferredK to follow it (if used correctly)
Changes practically nothing from the public api.

I'll add some info to the docs later because if used wrong it might produce unexpected results.
Should the methods in DeferredK (well the public ones) have api docs (runnable ones) as well?

@pakoito I have skipped the invoke constructor for Wrapped because its syntax conflicts with the other one too much (the compiler was only able to infer which one to choose after passing more and more arguments which defeats the point of default argumens there), one should just use the other one or `.k()` instead. (And going from `() -> Deferred<A>` to `suspend () -> A` is trivial)

### DIFF
--- a/modules/ank/arrow-ank/build.gradle
+++ b/modules/ank/arrow-ank/build.gradle
@@ -1,9 +1,10 @@
 apply plugin: 'kotlin'
 
 dependencies {
-    compile project(":arrow-instances-free")
     compile project(":arrow-instances-core")
     compile project(":arrow-instances-data")
+    compile project(":arrow-effects-instances")
+
     compile "org.jetbrains:markdown:0.1.25"
     compile "com.github.born2snipe:cli-progress:0.4"
     compile "org.jetbrains.kotlin:kotlin-compiler:$kotlinVersion"
--- a/modules/ank/arrow-ank/src/main/kotlin/arrow/ank/algebra.kt
+++ b/modules/ank/arrow-ank/src/main/kotlin/arrow/ank/algebra.kt
@@ -1,33 +1,34 @@
 package arrow.ank
 
+import arrow.Kind
+import arrow.core.Option
+import arrow.core.Try
 import arrow.core.Tuple2
+import arrow.data.Nel
+import arrow.typeclasses.MonadThrow
 import org.intellij.markdown.ast.ASTNode
-import java.io.File
+import java.nio.file.Path
 
-fun createTarget(source: File, target: File): File =
-  createTargetImpl(source, target)
+interface AnkOps<F> {
+  fun MF(): MonadThrow<F>
 
-fun getFileCandidates(target: File): List<File> =
-  getFileCandidatesImpl(target)
+  fun createTargetDirectory(source: Path, target: Path): Kind<F, Try<Path>>
 
-fun readFile(source: File): String =
-  readFileImpl(source)
+  fun getCandidatePaths(root: Path): Kind<F, Option<Nel<Path>>>
 
-fun preProcessMacros(source: Tuple2<File, String>): String =
-  preProcessMacrosImpl(source)
+  fun readFile(path: Path): Kind<F, Try<String>>
 
-fun parseMarkdown(markdown: String): ASTNode =
-  parseMarkDownImpl(markdown)
+  fun preProcessMacros(pathAndContent: Tuple2<Path, String>): String
 
-fun extractCode(source: String, tree: ASTNode): List<Snippet> =
-  extractCodeImpl(source, tree)
+  fun parseMarkdown(markdown: String): Kind<F, Try<ASTNode>>
 
-fun compileCode(snippets: Map<File, List<Snippet>>, compilerArgs: List<String>): List<CompiledMarkdown> =
-  compileCodeImpl(snippets, compilerArgs)
+  fun extractCode(content: String, ast: ASTNode): Kind<F, Option<Nel<Snippet>>>
 
-fun replaceAnkToLang(compilationResults: CompiledMarkdown): String =
-  replaceAnkToLangImpl(compilationResults)
+  fun compileCode(snippets: Tuple2<Path, Nel<Snippet>>, compilerArgs: List<String>): Kind<F, Try<Nel<Snippet>>>
 
-fun generateFile(candidate: File, newContents: String): File =
-  generateFileImpl(candidate, newContents)
+  fun replaceAnkToLang(content: String, compiledSnippets: Nel<Snippet>): String
 
+  fun generateFile(path: Path, newContent: String): Kind<F, Try<Unit>>
+
+  fun printConsole(msg: String): Kind<F, Unit>
+}
--- a/modules/ank/arrow-ank/src/main/kotlin/arrow/ank/algebra.kt
+++ b/modules/ank/arrow-ank/src/main/kotlin/arrow/ank/algebra.kt
@@ -2,7 +2,6 @@ package arrow.ank
 
 import arrow.Kind
 import arrow.core.Option
-import arrow.core.Try
 import arrow.core.Tuple2
 import arrow.data.Nel
 import arrow.typeclasses.MonadThrow
@@ -12,23 +11,23 @@ import java.nio.file.Path
 interface AnkOps<F> {
   fun MF(): MonadThrow<F>
 
-  fun createTargetDirectory(source: Path, target: Path): Kind<F, Try<Path>>
+  fun createTargetDirectory(source: Path, target: Path): Kind<F, Path>
 
   fun getCandidatePaths(root: Path): Kind<F, Option<Nel<Path>>>
 
-  fun readFile(path: Path): Kind<F, Try<String>>
+  fun readFile(path: Path): Kind<F, String>
 
   fun preProcessMacros(pathAndContent: Tuple2<Path, String>): String
 
-  fun parseMarkdown(markdown: String): Kind<F, Try<ASTNode>>
+  fun parseMarkdown(markdown: String): Kind<F, ASTNode>
 
   fun extractCode(content: String, ast: ASTNode): Kind<F, Option<Nel<Snippet>>>
 
-  fun compileCode(snippets: Tuple2<Path, Nel<Snippet>>, compilerArgs: List<String>): Kind<F, Try<Nel<Snippet>>>
+  fun compileCode(snippets: Tuple2<Path, Nel<Snippet>>, compilerArgs: List<String>): Kind<F, Nel<Snippet>>
 
   fun replaceAnkToLang(content: String, compiledSnippets: Nel<Snippet>): String
 
-  fun generateFile(path: Path, newContent: String): Kind<F, Try<Unit>>
+  fun generateFile(path: Path, newContent: String): Kind<F, Unit>
 
   fun printConsole(msg: String): Kind<F, Unit>
 }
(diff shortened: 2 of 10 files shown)


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
```diff
@@ -146,25 +143,25 @@ object MonadDeferLaws {
     df.flatMap { df }.flatMap { df }.equalUnderTheLaw(just(3), EQ) shouldBe true
   }
 
-  fun <F> MonadDefer<F>.stackSafetyOverRepeatedLeftBinds(iterations: Int = 5000, EQ: Eq<Kind<F, Int>>): Unit {
+  fun <F> MonadDefer<F>.stackSafetyOverRepeatedLeftBinds(iterations: Int = 50000, EQ: Eq<Kind<F, Int>>): Unit {
```
pakoito: 5k is enough to trigger a SO, 50k makes them too slow maybe.

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt
```diff
@@ -127,13 +165,13 @@ fun <A> DeferredKOf<A>.unsafeRunSync(): A =
   runBlocking { await() }
 
 fun <A> DeferredKOf<A>.runAsync(cb: (Either<Throwable, A>) -> DeferredKOf<Unit>): DeferredK<Unit> =
-  DeferredK.invoke(scope(), Dispatchers.Unconfined, CoroutineStart.DEFAULT) {
+  DeferredK.invoke(fix().scope(), Dispatchers.Unconfined, CoroutineStart.DEFAULT) {
```
pakoito: Fix shouldn't be necessary here, `scope()` is defined as an alias to `fix().scope` atop of the file.

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt
```diff
@@ -43,42 +82,41 @@ data class DeferredK<out A>(private val deferred: Deferred<A>, val scope: Corout
...
-
-    fun <A> failed(t: Throwable): DeferredK<A> =
-      CompletableDeferred<A>().apply { completeExceptionally(t) }.k()
+    operator fun <A> invoke(scope: CoroutineScope = GlobalScope, ctx: CoroutineContext = Dispatchers.Default, start: CoroutineStart = CoroutineStart.LAZY, f: suspend () -> A): DeferredK<A> =
+      Generated(ctx, start, scope, f)
 
     fun <A> raiseError(t: Throwable): DeferredK<A> =
       failed(t)
 
+    fun <A> failed(t: Throwable): DeferredK<A> =
```
pakoito: *squint* why is this here? why do we need an alias?

pakoito: I know I wrote it, I cannot remember why.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
1Jajen1: Done

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt
1Jajen1: :+1:

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt
1Jajen1: Other than providing a more intuitive constructor for those that have never used raiseError I don't see the point either :) I'll remove it for now

1Jajen1:
It's not really required to distinguish between instantiation strategies (only for .k(), all others will work fine if a user does not use a deferred from outside that deferredK). So I think a section in the docs for DeferredK would be enough there. I can (and probably will in the next hour or so) add some kdoc to the api methods of DeferredK with examples. But except for .k() I don't think its worth mentioning retry-ability every time.

pakoito:
Before merging I would hide the internal nodes behind a constructor and mark them as `internal data class`, else it's an API leak the users don't care about.

1Jajen1:
You mean marking Wrapped and Generated as Internal? They are already afaik. Changed them to data classes as well now. Do data classes override equals and stuff from their parent? o.O If so I may need to add equals/hashCode to those.

pakoito:
Don't make them into `data`, you were right all along :D

1Jajen1:
Good thing git makes it nice and easy to revert stuff :)

pakoito:
>Could you please add one more override of `invoke` that works as `.k()`, taking the scope too?

That and I'm good to merge :D

raulraja:
@1Jajen1 yes ideally all methods are documented with at lest one example and a brief description of what they do. We consider this extremely important and are trying to enforce it lately in all PRs to ensure new users have a good experience when visiting the Arrow documentation. Each one of your methods then will have a dedicated page like this one http://arrow-docs.s3-website-us-east-1.amazonaws.com/docs/apidocs/arrow-effects/arrow.effects.typeclasses/-bracket/bracket-case.html

1Jajen1:
@raulraja Yeah I saw those, built it locally. The jekyll command took a while but worked out fine o.O
Will add it then. :+1: 

@pakoito So like an invoke that takes a Deferred? As in a convenience method over `defer { deferred.k() }` I'll do that, but I don't think the deferred can be supplied by a function, as the type signature of that will be very similar to the other one, thus when invoking it'll be annoying the user that it does not know if it has a `suspend () -> Deferred<A>` or a `() -> Deferred<A>`.

pakoito:
The function signature is `operator fun <A> invoke(scope: CoroutineScope = GlobalScope, f: Deferred<A>): DeferredK<A>`, with a value instead of a function.

1Jajen1:
Added a lot of kdoc to almost all methods of DeferredK and the invoke constructor.

Btw. can you give generics a different color, like orange or something, in the api docs? Would make it so much more readable :)

raulraja:
@1Jajen1 yes, feel free to create an issue and assign it to Israel Perez

raulraja:
It's the number of libraries. As we move past 1.0 most of them will be in their own repository. Effects, streams, recursion schemes etc... we'll all be able to evolve differently and then the build times will be much smaller. In the mean-time, if you find areas where the build times can improve feel free to raise issues.

1Jajen1:
I'll look into speeding up gradle somehow and what exactly takes how much time... Don't know if I find stuff to improve, but worth trying :)

1Jajen1:
So the build itself consistently takes about 20 mins (15-18 on my laptop) with almost all time spent on tests (kapt is also pretty slow, but its nothing compared to the tests). 
Running `dokka :arrow-docs:runAnk` takes me about 3 mins, travis does not even get it done in 30 and my guess is that it's memory consumption. I frequently go up beyond 8-9 GB for that java task running it. (Which in turn slows down travis that does not have nearly as much memory available... Add another 10k files and let it run 2-3 hours and it'll probably throw a gc exception again)
So thats what needs to be dealt with. At that stage it's either dokka, ank or gradle. And considering memory consumption jumps up only when ank is processing it might still be it. I know it reports mem usage as ~1-2Gb all the time but something is obviously wrong :/

raulraja:
I fixed ank memory problems in master a while back, did you merge since then?

1Jajen1:
Yes. It reports memory usage to be arount 1.5GB.
The process that runs ank (not confirmed, just the only process actually doing something when ank prints output) however consumes a ton of memory, way more than printed out (usually going to 6-8GB).

Some ideas I had, but could not test yet:
- The 1.5GB is inaccurate (which I don't think because it's read out correctly)
- The script engine runs in a seperate process and consumes a ton of memory which the parent just does not include (just an idea, don't know how it works)
- The ank plugin consumes a ton of memory. (Since this delegates to ank this seems unlikely)

1Jajen1:
I can confirm the reported memory usage to be the actual heap memory usage, but also that the process running ank is actually consuming up to 4-5 times that, which means stuff not on the heap is leaking. Although I have no idea what causes that.

modules/ank/arrow-ank/build.gradle
```diff
@@ -1,9 +1,10 @@
 apply plugin: 'kotlin'
```
pakoito: Is this a refactor or a merge? O.O

1Jajen1:
So I ended up not finding the reason why so much off-heap memory was used and resorted to rewriting ank to use `java.nio` instead of `java.io` (for the most part). I also changed and optimised a few things here and there. Most logic for processing input is still the same, just the way its invoked changed a bit.

It now consumes up to ~3GB memory at most. (Heap ~2GB, rest is probably memory mapped io)
It is also considerably faster (at least it feels faster, maybe that has to do with my laptop not freezing and putting stuff on swap to keep running o.O)

I ended up putting the changes on this branch to test it on travis as well, if this should be in a seperate pr please let me know and I'll revert that :) Either way please check if this code is correct or produces good ouput, I couldn't find differences from running ank on master but you never know :see_no_evil:

modules/ank/arrow-ank/build.gradle
1Jajen1: Too fast again :)
Explained this below.

modules/ank/arrow-ank/src/main/kotlin/arrow/ank/algebra.kt
```diff
@@ -1,33 +1,34 @@
...
+import java.nio.file.Path
 
-fun createTarget(source: File, target: File): File =
-  createTargetImpl(source, target)
+interface AnkOps<F> {
+  fun MF(): MonadThrow<F>
 
-fun getFileCandidates(target: File): List<File> =
-  getFileCandidatesImpl(target)
+  fun createTargetDirectory(source: Path, target: Path): Kind<F, Try<Path>>
```
raulraja: There is no point on having concrete `Kind<F, Try<Path>` return type since they can just be `Kind<F, Path>` given you already have in scope a constrain on `MonadThrow<F>`. Why is the internal `Try` needed? You may just call `MF().raiseError` or use `MF().catch { ... }` if you just need to control synchronous error control. We can make the entire set of ops return `IO<A>` if you'd rather use an explicit concrete data type.

1Jajen1: Yeah, I kinda forgot when defining the algebra and then did not change it when implementing ...
Will do in a sec. Just removing a bunch of them should be fine since `MonadDefer.delay` also catches errors.
I had the type as `Kind<F, A>` because I was starting with only Monad as the constraint on F. That would make testing with something like an id interpreter nice. Now with MonadThrow this may not be valid anymore. (Although Try might be a good fit there)

1Jajen1:
Changed the signatures. (They really were kinda dumb...)

At least the whole build gets done in like 35 mins now. That leaves about 15 mins leeway for new tests and more kdoc to process before something has to be done about builds again, and if the repo gets split before that it likely won't be a problem ever :)

> The ank refactor is fine if it solves the memory issue for you but in whatever case we need to fix the nested Try signatures. You don't need those if you already have a constrain in MonadThrow<F>

Did you never get memory issues running it? o.O Maybe its a os specific thing then.

raulraja:
No issues here and the latest in master is gonna push those to s3 instead of github pages because jekyll times out in github

pakoito:
Soooooo...is anyone pushing this while the builds aren't broken and master is stable? :D

1Jajen1:
This should be good to go.
There is a bug in generating diagrams right now. Typeclasses which don't inherit from any other typeclass don't have a diagram atm. This is happening on master as well and I don't really know how and where they are generated or if this even may be intentional. For now I think merging this and creating a ticket should be fine.

PR #1166, opened 2018-11-29
Fix docs type class data class

### LINKED ISSUE #1120: Fix all docs type class and data types links
All type classes and data type in the documentation site should follow the same path as the package they declared in terms of permalinks in the docs.
For example:

`arrow.core.Option` -> `docs/arrow/core/option`

All permalinks need to be modified and if possible preserve existing redirects.
This is necessary because we plan on using reflection and other introspection techniques to have some parts of the docs autogenerated at build time. Links for a given data type or type class need to be predictable so the generator can assume those paths to be there.

Tylos:
Close #1120 

Change all permalinks in modules/docs/arrow-docs/docs/docs/typeclasses and modules/docs/arrow-docs/docs/docs/datatypes documentation folders, update all internal links within the documentation to use the aforementioned permalinks and provide redirection for previous ones.

Most important commits would be ca139db which adds `jekyll-redirect-from` plugin configuration allowing to define in jekyll's format matter header links which must redirect to or from the current page. This plugin is supported by github pages so no additional configuration must be done apart from declaring it in the _config.yml

As it is an external plugin, it is a different dependency that must be suffied. Hence, Gemfile is added so the dependency management is easier. Before generating the documentation, invoke bundle install command so the dependencies are updated. The Gemfile does not define any dependency itself, but uses github-pages gem which sets up and maintains a local Jekyll environment in sync with GitHub Pages

Also 785bb86 updates the CONTRIBUTING.md file to reflect the changes in builds


CONTRIBUTING.md
```diff
@@ -31,16 +31,18 @@ $ ./gradlew clean dokka :arrow-docs:runAnk
...
+$ gem install bundler
+$ bundle install --gemfile=modules/docs/arrow-docs/Gemfile
 ```
 
-Now you have jekyll, then you can preview the documentation by serving the artifact produced by ΛNK.
+After all the dependencies are installed, you can preview the documentation by serving the artifact produced by ΛNK.
 
 ```
-$ jekyll serve --source modules/docs/arrow-docs/build/site/
+$ cd modules/docs/arrow-docs/
```
Tylos: ☝️ Have not been able to make this work without switching folder first :(

CONTRIBUTING.md
```diff
@@ -31,16 +31,18 @@ $ ./gradlew clean dokka :arrow-docs:runAnk
...
+$ bundle exec jekyll serve --source build/site/
```
Tylos: @calvellido I have tried to run this in my machine... but the local webpage did not start! Could you double check? Not sure if it is something I have done wrong with the latest changes to documentation

CONTRIBUTING.md
calvellido: 🤔 weird no? Will give it a try

Tylos: Notice I had to install bundler so it manages the dependencies

JorgeCastilloPrz:
Huge effort @Tylos 👏 👏 , thanks for it!

CONTRIBUTING.md
calvellido: I think what's happening is that considering you are installing the gem dependencies on that docs directory, when you run `bundle exec` you are calling the installed dependency, so you need to be in the directory where it is installed.

I propose to change the order to be:

```sh
$ gem install bundler

$ cd modules/docs/arrow-docs/
$ bundle install --path vendor/bundle
$ bundle exec jekyll serve --source build/site
```

This will install the dependencies locally, which I also feel safer, so we will be needing to update the `.gitignore` file to something like:

```
build

## Jekyll
_site
.sass-cache
.jekyll-metadata

## OSX shit
.DS_Store

## Ruby environment normalization:
/.bundle/
/vendor/bundle
/lib/bundler/man/
```

PR #1170, opened 2018-11-30
Fix Travis timeouts because no output was received

calvellido:
* Fix a problem that may occur when building and deploying docs where the Travis job would fail when no output is received after 10 minutes. `jekyll build` is north of 9 minutes on my machine so there's a potential risk of the Travis job eventually failing due to this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,7 @@ env:
  global:
  # include $HOME/.local/bin for `aws`
  - PATH=$HOME/.local/bin:$PATH
+ - JAVA_OPTS="-Xms512m -Xmx1024m"
 
 before_install:
 - pip install --user awscli
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,6 +49,6 @@ jobs:
    script:
    - aws s3 sync s3://$S3_BUCKET/site ~/site > aws_sync_jekyll.log
    - rvm use 2.5.3 --install --fuzzy
-   - bundle install --gemfile modules/docs/arrow-docs/Gemfile --path vendor/bundle   
-   - bundle exec jekyll build -s ~/site -d ~/_site
+   - bundle install --gemfile modules/docs/arrow-docs/Gemfile --path vendor/bundle
+   - travis_wait BUNDLE_GEMFILE=modules/docs/arrow-docs/Gemfile bundle exec jekyll build -s ~/site -d ~/_site
    - aws s3 sync ~/_site s3://$S3_BUCKET > aws_sync_site.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,11 +32,11 @@ jobs:
    script:
    # redirecting outputs to files to avoid exceeding travis' logs limit
    - ./gradlew clean build > build.log
-   - ./gradlew dokka :arrow-docs:runAnk > run_ank.log
+   - travis_wait 30 ./gradlew dokka :arrow-docs:runAnk > run_ank.log
  - stage: build-and-dokka-transfer
    script:
    - ./gradlew clean build > build.log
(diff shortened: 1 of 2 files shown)


dominv:
:godmode:

raulraja:
We may also try caching the jekyll build folder in travis and run jekyll in incremental mode in case that helps avoiding an entire site build with the ~14K pages

calvellido:
Sorry for the noise on builds, Travis and such.

`travis_wait`  seems to be broken then? :man_facepalming:  I am trying to use it because if there's no response (output) on a process for more than 10 minutes Travis fails. We could fix that by setting some verbose option to the log for the tasks that doesn't print anything while processing (like `jekyll build`), but then if the log surpasses a 4MB size limit (which some current tasks can easily reach it) Travis fails.

For example, due to that 4MB log limit, we are sending the output of dokka to a file:

`./gradlew dokka :arrow-docs:runAnk > run_ank.log`

But then we are hit by the 10 minutes thing :disappointed: 

@raulraja while the incremental option would help with the Jekyll build I think that we'd still need some management of the dokka task, for that same reason.

calvellido:
If `travis_wait` is broken, could we maybe simulate its behaviour through a shell script?

1Jajen1:
What is the reason for sending the log to special files anyway? I mean yeah it hit the 10k line limit, but the raw log was always available. The last build before that change had 2.6Mb of logs with most of that  being uploading to github (removing all "uploading" lines brought it down to 860Kb), now assuming an s3 upload does not produce as many logged lines it should be fine. Also the logs will be accessible, with them being in files for example I have no idea how to actually get them o.O

The only need for travis wait would be for the jekyll task. Btw are you not missing a time in between `travis_wait BUNDLE_GEMFILE=modules/docs/arrow-docs/Gemfile bundle exec jekyll build -s ~/site -d ~/_site` :see_no_evil:

calvellido:
It's a way to prevent clogging the log by not printing it to the standard output to not reach the 4MB limit per build (https://docs.travis-ci.com/user/common-build-problems/#log-length-exceeded). AFAIK this is not related to the raw or display log, it can't reach 4MB total log or else will fail.

There's no GitHub upload, as we reached the limit for that. We are now serving the web page from Amazon S3. As there's no state saving between jobs, we first upload the dokka generated files to the Amazon S3 bucket, and as we need that to `jekyll_build` the page, we recover (download it) in its own job.

This upload/download is done that way because there's a 50 minutes time per job and we couldn't complete the Arrow build -> dokka -> `jekyll_build` in time for a single job.

Also, as far as I know, the default interval for `travis_wait` with no time is 10 minutes (https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received).

1Jajen1:
Well that is why I got the last logs that were not piped to its own files. The jobs that did get suppressed ran fine in there producing about 860kb of logs. No way that is ever gonna crack the 4Mb limit. That build also still had github upload (which had really verbose logging, so I removed that because I was assuming s3 was better).

All other jobs still have output that does not get sent to logs for some reason, I actually think that is the only reason they don't get terminated by the 10min limit.

Imo just printing the logs to stdout again (so everyone can easily see and access it) and putting travis wait for ~20 mins on jekyll build (because that is actually the only one without any display log for the whole build time) should be enough. Maybe thats worth a test :see_no_evil: 

Btw the default interval is 20 minutes, You did not actually need the time there, I was wrong :)

calvellido:
Anyway, now it fails with a

```Java HotSpot(TM) 64-Bit Server VM warning: INFO: os::commit_memory(0x00000000c1980000, 263716864, 0) failed; error='Cannot allocate memory' (errno=12)```

:sob:

Issue #284, opened 2017-09-16
Apply and Divide typeclasses


pakoito:
https://leanpub.com/fpmortals/read#leanpub-auto-everything-but-pure

https://leanpub.com/fpmortals/read#leanpub-auto-divide-and-conquer

1Jajen1:
The divide/divisible/decidable part of this is done in #1311

pedrovgs:
@pakoito @raulraja I could take care of the missing part of this issue tomorrow. I think the missing typeclass is Apply, am I wrong? If so, could you please elaborate what's the expected result of this task? I was thinking in something like:

* Implement the missing ``Apply`` typeclass.
* Refactor ``Applicative`` typeclass to extend from ``Apply``.
* Create the Apply laws.
* Create the documentation associated with this new typeclass.

The implementation should be backward compatible if I'm not wrong

pakoito:
Yeah, that's pretty much it

pedrovgs:
Ok, let's see if I can send a PR today with the implementation 😃

truizlop:
In case it helps, this is the implementation of Selective that we included in Bow: https://github.com/bow-swift/bow/pull/183

pedrovgs:
Thanks @truizlop 😃

JorgeCastilloPrz:
I think we'd also need to double check the data types that were supporting `Applicative` until now (aka: provided an instance for `Applicative`) 🤔. I'd say `Apply` was split from `Applicative` in Cats for a given reason, probably because they make sense as separate contracts since there could be data types that satisfy the constraints of `Apply` but not the ones from `Applicative`. @raulraja should be able to provide more clues about which types to review. I'm afraid this could bring some side refactors in. We'd also need to ensure everything still makes sense from the law tests perspective (as in we're still using the proper data types to run them).

pedrovgs:
I've been reviewing laws and the implementation for a while and the most important code changes will be related to the laws right now. The applicative instances shouldn't change at all for now because ``Applicative`` will extend ``Apply`` with the same signature. The good part of having ``Apply`` is you can decide when to provide an instance or not for this class. My idea is to provide the implementation change in a backward compatible way and then modify the rest of the code if needed.

raulraja:
I believe all instances of `Apply` will need to be added since the `@extension` processor will only export syntax for declared members in the type class interface. This means that a bunch of members currently exported from `Applicative` will disappear as you move them to apply. For example:

```
[...]

PR #1181, opened 2018-12-07
[WIP] Divide typeclasses

1Jajen1:
Finally got around to doing something :)
Partially solves #284 (Only divide not apply, because apply is a lot of refractoring and has some annoyances when it comes to defaults for map etc.)

Docs are mostly wip atm.

Few things:
- There is no test for StateT because it explicitly requires a monad that is divisible and we don't have that afaik)
- There are no laws for decidable atm, there are some [here](http://hackage.haskell.org/package/contravariant-1.5/docs/src/Data.Functor.Contravariant.Divisible.html), but I don't really understand them well enough to attempt to port them :/ Scala cats and/or scalaz don't include them either afaik.
- Docs for Divisible are a bit annoying as tbh I don't quite see a good example usecase for conquer...
- Should the typeclass docs include an entry for each typeclass or would also work in one file as they share a lot of functionality (or better they share usecases)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -126,3 +126,4 @@ reports/
 .DS_Store
 
 target/
+_site/
--- a/.gitignore
+++ b/.gitignore
@@ -126,4 +126,3 @@ reports/
 .DS_Store
 
 target/
-_site/
--- a/modules/core/arrow-core/src/test/kotlin/arrow/core/ConstTest.kt
+++ b/modules/core/arrow-core/src/test/kotlin/arrow/core/ConstTest.kt
@@ -1,0 +1,23 @@
+package arrow.core
+
+import arrow.instances.const.divisible.divisible
+import arrow.instances.monoid
+import arrow.test.UnitSpec
+import arrow.test.laws.DivisibleLaws
+import arrow.typeclasses.Const
+import arrow.typeclasses.Eq
+import arrow.typeclasses.fix
+import io.kotlintest.KTestJUnitRunner
+import org.junit.runner.RunWith
+
+@RunWith(KTestJUnitRunner::class)
+class ConstTest : UnitSpec() {
+
+  init {
+    testLaws(
+      DivisibleLaws.laws(Const.divisible(Int.monoid()), { i -> Const(i) }, Eq { a, b ->
+        a.fix().value == b.fix().value
+      })
+    )
+  }
+}
--- a/modules/core/arrow-core/src/test/kotlin/arrow/core/Function1Test.kt
+++ b/modules/core/arrow-core/src/test/kotlin/arrow/core/Function1Test.kt
@@ -1,19 +1,13 @@
 package arrow.core
 
 import arrow.Kind
-import arrow.core.ForFunction1
-import arrow.core.Function1
-import arrow.core.Function1Of
-import arrow.core.invoke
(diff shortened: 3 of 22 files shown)


1Jajen1:
Before I forget:

There are a few instances that would be really nice to have: Eq, Hash, Order.
But that does not work since they are a) not higherkind and b) even if they would be, since they inherit from each other not all of them can be.

Scala seems to not have that problem afaik (never used it but looking at it higher kinds can extend each other over there)

Haskell has a different problem (TypeClasses cannot be concrete types) and solves it by providing wrappers.
The wrapper solution to me sounds like the best one here, as one could easily add extension functions to convert back and forth (if needed, since the wrapper would delegate to the wrapped typeclass anyway).

Any ideas/thoughts?

raulraja:
@1Jajen1 do you have an example of those instances that illustrates the problem?

1Jajen1:
The problem is not with instances, but more with higher kinded types and how they are emulated atm. (assuming thats what you meant :) )
If `Eq<A>` were to inherit from `EqOf<A>` then `Order<A>` which inherits from `Eq<A>` cannot be higher kinded on its own because it can't inherit from `OrderOf<A>` (`Kind<ForOrder, A>` can't extend `Kind<ForEq, A>` afaik).
So for these typeclasses to have instances they need to be higher kinded, which they cannot be if I understand it correctly.

1Jajen1:
I can write up an actual code example later (or maybe tomorrow) when I am back home.

raulraja:
Monoid is higher kinded and it has an instance for `IO` as well as having one for `String` where `IO` is a type constructor and `String` is a monomorphic type, in case that is an indication that the issue you describe may be possible to encode.

1Jajen1:
Well yes, but monoid does not inherit from another higher kinded type does it? o.O
If now `Semigroup` would be changed to be higher kinded as well I don't think it'll work.

Making a typeclass higher kinded and writing instances for them worked fine, the only problem is that inheriting from a higher kinded type while being a higher kinded type as well does not work.

1Jajen1:
Code that hopefully outlines this properly.
```kotlin
class ForEq private constructor() { companion object }
typealias EqOf<A> = arrow.Kind<ForEq, A>
fun <A> EqOf<A>.fix(): Eq<A> = this as Eq<A>

interface Eq<A> : EqOf<A>

class ForOrder private constructor() { companion object }
typealias OrderOf<A> = arrow.Kind<ForOrder, A>
fun <A> OrderOf<A>.fix(): Order<A> = this as Order<A>

interface Order<A> : Eq<A>, OrderOf<A> // Type parameter F of Kind has inconsistent values: ForEq, ForOrder
```

raulraja:
For what is worth we will send a KEEP for higher kinds after the one we sent for extensions.

1Jajen1:
Aren't those two in keep 87 atm? Or is that going to be split? 
Anyway, in the meantime should this be implemented by a wrapper type or just ignore it for now?

raulraja:
They will be separated. I think ignore for now is probably best so we have less to port over later.

1Jajen1:
I'll close this for now. Will reopen a new pr for this tomorrow (I messed something up with git so I can't really salvage this :see_no_evil: )

Issue #1179, opened 2018-12-06
Allow Either to have a filter function that allows for the value to be used in the 'default' case


mikezx6r:
This is very helpful if the Either and `filterOrElse` is being used for validation like processing.

Currently, one needs to do this

```
val someValue = getTheProperty()
someValue.right()
.filterOrElse( {it.length == 12},{MyError("some message with context: ${someValue.length}") })
```

The proposal (after discussion on Slack) is to introduce a new function as below:

```
inline fun <A, B> EitherOf<A, B>.filterOrOther(predicate: (B) -> Boolean, default: (B) -> A): Either<A, B> =
  flatMap { if (predicate(it)) Right(it) else Left(default(it)) }
```

This will allow for the example at top to be rewritten as

```
someValue.right()
.filterOrOther({it.length == 12}, {Error("some message with context: ${it.length}")})
```

I will be creating a PR for this later today.

pakoito:
I believe this has been merged already? Closing!

PR #1183, opened 2018-12-08
(#1179) Add function to Either similar to filterOrElse but that provi…

mikezx6r:
…des value to 'default' for error context

filterOrElse is very useful for validating (filtering) a value and returning error info in the Left, BUT
it's nice to include context from the value in the Left error object. filterOrElse doesn't pass anything to
the 'default' function, so this PR adds `filterOrOther` that functions exactly like `filterOrElse` but also
passes the value to the default function so that context can be extracted from the value without having to
explicitly define it in an outer scope

### DIFF
--- a/modules/core/arrow-core/src/main/kotlin/arrow/core/Eval.kt
+++ b/modules/core/arrow-core/src/main/kotlin/arrow/core/Eval.kt
@@ -2,6 +2,8 @@ package arrow.core
 
 import arrow.higherkind
 
+fun <A> EvalOf<A>.value(): A = this.fix().value()
+
 /**
  * Eval is a monad which controls evaluation of a value or a computation that produces a value.
  *
--- a/modules/core/arrow-core/src/main/kotlin/arrow/core/Id.kt
+++ b/modules/core/arrow-core/src/main/kotlin/arrow/core/Id.kt
@@ -2,29 +2,31 @@ package arrow.core
 
 import arrow.higherkind
 
-fun <A> IdOf<A>.value(): A = this.fix().value
+fun <A> IdOf<A>.value(): A = this.fix().value()
 
 @higherkind
-data class Id<out A>(val value: A) : IdOf<A> {
+data class Id<out A>(private val value: A) : IdOf<A> {
 
-  inline fun <B> map(f: (A) -> B): Id<B> = Id(f(value))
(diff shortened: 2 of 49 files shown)


modules/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -266,6 +266,26 @@ inline fun <A, B> EitherOf<A, B>.getOrHandle(default: (A) -> B): B =
...
+/**
+ * * Returns [Either.Right] with the existing value of [Either.Right] if this is a [Either.Right] and the given
+ * predicate holds for the right value.
+ * * Returns `Left(default({right}))` if this is a [Either.Right] and the given predicate does not
+ * hold for the right value. Useful for error handling where 'default' returns a message with context on why the value
+ * did not pass the filter
+ * * Returns [Either.Left] with the existing value of [Either.Left] if this is a [Either.Left].
+ *
+ * Example:
+ * ```
```
raulraja: We should use here a Kotlin ank executable doc snippet so it's type checked and runnable in the docs . See the docs module for examples

mikezx6r: Oops. Obviously didn't read the comment well enough.

Updated to reflect function it represents, with an example that returns Left and uses the Right's value.

pakoito: What Raul is saying is that the docs tooling we use can execute the results of these snippets, rather than us hardcoding them. See for example:

https://github.com/arrow-kt/arrow/edit/master/modules/docs/arrow-docs/docs/docs/datatypes/option/README.md

What you can do is split this snippet in two sub-snippets, each starting with ```kotlin:ank

raulraja: I was actually referring to the executable ones not just regular ank snippets. https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Bracket.kt#L40-L72

mikezx6r: Ok. I'll take a look at the examples, figure out what's required, and update accordingly

PR #1186, opened 2018-12-10
(#1179) Add function to Either similar to filterOrElse but that provi…

mikezx6r:
…des value to 'default' for error context

filterOrElse is very useful for validating (filtering) a value and returning error info in the Left, BUT
it's nice to include context from the value in the Left error object. filterOrElse doesn't pass anything to
the 'default' function, so this PR adds `filterOrOther` that functions exactly like `filterOrElse` but also
passes the value to the default function so that context can be extracted from the value without having to
explicitly define it in an outer scope

@raulraja @pakoito 
Note that I accidentally closed the other PR, and this comes from a branch in my fork rather than from master.

I've added the ank, and an executable block into the comments. If that isn't quite what's meant, then I'll require a bit more guidance.

### DIFF
--- a/modules/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
+++ b/modules/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
@@ -267,6 +267,36 @@ inline fun <A, B> EitherOf<A, B>.filterOrElse(predicate: (B) -> Boolean, default
   flatMap { if (predicate(it)) Right(it) else Left(default()) }
 
 /**
+ * * Returns [Either.Right] with the existing value of [Either.Right] if this is a [Either.Right] and the given
+ * predicate holds for the right value.
+ * * Returns `Left(default({right}))` if this is a [Either.Right] and the given predicate does not
+ * hold for the right value. Useful for error handling where 'default' returns a message with context on why the value
+ * did not pass the filter
+ * * Returns [Either.Left] with the existing value of [Either.Left] if this is a [Either.Left].
+ *
+ * Example:
+ *
+ * {: data-executable='true'}
+ * ```kotlin:ank
+ * import arrow.core.*
+ *
+ * Right(12).filterOrOther({ it > 10 }, { -1 })
+ * ```
+ *
+ * {: data-executable='true'}
+ * ```kotlin:ank
+ * Right(7).filterOrOther({ it > 10 }, { "Value '$it' not greater than 10" })
+ * ```
+ *
+ * {: data-executable='true'}
+ * ```kotlin:ank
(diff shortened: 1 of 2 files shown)


modules/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -266,6 +266,28 @@ inline fun <A, B> EitherOf<A, B>.getOrHandle(default: (A) -> B): B =
...
+ * {: data-executable='true'}
+ * ```kotlin:ank
+ *
+ * Right(12).filterOrOther({ it > 10 }, { -1 }) // Result: Right(12)
+ * Right(7).filterOrOther({ it > 10 }, { "Value '$it' not greater than 10" })  // Result: Left("Value '7' not greater than 10")
+ *
+ * val left: Either<Int, Int> = Left(12)
+ * left.filterOrOther({ it > 10 }, { -1 })      // Result: Left(12)
+ * ```
+ */
```
pakoito: There are a couple of issues with the docs we've not flagged, let me give you the comment to have:

```
/**
 * * Returns [Either.Right] with the existing value of [Either.Right] if this is a [Either.Right] and the given
 * predicate holds for the right value.
 * * Returns `Left(default({right}))` if this is a [Either.Right] and the given predicate does not
 * hold for the right value. Useful for error handling where 'default' returns a message with context on why the value
 * did not pass the filter
 * * Returns [Either.Left] with the existing value of [Either.Left] if this is a [Either.Left].
 *
 * Example:
 *
 * {: data-executable='true'}
 * ```kotlin:ank
 * Right(12).filterOrOther({ it > 10 }, { -1 }) // Result: Right(12)
 * Right(7).filterOrOther({ it > 10 }, { "Value '$it' not greater than 10" }) 
 * ```
 *
 * {: data-executable='true'}
 * ```kotlin:ank
 * val left: Either<Int, Int> = Left(12)
 * left.filterOrOther({ it > 10 }, { -1 })
 * ```
 */
```

mikezx6r: Ok, this is beginning to make sense. So if there can only be one result/block, I assume that the two `Right` examples need to be split also?

I'll assume that's correct, and update the PR accordingly.

Thank you

pakoito: Correct, I missed it!

mikezx6r:
Updated and build passes. Let me know if there's anything else required for this.

pakoito:
Thanks @mikezx6r for the contribution, I hope you understand our docs tooling better :D The results will look awesome in the docsite, you'll see ;)

PR #1189, opened 2018-12-11
Incorrect Bracket instances

nomisRev:
@raulraja @pakoito @JorgeCastilloPrz The behavior currently encoded in the bracket instances is incorrect for all instances except `IO`.

As you can see in the current code the task returned from `release` is never executed. I rectified this behavior but it's quite obvious that currently the `ExitCase.Cancelled` case is not supported by any of these types.

Should we leave that as is? Before we merge I want to investigate how it was possible for all the laws & test to passed and for this to still sneak into a release.


pakoito:
How would we test that the behavior was originally wrong, and that it doesn't regress?

nomisRev:
@pakoito I still need to investigate the laws for Bracket.

By adding a law like below we can assure we don't regress in the future.

```
Ref.of(0, IO.monadDefer()).flatMap { ref ->
  IO.just(1).bracketCase(use = { IO.unit }, release = { i, _ ->
    ref.set(i + 1)
  }).flatMap { ref.get }
}.unsafeRunSync() shouldBe 2
```

nomisRev:
@JorgeCastilloPrz First of all the `doOnXXX` operators do not run the returned tasks by `release`.

Signature of `Single<B>#doOnSuccess` is `(B) -> Unit` so we still have to execute the `SingleK<Unit>` returned by `release`. To do so you need `subscribe()` but this is not safe since it can throw, etc etc. The existing operators don't have this problem, and that is why I choose them here.

pakoito:
Can we get @gildor to lend a hand for Deferred? I have my ideas of how it works, he'll know better.

JorgeCastilloPrz:
I see, I definitely think blowing the stack is not an option either. In the other hand you lose the already working support for cancellation, error and success callbacks for those data types (except for Deferred which has the problem we know about cancellation) 🤔 

About the laws it's interesting that there was a test for `cancellation` being properly called in Flowable, Mono and a few more, and it's actually passing. Maybe the test is wrong also, but we should double check how's that passing.

gildor:
@nomisRev Deferred is a child of Job, so you can add callback using `Job.invokeOnCompletion()` function to clean up resources. also you can check `Job.isCancelled` property

pakoito:
@gildor How can you inherit a cancellation Job and the scope from an existing Deferred?

nomisRev:
This branch builds fine locally using `./gradlew clean build test dokka runAnk` 🤷‍♂️

pakoito:
It's OOMing while generating the logs

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/typeclasses/Bracket.kt
```diff
@@ -47,30 +47,31 @@ interface Bracket<F, E> : MonadError<F, E> {
    *
    * class File(url: String) {
    *   fun open(): File = this
-   *   fun close(): Unit {}
```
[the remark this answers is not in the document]
nomisRev: Ow that must've sneaked in. Typo

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredKConnection.kt
```diff
@@ -20,19 +21,19 @@ typealias DeferredKProc<A> = (DeferredKConnection, (Either<Throwable, A>) -> Uni
...
   override fun <A> DeferredKOf<A>.handleErrorWith(f: (Throwable) -> DeferredKOf<A>): DeferredK<A> =
-    fix().handleErrorWith(f)
+    handleErrorW(f)
 
   override fun <A> just(a: A): DeferredK<A> =
     DeferredK.just(a)
 
   override fun <A, B> DeferredKOf<A>.flatMap(f: (A) -> DeferredKOf<B>): DeferredK<B> =
-    fix().flatMap(f)
+    fix().flatMap { f(it) }
```
pakoito: This extra allocation can be avoided

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredKConnection.kt
```diff
@@ -20,19 +21,19 @@ typealias DeferredKProc<A> = (DeferredKConnection, (Either<Throwable, A>) -> Uni
 @Suppress("FunctionName")
 fun DeferredKConnection(dummy: Unit = Unit): KindConnection<ForDeferredK> = KindConnection(object : MonadDefer<ForDeferredK> {
   override fun <A> defer(fa: () -> DeferredKOf<A>): DeferredK<A> =
-    DeferredK.defer(fa = fa)
+    DeferredK.defer(fa = { fa() })
```
pakoito: Unnecessary extra allocation

modules/effects/arrow-effects-kotlinx-coroutines-instances/src/test/kotlin/arrow/effects/DeferredTest.kt
```diff
@@ -170,35 +170,67 @@ class DeferredKTest : UnitSpec() {
...
+        DeferredK.async<Unit> { conn, cb ->
+          conn.push(latch.complete(Unit))
+          cb(Right(Unit))
+        }.flatMap {
+          DeferredK.async<Unit> { conn, _ ->
+            conn.cancel().fix().unsafeRunAsync { }
+          }
+        }.unsafeRunAsyncCancellable { }
+
+        latch.get
```
pakoito: Can we add a timeout here to not block the tests in case of failure?

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/Utils.kt
```diff
@@ -74,6 +74,19 @@ object Platform {
       is Either.Right -> Some(eitherRef.b)
     }
   }
+
+  /**
+   * Composes multiple errors together, meant for those cases in which error suppression, due to a second error being
+   * triggered, is not acceptable.
+   *
+   * On top of the JVM this function uses Throwable#addSuppressed, available since Java 7. On top of JavaScript the
```
pakoito: Does this mean it may not work on older versions of Android? Or was this pat of the "Java 6.5" update too?

pakoito: https://developer.android.com/reference/java/lang/Throwable#addSuppressed(java.lang.Throwable)

Minimum API 19. What's our expected level of support these days?

modules/effects/arrow-effects-kotlinx-coroutines-instances/src/main/kotlin/arrow/effects/DeferredKInstances.kt
```diff
@@ -81,7 +81,7 @@ interface DeferredKBracketInstance : Bracket<ForDeferredK, Throwable>, DeferredK
 @extension
 interface DeferredKMonadDeferInstance : MonadDefer<ForDeferredK>, DeferredKBracketInstance {
   override fun <A> defer(fa: () -> DeferredKOf<A>): DeferredK<A> =
-    DeferredK.defer(fa = fa)
+    DeferredK.defer(fa = { fa() })
```
pakoito: Unnecessary allocation. We can just pass `fa`, without the named parameter right?

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt
```diff
@@ -533,6 +616,59 @@ sealed class DeferredK<A>(
         }
       }
   }
+
+  override val children: Sequence<Job>
```
pakoito: Isn't this covered by `Deferred<A> by memoized`?

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt
```diff
@@ -633,9 +773,17 @@ fun <A> DeferredKOf<A>.unsafeRunSync(): A =
  * ```
  */
 fun <A> DeferredKOf<A>.runAsync(cb: (Either<Throwable, A>) -> DeferredKOf<Unit>): DeferredK<Unit> =
-  DeferredK(scope(), Dispatchers.Unconfined, CoroutineStart.DEFAULT) {
-    fix().forceExceptionPropagation()
-    unsafeRunAsync(cb.andThen { it.unsafeRunAsync { } })
+  DeferredK(scope(), Dispatchers.Unconfined, CoroutineStart.LAZY) {
```
pakoito: Nice catch with the `CoroutineStart.LAZY`. Only `unsafeRunAsync` should be `default`.

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt
```diff
@@ -633,9 +773,17 @@ fun <A> DeferredKOf<A>.unsafeRunSync(): A =
...
+    val result: A? = try {
+      await()
+    } catch (t: Throwable) {
+      cb(Left(t)).await()
+      null
+    }
+
+    result?.let { cb(Right(it)) }?.await()
+
+    Unit
```
pakoito: This shouldn't be necessary, right? Put elvish on the previous line, or `.let { }`

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt
```diff
@@ -146,22 +155,78 @@ sealed class DeferredK<A>(
...
+      /**
+       * Returns either the memoized [Deferred] if it has not been run yet. Or creates a new one.
+       */
+      override suspend fun await(): A = value().await()
+
+      override fun start(): Boolean = value().start()
+
+      override fun value(): Deferred<A> {
+        if (isCompleted || isActive || isCancelled) {
+          memoized = scope.async(ctx, coroutineStart) {
```
pakoito: This looks thread unsafe. Should it be at least wrapped in synchronized?

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt
nomisRev: Not since we had to change `memoized` from `val` to `var`. Kotlin only has support for `val` but it also possible to use with `var`... It's a little bit nasty tho.

A short example of what kotlinc spits out:

```
public final class Sample implements Deferred {
   @NotNull
   private Deferred memoized;
   // $FF: synthetic field
   private final Deferred $$delegate_0;

   @NotNull
   public final Deferred getMemoized() {
      return this.memoized;
   }

   public final void setMemoized(@NotNull Deferred var1) {
      Intrinsics.checkParameterIsNotNull(var1, "<set-?>");
      this.memoized = var1;
   }

   public Sample(@NotNull Deferred memoized) {
      Intrinsics.checkParameterIsNotNull(memoized, "memoized");
      super();
      this.$$delegate_0 = memoized;
      this.memoized = memoized;
   }

   @NotNull
   public Sequence getChildren() {
      return this.$$delegate_0.getChildren();
   }

   public boolean isActive() {
      return this.$$delegate_0.isActive();
   }

   public boolean isCancelled() {
      return this.$$delegate_0.isCancelled();
   }
```

modules/effects/arrow-effects-kotlinx-coroutines-instances/src/test/kotlin/arrow/effects/DeferredTest.kt
nomisRev: The problem we have now is that we're kind of in an in-between situation where we already have a lot of `cancellable` logic but we don't have any typeclass support for it yet. So all the current concurrency primitives like `Ref`, `Promise`, `Semaphore` etc all have non-cancellable operations since that's the only support we currently get by typeclasses. This is why I am trying to merge these PRs asap so I can write proper laws for this logic, at that point, we will remove all these tests from the concrete types.

modules/effects/arrow-effects/src/main/kotlin/arrow/effects/internal/Utils.kt
nomisRev: Good question, I completely missed this when working on it.

Since `minSdk` tweeted in Nov 2016 that they recommend min API 19, I'd think we're good. [source](https://twitter.com/minSdkVersion/status/795702388774567936). But perhaps we should ask around? I'll create a ticket for this.

modules/effects/arrow-effects-kotlinx-coroutines-instances/src/main/kotlin/arrow/effects/DeferredKInstances.kt
nomisRev: This was necessary because I changed signature of `defer` to `f: CoroutineScope.() -> DeferredK<A>` but this seems to not be necessary.

When reverting this back I solved the actual issue. I will create a seperate PR with the fix, since this is already merged and I don't want to do it in the `asyncF` PR. I will also add an explanation there.

modules/effects/arrow-effects-kotlinx-coroutines/src/main/kotlin/arrow/effects/DeferredK.kt
nomisRev: Good catch

PR #1194, opened 2018-12-12
Update and rename custom.md to feature_request.md

pakoito:
The current custom.md didn't make sense when making a new issue.

screen shot 2018-12-12 at 11 40 28 pm


MaureenElsberry:
@pakoito - looks like it added an extra blank one when I created the other two, it can be deleted!

pakoito:
@MaureenElsberry I thought about that, had it in a branch and everything, but actually I don't mind making it a new template instead :D

Feel free to change the copy, I just threw the first few things I thought could be valuable.

PR #1221, opened 2018-12-25
Added arrow-query-language in the dependencies

aasaandinesh:
arrow-query-language was missing from the list of dependencies. Added it in the readme.


pakoito:
Thanks for the PR!

PR #1231, opened 2019-01-04
Add DeferredKMonadThrow

daneko:


modules/effects/arrow-effects-kotlinx-coroutines-extensions/src/main/kotlin/arrow/effects/coroutines/extensions/deferredk.kt
```diff
@@ -71,6 +71,9 @@ interface DeferredKMonadError: MonadError<ForDeferredK, Throwable>, DeferredKMon
     deferredHandleErrorWith { f(it).fix() }
 }
 
+@extension
+interface DeferredKMonadThrow: MonadThrow<ForDeferredK>, DeferredKMonadError
+
 @extension
 interface DeferredKBracket: Bracket<ForDeferredK, Throwable>, DeferredKMonadError {
```
pakoito: `interface DeferredKBracket: Bracket<ForDeferredK, Throwable>, DeferredKMonadThrow {`

pakoito: Or maybe this goes in MonadDefer. Please check the other implementations :D

daneko: thinks poiting out, and fixing and check other impl.
DeferredKMonadDeferred already used DeferredKBracket, so I fixed DeferredKBracket using DeferredKMonadThrow.

(compare to [io.kt](https://github.com/arrow-kt/arrow/blob/0d261e6d7c961d67a3143f036ce0871f78f778d0/modules/effects/arrow-effects-extensions/src/main/kotlin/arrow/effects/extensions/io.kt) and [singlek.kt](https://github.com/arrow-kt/arrow/blob/bac9c1b0f16ce1ebdf00bf173066f91608ed3b13/modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/singlek.kt))

daneko: by the way, I have some question.

1. [DeferredKAsync#invoke](https://github.com/arrow-kt/arrow/blob/64d55d71cd5c3715d102985f9a71d1164d3c0881/modules/effects/arrow-effects-kotlinx-coroutines-extensions/src/main/kotlin/arrow/effects/coroutines/extensions/deferredk.kt#L106) is deprecated, 
so can I deleted it in this PR? or make other PR?

2. Which do you prefer implementation io.kt or singlek.kt (or else) ? (when I implement it) 
    - In case of Bracket, [singlek (and other rx ext) override only bracketCase](https://github.com/arrow-kt/arrow/blob/bac9c1b0f16ce1ebdf00bf173066f91608ed3b13/modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/singlek.kt#L75), on the other hand, [io override bracketCase, bracket, guaranteea and guaranteeCase](https://github.com/arrow-kt/arrow/blob/0d261e6d7c961d67a3143f036ce0871f78f778d0/modules/effects/arrow-effects-extensions/src/main/kotlin/arrow/effects/extensions/io.kt#L82). (I can't choose it...)
    - In case of MonadError, [io impl applicative method](https://github.com/arrow-kt/arrow/blob/0d261e6d7c961d67a3143f036ce0871f78f778d0/modules/effects/arrow-effects-extensions/src/main/kotlin/arrow/effects/extensions/io.kt#L58), other side [singlek impl applicative-error method](https://github.com/arrow-kt/arrow/blob/bac9c1b0f16ce1ebdf00bf173066f91608ed3b13/modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/singlek.kt#L61). ( I like singlek's ver ^_^ )


(Please excuse my poor English and these question are little related this PR.)

nomisRev: For the typeclasses, the derived functions are often less optimal than a concrete version. And thus if a data type has a concrete impl is available, it's always beneficial to override it.

The only important thing is that laws don't break, which is checked by tests (laws).

`MonadThrow` sits between `MonadError` and `Bracket` so that look goods!

nomisRev: About deprecating [DeferredKAsync#invoke](https://github.com/arrow-kt/arrow/blob/64d55d71cd5c3715d102985f9a71d1164d3c0881/modules/effects/arrow-effects-kotlinx-coroutines-extensions/src/main/kotlin/arrow/effects/coroutines/extensions/deferredk.kt#L106) I am fine with removing it, but it's probably best to keep that for another PR. That way you'll also have the joy of closing and contributing 2 PRs ;)

About your second questions.
It seems that the `SingleK` instance is incorrect, since `MonadError` should be a valid instance for both `Monad` and `ApplicativeError`. Which is not the case for `SingleK`.

`interface SingleKMonadError : MonadError<ForSingleK, Throwable>, SingleKMonad`

As you can see it does not implement `ApplicativeError<ForSingleK, Throwable>` (`SingleKApplicativeErrorInstance`).

This is the reason [SingleK](https://github.com/arrow-kt/arrow/blob/bac9c1b0f16ce1ebdf00bf173066f91608ed3b13/modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/singlek.kt#L61) implements the `applicative-error` methods here.

The reason [io](https://github.com/arrow-kt/arrow/blob/0d261e6d7c961d67a3143f036ce0871f78f778d0/modules/effects/arrow-effects-extensions/src/main/kotlin/arrow/effects/extensions/io.kt#L58) (re)implements applicative methods is because they're both implemented by `Monad` & `ApplicativeError` and thus need to explitely redefine it to avoid the [diamond problem](https://en.wikipedia.org/wiki/Multiple_inheritance#The_diamond_problem)

PR #1243, opened 2019-01-13
Fix recursive docs

### LINKED ISSUE #1240: Cannot get recursion example to work
I'm trying to use run the recursion scheme example locally but I'm having a few problems.

First I can't get the `@instance` annotation to work. I've added `kapt` in maven and it works fine for `@higherkind`.

I got around it by making `IntListPatternFunctor` an object:

```kotlin
object IntListPatternFunctor : Functor<ForIntListPattern> {
    override fun <A, B> IntListPatternOf<A>.map(f: (A) -> B): IntListPatternOf<B> {
        val lp = fix()
        return when (lp) {
            NilPattern -> NilPattern
            is ConsPattern -> ConsPattern(lp.head, f(lp.tail))
        }
    }
}
```

However I get stuck when trying to define the `multiply` function:

```kotlin
fun multiply(list: IntFixList): Int = Fix.recursive().run {
    IntListPatternFunctor.cata(list, multiply)
}
```

The compiler cannot find `recursive` in `Fix` and cannot find `cata` in the `IntListPattern` functor. However I can see `IntListPatternFunctor.hylo` function.

Is there something I am missing?

AdrianRaFo:
Fixes #1240 

Updates and add `ank` to recursive sample and glossary pattern


modules/docs/arrow-docs/docs/docs/recursion/intro/README.md
```diff
@@ -185,12 +189,12 @@ The [Recursive]({{ '/docs/recursion/recursive' | relative_url }}) typeclass prov
...
 similar to fold and unfold.
 
-```kotlin
+```kotlin:ank:silent
 typealias Algebra<F, A> = (Kind<F, A>) -> A     // fold
 typealias Coalgebra<F, A> = (A) -> Kind<F, A>   // unfold
 
-fun <F, A> Functor<F>.cata(f: Fix<F>, alg: Algebra<F, Eval<A>>): A
-fun <F, A> Functor<F>.ana(a: A, coalg: Coalgebra<F, A>): Fix<F>
+fun <F, A> Functor<F>.cata(f: Fix<F>, alg: Algebra<F, Eval<A>>): A = TODO()
```
cesartl: What do you mean by `TODO`?
Does that mean that it's a feature that is yet to come?

AdrianRaFo: not at all, it just to compile the doc because I cannot declare an abstract method at the top of the file

cesartl: ok thanks. Any reason why my compiler wouldn't see the `cata` and `ana` method of Functor? I can see `hylo` without problem

AdrianRaFo: those are in separate files/imports:
hylo -> import arrow.recursion.*
cata -> import arrow.recursion.extensions.fix.recursive.*
ana -> import arrow.recursion.extensions.fix.corecursive.*

cesartl: ok. Unfortunately, I cannot resolve those import. I'm using `0.8.2`.
I don't see any `extensions` subpackage under `arrow.recursion`. Is there an additional module to be imported?

AdrianRaFo: From arrow main page:

    compile "io.arrow-kt:arrow-recursion:$arrow_version"
    compile "io.arrow-kt:arrow-extensions-recursion:$arrow_version"

cesartl: oh that was the missing bit! Unfortunately that wasn't in `https://arrow-kt.io/docs/quickstart/libraries/` so I missed it. I think it's finally working thanks!

AdrianRaFo: you're welcome

pakoito: @cesartl that's on me, apologies!

PR #1245, opened 2019-01-14
fix apidoc bad format

### LINKED ISSUE #1228: Formatting issue on data type generated docs
On data type docs the extension functions section is completely unreadable at the moment. [Example](https://arrow-kt.io/docs/apidocs/arrow-core/arrow.core/-either/index.html#extension-functions)

Can we do something to fix this @israelperezglez?

AdrianRaFo:
Fix bad format due to markdown list on table style

Fixes #1228

### DIFF
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -248,11 +248,13 @@ inline fun <A, B> EitherOf<A, B>.getOrHandle(default: (A) -> B): B =
   fix().fold({ default(it) }, ::identity)
 
 /**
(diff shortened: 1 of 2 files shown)


calvellido:
I don't think this fixes #1228 as a whole, does it?

AdrianRaFo:
I tested locally it does

calvellido:
I was under the assumption that #1228 was caused by a bad formatting generated by Dokka, @israelperezglez could you please clarify?

AdrianRaFo:
that could make sense if all the file was a bad format but only one section has it. Also, I found really weird the return lists only in that section so after delete the `*` which gives the bad list format it was fixes

JorgeCastilloPrz:
This is welcome but does not fix the issue. The issue is about wrongly rendered tables in the autogenerated sections of the doc. @israelperezglez was working on that.

PR #1251, opened 2019-01-16
Update analytics config

### LINKED ISSUE #1218: Configure Google Analytics in the docs site
We've currently lost all signals for traffic and referrals on Analytics. I check them regularly to see who's talking about us, and provide support where needed :D

See #482 for the original configuration.

calvellido:
* Update Google Analytics config.
* Set correct Jekyll environment when building through Bitrise (this is actually what made us lost it before, as when the website was built through GitHub, the environment was being set correctly automatically, but we didn't specify it on the change of building flow through Travis and/or Bitrise)
* Extrafix of extra quotation marks in the docs README

This closes #1218


modules/docs/arrow-docs/docs/_data/commons.yml
```diff
@@ -2,5 +2,5 @@ name: "Arrow"
 title: "Λrrow"
 description: "Functional companion to Kotlin's Standard Library"
 keywords: "functional-programming, kotlin-library, monads, monad-transformers, functional-data-structure, kotlin, fp-types, adt, free-monads, tagless-final, mtl, for-comprehension, category-theory"
-google_analytics: "UA-110153729-1"
+google_analytics: UA-18433785-18
```
pakoito: Shouldn't this be a string?

calvellido: It is a string, but YAML string syntax is a bit all over the place, because in general you don't need quotes unless you want to force convert something that doesn't look like a string (e.g. a number: `10`).

But then you can read different opinions on how correct should be to always use quotation marks, or not at all. Anyway, probably, for the sake of consistency with the rest of the file this should be in quotation marks, really.

I can correct it in a future PR.

PR #1252, opened 2019-01-16
Move cancelable to Concurrent

nomisRev:
This PR moves the `cancelable` and `cancelableF` builders to the `Concurrent` typeclass.

I originally placed them in `Async` since they can be derived with all combinators available in `Async` but after working on `Concurrent` I feel like this might confuse users regarding cancellation.

This PR also fixes all the current warnings for the effects modules, except our own currently deprecated methods.

### DIFF
--- a/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/internal/UncancelablePromise.kt
+++ b/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/internal/UncancelablePromise.kt
@@ -6,7 +6,7 @@ import arrow.effects.Promise
 import arrow.effects.typeclasses.Async
 import java.util.concurrent.atomic.AtomicReference
 
-internal class UncancelablePromise<F, A>(AS: Async<F>) : Promise<F, A>, Async<F> by AS {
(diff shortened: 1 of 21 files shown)


modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/internal/UncancelablePromise.kt
```diff
@@ -6,7 +6,7 @@ import arrow.effects.Promise
 import arrow.effects.typeclasses.Async
 import java.util.concurrent.atomic.AtomicReference
 
-internal class UncancelablePromise<F, A>(AS: Async<F>) : Promise<F, A>, Async<F> by AS {
+internal class UncancelablePromise<F, A>(private val AS: Async<F>) : Promise<F, A>, Async<F> by AS {
```
pakoito: This shouldn't be necessary, it should be an import alias instead, right?

nomisRev: There are multiple inheritance problems from `: Async`, if you don't explicitly override those methods it'll give a warning.

PR #1254, opened 2019-01-17
Concurrent defaults EmptyCoroutineContext 

nomisRev:


nomisRev:
> You can't have parallelism with an empty coroutine context, by its pure definition :D

Wut!? 😫

raulraja:
@pakoito that is true if the default is the EmptyCoroutineContext. What I'm proposing is that we have our own Default/IO dispatchers and make all concurrent ops default to the Default dispatcher based in the common pool so users can get the right behavior with defaults and optionally pass their own context if needed.

pakoito:
@raulraja 'bout time we did :D Agreed.

nomisRev:
Alright, shall I just close this then? @raulraja @pakoito

JorgeCastilloPrz:
Do we have parallelization tests? Were those passing with this change?

nomisRev:
@JorgeCastilloPrz We do have test for parallelization but none with the default value for ctx. They wouldn't pass I just tested locally.

raulraja:
@nomisRev Yeah we need a couple of dispatchers one for CPU bound ops and other for IO blocking ops and have all our context dependent ops wherever makes sense default to the CPU common pool style one.

nomisRev:
Closed and created ticket @ #1256

PR #1255, opened 2019-01-17
Solve update kotlintest problems

AdrianRaFo:
Fix the problem because the laws are not running.
Also fix some deprecations and a bad merge from master


modules/core/arrow-test/src/main/kotlin/arrow/test/UnitSpec.kt
```diff
@@ -11,28 +10,19 @@ import io.kotlintest.specs.AbstractStringSpec
...
+  private val lawTestCases = mutableListOf<TestCase>()
+
   fun testLaws(vararg laws: List<Law>): List<TestCase> {
-   laws
+    return laws
      .flatMap { list: List<Law> -> list.asIterable() }
      .distinctBy { law: Law -> law.name }
      .map { law: Law ->
-       createTestCase(law.name, law.test, defaultTestCaseConfig, TestType.Test)
+       val lawTestCase = createTestCase(law.name, law.test, defaultTestCaseConfig, TestType.Test)
```
nomisRev: Shouldn't we use `addTestCase` here?

modules/core/arrow-test/src/main/kotlin/arrow/test/UnitSpec.kt
```diff
@@ -11,28 +10,19 @@ import io.kotlintest.specs.AbstractStringSpec
...
+       lawTestCases.add(lawTestCase)
+       lawTestCase
      }
-   return testCases()
   }
 
+  override fun testCases(): List<TestCase> = lawTestCases
```
nomisRev: Are we still running the regular tests now?

modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/IOTest.kt
```diff
@@ -232,7 +232,7 @@ class IOTest : UnitSpec() {
       val order = mutableListOf<Long>()
 
       fun makePar(num: Long) =
-        IO(newSingleThreadContext("$num")) {
+        IO(newCoroutineDispatcher("$num")) {
```
nomisRev: This is certainly not correct because `newCoroutineDispatcher` underlying uses a `ThreadPoolExecutor` with a corePoolSize = 10 and a maxPoolSize = 50. We're using `newSingleThreadContext` to test that parallel code still works correcrtly when running on a single thread.

modules/optics/arrow-optics/src/test/kotlin/arrow/optics/std/ListTest.kt
```diff
@@ -35,9 +35,9 @@ class ListTest : UnitSpec() {
 
     testLaws(OptionalLaws.laws(
       optional = ListK.tail(),
-      aGen = Gen.list(Gen.int()),
-      bGen = Gen.list(Gen.int()),
-      funcGen = genFunctionAToB(Gen.list(Gen.int())),
+      aGen = genNonEmptyList(Gen.int()).map { it.all },
```
nomisRev: ?

modules/core/arrow-test/src/main/kotlin/arrow/test/UnitSpec.kt
AdrianRaFo: the add test case requires unique names and the test laws are duplicated in some many places

modules/optics/arrow-optics/src/test/kotlin/arrow/optics/std/ListTest.kt
AdrianRaFo: set is not consistent with modify on a empty list due to:
set add a value to an empty list
modify does nothing over an empty list

nomisRev: Did these test break in this PR?

modules/core/arrow-test/src/main/kotlin/arrow/test/UnitSpec.kt
nomisRev: We're missing tests this way, right? I.e.

`FunctorLaws.laws(IO.functor())` & `TraverseLaws.laws(IO.traverse())` the functor laws aren't tested for `Traverse<ForIO>` but the functor logic might not be defined by the same functor instance.

modules/optics/arrow-optics/src/test/kotlin/arrow/optics/std/ListTest.kt
AdrianRaFo: Yes, sorry

modules/core/arrow-test/src/main/kotlin/arrow/test/UnitSpec.kt
AdrianRaFo: For those cases, you can have multiple testLaws declarations and kotlintest will execute them without errors even with duplicated names, the best example is the [FlowableKTests](https://github.com/arrow-kt/arrow/blob/arf-solve-update-kotlintest-problems/modules/effects/arrow-effects-rx2-data/src/test/kotlin/arrow/effects/FlowableKTests.kt) where we execute multiple times the same laws with different parameters. This works now because, only for law test, I deleted the unique name restriction

raulraja:
@nomisRev I think the IO strategy bundling dispatchers is the best one we have so far. Are you suggesting we merge this as is? and continue in a different PR for that?

nomisRev:
@raulraja well, if we're going with that strategy than `DeferredK` becomes deprecated? Which I would be in favor of given how much time I've spent trying to make it work, and it seems I keep failing :D 

So if we're going to provide an implementation for `interface Deferred<A>` based off `IO` than I don't really see the point doing it in this PR? No, point in blocking this PR for issues we already have in master and even released. Not running laws on master is a bigger issue IMO.

raulraja:
Merge at will

gradle.properties
```diff
@@ -6,7 +6,7 @@ org.gradle.jvmargs=-Xmx4g
 # Kotlin configuration
 kotlin.incremental=true
 # Kotlin Test configuration
-kotlintest.parallelism=2
+kotlintest.parallelism=1
```
pakoito: I'd add a comment here so we know next time :D

modules/core/arrow-test/src/main/kotlin/arrow/test/UnitSpec.kt
```diff
@@ -11,28 +10,19 @@ import io.kotlintest.specs.AbstractStringSpec
  */
 abstract class UnitSpec : AbstractStringSpec() {
 
+  private val lawTestCases = mutableListOf<TestCase>()
+
   fun testLaws(vararg laws: List<Law>): List<TestCase> {
```
pakoito: Can this be an expression body like

`fun testLaws(vararg laws: List<Law>): List<TestCase> = laws.flatMap { ...`

AdrianRaFo:
@pakoito comments addressed.
Thanks @nomisRev for your help

raulraja:
@AdrianRaFo https://api.travis-ci.org/v3/job/483637513/log.txt

nomisRev:
@pakoito could you take a look st

raulraja:
@nomisRev ready to merge when you and @pakoito  are

PR #1257, opened 2019-01-19
Update README.md with the core/data split

wiyarmir:
Just noticed it setting up a new project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,8 @@ Add the dependencies into the project's `build.gradle`
 ```groovy
 def arrow_version = "0.8.2"
 dependencies {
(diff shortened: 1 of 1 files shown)


pakoito:
Data is no more. Raul has more context.

pakoito:
Also what are you setting up, ladrón :*

raulraja:
sorry, this has not been released yet, therefore the confusion, but in 0.9.0 all those changes in the README are taken into consideration. We will support multi-version docs after 1.0

wiyarmir:
😄 that serves me well for being disconnected

PR #1260, opened 2019-01-22
Docs contributions section revamp

calvellido:
This PR revamp the media/contributions section giving it a more concise style:

screenshot_2019-01-22 arrow
 
_The clipped title and the bad setting of the Gitter button are due to the capturing technique, don't worry about them_

This closes #626
This closes #797

PR #1263, opened 2019-01-23
Arrow Fx

raulraja:
## Arrow Fx

Arrow Fx is a library that brings purity, referential transparency and direct syntax for effectful programs denoted as suspended functions.

Arrow Fx removes the need for all the combinators in the Functor hierarchy empowering programmers to express equivalent programs for all suspended capable runtimes like Kotlinx Coroutines, Reactor Framework, Rx2, etc bridging the gap of FP features a user must learn before
is productive defining algebras and interpreters.

The below program illustrates the *Hello World* of effectful programming with Arrow Fx:
 
### Pure functions

```kotlin
val helloWorld: String =
  "Hello World"
  
helloWorld
```

### Side effects

```kotlin
suspend fun sayHello(): Unit =
  println(helloWorld)
```

A `suspend` function denotes a suspended computation. All side effects should be suspended and non-evaluated in compositions to ensure purity and referential transparency of the effectful expressions.

Attempting to run a side effect in the pure environment is disallowed by the Kotlin compiler unless we prove that we know how to handle success and error outcomes from evaluating the effect once the suspension is resumed. 

```kotlin
sayHello() //fails to compile
```

The Kotlin compiler won't allow side-effects denoted as `suspend` to run or compile until you provide a `Continuation<A>` that handles success and error cases.

### Composing effects and pure expressions

```kotlin
import arrow.effects.extensions.io.fx.fx

fx {
  sayHello() // Does not compile because Arrow Fx forces uses to denote side effects in `effect` blocks
}
val x = sayHello() // Does not compile because suspended functions need to be in a continuation
```

```kotlin
import arrow.effects.extensions.io.fx.fx

val program = fx {
  effect { sayHello() } 
}

program
```

### Executing effectful programs

All programs encapsulated in `fx` block yield a value at the end wrapped in the concurrent data type the user wishes to make its program concrete to. 

Running effectful computations is restricted to runtime extensions of the `Unsafe` type class which `IO` provides extensions for.
Usage of `unsafe` is reserved to the end of the world and may be the only impure execution of a well-typed functional program:

```kotlin
import arrow.effects.unsafe
import arrow.effects.extensions.io.unsafeRun.runBlocking

unsafe { runBlocking { program } }
```

Arrow Fx makes emphasis in guaranteeing users understand where they are performing side effects in their program declaration.

## Fx eliminates the need for `F` parameterized tagless style programs.

The following combinators illustrate how the Functor hierarchy functions are pointless in the environment given we can declare programs that respect the same semantics thanks to the Kotlin suspension system. These below are examples of a few of the most well known combinators that would disappear from your day to day FP programming and what others would look like flattened in the environment via automatic suspended binding after each combinator is applied in the suspended environment:

| TypeClass           | Signature | Arrow Fx signature |
|---------------------|-----------|--------------------|
| Functor.map         | `just(1).map { it + 1 }` | `1 + 1`  |
| Applicative.just    | `just(1)` | `1` |
| Applicative.mapN    | `mapN(just(1), just(2), ::Tuple2)` | `1 toT 2` |
| Applicative.tupled  | `tupled(just(1), just(2))` | `1 toT 2` |
| Monad.flatMap       | `IO { 1 }.flatMap { n -> IO { n + 1 } }` | `1 + 1` |
| Monad.flatten       | `IO { IO { 1 } }.flatten()` | `1` |
| MonadDefer.delay    | `IO.delay { 1 }` | `effect { 1 }` |
| MonadDefer.defer    | `IO.defer { IO { 1 } }` | `effect { 1 }` |

### Enhancing the ergonomics of Concurrent and Async FP Programs

The examples below demonstrates the expressive power that suspended functions and Arrow Fx brings to reactive programming by eliminating
effects nesting in a simple program that uses parallelism to process effectful computations in parallel and non-blocking.

#### Concurrent parMap

```kotlin
import arrow.effects.extensions.io.fx.fx
import arrow.effects.unsafe
import arrow.effects.extensions.io.unsafeRun.runBlocking
import kotlinx.coroutines.Dispatchers

/** A user declared side effect **/
//sampleStart
suspend fun getThreadName(): String =
  Thread.currentThread().name

val program = fx {
  // note how the receiving value is typed in the environment and not inside IO despite being effectful and
  // non-blocking parallel computations
  val result: List<String> = 
    Dispatchers.Default.parMap(
      { getThreadName() }, // each one of these runs in parallel
      { getThreadName() }
    ) { a, b -> 
      listOf(a, b) 
    }
}
//sampleEnd
fun main() {
  unsafe { runBlocking { program } }
}
```

- [x] Includes new `ank` operator `ank:fail` which prints errors in the markdown. /cc @Guardiola31337


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadLaws.kt
```diff
@@ -95,25 +95,25 @@ object MonadLaws {
...
   fun <F> Monad<F>.monadComprehensions(EQ: Eq<Kind<F, Int>>): Unit =
     forAll(Gen.int()) { num: Int ->
-      binding {
-        val a = just(num).bind()
-        val b = just(a + 1).bind()
-        val c = just(b + 1).bind()
+      fx {
+        val (a) = just(num)
+        val (b) = just(a + 1)
+        val (c) = just(b + 1)
```
AdrianRaFo: is this syntax an option to bind or the new only one? and how that's works?

raulraja: It's an option to bind. Works by overriding `component1` for all `Kind<F, A>` resulting in a suspended `A`

modules/docs/arrow-docs/docs/_data/menu.yml
```diff
@@ -18,482 +18,5 @@ tab:
     - title: Media
       url: /blog/media
 
-  - title: Arrow Fx
-
-    nested_options:
```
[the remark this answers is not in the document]
raulraja: it's on sidebar.yml now since #1262 and #1260

modules/core/arrow-extras/src/main/kotlin/arrow/data/Validated.kt
```diff
@@ -220,14 +220,14 @@ fun <E, A> ValidatedOf<E, A>.combineK(SE: Semigroup<E>, y: ValidatedOf<E, A>): V
...
-fun <E, A> E.invalid(): Validated<E, A> =
+fun <E> E.invalid(): Validated<E, Nothing> =
   Invalid(this)
 
-fun <E, A> A.validNel(): ValidatedNel<E, A> =
+fun <A> A.validNel(): ValidatedNel<Nothing, A> =
   Validated.validNel(this)
 
-fun <E, A> E.invalidNel(): ValidatedNel<E, A> =
+fun <E> E.invalidNel(): ValidatedNel<E, Nothing> =
```
pakoito: 👍

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadContinuations.kt
```diff
@@ -12,8 +13,8 @@ interface BindingInContextContinuation<in T> : Continuation<T> {
 }
 
 @RestrictsSuspension
-open class MonadContinuation<F, A>(M: Monad<F>, override val context: CoroutineContext = EmptyCoroutineContext) :
-  Continuation<Kind<F, A>>, Monad<F> by M {
+open class MonadContinuation<F, A>(val M: Monad<F>, override val context: CoroutineContext = EmptyCoroutineContext) :
```
pakoito: Why is `M` necessary as a field if it's already `this` via delegation?

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadError.kt
```diff
@@ -131,4 +131,14 @@ interface MonadThrow<F> : MonadError<F, Throwable> {
...
     return continuation.returnedMonad()
   }
+
+  override fun <B> binding(c: suspend MonadContinuation<F, *>.() -> B): Kind<F, B> =
+    bindingCatch { c() }
+
+  override fun <A> fx(f: suspend MonadContinuation<F, *>.() -> A): Kind<F, A> =
+    bindingCatch { f() }
+
+  fun <A> fx(f: suspend MonadErrorContinuation<F, *>.() -> A, unit: Unit = Unit): Kind<F, A> =
```
pakoito: How do you desambiguate them later? when the user does their first raiseError inside the block?

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadContinuations.kt
pakoito: I'd say extract a couple of interfaces from MonadContinuation and MCError, and have this as the default implementation. That way we'll make it a typeclass instead of a class where users have full access to implementation details.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/suspended/prelude.kt
```diff
@@ -0,0 +1,38 @@
...
+  fun <A, B> (suspend (A) -> B).k(): (Kind<F, A>) -> Kind<F, B> =
+    { suspend { this(it.bind()) }.k() }
+
+  fun <A, B, C> (suspend (A, B) -> C).k(): (Kind<F, A>, Kind<F, B>) -> Kind<F, C> =
+    { ka, kb -> suspend { this(ka.bind(), kb.bind()) }.k() }
+
+  fun <A, B, C, D> (suspend (A, B, C) -> D).k(): (Kind<F, A>, Kind<F, B>, Kind<F, C>) -> Kind<F, D> =
+    { ka, kb, kc -> suspend { this(ka.bind(), kb.bind(), kc.bind()) }.k() }
+
+  fun <A, B> (suspend (A) -> B).kr(unit: Unit = Unit): (A) -> Kind<F, B> =
```
pakoito: what does `kr` mean here?

modules/docs/arrow-docs/docs/docs/arrow/data/optiont/README.md
```diff
@@ -144,23 +144,23 @@ Let's look at how a similar implementation would look like using monad comprehen
...
-        { ObservableK.just(it) }
-      ).bind()
-      val maybeCountry = findCountry(address.id).bind()
-      val country = maybeCountry.fold(
-        { ObservableK.raiseError<Country>(NoSuchElementException("...")) },
-        { ObservableK.just(it) }
-      ).bind()
-      country.code
-    }              
+       binding {
```
pakoito: We use 2 spaces, this was correct.

modules/docs/arrow-examples/src/test/kotlin/arrow/FpToTheMax.kt
```diff
@@ -63,31 +63,31 @@ object FpToTheMax {
 
     fun parseInt(s: String): Option<Int> = Try { s.toInt() }.toOption()
 
-    fun <F> MonadAndConsoleRandom<F>.checkContinue(name: String): Kind<F, Boolean> = binding {
-        putStrLn("Do you want to continue, $name?").bind()
-        val input = getStrLn().map { it.toLowerCase() }.bind()
+    fun <F> MonadAndConsoleRandom<F>.checkContinue(name: String): Kind<F, Boolean> = fx {
+        val (_) = putStrLn("Do you want to continue, $name?")
```
pakoito: Not a fan. Can we have an alias for this instead? Or keep `bind()`

modules/effects/arrow-effects-data/build.gradle
```diff
@@ -1,5 +1,5 @@
 dependencies {
-    compile project(':arrow-typeclasses')
+    compile project(':arrow-extras-extensions')
```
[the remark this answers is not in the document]
pakoito: The old `arrow-data`. It doesn't feel like that anymore.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/ConcurrentCancellableContinuations.kt
```diff
@@ -0,0 +1,23 @@
+package arrow.effects.typeclasses
+
+import arrow.Kind
+import arrow.effects.typeclasses.suspended.ConcurrentSyntax
+import arrow.typeclasses.MonadContinuation
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.coroutines.RestrictsSuspension
+
+//typealias Fx<F> = ConcurrentCancellableContinuation<F, *>
```
pakoito: Ded code. Please delet.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDeferCancellableContinuations.kt
```diff
@@ -17,13 +19,16 @@ typealias Disposable = () -> Unit
 
 @RestrictsSuspension
 @Suppress("DELEGATED_MEMBER_HIDES_SUPERTYPE_OVERRIDE")
-open class MonadDeferCancellableContinuation<F, A>(SC: MonadDefer<F>, override val context: CoroutineContext = EmptyCoroutineContext) :
-  MonadErrorContinuation<F, A>(SC), MonadDefer<F> by SC {
+open class MonadDeferCancellableContinuation<F, A>(val SC: MonadDefer<F>, override val context: CoroutineContext = EmptyCoroutineContext) :
```
pakoito: Doesn't need to be a field as it's available as `this` via delegation.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/UnsafeRun.kt
```diff
@@ -0,0 +1,11 @@
+package arrow.effects.typeclasses
+
+import arrow.Kind
+import arrow.core.Either
+import arrow.effects.unsafe
+
+interface UnsafeRun<F> {
+  suspend fun <A> unsafe.runBlocking(fa: () -> Kind<F, A>): A
+  suspend fun <A> unsafe.runNonBlocking(fa: () -> Kind<F, A>, cb: (Either<Throwable, A>) -> Unit): Unit
```
pakoito: We call this `runAsync` even if it's not multithreaded.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/AsyncSyntax.kt
```diff
@@ -0,0 +1,41 @@
...
+import arrow.Kind
+import arrow.core.Either
+import arrow.core.Left
+import arrow.core.Right
+import arrow.effects.typeclasses.Async
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.coroutines.startCoroutine
+
+interface AsyncSyntax<F> : MonadDeferSyntax<F>, Async<F> {
```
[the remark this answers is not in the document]
pakoito: I'd say we can join them if the idea is well received.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/AsyncSyntax.kt
```diff
@@ -0,0 +1,41 @@
...
+
+  override fun <A> f(fa: suspend () -> A): Kind<F, A> =
+    super.async { cb ->
```
pakoito: Why `super`? It's redundant, as MonadDefer doesn't have it.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/AsyncSyntax.kt
```diff
@@ -0,0 +1,41 @@
...
+    asyncOp { defer(this@defer) { f.k() } }
+
+  suspend fun continueOn(ctx: CoroutineContext): Unit =
+    asyncOp { ctx.shift() }
+
+}
+
+internal open class AsyncContinuation<A>(val cb: (Either<Throwable, A>) -> Unit) : kotlin.coroutines.Continuation<A> {
+  override val context: CoroutineContext
+    get() = EmptyCoroutineContext // TODO this should probably use the default dispatcher by default.
```
pakoito: Yep, adding the ConcurrentRuntime soon :)

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/SuspendToKindSyntax.kt
```diff
@@ -0,0 +1,8 @@
+package arrow.effects.typeclasses.suspended
+
+import arrow.typeclasses.suspended.BindSyntax
+
+interface SuspendToKindSyntax<F> : BindSyntax<F> {
+
+
```
pakoito: ??? TODO or something

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/concurrent/Fx.kt
```diff
@@ -0,0 +1,16 @@
+package arrow.effects.typeclasses.suspended.concurrent
+
+import arrow.Kind
+import arrow.core.Tuple2
+import arrow.effects.typeclasses.Concurrent
+import arrow.effects.typeclasses.ConcurrentCancellableContinuation
+import arrow.effects.typeclasses.Disposable
+
+interface Fx<F> {
```
pakoito: Why is every Fx called Fx even if they're for different typeclasses? Are we relying on IJ to pull the right one?

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/unsafe.kt
```diff
@@ -0,0 +1,35 @@
...
+
+import arrow.core.Continuation
+import java.util.concurrent.atomic.AtomicReference
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.coroutines.RestrictsSuspension
+import kotlin.coroutines.startCoroutine
+
+private class UnsafeContinuation<A>(
+  val result: AtomicReference<A> = AtomicReference()
```
pakoito: Use `Promise.uncancellable()` instead.

modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/EffectsSuspendDSLTests.kt
```diff
@@ -0,0 +1,373 @@
...
+          val (result) = f { sideEffect() }
+          result
+        }
+      } shouldBe ""
+    }
+
+  }
+}
+
+fun <A> fxTest(f: () -> IO<A>): A =
```
pakoito: This is the `unsafeRunBlocking` alias I told you about the other day may be useful for cases like this :)

modules/meta/arrow-meta/src/main/java/arrow/common/utils/AbstractProcessor.kt
```diff
@@ -47,11 +48,13 @@ abstract class AbstractProcessor : KotlinAbstractProcessor(), ProcessorUtils, Ko
...
-        Files.createDirectories(kDocLocation.toPath().parent)
-        Files.createFile(kDocLocation.toPath())
+        val path = kDocLocation.toPath()
+        Files.createDirectories(path.parent)
+        Files.delete(path)
+        Files.createFile(path)
         kDocLocation.writeText(doc)
       } catch (x: IOException) {
-        logE("Failed to generate kdoc file location: $kDocLocation", e)
+        logW("Failed to generate kdoc file location: $kDocLocation", e)
```
pakoito: oh thank god

modules/core/arrow-core-data/src/test/kotlin/arrow/core/OptionTest.kt
```diff
@@ -137,6 +139,27 @@ class OptionTest : UnitSpec() {
       None or None shouldBe None
 
     }
+
+    "fx can turn effects into pure kinded values" {
+      suspend fun sideEffect(): Int =
+        1
+      fx {
+        val (result) = f { sideEffect() }
```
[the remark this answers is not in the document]
raulraja: I got private feedback from some people I asked from the FP community in Scala, Haskell and Kotlin.
`f` turns a suspended function that denotes a side effect into a `Kind<F, A>` and it's available in all `fx` blocks. It then lifts the side effect to the monad in question, in this case `Option`. When it injests the side effect it tries to lift `A` resulting from the effect into the monadic context.
In option it will throw an exception if it failed, in other data types capable of monad error captures the exception and raises it to F and in others that are suspend capable like MonadDefer, Async or Concurrent it does the appropriate defer op to suspend side effects.
```kotlin
fun <A> f(fa: suspend () -> A): Kind<F, A>
```
We may rename it or alias it to something more meaningful. currently the Fx primitives are:
- `fx`. A monadic binding block 
- effect. Injest an external side effect and bind
- f. Injest a side effect and not bind but instead get the `Kind<F, A>` value
- unsafe. explicitly perform side effects at the edge of the world

modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/either.kt
```diff
@@ -168,3 +170,8 @@ interface EitherHash<L, R> : Hash<Either<L, R>>, EitherEq<L, R> {
     HR().run { it.hash() }
   })
 }
+
+@extension
+interface EitherFx<A> : Fx<EitherPartialOf<A>> {
```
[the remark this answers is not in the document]
raulraja: Fx is a family of type classes all with the same name and different package.
Each one of them represents a power in effectful monads:
```
arrow.typeclasses.suspended.monad.Fx
arrow.typeclasses.suspended.monaderror.Fx
arrow.effects.typeclasses.suspended.monaddefer.Fx
arrow.effects.typeclasses.suspended.concurrent.Fx
```
This allows us to only implement one of them for a given data type. For example we want the `IO` `fx` to be `arrow.effects.typeclasses.suspended.concurrent.Fx` so when users import it it already has all the concurrent combinators for races and parallelism instead of just the lower powers of `arrow.typeclasses.suspended.monad.Fx`. In other langs it may not make sense but in Kotlin this is a practical technique to divide the same type class concept which in this case is side effect suspension in different types in a hierarchy to avoid exposing lowered powered versions of it.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadLaws.kt
```diff
@@ -76,16 +76,16 @@ object MonadLaws {
   fun <F> Monad<F>.equivalentComprehensions(EQ: Eq<Kind<F, Int>>) {
     val M = this
     forAll(Gen.int()) { num: Int ->
-      val aa = binding {
-        val a = just(num).bind()
-        val b = just(a + 1).bind()
-        val c = just(b + 1).bind()
+      val aa = fx {
+        val (a) = just(num)
```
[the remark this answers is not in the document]
raulraja: yes, except `effect` will take a `suspend () -> A` and bind directly to `A`. 
`effect` is just a compile time trick and can be inlined to the `effectIdentity`. This is how it's implemented:
```kotlin
suspend fun <A> effect(f: suspend () -> A): A = f()
```
It does nothing but force the user to use it whenever a side effect is denoted to be injested from the environment into the continuation. At the moment this is only available for MonadDefer and up because those are the only ones that can suspend side effects.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadError.kt
```diff
@@ -131,4 +131,14 @@ interface MonadThrow<F> : MonadError<F, Throwable> {
     wrapReturn.startCoroutine(continuation, continuation)
     return continuation.returnedMonad()
   }
+
+  override fun <B> binding(c: suspend MonadContinuation<F, *>.() -> B): Kind<F, B> =
```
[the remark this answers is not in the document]
raulraja: No, but this does not need to be annotated with `@Deprecated` because this is an override and the most super declaration already is annotated as `@Deprecated`

modules/docs/arrow-docs/docs/docs/arrow/data/nonemptylist/README.md
```diff
@@ -136,7 +136,7 @@ map(nelId, nelName, nelYear) { (id, name, year) ->
 - `NonEmptyList` is __used to model lists that guarantee at least one element__
 - We can easily construct values of `NonEmptyList` with `NonEmptyList.of`
 - `foldLeft`, `map`, `flatMap` and others are used to compute over the internal contents of a `NonEmptyList` value.
-- `binding { ... } Comprehensions` can be __used to imperatively compute__ over multiple `NonEmptyList` values in sequence.
+- `fx { ... }  and binding { ... } Comprehensions` can be __used to imperatively compute__ over multiple `NonEmptyList` values in sequence.
```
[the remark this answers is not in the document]
raulraja: It's deprecated everywhere because it's annotated in its definition. I don't believe we should remove the reference to `binding` from that doc because it's deprecated but not gone and we already know the issues we have with doc versioning which @dominv and Watt from 47 are gonna help us address for 0.9.0

modules/effects/arrow-effects-data/build.gradle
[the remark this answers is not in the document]
raulraja: `data` know implies data types and what we called before instances is now `extensions` to be inline with Kotlin idioms and Keep 87.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/AsyncSyntax.kt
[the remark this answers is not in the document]
raulraja: We can definitely move them to each of their correspondent type classes in the next PR for the cases where `bind()` is in scope, otherwise direct syntax does not work but I believe all the candidates are already `Monad`

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/prelude.kt
```diff
@@ -0,0 +1,5 @@
+package arrow.effects.typeclasses.suspended
```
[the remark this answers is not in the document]
raulraja: `prelude` and `predef` are common names in fp for the file with global static functions that are part of the base of the program for example `identity`. If anyone has a better suggestion for Kotlin it can be changed

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/AsyncSyntax.kt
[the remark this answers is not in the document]
raulraja: Something to be determined would be if we want the suspended functions also exported as the rest of the type class. Putting them on the type class will increase the size considerably whereas now since in the suspended world there are no `Kind` we don't have to worry about `fix` and they are just aggregated in the continuation.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadContinuations.kt
raulraja: it shouldn't be necessary. Addressed in https://github.com/arrow-kt/arrow/pull/1263/commits/b971abecd65ad59dd767cfe1f59b044faad0ed2c

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadError.kt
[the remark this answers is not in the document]
JorgeCastilloPrz: Would I get the deprecation warning on the IDE if I reference it directly as Try.monadError().binding {} ?

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadContinuations.kt
raulraja: Tried that in the original implementation but the issue is that `A` needs to be carried around always because its the `A` from `Continuation<A>` injected by the compiler in suspended functions so all attempts to turn it into a type class are frustrating since there will always be `F` and `A` in the top signature. If anyone else wants to give it another shot I'll be happy to help.

modules/effects/arrow-effects-data/build.gradle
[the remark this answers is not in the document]
JorgeCastilloPrz: Mmmm ok, I think I need to make my mind a bit more around the new structure

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/prelude.kt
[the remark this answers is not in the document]
JorgeCastilloPrz: Right now it looks like there's just a callback inside. Maybe we can put that somewhere else or alternatively name the file to something more semantic related to what the callback is used for

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/concurrent/Fx.kt
JorgeCastilloPrz: I guess where strongly relying on imports

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/suspended/prelude.kt
raulraja: `k` is `liftM` which we don't have in `Arrow` in monad, there is also `liftA` versions of it for applicative. https://www.haskell.org/hoogle/?hoogle=liftM . It supports N arities we can provide up to 3 or 4 because otherwise the number of methods in the extensions will increase considerably. I renamed them in this commit to `liftM` and `flatLiftM` until we come up with better names in https://github.com/arrow-kt/arrow/pull/1263/commits/cb6a3aa2af3beba7c96b7efea7c9d9458625bbc2

modules/docs/arrow-examples/src/test/kotlin/arrow/FpToTheMax.kt
raulraja: `bind` is not deprecated. This is just another supported style. The need for `bind` or `component1` destructuring only arises when a user receives a kinded value. A denoted side effect with `suspend` has nothing to destructure just function applications. This allows for programs to be defined in 3 different ways:

1. All suspended continuations denoted as effects:
```kotlin
/**
 * All side effects are marked as `suspend` functions
 */
suspend fun shouldCreateDir(file: File): Boolean =
  file.exists() && file.isDirectory

suspend fun mkdir(file: File): Boolean =
  file.createNewFile()

suspend fun tmpFile(file: File): File =
  File.createTempFile("", "", file)

suspend fun writeInto(file: File, s: String): Unit =
  file.writeText(s)

suspend fun threadName(): String =
  Thread.currentThread().name

/**
 * Direct style monadic version
 * 
 * Composisition of side effects is direct when they are encapsulated in other side effects
 * 
 * Functor, Applicative and Monad are gone from the syntax. 
 * The entire computation can't compile in a pure environment unless in the pressence of a suspend
 * capable data type.
 */
suspend fun programDirectStyle(): Unit {
  val file = File.createTempFile("tmp", "tmp")
  if (shouldCreateDir(file)) mkdir(file)
  val fh = tmpFile(file)
  writeInto(fh, threadName())
}
```

Or with wrapped pure values that encapsulate effects : 
```kotlin
/**
 * Direct style monadic version
 * 
 * Composition of side effects in `fx` blocks requires explicit `effect` denotation.
 *
 * `effect` is a compiler proof that raw suspended function are accepted in our monad continuation
 * and we have continuations to handle those unlike the raw Kotlin or Scala environment where suspended functions can't compile.
 *
 * It's implemented as:
 *
 * suspend fun <A> effect(f: suspend () -> A): A = f()
 * 
 */
fun programMonadic(): IO<Unit> = fx { // presence of an IO continuation
  effect { // denotes a block as effectful otherwise this code would not have compiled inside `fx`
    val file = File.createTempFile("tmp", "tmp")
    if (shouldCreateDir(file)) mkdir(file)
    val fh = tmpFile(file)
    writeInto(fh, threadName())
  }
}
```
or fully wrapped with all effects lifted into pure values individually:
```kotlin
/**
 * Kinded style monadic version with enhanced destructuring syntax
 * 
 * `f` lifts a `suspend () -> A` function into `Kind<F, A>` or `IO<A>`.
 *
 * Monads in Kotlin can be destructured non blocking like tuples thanks to the suspension system
 * The following expressions are equivalent and produce the effect in a controlled way:
 * val p : IO<Unit> = f { println("a") }
 * val (x : Unit) = f { println("a") }
 * val x: Unit = effect { println("a") }
 */
fun programMonadicKinded(): IO<Unit> = fx {
  val (file) = f { File.createTempFile("tmp", "tmp") }
  val (createDir) = f { shouldCreateDir(file) }
  if (createDir) f { mkdir(file) }
  val (fh) = f { tmpFile(file) }
  val (_) = f { writeInto(fh, threadName()) }
}
```

All 3 programs are equivalen and suspend effects in the same way. That is because `suspend -> A` and `F<A>` are the same thing in an `fx` block.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/ConcurrentCancellableContinuations.kt
raulraja: Addressed in https://github.com/arrow-kt/arrow/pull/1263/commits/6ab8081ad7b6bb3d605f7eed806ab8660e8ef8ad

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDeferCancellableContinuations.kt
raulraja: Already addressed

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/UnsafeRun.kt
raulraja: yes, and it's been changed to make sure of 2 added compile time guarantees:
1. This removes the need to have unsafe functions in the environment which means we can deprecate `unsafeRunSync` etc.
2. Calling `runBlocking` and `runNonBlocking` does not compile unless the user has explicitly imported the `unsafe` function which is the only function in `Arrow` that will be allowed to run Fx programs.

The names have been changed because these are more familiar idioms to Kotlin users and we have extra compile guarantees they are well aware that effects are performed at that point.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/AsyncSyntax.kt
raulraja: There is another `async` function in this interface that classes with `async` unless you qualify the call like this since it needs to infer the type signature of `cb`.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/SuspendToKindSyntax.kt
raulraja: addressed in https://github.com/arrow-kt/arrow/pull/1263/commits/abde3ac466da7379d06e77dea1986c89ee69b57d

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/concurrent/Fx.kt
raulraja: See https://github.com/arrow-kt/arrow/pull/1263#discussion_r

modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/EffectsSuspendDSLTests.kt
raulraja: Yes, but this goes agains't the idea that effect can only run on `unsafe` via explicit imports. Since the goal of the Fx system is to provide higher compile time guarantees when dealing with effects we aim for a single way to run effects. Users can create such functions and name them `runThisIsFineNothingWillHappen()` if they want to but we shouldn't provide alias for unsafe runs in the environment. The Arrow Fx system treats the environment as `pure` an `suspend` as impure therefore running should be suspended and only materialized in one place by the user.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Foldable.kt
```diff
@@ -95,6 +95,9 @@ interface Foldable<F> {
     foldLeft(MN.empty()) { b, a -> b.combine(f(a)) }
   }
 
+  fun <A> orEmpty(AF: Applicative<F>, MA: Monoid<A>): Kind<F, A> =
```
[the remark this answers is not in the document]
raulraja: @nomisRev question regarding `binding` vs `fx` are addressed in https://github.com/arrow-kt/arrow/pull/1263#discussion_r

Regarding `orEmpty` vs `orElse` I believe `orElse` does not provide enough information for the user to understand the returned value is the empty value for A. To me `Else` means `or this other value` which I'm responsible to provide. Here the user is just providing evidence of type classes extensions but not an actual user-provided function. `orEmpty` is also consistent with `arrow.mtl.typeclasses.MonadFilter#empty` for a parameterless return of `Kind<F, A>`

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadErrorContinuations.kt
```diff
@@ -1,15 +1,24 @@
...
 
+import arrow.Kind
+import arrow.typeclasses.suspended.MonadThrowSyntax
 import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.EmptyCoroutineContext
 import kotlin.coroutines.RestrictsSuspension
 
 @RestrictsSuspension
 @Suppress("DELEGATED_MEMBER_HIDES_SUPERTYPE_OVERRIDE")
 open class MonadErrorContinuation<F, A>(val ME: MonadError<F, Throwable>, override val context: CoroutineContext = EmptyCoroutineContext) :
```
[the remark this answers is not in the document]
raulraja: Yes but since we have ambitions to introduce `BIO` and I did not restructure continuation dependencies in this PR I figured they are all gonna be parametric to E when we have E in the Fx hierarchy. Feel free to jump in and rename at will.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/ConcurrentSyntax.kt
```diff
@@ -0,0 +1,103 @@
...
+import arrow.core.Tuple3
+import arrow.core.Tuple4
+import arrow.core.Tuple5
+import arrow.effects.typeclasses.Concurrent
+import arrow.effects.typeclasses.Fiber
+import arrow.effects.typeclasses.RacePair
+import arrow.effects.typeclasses.RaceTriple
+import kotlin.coroutines.CoroutineContext
+
+interface ConcurrentSyntax<F> : AsyncSyntax<F>, Concurrent<F>, ListParTraverseSyntax<F> {
```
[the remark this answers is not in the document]
raulraja: yes, I like startFiber better but let's do that in a separate PR

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/ConcurrentSyntax.kt
```diff
@@ -0,0 +1,103 @@
...
+    fb: suspend () -> B,
+    fc: suspend () -> C,
+    fd: suspend () -> D,
+    fe: suspend () -> E,
+    f: (A, B, C, D, E) -> G
+  ): G =
+    concurrently { parMapN(this@parMap, fa.k(), fb.k(), fc.k(), fd.k(), fe.k(), f) }
+
+
+  suspend fun <A, B> CoroutineContext.parallel(
```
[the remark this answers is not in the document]
raulraja: The spirit of the Fx DSL is to offer even friendlier names. The use of this function does not even expose tuple to the end user because destructuring is automatic in Tuple. I think a new user may not even understand tuples all around but can understand the functions passed into `parallel` run in parallel and return their result independently for as many functions you use as args:
```kotlin
val (ta, tb) = parallel({ Thread.currentThread.name }, { Thread.currentThread.name })
```

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadError.kt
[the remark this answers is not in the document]
raulraja: screen shot 2019-01-24 at 3 09 15 pm
Yep

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/suspended/MonadSyntax.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import arrow.Kind
+import arrow.typeclasses.Applicative
+import arrow.typeclasses.Monad
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.coroutines.startCoroutine
+
+interface MonadSyntax<F> : ApplicativeSyntax<F>, Monad<F> {
+  override fun <A> f(fa: suspend () -> A): Kind<F, A> =
```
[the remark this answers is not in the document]
raulraja: @JorgeCastilloPrz yes, each monad level understand how to inject a suspend function which is what denotes a side effect. 

@nomisRev `effect` is only available for `MonadDefer` and up. `effect` is the same as `::identity` for suspended functions and it's only available there because is the only place where you can truly suspend and bind side effects. `effect` returns `A` but `f` which is the same as `liftM` for which other versions called `liftM` are in this PR returns `Kind<F, A>`.

You need these two functions to go from side effects to wrapped pure values depending on how you assemble the composition of your side effects and if you go back and forth between `suspend () -> A` and `Kind<F, A>`.

We need to get together to better name all of these so they are friendlier to users.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/prelude.kt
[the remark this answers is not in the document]
raulraja: Prelude is not `Utils`. Prelude is what comes before something, the beginning and building blocks needed. in the global scope to compose programs. I renamed it for now to `Predef` so it's consistent with optics and Arrow core which is where the `identity` lives now. If we want a different strategy it would have to be applied consistently to all modules.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadError.kt
raulraja: With the multiple Fx instances. Each monad targets a single `Fx` instance with the right powers. That allows us to have for types like `IO` only the `Concurrent<F>` `fx` version available.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/suspended/prelude.kt
raulraja: Let's discuss over slack names for all the functions of the DSL

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/prelude.kt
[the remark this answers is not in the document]
pakoito: I'm up for unifying names as `predef`.

modules/docs/arrow-examples/src/test/kotlin/arrow/FpToTheMax.kt
pakoito: My point is that `val (_)` for a side-effect is just meh. Can't we have an specialization of `bind()` for `Unit` that's like `sideEffect()` and can be grepped for.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/unsafe.kt
pakoito: Some other construct we have that's not AtomicReference then? I believe we had an alias somewhere.

gradle.properties
```diff
@@ -6,7 +6,8 @@ org.gradle.jvmargs=-Xmx4g
 # Kotlin configuration
 kotlin.incremental=true
 # Kotlin Test configuration
-kotlintest.parallelism=2
+#Parallelism needs to be set to 1 since the concurrent tests in arrow-effects become flaky otherwise
+kotlintest.parallelism=1
```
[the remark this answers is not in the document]
raulraja: This comes from a master merge, not part of this PR

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/ConcurrentSyntax.kt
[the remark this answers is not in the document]
raulraja: Once we move all the syntax to type classes then this will moved with it.

modules/docs/arrow-examples/src/test/kotlin/arrow/FpToTheMax.kt
raulraja: what would it look like when used?

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/suspended/prelude.kt
```diff
@@ -0,0 +1,38 @@
+package arrow.typeclasses.suspended
+
+import arrow.Kind
+
+interface Prelude {
```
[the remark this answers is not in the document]
raulraja: so it's aggregated and the function is defined in the restricted scope of the continuation otherwise `effectIdentity` would need to also become lifted each time into the monad context.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
```diff
@@ -688,6 +693,27 @@ interface Concurrent<F> : Async<F> {
...
+   * Entry point for monad bindings which enables for comprehensions. The underlying impl is based on coroutines.
+   * A coroutines is initiated and inside [ConcurrentCancellableContinuation] suspended yielding to [Monad.flatMap]. Once all the flatMap binds are completed
+   * the underlying monad is returned from the act of executing the coroutine
+   *
+   * This one operates over [Concurrent] instances
+   *
+   * This operation is cancellable by calling invoke on the [Disposable] return.
+   * If [Disposable.invoke] is called the binding result will become a lifted [BindingCancellationException].
+   */
+  fun <B> bindingConcurrent(c: suspend ConcurrentCancellableContinuation<F, *>.() -> B): Tuple2<Kind<F, B>, Disposable> {
```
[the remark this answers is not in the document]
raulraja: The alternative is to define `putStrnLn` as `suspend` and just run it as `effect { putStrnLn("") }`

modules/docs/arrow-examples/src/test/kotlin/arrow/FpToTheMax.kt
pakoito: `putStrLn("Do you want to continue, $name?").sideEffect()`

PR #1268, opened 2019-01-26
Effects Environment and Dispatchers

pakoito:
Add typeclasses for injecting Environment needs such as a global async error handler, and a set of dispatchers


modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
```diff
@@ -186,8 +188,7 @@ interface Concurrent<F> : Async<F> {
    *
    * @see raceN for a simpler version that cancels loser.
    */
-  fun <A, B> racePair(ctx: CoroutineContext, fa: Kind<F, A>, fb: Kind<F, B>): Kind<F, RacePair<F, A, B>>
-
+  fun <A, B> racePair(ctx: CoroutineContext = environment().dispatchers().default(), fa: Kind<F, A>, fb: Kind<F, B>): Kind<F, RacePair<F, A, B>>
```
pakoito: Calling ms liskov

modules/effects/arrow-effects-io-extensions/src/main/kotlin/arrow/effects/extensions/io.kt
```diff
@@ -0,0 +1,63 @@
...
+    IOD.CommonPool
+}
+
+@extension
+interface IOEnvironment : Environment<ForIO> {
+  override fun dispatchers(): Dispatchers<ForIO> =
+    IO.dispatchers()
+
+  override fun handleAsyncError(e: Throwable): IO<Unit> =
+    IO { println(e.message) }
```
pakoito: I'm happy to discuss this one.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
```diff
@@ -191,8 +193,7 @@ interface Concurrent<F> : Async<F> {
    *
    * @see raceN for a simpler version that cancels loser.
    */
-  fun <A, B> racePair(ctx: CoroutineContext, fa: Kind<F, A>, fb: Kind<F, B>): Kind<F, RacePair<F, A, B>>
-
+  fun <A, B> racePair(ctx: CoroutineContext = dispatchers().default(), fa: Kind<F, A>, fb: Kind<F, B>): Kind<F, RacePair<F, A, B>>
```
raulraja: 👏

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
```diff
@@ -546,7 +547,7 @@ interface Concurrent<F> : Async<F> {
    * @see racePair for a version that does not automatically cancel the loser.
    */
   fun <A, B> raceN(
-    ctx: CoroutineContext,
+    ctx: CoroutineContext = dispatchers().default(),
```
raulraja: `ConcurrentSyntax` and `AsyncSyntax` which Fx uses should also be updated so the `parMap` versions or anything there that uses a `CoroutinesContext` also uses the default one

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Environment.kt
```diff
@@ -0,0 +1,9 @@
+package arrow.effects.typeclasses
+
+import arrow.Kind
+
+interface Environment<F> {
+  fun dispatchers(): Dispatchers<F>
+
+  fun handleAsyncError(e: Throwable): Kind<F, Unit>
```
raulraja: What is this needed for?

modules/effects/arrow-effects-extensions/src/main/kotlin/arrow/effects/extensions/io.kt
```diff
@@ -143,18 +144,27 @@ interface IOConcurrent : Concurrent<ForIO>, IOAsync {
...
 @extension
-interface IOEffect: Effect<ForIO>, IOAsync {
+interface IOEffect : Effect<ForIO>, IOAsync {
   override fun <A> IOOf<A>.runAsync(cb: (Either<Throwable, A>) -> IOOf<Unit>): IO<Unit> =
     fix().runAsync(cb)
 }
 
-@extension
-interface IOConcurrentEffect: ConcurrentEffect<ForIO>, IOEffect, IOConcurrent {
+// FIXME default @extension are temporarily declared in arrow-effects-io-extensions due to multiplatform needs
```
raulraja: What is the actual limitation why `@extension` can't be used here?

modules/effects/arrow-effects-io-extensions/src/main/kotlin/arrow/effects/extensions/io.kt
```diff
@@ -0,0 +1,38 @@
...
+    IOD.CommonPool
+}
+
+@extension
+interface IOEnvironment : Environment<ForIO> {
+  override fun dispatchers(): Dispatchers<ForIO> =
+    IO.dispatchers()
+
+  override fun handleAsyncError(e: Throwable): IO<Unit> =
+    IO { println("Found uncaught async exception!"); e.printStackTrace() }
```
raulraja: When would this ever happen that is not handled in `Async<F>`?

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Environment.kt
pakoito: This will be used in `BIO` for async errors that do not belong in the user domain. `cats-bio` logs them. I give the user a choice to do whatever they want with it. (B)IO defaults to log.

modules/effects/arrow-effects-extensions/src/main/kotlin/arrow/effects/extensions/io.kt
pakoito: There's no value for Dispatchers in this module, as it's not JVM-dependent. This is here so users can do `IO.concurrent(MyDispatcher)` where they plug the kx.coroutines, reactor, or the rx ones.

modules/effects/arrow-effects-io-extensions/src/main/kotlin/arrow/effects/extensions/io.kt
pakoito: With BIO, whenever there's an async exception that's caught but not on the user's domain.

modules/effects/arrow-effects-extensions/src/main/kotlin/arrow/effects/extensions/io.kt
nomisRev: @pakoito what is to fix here?

PR #1275, opened 2019-01-31
fixed links to Try translation

dobrowins:
Fixed link to Russian translation (and from Russian to English) — it hasn't been visible the first time.
My bad, my fix.


pakoito:
Thank you for the find!

PR #1285, opened 2019-02-04
Ensure paths are compatible with windows

pablisco:
Windows is not very good with having `?` characters as part of a path. This, hopefully, fixes the issue for people trying to use the Annotation processor with Windows.


raulraja:
thanks @pablisco !

PR #1289, opened 2019-02-06
Ank and Doc Playground improvements

raulraja:


modules/meta/arrow-meta/src/main/java/arrow/common/utils/AbstractProcessor.kt
```diff
@@ -50,9 +50,15 @@ abstract class AbstractProcessor : KotlinAbstractProcessor(), ProcessorUtils, Ko
...
-        Files.createFile(path)
+        @Suppress("SwallowedException")
+        try { Files.createDirectories(path.parent) } catch (e : IOException) {
+        }
+        @Suppress("SwallowedException")
+        try { Files.delete(path) } catch (e : IOException) {
+        }
+        @Suppress("SwallowedException")
+        try { Files.createFile(path) } catch (e : IOException) {
+        }
```
pakoito: *squint* what

raulraja: yep, tried with file permissions but it does not work so we have to silently fail if it can't create the file because a previous annotation processor already created it.

raulraja:
@pablisco the issue wasn't your PR #1285 but something else related to Ank. I fixed it and kept your commits.

raulraja:
/cc @calvellido @dominv This PR should fix the docs issue where the extra executable directive was being added and other issues related to type classes not exporting kdocs correctly.
@dominv Do we have the remote Arrow playgrounds server stable and deploying on each snapshot? I need to be able to run the latest in master before my presentation on Arrow Fx on Friday. Is it reliable?

dominv:
@raulraja So far, there is a cron job that daily deploys the last changes in master in our **tryArrow** aws instance. We can also manually deploy it from Travis. 
I need to finish a couple of issues in tryArrow to have it deploying every time a snapshot is published.

raulraja:
thanks @dominv can you and Calvellido be online while the presentation in case there are any issues with the server? The presentation relies on the server to run the snippets and this is the first time someone uses it for this purpose.

calvellido:
Yeah, I'll be around :+1:

captura de pantalla de 2019-02-06 15-51-21

dominv:
Me too 👍

PR #1290, opened 2019-02-06
Mre doc edits

MaureenElsberry:
Small edits for the Arrow Fx, async, and polymorphism readmes

### DIFF
--- a/modules/docs/arrow-docs/docs/docs/effects/fx/polymorphism/README.md
+++ b/modules/docs/arrow-docs/docs/docs/effects/fx/polymorphism/README.md
@@ -16,12 +16,11 @@ Fx programs can be declared in a polymorphic style and made concrete to your fra
 
 ## Creating polymorphic programs
 
-So far all programs we've created in previous examples where built with `Arrow`'s `IO`.
-Fx is not restricted to `IO`. Since `IO` provides extensions for `Concurrent<F>` we can also create extensions for the `fx` DSL directly as in all previous examples.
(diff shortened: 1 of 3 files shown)


modules/docs/arrow-docs/docs/docs/effects/fx/polymorphism/README.md
```diff
@@ -158,10 +157,10 @@ The following combinators illustrate how the Functor hierarchy functions are poi
 | MonadDefer.delay    | `IO.delay { 1 }` | `effect { 1 }` |
 | MonadDefer.defer    | `IO.defer { IO { 1 } }` | `effect { 1 }` |
 
-This is in general true for effectful data types that are non-commutative. 
-Note that implicit CPS style with non-blocking direct binding has the disadvantage that for non-commutative monads where the order of effects matter you can't apply substitution based on referential transparency.
+This is, in general, true for effectful data types that are non-commutative.
```
pakoito: I would add a new header here that reads:

`Effects with non-commutative monads`

pakoito: Also, the line should read `are commutative.`

raulraja: Addressed as part of the last master merge

pakoito:
I had fixed a bunch of links already on master, this diff has smashed over them 🙈

For example,

```
[`parMapN`](#-parmapn-)
```

is a regression.

raulraja:
Let's get rid of those TOC at the top. They are a burden to maintain unless we automate it and then they would not be there in the first place.

PR #1291, opened 2019-02-06
Update Coproducts

### LINKED ISSUE #1284: Coproducts type inference issue
`Coproducts` created by @abergfeld are awesome, but they have a problem of type inference. Let's say I have this network service call:

```kotlin
fun getUserVerified(userId: UserId): IO<Coproduct4<Boolean, UnauthorizedError, UnenrolledError, UnknownError>> =
    userApi.getUserVerified(userId)
        .map { response ->
            when (response.code) {
                200 -> with(response.body<GetUserVerifiedResponse>()) {
                    this.enrolled.cop<Boolean, UnauthorizedError, UnenrolledError, UnknownError>()
                }
                else -> Unknown.cop<Boolean, UnauthorizedError, UnenrolledError, UnknownError>()
            }
        }
```

I would expect all the `cop()` call types to be inferred, but it's not able to, so `Coproducts` become a bit of a pain to use everywhere.

abergfeld:
Fixes #1284 

- Deprecates `cop` and `coproductOf` methods as they always need generic types declared due to ambiguity 
- Add `first`, `second`, `third` etc constructor methods that don't require generic types declared unless they cannot be inferred
- Make the data classes public as an alternative constructor 
- Update documentation to include runnable code examples
- Update documentation to include example of flattening hierarchies
- Update code generation to include class level and method level documentation


abergfeld:
@JorgeCastilloPrz @raulraja Just some questions:

- Should we use the generic letter for extension method names `"String".a()` or the name of the data class `"String".first()`?

- Should `coproductOf` also be deprecated? It should suffer from the safe inference issue

- Should I add something to the docs about the data classes being a valid way to construct a Coproduct?

- Is there something more specific we should use for the deprecation message?

raulraja:
I'd make the extensions the same name as the data class constructors. 
`corpoducOf` should also be deprecated as you noted. 
It'd be nice if the docs showcased both construction styles. ADT constructors and syntax.
The docs should be moved to kdocs and relinked in the menu to the generated api doc, ideally with runnable snippets as it is done in other places like functor, Bracket, etc.

abergfeld:
@raulraja Updated to use the `first`, `second` etc naming. Also deprecated the `coproductOf` functions as well in favor of just using the data classes. 

I added the ADT examples to the docs and added kdocs for all the non-deprecated things. 

I have a couple things I'm not sure of:
1) Given that there's so many functions related to Coproducts (2-22 sealed classes, etc) I just added runnable snippets to the current docs page. I assume it's not worth linking the same kind of runnable snippets over and over if it's generated off the kdocs. Just wanted to validate that it's fine to leave that out of the kdocs.

2) I'm having some trouble building the documentation locally. The code snippets all show up correctly but when I run them, it says that they don't compile on things that didn't break the build. Does the code snippet running happen with the current master version of Arrow? I had compilation failures running Ank locally and fixed them but the snippets still failed after I ran the jeykll command

JorgeCastilloPrz:
Are you following the steps described [here](https://github.com/arrow-kt/arrow/blob/master/modules/docs/arrow-docs/README.md#how-to-deploy-docs-to-a-local-server)?
 (dokka + ank + deploy)

modules/core/arrow-generic/src/test/kotlin/arrow/generic/CoproductTest.kt
```diff
@@ -82,5 +82,37 @@ class CoproductTest : UnitSpec() {
                     { "Third" }
             ) shouldBe "Third"
         }
+
+        "cop replacement functions are equivalent" {
+            val firstValue = "String"
+            val secondValue = 100L
+            val thirdValue = none<String>()
+
+            firstValue.cop<String, Long, Option<String>>() shouldBe firstValue.first<String, Long, Option<String>>()
```
JorgeCastilloPrz: didn't we remove cop ext funs?

modules/core/arrow-generic/src/test/kotlin/arrow/generic/CoproductTest.kt
```diff
@@ -82,5 +82,37 @@ class CoproductTest : UnitSpec() {
...
+            secondValue.cop<String, Long, Option<String>>() shouldBe secondValue.second<String, Long, Option<String>>()
+            thirdValue.cop<String, Long, Option<String>>() shouldBe thirdValue.third<String, Long, Option<String>>()
+        }
+
+        "types can be inferred with the cop replacement functions" {
+            fun typeInferenceTest(input: String?): Coproduct3<String, Int, Option<String>> {
```
JorgeCastilloPrz: would we need to add a test for inference when direcly using the ADT constructors in the function returns?

modules/core/arrow-generic/src/test/kotlin/arrow/generic/CoproductTest.kt
```diff
@@ -82,5 +82,37 @@ class CoproductTest : UnitSpec() {
...
+
+            typeInferenceTest(null) shouldBe Third<String, Int, Option<String>>(none())
+        }
+
+        "Coproduct data classes should be available" {
+            val firstValue = "String"
+            val secondValue = 100L
+            val thirdValue = none<String>()
+
+            First<String, Long, Option<String>>(firstValue) shouldBe firstValue.first<String, Long, Option<String>>()
```
JorgeCastilloPrz: A similar one but with a function to test type inference by using the ADT constructors would be nice

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -37,6 +38,18 @@ fun toDisplayValues(items: List<Coproduct3<Car, Dealership, Salesperson>>): List
...
+                Car(Speed(100)).first(),
+                Dealership("Cedar Falls, Iowa").second(),
+                Salesperson("Car McCarface").third()
+            )
+        )
+    )
+}
 ```
 
 Let's say we have an api. Our api operates under the following conditions:
```
JorgeCastilloPrz: I think we could also explain here what's the most effective benefit we've found of Coproducts over sealed classes: to flat the hierarchy. The alternative is composed sealed classes, which doesn't leave you in a pretty neat scenario (sealed classes don't support inheritance outside of their own hierarchy). Describing both scenarios and how they differ would be very enlightening for the reader.

JorgeCastilloPrz:
@raulraja don't you think the blogpost style would work good for some ideas here like what coproducts provide versus nested sealed classes? or would you completely move everything to Kdoc?

raulraja:
Yes but that also goes well in the class header unless it's tutorial style in which case is fine in the regular docs and should be linked from the class kdoc headers

abergfeld:
@JorgeCastilloPrz @raulraja This PR currently has docs being generated with the code:
```
/**
 * Runs the function related to the actual value of the Coproduct and returns the result
 *
 * Parameters: A function for every possible value the Coproduct could be.
 * Returns: RESULT generated by the input functions */
fun <A, B, C, RESULT> Coproduct3<A, B, C>.fold(
    a: (A) -> RESULT,
    b: (B) -> RESULT,
    c: (C) -> RESULT
): RESULT = when (this) {
    is First -> a(this.a)
    is Second -> b(this.b)
    is Third -> c(this.c)
}

```

Is that sufficient to auto link to the API DOCs section?

@raulraja Does the runnable code examples in the docs run using only the latest master version of Arrow? Even compiling the docs locally seems like it should work but they don't run correctly as if they're using the current master version of Coproduct

modules/core/arrow-generic/src/test/kotlin/arrow/generic/CoproductTest.kt
abergfeld: No I think the intent is to deprecate them and remove them in a later version? https://github.com/arrow-kt/arrow/issues/1284#issuecomment-460337722

I'm not sure what the next Arrow version will be I guess, I may have misunderstood

abergfeld:
@JorgeCastilloPrz I think I addressed all your comments except removing the extension functions. I'm not sure if we should just remove those functions now or not.

I've updated the kdocs to be more inlined with what Option is doing and the locally generated docs look like it's all linking up correctly and showing descriptions / params / returns. 

The only part I'm still not sure works is the runnable code examples but it seems like those are just based off the master version of arrow and the examples compile so I'd assume they'll work once merged?

Here's the generated output for Coproduct2 since it's generated and doesn't show as part of the diff:
```kotlin
package arrow.generic.coproduct2

import arrow.core.Option
import arrow.core.toOption
import kotlin.Deprecated
import kotlin.Suppress
import kotlin.Unit

/**
 * Represents a sealed hierarchy of 2 types where only one of the types is actually present.
 */
sealed class Coproduct2<A, B>

/**
 * Represents the first type of a Coproduct2
 */
data class First<A, B>(val a: A) : Coproduct2<A, B>()

/**
 * Represents the second type of a Coproduct2
 */
data class Second<A, B>(val b: B) : Coproduct2<A, B>()

@Deprecated(message = "This has issues with type inference, use First() instead",
replaceWith = ReplaceWith("First<A, B>(a)", "arrow.generic.coproduct2.First")
)
fun <A, B> coproductOf(a: A): Coproduct2<A, B> = First(a)

@Deprecated(message = "This has issues with type inference, use Second() instead",
replaceWith = ReplaceWith("Second<A, B>(b)", "arrow.generic.coproduct2.Second")
)
@Suppress("UNUSED_PARAMETER")
fun <A, B> coproductOf(b: B, dummy0: Unit = Unit): Coproduct2<A, B> = Second(b)

@Deprecated(message = "This has issues with type inference, use .first() instead",
replaceWith = ReplaceWith("this.first<A, B>()")
)
fun <A, B> A.cop(): Coproduct2<A, B> = coproductOf<A, B>(this)

@Deprecated(message = "This has issues with type inference, use .second() instead",
replaceWith = ReplaceWith("this.second<A, B>()")
)
@Suppress("UNUSED_PARAMETER")
fun <A, B> B.cop(dummy0: Unit = Unit): Coproduct2<A, B> = coproductOf<A, B>(this)

/**
 * Creates a Coproduct from the A type
 *
 * @return A Coproduct2<A, B> where the receiver is the A
 */
fun <A, B> A.first(): Coproduct2<A, B> = First(this)

/**
 * Creates a Coproduct from the B type
 *
 * @return A Coproduct2<A, B> where the receiver is the B
 */
fun <A, B> B.second(): Coproduct2<A, B> = Second(this)

/**
 * Transforms the Coproduct into an Option based on the actual value of the Coproduct
 *
 * @return None if the Coproduct was not the specified type, Some if it was the specified type
 */
fun <A> Coproduct2<A, *>.select(): Option<A> = (this as? First)?.a.toOption()

/**
 * Transforms the Coproduct into an Option based on the actual value of the Coproduct
 *
 * @return None if the Coproduct was not the specified type, Some if it was the specified type
 */
@Suppress("UNUSED_PARAMETER")
fun <B> Coproduct2<*, B>.select(dummy0: Unit = Unit): Option<B> = (this as? Second)?.b.toOption()

/**
 * Runs the function related to the actual value of the Coproduct and returns the result
 *
 * @param a The function used to map A to the RESULT type
 * @param b The function used to map B to the RESULT type
 *
 * @return RESULT generated by one of the input functions
 */
fun <A, B, RESULT> Coproduct2<A, B>.fold(a: (A) -> RESULT, b: (B) -> RESULT): RESULT = when (this) {
    is First -> a(this.a)
    is Second -> b(this.b)
}

```

JorgeCastilloPrz:
Deprecating the cop() functions makes sense 👍

modules/core/arrow-generic/src/test/kotlin/arrow/generic/CoproductTest.kt
raulraja: that is correct @abergfeld they will be deprecated in 0.9.0 coming out now and will be removed sometime before 1.0 final.

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -163,10 +235,13 @@ Here we're able to return the result of the `fold` and since it's exhaustively e
 
 We're able to take a Coproduct and `select` the type we care about from it. `select` returns an `Option`, if the value of the Coproduct was for the type you're trying to `select`, you'll get `Some`, if it was not the type used with `select`, you'll get `None`.
 
+{: data-executable='true'}
```
raulraja:
```suggestion
```
This is not needed if we use `kotlin:ank:playground`

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -163,10 +235,13 @@ Here we're able to return the result of the `fold` and since it's exhaustively e
...
+
 ```kotlin:ank
```
raulraja:
```suggestion
```kotlin:ank:playground
```

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -178,16 +253,28 @@ fun handleApiResult(
                     { database.insertRegistration(it) }
             )
 }
+
+fun main() {
```
raulraja: you can use `//sampleStart` and `//sampleEnd` to remove noise from the examples such as imports main etc.

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -178,16 +253,28 @@ fun handleApiResult(
...
+    println(
+        ServerError.first<CommonServerError, RegistrationError, Registration>()
+            .select<RegistrationError>()
+    )
+}
 ```
 
 `select` can only be called with a type that exists on the Coproduct, if the type doesn't exist, it won't compile:
+
+{: data-executable='true'}
```
raulraja:
```suggestion

```

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -178,16 +253,28 @@ fun handleApiResult(
...
+
 ```kotlin:ank
```
raulraja:
```suggestion
```kotlin:ank:fail
```
You can use `fail` to let it fail and display the compilation errors in the text.

modules/docs/arrow-docs/docs/docs/generic/coproduct/README.md
```diff
@@ -178,16 +253,28 @@ fun handleApiResult(
...
 import arrow.generic.*
 import arrow.generic.coproduct3.Coproduct3
 import arrow.generic.coproduct3.select
 
-fun handleApiResult(apiResult: Coproduct3<CommonServerError, RegistrationError, Registration>): Unit {
-//    apiResult.select<String>()
-//error: type mismatch: inferred type is Coproduct3<CommonServerError, RegistrationError, Registration> but Coproduct3<String, *, *> was expected
+fun main() {
+//    println(
```
raulraja: if you end up using `fail` you'd need to uncomment these all these `//`

modules/meta/arrow-meta/src/main/java/arrow/generic/CoproductFileGenerator.kt
```diff
@@ -171,6 +220,28 @@ private fun FileSpec.Builder.addFoldFunction(generics: List<String>): Unit {
...
+    val parameters = if (input.isEmpty()) {
+        ""
+    } else {
+        input.joinToString(separator = "") { "@param ${it.first} ${it.second}\n" } + "\n"
+    }
+
+    return "$description\n\n$parameters@return $output\n"
+}
+
+private fun String.toCamelCase(): String {
```
raulraja: Nothing wrong with this implementation style but soon we will enforce across all arrow modules a style that is more succinct and expression oriented:
```kotlin
private fun String.toCamelCase(): String =
  if (isEmpty()) this 
  else first().toLowerCase() + substring(1)
```
Also this function may be public and placed in the arrow meta String utils file.

abergfeld:
@raulraja I think I addressed all your comments 🎉 Let me know if I should update anything else!

JorgeCastilloPrz:
So no plans on generating runnable snippets with dokka anymore? Can we just merge like this @raulraja? (coproduct code is compile time generated, we'd get those snippets generated 26 times, which might be too much)

raulraja:
it's fine to get the snippets generated N times. All pages need to documented and some users may land straight in coproduct5 for example.

JorgeCastilloPrz:
Can we get this in on 0.9.0 or 0.9.1 or something?

raulraja:
This is coming out in 0.9.0. @nomisRev and I are making some final touches to Fx and then we will release. Thanks for your patience.

pakoito:
@raulraja make sure to double-check the crash!

raulraja:
any way to repro on tests or does it require an Android runtime? I may need help with this once fx is ready. I'll ping one of you guys that would not have to download the internet to test it on Android.

JorgeCastilloPrz:
What crash are we talking about?

PR #1293, opened 2019-02-08
Added Either doc in Russian

dobrowins:
1. Added Russian translation of Either class
2. Moved Try's translation to respective folder. So anyone who want to contribute would be redirected to proper page after clicking «Edit» button in the header.


pakoito:
Tell us when it's not WIP anymore so we can merge!

dobrowins:
> Tell us when it's not WIP anymore so we can merge!

Dear @pakoito I've removed an WIP status from the PR.
Translation is correct. I've checked the links. All seems fine.

pakoito:
Into the repo it goes then! Thank you :D

PR #1305, opened 2019-02-16
[#153] Control NonFatal Throwables

### LINKED ISSUE #153: Allow `NonFatal` control over exceptions for MonadError and Monadic comprehensions
Currently comprehending over Try in an stack unsafe way returns a `Failures(SOE)` we should never capture fatal exceptions. See for reference http://www.scala-lang.org/api/2.12.0/scala/util/control/NonFatal$.html

kartoffelsup:
As discussed in https://kotlinlang.slack.com/archives/C5UPMM0A0/p1550319070097800
Port of the Scala NonFatal.

I've hopefully found all places where it should be applied. I've searched with `catch \(.*:\s+Throwable\)` in all Production Files using IntelliJ.

Closes #153

### DIFF
--- a/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/ApplicativeError.kt
+++ b/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/ApplicativeError.kt
@@ -1,7 +1,14 @@
 package arrow.typeclasses
 
 import arrow.Kind
-import arrow.core.*
+import arrow.core.Either
+import arrow.core.Left
+import arrow.core.NonFatal
+import arrow.core.OptionOf
+import arrow.core.Right
+import arrow.core.TryOf
+import arrow.core.fix
+import arrow.core.identity
 
 /**
  * ank_macro_hierarchy(arrow.typeclasses.ApplicativeError)
@@ -36,7 +43,11 @@ interface ApplicativeError<F, E> : Applicative<F> {
     try {
       just(f())
     } catch (t: Throwable) {
-      raiseError<A>(recover(t))
+      if (NonFatal(t)) {
+        raiseError<A>(recover(t))
+      } else {
+        throw t
+      }
     }
 
   fun <A> ApplicativeError<F, Throwable>.catch(f: () -> A): Kind<F, A> =
--- a/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/internal/IOBracket.kt
+++ b/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/internal/IOBracket.kt
@@ -1,6 +1,7 @@
 package arrow.effects.internal
 
 import arrow.core.Either
+import arrow.core.NonFatal
 import arrow.effects.*
 import arrow.effects.typeclasses.ExitCase
 import java.util.concurrent.atomic.AtomicBoolean
@@ -61,7 +62,11 @@ internal object IOBracket {
               val fb = try {
                 use(a)
               } catch (e: Throwable) {
-                IO.raiseError<B>(e)
+                if (NonFatal(e)) {
+                  IO.raiseError<B>(e)
+                } else {
+                  throw e
+                }
               }
 
               IO.Bind(fb.fix(), frame)
--- a/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Async.kt
+++ b/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Async.kt
@@ -2,6 +2,7 @@ package arrow.effects.typeclasses
 
 import arrow.Kind
 import arrow.core.Either
+import arrow.core.NonFatal
 import arrow.core.Right
 import arrow.documented
 import arrow.typeclasses.MonadContinuation
@@ -142,7 +143,7 @@ interface Async<F> : MonadDefer<F> {
       try {
         just(f())
       } catch (t: Throwable) {
-        raiseError<A>(t)
+        t.raiseNonFatal<A>()
       }
     }
 
--- a/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Async.kt
+++ b/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Async.kt
@@ -2,7 +2,6 @@ package arrow.effects.typeclasses
 
 import arrow.Kind
 import arrow.core.Either
-import arrow.core.NonFatal
 import arrow.core.Right
 import arrow.documented
 import arrow.typeclasses.MonadContinuation
--- a/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDefer.kt
+++ b/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDefer.kt
@@ -2,7 +2,6 @@ package arrow.effects.typeclasses
 
 import arrow.Kind
 import arrow.core.Either
-import arrow.core.NonFatal
 import arrow.core.Tuple2
 import arrow.core.toT
 import arrow.effects.data.internal.BindingCancellationException
--- a/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDefer.kt
+++ b/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDefer.kt
@@ -2,6 +2,7 @@ package arrow.effects.typeclasses
 
 import arrow.Kind
 import arrow.core.Either
+import arrow.core.NonFatal
 import arrow.core.Tuple2
 import arrow.core.toT
 import arrow.effects.data.internal.BindingCancellationException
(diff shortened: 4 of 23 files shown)


modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/ApplicativeError.kt
```diff
@@ -36,7 +43,11 @@ interface ApplicativeError<F, E> : Applicative<F> {
     try {
       just(f())
     } catch (t: Throwable) {
-      raiseError<A>(recover(t))
+      if (NonFatal(t)) {
```
raulraja: We can encapsulate this pattern For All `ApplicativeError<F>` which constitutes most of the changes in this PR.
If you place this function as part of the MonadThrow<F> type class
```kotlin
fun <F, A> Throwable.raiseNonFatal(): Kind<F, A> =
  if (NonFatal(this)) raiseError<A>(this) else throw e
```

This functions is scoped to the MonadThrow receiver and will be available in all instances of all the data types as extension functions and would allow you to remove the if else boilerplate in most cases below.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/internal/IOBracket.kt
```diff
@@ -61,7 +62,11 @@ internal object IOBracket {
               val fb = try {
                 use(a)
               } catch (e: Throwable) {
-                IO.raiseError<B>(e)
+                if (NonFatal(e)) {
```
raulraja: `e.raiseNonFatal()`

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Async.kt
```diff
@@ -142,7 +143,11 @@ interface Async<F> : MonadDefer<F> {
       try {
         just(f())
       } catch (t: Throwable) {
-        raiseError<A>(t)
+        if (NonFatal(t)) {
```
raulraja: `t.raiseNonFatal()`

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDefer.kt
```diff
@@ -25,7 +26,11 @@ interface MonadDefer<F> : MonadThrow<F>, Bracket<F, Throwable> {
       try {
         just(f())
       } catch (t: Throwable) {
-        raiseError<A>(t)
+        if (NonFatal(t)) {
```
raulraja: `t.raiseNonFatal()`

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDefer.kt
```diff
@@ -38,7 +43,11 @@ interface MonadDefer<F> : MonadThrow<F>, Bracket<F, Throwable> {
...
```
raulraja: `t.raiseNonFatal()`

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDeferCancellableContinuations.kt
```diff
@@ -59,7 +60,11 @@ open class MonadDeferCancellableContinuation<F, A>(val SC: MonadDefer<F>, overri
       val datatype = try {
         just(m())
       } catch (t: Throwable) {
-        raiseError<B>(t)
+        if (NonFatal(t)) {
```
raulraja: `t.raiseNonFatal()`

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/ApplicativeError.kt
kartoffelsup: Thanks for the review and the great suggestion. :) I've applied it to all but IOBracket, it didn't work there (I guess because it's not running in the MonadThrow context?)

There is also a AbstractMethodError in `IOTest MonadDefer bind: binding failure` that I can't quite figure out.

pakoito: Change the genError code to throw something else instead, maybe that's the issue. I'd add a new generator that throws one of the fatal errors and test for that.

kartoffelsup: After updating graalvm to rc12 (my jdk), the test no longer fails.
Should I add the fatal error test to MonadDefer laws?

pakoito: I'd add a MonadThrow laws altogether that have all MonadErrorLaws, put that test in them, and make MonadDefer do MonadThrowLaws instead of MonadErrorLaws.

That way we have tests where the functions are defined in the hierarchy :D

kartoffelsup: Mmh, does this test even work inside MonadThrow? There is no delay/defer that actually does a try {} catch {} or am I missing something?

pakoito: Then the function should live in MonadDefer :D @raulraja opinions?

raulraja: The function can be expressed in `MonadThrow` and override `catch` to use it:
```kotlin
override fun <A> catch(recover: (Throwable) -> Throwable, f: () -> A): Kind<F, A> 
override fun <A> ApplicativeError<F, Throwable>.catch(f: () -> A): Kind<F, A> 
```
It has application outside suspensions of effects given the loose approach we take with `Try` at the moment. This is an example of a datatype that would benefit from it and that does not allow suspension and can't implement MonadDefer. Another one would be Either for partially applied Throwables.
IMO it should live in `MonadThrow`.

kartoffelsup: Sorry, I'm kind of lost now. :D I've added the suggested function `raiseNonFatal` to MonadThrow.
What else can I do? (and how can I/should I test it as MonadThrow Law?)

raulraja: @kartoffelsup this should cover it:

- override the two functions above using `raiseNonFatal` to implement them in `MonadThrow`
- Add a new law in the `arrow.test.laws.MonadErrorLaws` 
> Monad Error Laws: NonFatal is caught
- Add a new la  in the `arrow.test.laws.MonadErrorLaws` 
> Monad Error Laws: Fatal errors are thrown
- Add kdoc and a `kotlin:ank:playground` snippet similar to the one in `Functor` that illustrates the usage of NonFatal in the `NonFatal` object.

That should do it.

kartoffelsup: @raulraja thanks for the guidance!
Regarding the catch overrides, atm I've already overridden them in ApplicativeError like so:
```kotlin
 fun <A> catch(recover: (Throwable) -> E, f: () -> A): Kind<F, A> =
    try {
      just(f())
    } catch (t: Throwable) {
      if (NonFatal(t)) {
        raiseError<A>(recover(t))
      } else {
        throw t
      }
    }

  fun <A> ApplicativeError<F, Throwable>.catch(f: () -> A): Kind<F, A> =
    catch(::identity, f)
```
Do we really need the override in MonadThrow? 
Also, it looks to me that the `catch(recover: (Throwable) -> E, f: () -> A)` variant can't be done with `raiseNonFatal` as I am unable to call `#map` on it for the `recover` function.

raulraja: no, if you already did it there it's fine how you did it.

kartoffelsup: I've applied the changes you requested. Please let me know if there is anything else I can do. Thanks again for the help!

raulraja:
Looks like reactor Flowable is not stack safe https://app.bitrise.io/build/73c812fbc6ef9620.
Can you decrease that number.
You can check the docs run by running `./gradlew clean dokka :arrow-docs:runAnk`

raulraja:
I'll rerun the build and if it passes we will leave that for another PR

raulraja:
@pakoito any idea what is going on here and why this has not surfaced before? https://app.bitrise.io/build/288ec840d3cdc0ca

pakoito:
None of the Rx constructs are stack-safe.

kartoffelsup:
Looks like that is intended there? https://arrow-kt.io/docs/integrations/rx2/#stack-safety, the comment above states this is what's supposed to happen. But the StackOverflow was caught previously.

PR #1306, opened 2019-02-17
Add '-data' Suffix to Effects Modules in Getting Started Dependencies

kartoffelsup:
Add '-data' Suffix to Effects Modules in Getting Started Dependencies


pakoito:
Derp this should habe been fixed a while ago.

PR #1311, opened 2019-02-22
Add divide typeclasses

1Jajen1:
Partially solves #284 (Only the divide parts)

Again like stated in #1181 there some annoying parts:
* The docs are there, I don't really like them, but could not come up with better text/examples :see_no_evil: 
* There are no laws for decidable atm, there are some [here](http://hackage.haskell.org/package/contravariant-1.5/docs/src/Data.Functor.Contravariant.Divisible.html), but I don't really understand them well enough to attempt to port them :/ Scala cats and/or scalaz don't include them either afaik.
* Some monad transformer instances are missing tests because we don't have instances that fulfill the constraints for them (don't really know what to do there)

Any ideas for better examples/texts do describe them? I'll think about it some more, but don't know if I can come up with something better :/


1Jajen1:
Hmm I don't think I understand the error ank throws here. It happens locally as well, but not when the code is just run in a standalone file (copy pasted over basically). I am confused :/

pakoito:
I'll review this tonight or tomorrow!

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/DivideLaws.kt
```diff
@@ -0,0 +1,40 @@
...
+    Law("Divide laws: Associative") { DF.associative(cf, EQ) }
+  )
+
+  fun <A> delta(a: A): Tuple2<A, A> = a toT a
+
+  fun <F> Divide<F>.associative(
+    cf: (Int) -> Kind<F, Int>,
+    EQ: Eq<Kind<F, Int>>
+  ): Unit =
+    forAll(Gen.int().map(cf)) { fa ->
```
pakoito: Cute. I like it :)

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/DivideLaws.kt
```diff
@@ -0,0 +1,40 @@
...
+      EQ.run {
```
pakoito: I believe this should use `equalsUnderTheLaw` instead of `EQ` directly.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/DivisibleLaws.kt
```diff
@@ -0,0 +1,38 @@
...
+    Law("Divisible laws: Left identity") { DF.leftIdentity(cf, EQ) },
+    Law("Divisible laws: Right identity") { DF.rightIdentity(cf, EQ) }
+  )
+
+  fun <F> Divisible<F>.leftIdentity(
+    cf: (Int) -> Kind<F, Int>,
+    EQ: Eq<Kind<F, Int>>
+  ): Unit =
+    forAll(Gen.int().map(cf)) { fa ->
+      EQ.run {
```
pakoito: Same as above, `equalsUnderTheLaw`

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Decidable.kt
```diff
@@ -0,0 +1,368 @@
...
+ *      }
+ *   }
+ * }
+ *
+ * val stringSerializer = Serializer<String> { "STRING: $it" }
+ * val intSerializer = Serializer<Int> { "INT: $it" }
+ *
+ * fun main(args: Array<String>) {
+ *   //sampleStart
+ *   val stringOrInt: Serializer<Either<String, Int>> = Serializer.decidable()
```
pakoito: Powerful shit.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Decidable.kt
```diff
@@ -0,0 +1,368 @@
...
+    fb: Kind<F, B>,
+    fc: Kind<F, C>,
+    fd: Kind<F, D>,
+    fe: Kind<F, E>,
+    ff: Kind<F, FF>,
+    fg: Kind<F, G>,
+    fh: Kind<F, H>,
+    fi: Kind<F, I>,
+    fj: Kind<F, J>,
+    f: (Z) -> Either<A, Either<B, Either<C, Either<D, Either<E, Either<FF, Either<G, Either<H, Either<I, J>>>>>>>>>
```
pakoito: @raulraja is there any way of generalizing this for Product? Or convert between them?

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Divide.kt
```diff
@@ -0,0 +1,318 @@
+package arrow.typeclasses
+
+import arrow.Kind
+import arrow.core.*
+
+/**
+ * [Divide] is a typeclass that models the divide part of divide and conquer.
+ *
+ * [Divide] basically states: Given a Kind<F, A> and a Kind<F, B> and a way to turn C into an A and B it gives you a Kind<F, C>
```
pakoito: This doesn't seem correct: `(Z) -> Tuple2<A, B>`

pakoito: `into an A and B` may be better phrased as a `tuple of A and B`, even if it's an implementation detail it's easy to miss.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Divide.kt
```diff
@@ -0,0 +1,318 @@
...
+    f: (Z) -> Tuple10<A, B, C, D, E, FF, G, H, I, J>
+  ): Kind<F, Z> = divide(
+    fa,
+    fb.product(fc).product(fd).product(fe).product(ff).product(fg).product(fh)
+      .product(fi).product(fj)
+  ) {
+    val (a, b, c, d, e, fff, g, h, i, j) = f(it)
+    a toT Tuple9(b, c, d, e, fff, g, h, i, j)
+  }
+}
```
pakoito: NIT: additional empty line

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Divide.kt
```diff
@@ -0,0 +1,318 @@
...
+    other: Kind<F, J>,
+    dummy: Unit = Unit,
+    dummy2: Unit = Unit,
+    dummy3: Unit = Unit,
+    dummy4: Unit = Unit,
+    dummy5: Unit = Unit,
+    dummy6: Unit = Unit,
+    dummy7: Unit = Unit,
+    dummy8: Unit = Unit
+  ): Kind<F, Tuple10<A, B, C, D, E, FF, G, H, I, J>> =
```
pakoito: RIP IntelliJ

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/divide/README.md
```diff
@@ -0,0 +1,87 @@
...
+  - /docs/typeclasses/divide
+---
+
+## Divide
+
+{:.beginner}
+beginner
+
+`Divide` is a typeclass that models the divide part of divide and conquer.
+`Divide` basically states: Given a `Kind<F, A>` and a `Kind<F, B>` and a way to turn `C` into an `A` and `B` it provides you a `Kind<F, C>`
```
pakoito: Same as above, mention a coproduct or a tuple or something that helps with the association, else it's easy to miss.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Divide.kt
1Jajen1: I've done worse with a small project regarding generators for prop testing (generated ~1k lines of Tuple code for generating tuples and shrinkers for them) :) IntelliJ takes a while to load that file :D

1Jajen1:
Done the small stuff. Tests are missing for StateT (we have no monad that implements Divide etc) (although this may be possible with St

1Jajen1:
Done the small stuff.
Tests are missing for StateT (we have no monad that implements Divide etc) (although this may be possible with using like OptionT and Const like in their tests, which is already weird but might work) and Sum (again no comonads that implement divide etc)

1Jajen1:
Added a test for Sum. Now only StateT is missing, but since we have no instance that is Monad<F> and Divisible<F> that is kinda hard to do. (Function1 is almost it with `Monad<Function1PartialOf<A>>` but Divisible is defined as `Divisible<Conested<ForFunction1, A>>`)

1Jajen1:
So ank is being funny (or rather what executes the script). The code 100% compiles and runs fine but running in ank crashes with:
`
Backend Internal error: Exception during code generation
Cause: Back-end (JVM) Internal error: Don't know how to generate outer expression: Class: class Companion
Cause: Don't know how to generate outer expression: Class: class Companion
File being compiled at position: (17,19) in /Line_1.kts
The root cause was thrown at: CodegenContext.java:219
File being compiled at position: mock:///Line_1.kts
The root cause was thrown at: ExpressionCodegen.java:319
`
I don't really understand whats going on atm

pakoito:
Do you know what file is that for?

1Jajen1:
> Do you know what file is that for?

The scary internal error is for Decidable's README and the other type inference error happens for all snippets where Serializer's contramap is defined without this@contramap (every snippet in kdoc for divide etc)

pakoito:
I'd move the setup code to the arrow-docs project instead. You'll probably get codegen right with it too.

pakoito:
Here: https://github.com/arrow-kt/arrow/tree/master/modules/docs/arrow-docs/src/main/kotlin

1Jajen1:
Yeah this fixed it. Too bad the examples now cannot be viewed in full, but better than no examples at all

modules/docs/arrow-docs/src/main/kotlin/DivideHelper.kt
```diff
@@ -0,0 +1,44 @@
+package com.example.domain
```
pakoito: @1Jajen1 I didn't see this before merging. Could you please make it a nicer package?

PR #1313, opened 2019-02-23
Fixes for FX docs

aballano:
* Fixed some small errors causing the docs samples not to run properly.
  * `effects/fx/README.md` -> Removed ank from `suspend composition` example since there's no main function and the example doesn't need it.
  * `effects/fx/async/README.md` -> Fixed example in `Cancellation` by moving the destructuring to a function since "Destructuring declarations are only allowed for local variables/values".
  * `effects/fx/polymorphism/README.md` -> 
    * Fixed one of the examples by converting an expression body main function to block body. 
    * Fixed another by adding the `k()` call for the conversion from `List` to `ListK`. Also changed the binding to `!` as it's more explicit and less error-prone than destructuring, which also works for lists.


modules/docs/arrow-docs/docs/docs/effects/fx/async/README.md
```diff
@@ -241,7 +242,7 @@ suspend fun printThreadName(): Unit =
   println(Thread.currentThread().name)
 
 suspend fun program() = 
-  async { printThreadName() }
+  async { printThreadName() } //XXX This doesn't execute: "IllegalAccessError: tried to access field kotlinx.coroutines.Dispatchers.Default from class kotlinx.coroutines.DeferredKt"
```
aballano: This is throwing the exception shown above.

modules/docs/arrow-docs/docs/docs/effects/fx/async/README.md
```diff
@@ -261,7 +262,7 @@ suspend fun printThreadName(): Unit =
   println(Thread.currentThread().name)
 
 suspend fun program() = 
-  async(start = CoroutineStart.LAZY) { printThreadName() }
+  async(start = CoroutineStart.LAZY) { printThreadName() }  //XXX This doesn't execute: same as above
```
aballano: Same

modules/docs/arrow-docs/docs/docs/effects/fx/async/README.md
pakoito: This is new

modules/docs/arrow-docs/docs/docs/effects/fx/polymorphism/README.md
```diff
@@ -26,17 +26,17 @@ import arrow.effects.extensions.io.fx.fx
...
 
-/* for all `F` that provide an `UnsafeRun` extension define a main function
+/* for all `F` that provide an `UnsafeRun` extension define a main function */
 fun <F> UnsafeRun<F>.main(fx: Fx<F>): Int =
   unsafe { runBlocking { fx.program() } }
 
 /* Run program in the IO monad */
 fun main() =
-  IO.unsafeRun().main(IO.fx()) 
+  IO.unsafeRun().main(IO.fx()) //XXX This doesn't compile
```
aballano: There seems to be no extension function `fx()`.

There's a similar example in `EffectsSuspendDSLTests -> "FX supports polymorphism"` with the import `import arrow.effects.extensions.io.fx.fx` which cannot be resolved

modules/docs/arrow-docs/docs/docs/effects/fx/async/README.md
```diff
@@ -187,12 +187,13 @@ import arrow.effects.IO
 import arrow.unsafe
 import arrow.effects.extensions.io.unsafeRun.runBlocking
 import arrow.effects.extensions.io.fx.fxCancellable
+
+fun main() { // The edge of our world
 //sampleStart
-val (_, disposable) = fxCancellable {
-  !effect { println("BOOM!") }
-}
+  val (_, disposable) = fxCancellable {
```
aballano: Is not possible to do destructuring outside local variables, so I moved this inside the main

modules/docs/arrow-docs/docs/docs/effects/fx/polymorphism/README.md
raulraja: @dominv is this because the remote server is not including the right libs?

aballano: Sounds like that, because the import for `arrow.effects.extensions.io.fx.fx` is not properly resolved on my machine as well.

raulraja:
is this still relevant with latest from master?

modules/docs/arrow-docs/docs/docs/effects/fx/polymorphism/README.md
```diff
@@ -194,8 +195,8 @@ import arrow.data.extensions.list.fx.fx
 //sampleStart
 val result1 = unsafe { 
   fx {
-    val (a) = listOf(1, 2)
-    val (b) = listOf(true, false)
+    val (a) = listOf(1, 2)          // This is destructuring the list and is obtaining the first element,
+    val (b) = listOf(true, false)   // that's why the output is `ListK(list=[Tuple2(a=1, b=true)])`
```
aballano: @pakoito @raulraja I don't totally understand this example, is this behaviour on purpose?

raulraja: yeah, this is wrong, because List has it's own component1. we should use ! or bind here to fix the example and this will be ready to go. Thanks!

aballano: @raulraja Found the issue, the binding wasn't properly executed because the `.k()` call on the lists was missing.

raulraja:
Thanks @aballano!

PR #1318, opened 2019-02-25
Add arrow.syntax.unsafe Option.get() extension method

starkej2:
From a conversation in [gitter](https://gitter.im/arrow-kt/Lobby?at=5c73c266e5eeec0d9bc1ecee):

> @raulraja: Arrow won't have an unsafe Option.get in core but as mentioned if anyone wants to maintain an unsafe module we are not opposed to. It just has to be outside of core or any of the other safe Arrow modules.

Let me know if I need to move this extension to a different location or make any other changes/additions 😎

### DIFF
--- a/modules/core/arrow-syntax/src/main/kotlin/arrow/syntax/unsafe/option.kt
+++ b/modules/core/arrow-syntax/src/main/kotlin/arrow/syntax/unsafe/option.kt
@@ -1,0 +1,11 @@
+package arrow.syntax.unsafe
+
+import arrow.core.Option
+
+/**
+ * Returns the [Option]'s value. This method should only be used in cases where the [Option] is not
+ * expected to be empty.
+ *
+ * @throws NoSuchElementException if the [Option] is empty.
+ */
+fun <T> Option<T>.get(): T = fold({ throw NoSuchElementException("None.get") }, { it })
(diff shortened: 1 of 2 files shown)


starkej2:
> By module I think he meant a gradle module, as in a new library `arrow-syntax-unsafe`

Meaning I should move it to `modules/core/arrow-syntax-unsafe`?

> I will gladly help you refactor your code instead of accepting any PR to create such a library

We mostly just find `Option.get()` useful in tests. [I could use some advice](https://stackoverflow.com/questions/54836474/difficulty-using-arrow-kt-either-option-and-rxjava-together) using arrow with rxjava though if you have time (unrelated).

pakoito:
I remember that! I told you to use sealed classes. I'll message you in Slack in a second.

For JUnit testing you can do this:

```
fun <A> Option<A>.assertValue(a: A) = fold({ fail ("None found when expecting Some($a)") }, { assertEquals(it, a) })

fun <A> Option<A>.assertAnyValue() = fold({ fail ("No value found when expecting Some(...)") }, { })

fun <A> Option<A>.assertNoValue() = fold({ }, { fail ("Value found when expecting None") })
```

and for property-based testing

```
fun <A> Option<A>.assertValue(a: A) = fold({ false }, { it == a })

fun <A> Option<A>.assertAnyValue() = fold({ false }, { true })

fun <A> Option<A>.assertNoValue() = fold({ true }, { false })
```

starkej2:
Cool we'll use those approaches and/or add an extension method to our project. Thanks!

PR #1320, opened 2019-02-28
Fixes for Fx

pakoito:
EZPZ


modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/monaddefer/Fx.kt
```diff
@@ -1,16 +1,11 @@
...
-import arrow.core.Tuple2
-import arrow.effects.typeclasses.Disposable
 import arrow.effects.typeclasses.MonadDefer
 import arrow.effects.typeclasses.MonadDeferCancellableContinuation
 
 interface Fx<F> {
   fun monadDefer(): MonadDefer<F>
   fun <A> fx(f: suspend MonadDeferCancellableContinuation<F, *>.() -> A): Kind<F, A> =
     monadDefer().bindingCancellable(f).a
-  fun <A> fxCancellable(f: suspend MonadDeferCancellableContinuation<F, *>.() -> A): Tuple2<Kind<F, A>, Disposable> =
```
pakoito: Also this is impossible because defer does not imply concurrent. Even if it works, it's merely an accident of how suspension works :P

pakoito: Worth considering removing bindingCancellable from MonadDefer too.

raulraja: let's get rid of it. suspension implies async in kotlin

pakoito:
Master is broken, all in

PR #1321, opened 2019-02-28
Add semigroup instance for Tuple2

pakoito:


pakoito:
I know this is correct so by the power of Grayskull imma ignore the test errors

PR #1322, opened 2019-03-01
Improvements for Try

### LINKED ISSUE #1078: Should `Try.raise` have a `Nothing` generic?
Problem:

```kotlin
repo.fetchNameOfBank() // returns `Try<String>`
    .map { name: String -> name.capitalize() } // transformation on expected value
    .flatMap { name: String ->
        val isInsured = repo.isBankFdicInsured(name)
        if (isInsured) {
            Try.Success(name)
        } else {
            // I would like to do this, but it doesn't compile due to a type inference failure
            // Try.raise(Exception("Bank is not insured"))
            
            // so I do this, which is more verbose, but it compiles
            Try.raise<String>(Exception("Bank is not insured"))
        }
    }
```

The issue is having to define `Try.raise<String>(exception)` instead of just `Try.raise(exception)`.

```kotlin
// current implementation, that has the type inference issue
fun <A> raise(e: Throwable): Result<A> = Unexpected(e)

// proposed implementation, which would fix the type inference issue
fun raise(e: Throwable): Result<Nothing> = Unexpected(e)
```

I'm sure there's a reason this was designed this way, so changing it now may be too late...but what is that reasoning?

pakoito:
Closes #1078


pakoito:
Failures in the reflection test suite. We're okay. Merging.

PR #1325, opened 2019-03-01
Fix order super interfaces

nomisRev:


modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/tuple.kt
```diff
@@ -111,7 +111,7 @@ interface Tuple2Semigroup<A, B> : Semigroup<Tuple2<A, B>> {
 }
 
 @extension
-interface Tuple2Monoid<A, B> : Tuple2Semigroup<A, B>, Monoid<Tuple2<A, B>> {
```
nomisRev: `val typeClass = typeSignature.superinterfaceSignatures[0].fullyQualifiedClassName` this is how we resolve the typeclass with reflection.

Which we then use to figure out what data type it is defined for. In this case we were resolving to `A` instead of `Tuple2<A, B>` hence the reflection code crashed.

PR #1327, opened 2019-03-03
Add Monoidal type class

### LINKED ISSUE #1323: Add Monoidal type class
Request comes from [here](https://github.com/arrow-kt/arrow/pull/1280#discussion_r259910269).

juliankotrba:
- [x] Interface definition
- [x] Laws
- [x] Instances
    - [x] Option
    - [x] ListK
    - [x] SequenceK
    - [x] SetK
- [x] Tests
- [x] Documentation

closes #1323


modules/core/arrow-core-data/src/test/kotlin/arrow/core/OptionTest.kt
```diff
@@ -49,7 +50,8 @@ class OptionTest : UnitSpec() {
       TraverseFilterLaws.laws(Option.traverseFilter(), Option.applicative(), ::Some, Eq.any()),
       MonadFilterLaws.laws(Option.monadFilter(), ::Some, Eq.any()),
       HashLaws.laws(Option.hash(Int.hash()), Option.eq(Int.eq())) { it.some() },
-      SemigroupalLaws.laws(Option.semigroupal(), ::Some, ::bijection, associativeSemigroupalEq)
+      SemigroupalLaws.laws(Option.semigroupal(), ::Some, ::bijection, associativeSemigroupalEq),
```
pakoito: SemigroupalLaws shouldn't be necessary anymore, because they're included in MonoidalLaws.

juliankotrba: I have not included them yet but I could do something like `SemigroupalLaw.laws + listOf(..)` in `MonoidalLaws` if you want.

pakoito: Yeah, that works!

pakoito:
Ready to review whenever you're ready to publish :D

juliankotrba:
Ready to review but I have no idea how to change the PR to _Ready to review_ 😅. Maybe that's because a check hasn't finished yet?

pakoito:
I can see it here.

Screen Shot 2019-03-11 at 11 14 29 AM

I'll click the button!

pakoito:
Branch has conflicts, sadly

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonoidalLaws.kt
```diff
@@ -0,0 +1,46 @@
...
+          ASSOCIATIVE_SEMIGROUPAL_EQ: Eq<Kind<F, Tuple2<Int, Tuple2<Int, Int>>>>
+  ): List<Law> =
+    SemigroupalLaws.laws(MDAL, f, BIJECTION, ASSOCIATIVE_SEMIGROUPAL_EQ) + listOf(
+      Law("Monoidal Laws: Left identity") { MDAL.monoidalLeftIdentity(f, EQ) },
+      Law("Monoidal Laws: Right identity") { MDAL.monoidalRightIdentity(f, EQ) }
+    )
+
+  private fun <F> Monoidal<F>.monoidalLeftIdentity(f: (Int) -> Kind<F, Int>, EQ: Eq<Kind<F, Tuple2<Int, Int>>>): Unit =
+    forAll(Gen.int().map(f)) { fa: Kind<F, Int> ->
+      identity<Int>().product(fa).equalUnderTheLaw(identity(), EQ)
```
pakoito: I thought this would be equal to `fa`

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Monoidal.kt
```diff
@@ -0,0 +1,109 @@
...
+import arrow.core.Option
+
+/**
+ * ank_macro_hierarchy(arrow.typeclasses.Monoidal)
+ *
+ * The [Monoidal] type class adds an identity element to the [Semigroupal] type class by defining the function [identity].
+ *
+ * [identity] returns a specific identity `Kind<F, A>` value for a given type [F] and [A].
+ *
+ * Like any other Monoid type class, [Monoidal] also complies with the left and right identity law:
```
pakoito: Same as in the laws, I thought that the product of `empty` and `a` is `a`, and not `empty`

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonoidalLaws.kt
juliankotrba: Of course you are right, don't know why I implemented it like this. Thanks for pointing that out!

juliankotrba:
> I can see it here.
> 
> Screen Shot 2019-03-11 at 11 14 29 AM
> 
> I'll click the button!

The button is no longer visible to me .

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonoidalLaws.kt
juliankotrba: I think I remember why I did it like this. For example, if we look at ListK:
`fa = {1, 2} `
`fb = {}`
The current outcome of `fa x fb` is `{}`
`{}` is also the `identity` element of `ListK`, so the outcome within a `Monoidal` should actually be `fa x fb = fa` and not  `fa x fb = {}`. This kinda confuses me 😅. It would be great if you could shed some light on this.

pakoito: >The current outcome of fa x fb is {}

It shouldn't, should it? AFAIK the outcome should be the same as with monoid. That is, the concatenation of both lists.

juliankotrba: I am understanding the `Semigroupal` type class as an abstraction over the cartesian product so the current behaviour is `fa x {} = {}`

pakoito: Ooooooh. Never mind me then.

juliankotrba: But then I still have to change the documentation, right?

pakoito:
And in! Thank you again @juliankotrba for working on this :D Hope you had fun

Issue #1324, opened 2019-03-01
handleErrorWith for Either


pakoito:
Add it to Either.kt as a member function. It's currently defined inline in the instance so it's a case of copypasting it :D

https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/either.kt#L92-L98

kioba:
I am happy to take it, I will set up the project and finish on the weekend

pakoito:
Thanks @kioba for the fixxxx

kioba:
Thank you for letting me contribute! :)

PR #1328, opened 2019-03-05
handleErrorWith for Either

kioba:
fix for the task: #1324
-[ ] documentation does not contain example code to handling error ( `[Left]` ). Could be some example added.


modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -367,3 +367,16 @@ fun <A, B> B?.rightIfNotNull(default: () -> A): Either<A, B> = when (this) {
   null -> Either.Left(default())
   else -> Either.Right(this)
 }
+
+
+/**
+ * Applies the given function `f` if this is a [Left], otherwise returns this if this is a [Right].
+ * This is like `flatMap` for the exception.
```
kioba: instead of saying "exception" we could mention Left:
`This is like `flatMap` for the [Left].`
but the name mentions that this is an error type with "handleErrorWith" so I left the exception there.

PR #1332, opened 2019-03-07
AQL Effects

### LINKED ISSUE #1158: AQL instances for IO, Observable, Flux, Deferred...
We need to expand the AQL instance family :D

alissonfpmorais:
Add AQL instances for: IO, ObservableK, FlowableK, MaybeK, SingleK, FluxK and MonoK.

For each datatype a new set of operators was added:
- IO: select, from, where;
- ObservableK: select, from, where, count, sum, groupBy, orderBy, union;
- FlowableK: select, from, where, count, sum, groupBy, orderBy, union;
- MaybeK: select, from, where, count, sum, groupBy, orderBy, union;
- SingleK: select, from, where;
- FluxK: select, from, where, count, sum, groupBy, orderBy, union;
- MonoK: select, from, where;

This PR fixes #1158


modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -220,6 +235,25 @@ sealed class IO<out A> : IOOf<A> {
...
+    override fun <B> mapFilter(f: (A) -> Option<B>): IO<B> {
+      val h: (Option<A>) -> Option<B> = { it.fold({ None }, { a: A -> f(a) }) }
+
+      return when {
+        index != maxStackDepthSize -> MapFilter(source, g.andThen(h), index + 1)
+        else -> MapFilter(this, f, 0)
+      }
+    }
+
+    override fun unsafeRunTimedTotal(limit: Duration): Option<A> = throw AssertionError("Unreachale")
```
pakoito: Typo: Unreacha*b*le

modules/aql/arrow-query-language/build.gradle
```diff
@@ -1,6 +1,9 @@
 dependencies {
     compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
     compile project(':arrow-annotations')
+    compile project(':arrow-effects-extensions')
```
pakoito: I believe this should be the other way around, where the extension packages depend on AQL. Else this library will contain 3 different concurrency runtimes :D

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -96,6 +96,9 @@ sealed class IO<out A> : IOOf<A> {
   open fun <B> map(f: (A) -> B): IO<B> =
     Map(this, f, 0)
 
+  open fun <B> mapFilter(f: (A) -> Option<B>): IO<B> =
```
pakoito: YES

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -155,6 +158,15 @@ sealed class IO<out A> : IOOf<A> {
     // Pure can be replaced by its value
     override fun <B> map(f: (A) -> B): IO<B> = Suspend { Pure(f(a)) }
 
+    // Pure can be replaced by its value
+    override fun <B> mapFilter(f: (A) -> Option<B>): IO<B> =
```
pakoito: YES

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -165,6 +177,9 @@ sealed class IO<out A> : IOOf<A> {
     // Errors short-circuit
     override fun <B> map(f: (Nothing) -> B): IO<B> = this
 
+    // Errors short-circuit
+    override fun <B> mapFilter(f: (Nothing) -> Option<B>): IO<B> = this
```
pakoito: YEEEESSSS

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
alissonfpmorais: done

modules/aql/arrow-query-language/build.gradle
alissonfpmorais: Okey dokey, I'll move those files to right packages

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -220,6 +235,25 @@ sealed class IO<out A> : IOOf<A> {
...
+
+  internal data class MapFilter<E, out A>(val source: IOOf<E>, val g: (E) -> Option<A>, val index: Int) : IO<A>(), (E) -> IO<A> {
+    override fun invoke(value: E): IO<A> = g(value)
+      .fold(
+        { raiseError(IllegalArgumentException("IO execution should yield a valid result")) },
+        { just(it) }
+      )
+
+    override fun <B> mapFilter(f: (A) -> Option<B>): IO<B> {
+      val h: (Option<A>) -> Option<B> = { it.fold({ None }, { a: A -> f(a) }) }
```
pakoito: `it.map { a: A -> f(a) }` should suffice here

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -220,6 +235,25 @@ sealed class IO<out A> : IOOf<A> {
...
+
+      return when {
+        index != maxStackDepthSize -> MapFilter(source, g.andThen(h), index + 1)
+        else -> MapFilter(this, f, 0)
```
pakoito: Only define `h` if it's going to be used, else it's an unnecessary allocation

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -155,6 +158,15 @@ sealed class IO<out A> : IOOf<A> {
     // Pure can be replaced by its value
     override fun <B> map(f: (A) -> B): IO<B> = Suspend { Pure(f(a)) }
 
+    // Pure can be replaced by its value
+    override fun <B> mapFilter(f: (A) -> Option<B>): IO<B> =
+      Suspend {
+        f(a).fold(
+          { RaiseError(IllegalArgumentException("IO execution should yield a valid result")) },
```
pakoito: The error message here and below could use a bit more context, else it'll be hard to point what the origin is.

modules/effects/arrow-effects-reactor-data/src/main/kotlin/arrow/effects/reactor/FluxK.kt
```diff
@@ -24,6 +24,12 @@ data class FluxK<A>(val flux: Flux<A>) : FluxKOf<A>, FluxKKindedJ<A> {
   fun <B> map(f: (A) -> B): FluxK<B> =
     flux.map(f).k()
 
+  fun <B> mapFilter(f: (A) -> Option<B>): FluxK<B> =
+    flux
+      .filter { f(it).isDefined() }
+      .map { f(it).orNull()!! }
```
pakoito: .flatMap { f(it).fold({ Flux.empty() }, { Flux.just(it) }) }

modules/effects/arrow-effects-reactor-data/src/main/kotlin/arrow/effects/reactor/MonoK.kt
```diff
@@ -24,6 +26,13 @@ data class MonoK<A>(val mono: Mono<A>) : MonoKOf<A>, MonoKKindedJ<A> {
   fun <B> map(f: (A) -> B): MonoK<B> =
     mono.map(f).k()
 
+  fun <B> mapFilter(f: (A) -> Option<B>): MonoK<B> =
+    mono
+      .filter { f(it).isDefined() }
+      .map { f(it).orNull()!! }
+      .toMono()
```
pakoito: I don't believe this is a valid instance, as MonoK must always return a value. What do the laws say?

modules/effects/arrow-effects-rx2-data/src/main/kotlin/arrow/effects/rx2/SingleK.kt
```diff
@@ -24,6 +25,13 @@ data class SingleK<A>(val single: Single<A>) : SingleKOf<A>, SingleKKindedJ<A> {
   fun <B> map(f: (A) -> B): SingleK<B> =
     single.map(f).k()
 
+  fun <B> mapFilter(f: (A) -> Option<B>): SingleK<B> =
+    single
+      .filter { f(it).isDefined() }
+      .map { f(it).orNull()!! }
+      .toSingle()
+      .k()
```
pakoito: Same as MonoK, I'm surprised this is a valid instance

pakoito:
Really good job, including the new node for IO! For it to work you will need to add the relevant code to `IORunLoop` that run that specific node.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
pakoito: even `it.map(f)` and we avoid another allocation #perf

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
alissonfpmorais: Removed after refactor

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
alissonfpmorais: Ohhh great! Didn't see that
I just had to change to `it.flatMap { a: A -> f(a) }`

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
alissonfpmorais: Error message change to: "IO execution yield invalid result after mapFilter"

modules/effects/arrow-effects-reactor-data/src/main/kotlin/arrow/effects/reactor/FluxK.kt
pakoito: Same for ObservableK too.

alissonfpmorais: Really really better!

modules/effects/arrow-effects-rx2-data/src/main/kotlin/arrow/effects/rx2/SingleK.kt
alissonfpmorais: Changed to `flatMap` as suggested

modules/effects/arrow-effects-reactor-data/src/main/kotlin/arrow/effects/reactor/MonoK.kt
alissonfpmorais: I've changed to `flatMap` approach as suggested and all tests are successful!

alissonfpmorais:
> This is just awesome raised_hands , thanks so much @alissonfpmorais

I'm very glad to help :)


> Really good job, including the new node for IO! For it to work you will need to add the relevant code to `IORunLoop` that run that specific node.
> 
> The laws for FunctorFilter in some of these should fail I believe.

Thank you! I've added the node to `IORunLoop`, is this the right implementation?

raulraja:
@pakoito please merge at will if the loop impl looks good to you

pakoito:
The instances are missing passing `FunctorFilterLaws`. I don't believe several of them, like IO or Single, will pass them.

alissonfpmorais:
> The instances are missing passing `FunctorFilterLaws`. I don't believe several of them, like IO or Single, will pass them.
> 
> Go to `IOTest.ks`, add a test for `FunctorFilterLaws.laws(IO.functorFilter())`, and the same for Single and Mono. See whether they pass.

They're not passing :thinking: 
I'm going to investigate more and tell you my finds at night

pakoito:
@alissonfpmorais they cannot pass. FunctorFilter expects "zero" or "empty" baked into the datatype, and none of them have it. They require one return. `Maybe` for example, allows one return or none, and that's why it passes :D

alissonfpmorais:
@pakoito that's not exactly what tests reproduced.
Just `FluxK` and `MaybeK` passed `FunctorFilter` laws. Both `FlowableK` and `ObservableK` have a `zero` representation, but tests still fails. I don't know yet, investigating what's going on.
For `IO`, `SingleK` and `MonoK` they really have no `zero`, so I'm thinking what would be the best option to go forward. Maybe create a zero (like `None` for `Option` datatype)?

pakoito:
For IO, SingleK and MonoK just don't provide FunctorFilter and the corresponding AQL instances. You still have the possibility of wrapping over List or other AQL-ready datatype for it.

pakoito:
`mapFilter` with an error on one side can be moved here: https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/ApplicativeError.kt#L53

So it'll be available for IO and everyone else :D

alissonfpmorais:
Wow, that's amazing!
I'm going to work with this tomorrow morning. Thanks for sharing!! :D

pakoito:
Don't call it `mapFilter` because that's just the case for `(A) -> Option<B>`, use something _catchy_ like `mapConditional`.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -163,6 +166,15 @@ sealed class IO<out A> : IOOf<A> {
     // Pure can be replaced by its value
     override fun <B> map(f: (A) -> B): IO<B> = Suspend { Pure(f(a)) }
 
+    // Pure can be replaced by its value
+    override fun <B> mapFilter(f: (A) -> Option<B>): IO<B> =
+      Suspend {
+        f(a).fold(
+          { RaiseError(IllegalArgumentException("IO execution yield invalid result after mapFilter")) },
```
raulraja: This is not ok. We can;t have unsafe operations like this just because we can raise Throwables. This makes IO unsafe and non pure because IO can't provide an empty value.

raulraja:
I suggested alternatives here that I believe are more inline with the spirit of AQL https://kotlinlang.slack.com/archives/C8UK6RTHU/p1552854784155300 we just need a `WhereOr` operator in AQL and let IO and other `ApplicativeError` capable data types provide an extension for their `E`

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
alissonfpmorais: I'm going to remove this. Thanks for feedback :+1:

alissonfpmorais:
@pakoito I tried with `(A) -> Either<E, B>` and ended with: `fun <A, B> Kind<F, A>.mapIfRight(f: (A) -> Either<E, B>): Kind<F, Either<E, B>>`. Doesn't look the best option, but I don't know what else could be.
Applying same idea to `MonadError` instead of `ApplicativeError` this could be achived: `fun <A, B> Kind<F, A>.mapIfRight(f: (A) -> Either<E, B>): Kind<F, B>`
What you think?  Not sure if it makes sense

@raulraja I've read what you said about all those datatypes that can't implement `FunctorFilter` and `WhereOr<E, A>` typeclass. It all makes sense!
Finishing conditional map (as said above) and fixing IO, I'm going to start that! Thanks you for helping me, probably going to need some more tips to implement all those concepts

raulraja:
I'll be happy to help any time. Just ask here or in slack and we can
discuss the approach or any questions you may have.

JorgeCastilloPrz:
Hi @alissonfpmorais! I contacted you in Kotlinlang Slack with the same question with no response, I'll try here!

We’re cleaning some open PRs in Arrow and this one got stuck since many months ago. We’re closing all the ones older than 3 months or so, but this one seems interesting to me. Is this something you’re planning to redo given the current Arrow encodings? (many things changed lately). If you’re not planning to get back to it soon my suggestion is to close the PR then we can reopen it or read the diff at any point in the future if you feel like getting back to it. Thoughts?

alissonfpmorais:
Hi @JorgeCastilloPrz!
Sorry for not answering you on slack channel, I didn't see your messages.
I got really stuck on my job in these last months so I had no time to work on this PR, but my vacation starts on Christmas and I'm planning to use on this issue.
Since lot of code has changed after my last commit, what you think is a better option: close this PR and start another one or continue on this one?

JorgeCastilloPrz:
I suspect it'll be easier for you to branch off master again and start from scratch, but I don't have a very deep scope on your PR. Feel free to close it if you consider the same!

1Jajen1:
Since this pr is add only and the datatypes almost did not change it's probably best to try and merge. Throw away every conflict for master if it's too much but I don't think that's gonna be the case here.
Aql did not change either so you could have an easy merge here, definitly worth the try. ^^

alissonfpmorais:
@JorgeCastilloPrz @1Jajen1 I've checkout the branch that I was working on, but a while ago I messed up my fork and I ended up deleting and forking it again. Now this PR points to unknown repository.

It looks like that would be easier to start the work again so I'm gonna close this PR and start another one

JorgeCastilloPrz:
Sounds totally fine to me :) thanks for the heads-up!

PR #1336, opened 2019-03-08
Update kotlin to 1.3.21

LordRaydenMK:
I had to move the kapt plugin to the top to make the gradle sync work.

I was trying to update kotlint test which led me to updating kotlin and in the end it didn't work out. So here is this one instead.


pakoito:
There has been a couple of big changes to module structure today due to the 0.9 release. I'd suggest redoing this PR rather than fixing it.

LordRaydenMK:
Let's try again after 0.9 is published

PR #1341, opened 2019-03-09
Update READMEs after 0.9.0

pakoito:
Let's keep the current structure for one more version.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,9 +50,9 @@ allprojects {
 }
 ```
 
-# Dependency breakdown starting in Arrow 0.9.0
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -50,9 +50,9 @@ allprojects {
 }
 ```
 
-# Dependency breakdown starting in Arrow 0.9.0
+# Dependency breakdown starting in Arrow 0.9.1
```
jrgonzalezg: This was correct as it was since 0.9.0 already had the changes mentioned.

pakoito: We're missing the single package version tho

PR #1342, opened 2019-03-09
Updated artifact references to 0.9.0 and removed the coroutine artifacts

jrgonzalezg:


raulraja:
@jrgonzalezj thanks! This needs to get updated with latest from master since some of the Readme items were modified late last night by Paco.

jrgonzalezg:
That must have been merged minutes ago cause I took the last master from one hour ago 😓. I am withdrawing the PR then cause most changes are the same, and now the reference to 0.9.0 dependencies will not be something we can correct since the tag is stuck with the wrong reference to the coroutine artifacts.

raulraja:
The docs are updated on each master merge, if there are any improvements they can still get in, they are separate from the artifacts release

jrgonzalezg:
I think the doc would have been more correct merging the changes here instead of the ones merged on the other PR which lead to errors on the 0.9.0 documentation and also wrong information on when the artifact changes took effect. I don't think we can have both the tag reference and the correct docs at the same time and I see no way to get the updates right for 0.9.0 if not having the README as it was on this PR instead of the updated version in master which doesn't seem possible to fix on the way it is.

I could make another PR taking just the changes to remove the coroutines artifacts, but that won't fix the incorrect 0.9.0 doc reference.

raulraja:
That's not entirely accurate. Starting in 0.9.0 we have docs sites we can selectively build to any hash. For example https://0-8-2.arrow-kt.io/. That is the canonical way to get docs moving forward and the main site will contain links in which @AntonioMateoGomez @calvellido and @dominv are working on to have those live soon. 

This means that if you want improvements that will be published for the official 0.9.0 docs we could build a site for it even from a branch. 

True that the actual sources bundled in the tag currently wouldn't include those automatically but the Arrow docs home for 0.9.0 can show whatever we want it to at any point in time.

PR #1343, opened 2019-03-09
Updated Gradle and some dependencies including Kotlin

jrgonzalezg:
* Updated Gradle to 5.2.1 through a configured gradle wrapper task (running `gradle wrapper` updates it to the version configured in the task on the main `build.gradle` file).
* Changed several additional `build.gradle` files cause the `kapt` plugin was applied too late causing the builds to fail with latest stable Kotlin.
* For some reason the last Gradle Version plugin does not work with Arrow's repo. I have reported it here: https://github.com/ben-manes/gradle-versions-plugin/issues/295
* There are some additional dependencies that could still be updated but that probably requires some additional migration / restructuring so I have skipped them for this PR.


build.gradle
```diff
@@ -17,21 +17,21 @@
...
         kotlinxCollectionsImmutableVersion = '0.1'
-        kotlinPoetVersion = '1.0.0-RC1'
-        projectReactorVersion = '3.1.8.RELEASE'
-        rxJavaVersion = '2.2.1'
-        dokka_version = '0.9.17'
-        detekt_version = '1.0.0-RC12'
+        kotlinxCoroutinesVersion = '1.1.1'
+        projectReactorVersion = '3.2.6.RELEASE'
+        retrofitVersion = '2.5.0'
+        rxJavaVersion = '2.2.7'
```
pakoito: BERRY BERRY KLEAN

PR #1345, opened 2019-03-09
Migrated to latest KotlinPoet stable 1.1.0

jrgonzalezg:
* This tries to migrate to latest KotlinPoet 1.1.0.
* Code changes seem equivalent to previous code (please comment if you see any issue), but it seems KotlinPoet is adding some extra new lines that break compilation.

* Example generated code for `ConstFunctor` before with KotlinPoet 1.0.0-RC1:
```
fun <A> Kind<Kind<ForConst, A>, A>.unit(): Const<A, Unit> = arrow.typeclasses.Const.functor<A>().run {
   this@unit.unit() as arrow.typeclasses.Const<A, kotlin.Unit>
}
```

* Example generated code for `ConstFunctor` now with KotlinPoet 1.1.0:
```
fun <A> Kind<Kind<ForConst, A>, A>.unit(): Const<A, Unit> = arrow.typeclasses.Const.functor<A>().run
          {
    this@unit.unit() as arrow.typeclasses.Const<A, kotlin.Unit>
  }
```

The new line before the brace causes the following compilation errors:
`.../arrow/modules/core/arrow-core-extensions/build/generated/source/kaptKotlin/main/extension/arrow/core/extensions/const/functor/ConstFunctor.kt: (130, 9): Expecting a top level declaration`

### DIFF
--- a/modules/meta/arrow-meta/src/main/java/arrow/extensions/ExtensionProcessor.kt
+++ b/modules/meta/arrow-meta/src/main/java/arrow/extensions/ExtensionProcessor.kt
@@ -233,7 +233,7 @@ class ExtensionProcessor : MetaProcessor<extension>(extension::class), PolyTempl
     companionOrFactory: TypeName,
     impl: String
   ): String =
-    """|return ${+companionOrFactory}.${typeClass.name.simpleName.decapitalize()}${+instance.typeVariables}(${requiredParameters.code { +it.name }}).run {
+    """|return ${+companionOrFactory}.${typeClass.name.simpleName.decapitalize()}${+instance.typeVariables}(${requiredParameters.code { +it.name }}).run·{
            |  $impl
            |}
            |
--- a/build.gradle
+++ b/build.gradle
@@ -27,7 +27,7 @@ buildscript {
         daggerVersion = '2.17'
         kotlinxCoroutinesVersion = '1.1.0'
         kotlinxCollectionsImmutableVersion = '0.1'
-        kotlinPoetVersion = '1.0.0-RC1'
+        kotlinPoetVersion = '1.1.0'
         projectReactorVersion = '3.1.8.RELEASE'
         rxJavaVersion = '2.2.1'
         dokka_version = '0.9.17'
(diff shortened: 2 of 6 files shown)


jrgonzalezg:
The problem is probably related to the wrapping changes in KotlinPoet commented here: https://github.com/square/kotlinpoet/issues/598

jrgonzalezg:
The problem seemed to be due to Arrow using `CodeBlock::toString()` transformations for code that is going again through code generation. I have removed some usages of that and the code now compiles, but we should not be using that `toString()` at all cause it is lossy and changes special characters like the `·` character for non breaking spaces to a normal space which is then causing KotlinPoet to potentially add a new line there when it shouldn't.

A bigger rewrite to avoid the current usage of KotlinPoet would be a better fix but this seems to work for now. Although I'm not really sure if the internal `Code` model for the instances I changed is currently used for printing code somewhere else like the website or the KDoc. If that is the case the changes I made here may break that a bit cause it may be outputting the `·` instead of a normal whitespace if it does not go through KotlinPoet again for the final code printed.

modules/meta/arrow-meta/src/main/java/arrow/extensions/ExtensionProcessor.kt
```diff
@@ -233,7 +233,7 @@ class ExtensionProcessor : MetaProcessor<extension>(extension::class), PolyTempl
     companionOrFactory: TypeName,
     impl: String
   ): String =
-    """|return ${+companionOrFactory}.${typeClass.name.simpleName.decapitalize()}${+instance.typeVariables}(${requiredParameters.code { +it.name }}).run {
+    """|return ${+companionOrFactory}.${typeClass.name.simpleName.decapitalize()}${+instance.typeVariables}(${requiredParameters.code { +it.name }}).run·{
```
pakoito: Ese puntito que es lo que ess??

pakoito: ·???

jrgonzalezg: It's the way to specify non-breaking spaces for latest KotlinPoet versions so it doesn't insert a new line on that space.

jrgonzalezg: Longer explanation: https://github.com/square/kotlinpoet#spaces-wrap-by-default

PR #1350, opened 2019-03-11
Update predef.kt

Sergeydayneko:
Function for currying context


JorgeCastilloPrz:
This functions is [already available](https://github.com/arrow-kt/arrow/blob/7c35fad5d35ff45ad25cc1a3d0c76baba7daf2b2/modules/core/arrow-core-data/src/main/kotlin/arrow/core/predef.kt#L5) in Arrow. Docs are still needed, but this PR can be closed. Also note that in an effort to clean the PR list we've decided to close all the ones that are older than 2 months and consider them obsolete except for very specific circumstances that we will evaluate per PR. Thanks for your contribution! 🙏

PR #1351, opened 2019-03-11
Updated Quickstart Maven docs

Murraymint1804:
Updated the quickstart maven docs to the current Arrow version.


JorgeCastilloPrz:
In an effort to clean the PR list we've decided to close all the ones that are older than 2 months and consider them obsolete except for very specific circumstances that we will evaluate per PR.

In this case the PR is too subtle and that file has been changed many times since then, so we'll be closing it. Thanks for the contribution.

PR #1352, opened 2019-03-11
selective functor extras

richard-gibson:
- docs fix for selective hierarchy diagram
- remove redundant identity calls 
- added selective instance for validated (all other instances so far also have monad instances)
- distributivity and accociativity laws from Selective paper added


pakoito:
Nice cleanup. Thanks :D

PR #1353, opened 2019-03-11
missing imports added to keep ank happy

richard-gibson:


pakoito:
@richard-gibson could you please make it more side-effecty? Maybe mutating the variable :D

richard-gibson:
> @richard-gibson could you please make it more side-effecty? Maybe mutating the variable :D

haha, first time I've been asked to do that on a PR :D
Will do

PR #1355, opened 2019-03-12
Updated Maven section for 0.9.0

routis:
Updated the section of the Maven Basic setup to use the -new- jars released under 0.9.0 tag


routis:
@pakoito Can you please merge this?

pakoito:
Yep! Done :D

routis:
@pakoito Unfortunately, current doc page (https://arrow-kt.io/docs/) displays again the old version.

calvellido:
Yeah, sorry about that. Basically, if this was merged after 0.9.0 was released, we will need to add it manually to those docs, as the current page is the frozen docs state of the stable release.

Live docs resides now here: https://next.arrow-kt.io/

PR #1360, opened 2019-03-13
Docs Arrow version selector

### LINKED ISSUE #1309: Docs version selector
We need to implement a version selector between different published Arrow versions to consult on the website:

https://0-8-2.arrow-kt.io
https://arrow-kt.io
(...)
https://next.arrow-kt.io

AntonioMateoGomez:
This PR closes #1309 

That's how it would look. We have incorporated an information bar that highlights the version of the library in which we are. At the moment in which the user closes it, you will no longer appear. We do not want it to be a nuisance.

Captura de pantalla 2019-03-13 a las 10 42 05

We have chosen to give importance to the middle button (stable version) in the masthead using the highlight of colors instead of changing its size so that the design continues having a better disposition


pakoito:
Do we only want to keep 3 versions?

AntonioMateoGomez:
We will keep all the versions, but the top bar and the buttons will show just `PREVIOUS`, `STABLE` and `NEXT` . Clicking in `documentation` in the nav bar will show all the versions, we have got the same in the docs sidebar.  

Captura de pantalla 2019-03-13 a las 11 26 52

Captura de pantalla 2019-03-13 a las 11 27 57

The oldest versions will appear below the 3 that are shown in the screenshot as It shows below. `V.0.1.0` and `v.0.2.0` are not real data, it is just to have an idea of how it will look like.

Captura de pantalla 2019-03-13 a las 11 32 02

pakoito:
thanks!

JorgeCastilloPrz:
Great job 👏👏👏

pakoito:
If it's all green I'm happy for it to be sent for review and stamp it!

modules/docs/arrow-docs/docs/_sass/components/_navigation.scss
```diff
@@ -41,6 +41,59 @@
...
         }
+
+        .dropbtn {
+          @include links($white, $white, rgba($white, 0.5), rgba($white, 0.5));
+
+          cursor: pointer;
+          border: none;
+          background: transparent;
+          color: $white;
+          font-family: $brand-font-family;
```
calvellido: This ends up resolving to “Source Sans Pro”, when the rest of the items in the top menu are using “Titillium Web”. We need some consistency there.

modules/docs/arrow-docs/docs/_includes/_home-doc-versions.html
```diff
@@ -0,0 +1,32 @@
+<div class="masthead-doc-version">
+  <ul>
+    <li>
+      <a class="masthead-button secondary-button button"
+        href="{{ previous_version.url }}">
+        <span>{{ previous_version.title }}</span>
+        <span class="secondary-doc-version-text">
+          {{ site.data.doc-versions.previous }}
```
calvellido: Is this not pointing to the `docs` path deliberately?  As the previous CTA button behaviour was that way, I find it a bit weird that they don't lead me to the docs, but maybe it is what we want.

modules/docs/arrow-docs/docs/_includes/_home-doc-versions.html
```diff
@@ -0,0 +1,32 @@
...
+        </span>
+      </a>
+    </li>
+
+    <li>
+      <a class="masthead-button button" href="{{ stable_version.url }}">
```
calvellido: Is this not pointing to the `docs` path deliberately?  As the previous CTA button behaviour was that way, I find it a bit weird that they don't lead me to the docs, but maybe it is what we want.

modules/docs/arrow-docs/docs/_includes/_home-doc-versions.html
```diff
@@ -0,0 +1,32 @@
...
+        <span>{{ stable_version.title }}</span>
+        <span class="secondary-doc-version-text">
+          {{ site.data.doc-versions.stable }}
+        </span>
+      </a>
+    </li>
+
+    <li>
+      <a class="masthead-button secondary-button button"
+        href="{{ next_version.url }}">
```
calvellido: Is this not pointing to the `docs` path deliberately?  As the previous CTA button behaviour was that way, I find it a bit weird that they don't lead me to the docs, but maybe it is what we want.

modules/docs/arrow-docs/docs/js/docVersionsToogle.js
```diff
@@ -0,0 +1,49 @@
+/* When the user clicks on the navigation Documentation button,
+toggle between hiding and showing the dropdown content */
+function displayToogle() {
+  document.getElementById("docVersionDropdown").classList.toggle("show");
+}
+
+// Close the dropdown if the user clicks outside of it
+window.onclick = function(event) {
```
calvellido: This is a bit hacky. Overriding global objects could lead to other parts of the website going bananas, never knowing why. Also we are penalizing each and every click on the page for a simple dropdown closing.

It would be better if we implement what is trying to be achieved here through the lose of focus on the specific element that requires it, or if adding some code to a the global `onclick` event is needed, the best way would be through [addEventListener](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener).

modules/docs/arrow-docs/docs/js/docVersionsToogle.js
```diff
@@ -0,0 +1,49 @@
...
+    document.getElementById("topWarning").style.display = "none";
+  }
+}
+
+function closeTop() {
+  document.getElementById("topWarning").style.display = "none";
+  sessionStorage.setItem('showTopWarning', 1);
+}
+
+window.onload = highlightDocVersion;
```
calvellido: Same as before, even worse here as we are overriding the whole `onload`event.

modules/docs/arrow-docs/docs/js/docVersionsToogle.js
```diff
@@ -0,0 +1,49 @@
...
+          document.getElementById("top-next").style.color = "white";
+        }
+  }
+  else {
+    document.getElementById("topWarning").style.display = "none";
+  }
+}
+
+function closeTop() {
+  document.getElementById("topWarning").style.display = "none";
```
calvellido: As there's no way the element will be appearing again, it would be better just to dispose it from the DOM instead of hiding it.

modules/docs/arrow-docs/docs/js/docVersionsToogle.js
```diff
@@ -0,0 +1,49 @@
...
+          document.getElementById("top-previous").style.color = "white";
+        }
+        if (currentUrl.includes(document.getElementById("stable").textContent)) {
+          document.getElementById("top-stable").style.color = "white";
+        }
+        if (currentUrl.includes(document.getElementById("next").textContent)) {
+          document.getElementById("top-next").style.color = "white";
+        }
+  }
+  else {
```
calvellido: As there's no way the element will be appearing again, it would be better just to dispose it from the DOM instead of hiding it.

modules/docs/arrow-docs/docs/_sass/components/_navigation.scss
AntonioMateoGomez: Ok, I will remove that line.

modules/docs/arrow-docs/docs/js/docVersionsToogle.js
AntonioMateoGomez: Ok, this will be solved through addEventListener("DOMContentLoaded")

modules/docs/arrow-docs/docs/_sass/components/_navigation.scss
calvellido: :thinking: doesn't look like you did it in the end?

raulraja:
@calvellido @AntonioMateoGomez merge at will when this is ready to go

PR #1362, opened 2019-03-13
Fix shouldBeEq function

juliankotrba:
The passed matcher function of the should function currently only
creates a Result object, but does not do anything with it. To make a
unit test actually fail, a Matcher object should be passed to the
overloaded should function. This commit is changing this behaviour of
the shouldBeEq function.

After applying this fix, the kotlintest test

```
"shouldBeEq improved" {
      1.shouldBeEq(2, Int.eq())
}
```

will fail with the following error message:

> Expected: 2 but found: 1
> java.lang.AssertionError: Expected: 6 but found: 5
> 	at arrow.test.laws.LawKt.shouldBeEqAllNew(Law.kt:31)
> 	...


juliankotrba:
Hey @nomisRev, it would be nice if you could take a look at this PR, since you are the original author :)
Maybe I misunderstood something..

nomisRev:
@juliankotrba nice catch thanks! 👍

juliankotrba:
> @juliankotrba nice catch thanks! 👍

Thanks for taking your time :raised_hands:

PR #1364, opened 2019-03-15
Rename deferUnsafe to delayUnsafe (#1356)

### LINKED ISSUE #1356: Rename deferUnsafe to delayUnsafe
https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDefer.kt#L47

- [ ] Add a deprecation notice to deferUnsafe
- [ ] Add a new method delayUnsafe
- [ ] Make deferUnsafe call delayUnsafe
- [ ] Check that tests, examples, etc. call delayUnsafe now

In one or two versions

- [ ] Delete deferUnsafe

juliankotrba:
- [x]  Add a deprecation notice to deferUnsafe
- [x]  Add a new method delayUnsafe
- [x]  Make deferUnsafe call delayUnsafe
- [x]  Check that tests, examples, etc. call delayUnsafe now

closes #1356


juliankotrba:
What about [bindDeferUnsafe](https://github.com/arrow-kt/arrow/pull/1364/files#diff-880641dc110155d19cd0a437df74035fR40)?

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDefer.kt
```diff
@@ -44,6 +44,11 @@ interface MonadDefer<F> : MonadThrow<F>, Bracket<F, Throwable> {
 
   fun lazy(): Kind<F, Unit> = delay { }
 
+  fun <A> delayUnsafe(f: () -> Either<Throwable, A>): Kind<F, A> =
+    defer { f().fold({ raiseError<A>(it) }, { just(it) }) }
+
+  @Deprecated("Use delayUnsafe instead",
+          ReplaceWith("delayUnsafe(f)", "arrow.effects.typeclasses.MonadDefer"))
   fun <A> deferUnsafe(f: () -> Either<Throwable, A>): Kind<F, A> =
     defer { f().fold({ raiseError<A>(it) }, { just(it) }) }
```
pakoito: The implementation should just call `delayUnsafe`

pakoito:
>What about bindDeferUnsafe()?

Yikes. Rename and deprecate, same as the original API

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDefer.kt
juliankotrba: I implemented it like this so we do not have to copy the implementation when `deferUnsafe` gets deleted. But yes, it would be more clean to call `deferUnsafe` :+1:

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDefer.kt
```diff
@@ -44,8 +44,7 @@ interface MonadDefer<F> : MonadThrow<F>, Bracket<F, Throwable> {
 
   fun lazy(): Kind<F, Unit> = delay { }
 
-  fun <A> delayUnsafe(f: () -> Either<Throwable, A>): Kind<F, A> =
-    defer { f().fold({ raiseError<A>(it) }, { just(it) }) }
+  fun <A> delayUnsafe(f: () -> Either<Throwable, A>): Kind<F, A> = deferUnsafe(f)
```
pakoito: The other way around, the old implementation should call the new implementation :D My bad for not being clear about it!

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDeferCancellableContinuations.kt
```diff
@@ -37,8 +37,13 @@ open class MonadDeferCancellableContinuation<F, A>(val SC: MonadDefer<F>, overri
...
     defer { bindingCatch { bindIn(context, f) } }.bind()
 
+  suspend fun <B> bindDelayUnsafe(f: () -> Either<Throwable, B>): B =
+          delayUnsafe(f).bind()
+
+  @Deprecated("Use bindDelayUnsafe instead",
+          ReplaceWith("bindDelayUnsafe(f)", "arrow.effects.typeclasses.MonadDeferCancellableContinuations"))
   suspend fun <B> bindDeferUnsafe(f: () -> Either<Throwable, B>): B =
-    delayUnsafe(f).bind()
+    deferUnsafe(f).bind()
```
pakoito: delayUnsafe, as per above!

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDefer.kt
juliankotrba: omg completely my fault :see_no_evil:

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDefer.kt
```diff
@@ -44,9 +44,13 @@ interface MonadDefer<F> : MonadThrow<F>, Bracket<F, Throwable> {
 
   fun lazy(): Kind<F, Unit> = delay { }
 
-  fun <A> deferUnsafe(f: () -> Either<Throwable, A>): Kind<F, A> =
+  fun <A> delayUnsafe(f: () -> Either<Throwable, A>): Kind<F, A> =
```
raulraja: What is unsafe about this operation? If it were unsafe in any case it should be placed under the `arrow.unsafe` receiver to be consistent with other unsafe api's and runners post 0.9.0

pakoito: It's called unsafe because the function inside can fail with a Throwable, and that's forwarded via the left side of the Either. The way I see it, it's a "softer" version of flatMap for non-effectful code that can still fail.

pakoito: When MonadDefer gets parametrized to `E`, this'd be `f: () -> Either<E, A>`

raulraja: but if it takes already a Throwable inside the either and defer itself is safe this isn´t unsafe.
I think a better name for this is needed. This is also what `effect` takes in `fx` for what will become `BIO`. I think this is just a lazy version of `fromEither` and I'd call it `deferOrRaise`. It's not unsafe because it does not perform effects when invoked.

pakoito: I'm okay with `delayOrRaise`, yep. @juliankotrba, opinions?

raulraja: yeah, delayOrRaise is a better name

juliankotrba: delayOrRaise sounds good, should this be changed in a new PR?

PR #1365, opened 2019-03-15
Add SNAPSHOT so the build is successful

i-walker:
If you check the link https://github.com/arrow-kt/arrow/blob/0.9.0/README.md#next-development-version-090 it should be coherent with the changes


pakoito:
I'm not entirely sure I follow the logic here. 0.9.0 has been released already, although without too much promo.

i-walker:
Sorry I did not get the notification. 

I know, but if you visit the page which is referenced i
> ???

PR #1369, opened 2019-03-18
Rename delayUnsafe to delayOrRaise

juliankotrba:


raulraja:
thanks @juliankotrba !

PR #1370, opened 2019-03-18
Fix stack safety on async jumps through FxSyntax.effect

raulraja:


raulraja:
closing due to changes on strategy, this may break more than it fixes.

Issue #1299, opened 2019-02-11
[EPIC] - Arrow FX Benchmarks


fedefernandez:
Using jmh we want to port the [cats-effect](https://github.com/typelevel/cats-effect/tree/master/benchmarks) and [zio](https://github.com/scalaz/scalaz-zio/tree/master/benchmarks/src/main/scala/scalaz/zio) benchmarks to Arrow FX. This needs to be done in several tasks:

* [x] Create a new module in arrow for holding the benchmarks
* [x] Migrate the cats-effect benchmarks
* [x] Migrate de zio benchmarks

JorgeCastilloPrz:
I believe this is part of our work over Fx @raulraja @nomisRev ? Let me know if I'm wrong. I'm organizing goals a bit in 47 and wanna keep track of what's done / to be done around Fx for this quarter.

raulraja:
yes, though all the tasks @fedefernandez and @AdrianRaFo were working with related to Fx Benchmarks are already completed, it's just Simon and me for now improving the benchmarks and Fx stack safety as we go over the benchmarks.

JorgeCastilloPrz:
Sounds good, thanks.

PR #1371, opened 2019-03-18
Fx Trampoline Coroutine Context

raulraja:
Fixes stack safety on `!effect { foo() }` jumping async after 127 binds.
To be improved and configurable.


modules/effects/arrow-effects-io-extensions/src/main/kotlin/arrow/effects/IODispatchers.kt
```diff
@@ -55,7 +61,12 @@ object IODispatchers {
     override fun resumeWith(result: Result<T>) {
       if (iterations.get() > asyncBoundaryAfter) {
         iterations.set(0)
-        pool.execute { cont.resumeWith(result) }
+        println("Jumping async. cont.context: $context")
```
pakoito: Remove?

modules/effects/arrow-effects-io-extensions/src/main/kotlin/arrow/effects/internal/Trampoline.kt
```diff
@@ -0,0 +1,60 @@
...
+      underlying.execute(ResumeRun(head, rest))
+    }
+  }
+
+  private tailrec fun immediateLoop(task: Runnable): Unit {
+    try {
+      task.run()
+    } catch (ex: Throwable) {
+      forkTheRest()
+      throw ex.nonFatalOrThrow()
```
pakoito: Should we throw this? I thought it throwed internally

modules/effects/arrow-effects-io-extensions/build.gradle
```diff
@@ -1,6 +1,7 @@
 apply plugin: 'kotlin-kapt'
 
 dependencies {
+    compile project(':arrow-free-data')
```
pakoito: Free isn't used, it should not be necessary.

raulraja:
We are still working in the design in #1299 for a 0.9.1 solution in a different branch that is more performant than this one.

raulraja:
This is part of IO now in master

Issue #795, opened 2018-04-14
ManagedT datatype


pakoito:
https://www.iravid.com/posts/resource-management.html

https://github.com/iravid/managedt/blob/master/core/src/main/scala/com/iravid/managedt/ManagedT.scala

1Jajen1:
Is this still relevant? If so I could attempt this. I was looking for something to get familiar with arrow (and contributing to it) and this seems doable. If there is something else more recent (and easy to get started with) then please point me to that as well :) 
As for ManagedT implementing this with using bracket should not be too hard or am I missing something here?

pakoito:
`ManagedT` sounded like a fun datatype to add. It isn't a must, but rather a nice to have if you'll have fun writing it!

If you'd like some easier tickets that touch several parts of the codebase, there is:

Hash typeclass: https://github.com/arrow-kt/arrow/issues/687
MonadSuspendLaws: https://github.com/arrow-kt/arrow/issues/590, https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt#L23

1Jajen1:
I think I'll take a look at hash first and after that either #590 or this one. Thanks for the quick answer :)

pakoito:
Neat! Up next, some effects for you: #590

This one should be quicker than Hash, and at the same time allow you to experiment a bit with `arrow-effects`.

1Jajen1:
Ok so now with MonadDefer done, what would be something good to have next? This, an attempt at the Deferred issue or is there something else?

pakoito:
@1Jajen1 If you're not interested in Deferred, you can define the typeclass for destructuring: https://github.com/arrow-kt/arrow/issues/284

1Jajen1:
Closed by #1376

PR #1376, opened 2019-03-24
`Resource` data type

1Jajen1:
I got a bit side-tracked while writing and converting tests...

Solves #795.

Docs and kdoc is unfinished (kdoc is non-existant atm :sweat_smile:), will finish that tomorrow.


pakoito:
So far so good for me :D

raulraja:
what is the core algebra of `ManagedT` and where does it fit in the type classes expressed in the `Functor` hierarchy in which `Bracket` is already part of? https://arrow-kt.io/docs/apidocs/arrow-effects-data/arrow.effects.typeclasses/-bracket/index.html

essentially this can't be merged in until we understand where does this transformer fits in terms of Polymorphism and what it idioms for `fx` are. This looks a lot like `EnvFx` in the https://github.com/arrow-kt/arrow/blob/ff%2Fattempt-beanchmark/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/suspended/env/EnvFx.kt https://github.com/arrow-kt/arrow/blob/ff%2Fattempt-beanchmark/modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/EnvFxExample.kt but with a different purpose which is potentially just an addition over the bracket syntax which can also be achieved in direct style through `fx`. Let's explore this further please but let's focus on the polymorphism question.

raulraja:
For what I see all of this is done in terms of Bracket which leads me to believe that either this is missing in Bracket or it's poorly named or encoded as something else.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/Resource.kt
```diff
@@ -15,7 +15,6 @@ import arrow.typeclasses.Semigroup
  * When a resource is created one can make use of [invoke] to run a computation with the resource. The finalizers are then
  *  guaranteed to run afterwards in reverse order of acquisition.
  **/
-@documented
```
raulraja: why can't this be documented?

1Jajen1: idk I am not familiar enough with those annotations. The build failed because of it tho.

modules/docs/arrow-docs/docs/docs/effects/resource/README.md
```diff
@@ -0,0 +1,127 @@
+---
```
raulraja: this all should be in the kdocs, we no longer create markdown files for data types or type classes

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/Resource.kt
1Jajen1: At least that is what I read from the error

raulraja:
FTR we discussed in slack this is useful as a data type

1Jajen1:
I think this should be done :) I moved the docs to kdoc only and linked the api-docs in the menu.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/Resource.kt
```diff
@@ -0,0 +1,217 @@
...
+ * fun createDBHandle(): IO<Handle> = IO { println("Creating db handle"); Handle }
+ * fun createFancyService(consumer: Consumer, handle: Handle): IO<Service> = IO { println("Creating service"); Service(handle, consumer) }
+ *
+ * fun closeConsumer(consumer: Consumer): IO<Unit> = IO { println("Closed consumer") }
+ * fun closeDBHandle(handle: Handle): IO<Unit> = IO { println("Closed db handle") }
+ * fun shutDownFanceService(service: Service): IO<Unit> = IO { println("Closed service") }
+ *
+ * //sampleStart
+ * val program = fx {
+ *   val consumer = createConsumer().bind()
```
raulraja: Can we use ! instead of bind() that is the preferred way in fx

1Jajen1: Changed. That only goes for `fx` blocks right? I left the resource binding as is for now

1Jajen1: Oh and something I forgot: Don't ank:playground snippets require all code despite actually compiling with just the important bits? If so I need to add some stuff before merge :/

PR #1377, opened 2019-03-25
unnecessary AQL dependency

i-walker:


i-walker:
In the webpage they duplicated the imports. Was that intended?

pakoito:
Oh, I mean that we want to list all the dependencies on the docs, optionals too. AQL is one of those.

And yes, we do manually sync the main README and the one in the docs, although the later has more things on it.

i-walker:
I know, but I think you misunderstand me. Check this screenshot. 
There is one AQl import too much

Screenshot 2019-03-26 at 00 33 24

pakoito:
Ooooooooooooooh. Yeah I see it now! Thanks :D

i-walker:
Your welcome :)

PR #1384, opened 2019-03-28
Rework recursion schemes to be more modular

1Jajen1:
Mostly wip. Most of the implementation details come from http://hackage.haskell.org/package/recursion-schemes.

TODO:
- [x] Documentation for hylo and recursive folds
- [x] Documentation for corecursive unfolds
- [x] Missing some recursion schemes (especially those that use Free, Free in Arrow is kinda weird compared to haskell)
- [x] Missing type-aliases and helpers
- [x] Missing tests (even the current ones are a bit weird)
- [x] Stacksafety with eval


modules/recursion/arrow-recursion-data/src/main/kotlin/arrow/recursion/typeclasses/Recursive.kt
```diff
@@ -1,31 +1,89 @@
...
-interface Recursive<T> {
+interface Recursive<T, F> {
+
+  fun FF(): Functor<F>
+
   /**
    * Implementation for project.
    */
-  fun <F> Functor<F>.projectT(tf: Kind<T, F>): Kind<F, Kind<T, F>>
+  fun T.projectT(): Kind<F, T>
```
pakoito: I'm not sure these are equivalent, right? `T` isn't restricted to be a Kind like on the original.

1Jajen1: This basically allows any non-recursive type to be recursive if they can provide a functor f for which they can implement projectT

pakoito: If they have a `projectT` isn't that the `just` part of `Applicative`?

1Jajen1: Not really, applicative just will not convert structure correctly. Take lists for example (using ListF as a base functor): 
```
ListK<A>.projectT(): Kind<ListFPartialOf<A>, T> = when {
  isEmpty() -> ListF.NilF()
  else -> ListF.ConsF(first(), drop(1).k()) 
 }
```
Applicative would not be able to do that.

1Jajen1: as another example where applicative can't really work: The instance for long together with a fib alg.
```
interface LongBirecursive : Birecursive<Long, ForOption> {
  override fun FF(): Functor<ForOption> = Option.functor()

  override fun Kind<ForOption, Eval<Long>>.embedT(): Eval<Long> = binding {
    fix().fold({ 0L }, { it.bind() + 1L })
  }

  override fun Long.projectT(): Kind<ForOption, Long> = when {
    this < 0 -> throw IllegalArgumentException("IntBirecursive only works on natural numbers")
    this == 0L -> none()
    else -> (this - 1).some()
  }
}

fun Long.Companion.birecursive(): Birecursive<Long, ForOption> = object : LongBirecursive {}

fun fib(n: Long): Long = Long.birecursive().run {
  n.histo {
    it.fix().fold({
      0
    }, { co ->
      co.tail.value().fix().fold({
        1
      }, {
        println("(${co.head}, ${it.head})")
        co.head + it.head
      })
    })
  }
}

fun main() {
  fib(92).also(::println)
}
```

modules/recursion/arrow-recursion-data/src/main/kotlin/arrow/recursion/typeclasses/Recursive.kt
```diff
@@ -51,33 +51,42 @@ interface Recursive<T, F> {
...
+*/
 
-  fun <A> T.histo(alg: (Kind<F, Cofree<F, A>>) -> A): A =
-    gcata<A, CofreePartialOf<F>>(Cofree.comonad(), {
-      distHisto(FF(), FF().run { it.map { it.fix() } })
-    }, {
-      alg(FF().run { it.map { it.fix() } })
-    })
+  fun <A> T.histo(alg: (Kind<F, Cofree<F, A>>) -> A): A {
+    fun go(t: T): Cofree<F, A> = FF().run {
```
pakoito: The Android compiler may choke on this. Favor opening and closing the runs with the smallest sope possible.

modules/recursion/arrow-recursion-data/src/main/kotlin/arrow/recursion/typeclasses/Recursive.kt
```diff
@@ -51,33 +51,42 @@ interface Recursive<T, F> {
...
+      t.projectT().map { go(it) }.let {
+        Cofree(FF(), alg(it), Eval.now(it))
+      }
+    }
+    return go(this).extract()
+  }
+  /*
```
pakoito: Remember to fix/delet!

modules/recursion/arrow-recursion-data/src/main/kotlin/arrow/recursion/typeclasses/Recursive.kt
1Jajen1: Wait what? I never used android, what is special about it? o.O

pakoito: It doesn't like when there are two nested `this`-dependent calls.

pakoito: They run their own compiler/optimizer on top of the regular kotlin one.

1Jajen1: Hmm not quite sure I understand. Where exactly does this nest this calls? `go` only ever uses FF's this context right?

pakoito: I mean just 

```
FF().run { t.projectT().map { go(it) } }.let { ... } 
```

1Jajen1:
Making some recursion schemes stack-safe is a bit tricky. Take for example these two versions of histo:
```
fun <A> T.histo(alg: (Kind<F, Cofree<F, A>>) -> A): A {
    fun go(t: T): Cofree<F, A> = FF().run {
      t.projectT().map { go(it) }.let {
        Cofree(FF(), alg(it), Eval.now(it))
      }
    }
    return go(this).extract()
  }

  fun <A> T.histoStackSafe(TF: Traverse<F>, alg: (Kind<F, Cofree<F, A>>) -> A): A {
    fun go(t: T): Eval<Cofree<F, A>> = TF.run {
      t.projectT().traverse(Eval.monad()) { Eval.defer { go(it) } }.fix().map {
        Cofree(FF(), alg(it), Eval.now(it))
      }
    }
    return go(this).extract().extract()
  }
```

In the first one `go` cannot return `Eval<Cofree<F, A>>` because the tail argument for `Cofree` takes a `Kind<F, Cofree<F, A>>`. In the second one this is alleviated by using `traverse` to wrap the whole construction in `Eval` and then perform operations, but that is at the cost of an additional dependency (which for Traverse is not too bad but more than one might expect). Any ideas on how to work around this? For now I'll provide both versions stack-unsafe and stack-safe with more constraints. (Only applies to histo right now, but may happen in other parts as well)

1Jajen1:
This right now is as far as I am going to take it for now. Will provide docs and tests soon-ish (busy week ahead :/)

1Jajen1:
Stack safety issues are sorted out. Stack-safety is now only guaranteed when run with a monadic version (available for all folds and unfolds) and when that monad is stacksafe (like eval or io). I added some tests and will finish this up tomorrow. I'll also as a followup write some docs on how to use and understand recursion schemes (outside of the api-docs because that would not really fit there) but that will be on another pr

1Jajen1:
Hey, things have been catching up for me lately, so I have some time to finish things up finally :) I'd like to start with this pr, just a few questions:
- Is there anything non-obvious that I need to check, that changed in the past few months? (Like different formatting/build tools/documentation changes?)
- Has there been any change to Free? FreeF and Free in this pr have a weird relationship that I'd like to change, but if there has not been any change that will be another pr then

pakoito:
Welcome back :D I'm happy to hear that things are more stable now, and hope to see you back in the channels ^^

So, changes. We document inline on the files instead of in separate files. Those can still be there as blogpost-style docs.

Free is mostly untouched, maybe some small things around Fx. If you'd like to give it a pass to keep it fresh and documented it'd be awesome.

Compiler plugins for KEEP87 are around the corner (end of Q4), so a full library rewrite will happen afterwards. With plugins we can also do comprehensions without paying the coroutine cost!

1Jajen1:
> Welcome back :D I'm happy to hear that things are more stable now, and hope to see you back in the channels ^^

Thanks. I will be, won't do anything larger the next 2-3 months tho, but after that back in full force^^

> So, changes. We document inline on the files instead of in separate files. Those can still be there as blogpost-style docs.

So for now keep things as they are (missing a few bits here and there, but I'll see what I can do). Recursion schemes do need a longer tutorial at some point tho, although I'd usually just point at this awesome blog [series](https://blog.sumtypeofway.com/an-introduction-to-recursion-schemes/)

> Free is mostly untouched, maybe some small things around Fx. If you'd like to give it a pass to keep it fresh and documented it'd be awesome.

Yeah I probably will, Free is not really usefull outside of the more "excentric" recursion schemes but a different faster encoding (which also does not use weird type hacks) will be a good thing anyway. (Also something that just interests me :) )

> Compiler plugins for KEEP87 are around the corner (end of Q4), so a full library rewrite will happen afterwards. With plugins we can also do comprehensions without paying the coroutine cost!

Heard about that, I am very excited to see that happening^^

1Jajen1:
So I looked over it and I think all that is missing here is:
- Better tests and generators for recursive structures
- Documentation (although I'd like to skip examples for some recursion schemes, it's kind of hard to come up with good examples and bad examples don't really add value imo)

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/typeclasses/Composed.kt
```diff
@@ -97,20 +95,17 @@ interface ComposedTraverse<F, G> :
   companion object {
     operator fun <F, G> invoke(
       FF: Traverse<F>,
-      GF: Traverse<G>,
-      GA: Applicative<G>
```
1Jajen1: This was not used and overcomplicated some function signatures using it.
Removing it had no effect as far as I can tell

1Jajen1:
Api docs are now up to date and every instance except Mu/Nu is being tested now. I still don't like the tests, but for now they work.
I'd like to merge this in the current state, as more docs and tests are something I do want to do at some point but a) require quite a bit more work (time that I don't have right now) and b) the api and the most important recursion schemes are done and working.

1Jajen1:
@pakoito What do you think? Is this ready to go for now?

1Jajen1:
Thanks :) I'll quickly add a ticket to rewrite the recursion schemes doc page (so I don't forget :D) and then this should really be done

pakoito:
This'll go to 0.10.2, the docs pobably not

1Jajen1:
> This'll go to 0.10.2, the docs pobably not

Thanks! I am super excited to start using this in some projects (especially propcheck could use recursion schemes to simplify it's code)

Yeah, writing a good replacement for the current docs will take a while. I'll start something this weekend and see how far I get.

pakoito:
I need to learn how to use then effectively. I still write my own folds at work.

1Jajen1:
Have you read https://blog.sumtypeofway.com/an-introduction-to-recursion-schemes/ ? That blog series was what really got me into recursion-schemes.
Using them is made quite a bit harder outside of haskell because haskell automagically gives you the base-functor data-type and deals with lots of other small pain-points. Outside of haskell using para as a replacement for fold (for example) comes with a bit of boilerplate (not as much anymore here, but still) which makes it less convenient for small folds imo.
Another thing: if you use your "base-functor'd" datastructure (using Fix to get the normal structure back for example) outside of recursion-schemes there is more wrapping and unwrapping involved which can be a pain.
Lots of small issues caused by the language itself, but this pr helps with quite a few of them :)

pakoito:
Noted, I'll find time to read it 🎉

Issue #1357, opened 2019-03-13
Add delayOrRaise that jumps threads to Async


pakoito:
`delay` with a CoroutineContext is a nice helper for a common pattern. `delayUnsafe` should exist too.

Delay with thread jump:

https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Async.kt#L140

`delayOrRaise` (deferUnsafe until #1356 ) without thread jump:
https://github.com/arrow-kt/arrow/blob/master/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDefer.kt#L47

pakoito:
Closed by @i-walker !!

PR #1387, opened 2019-03-29
delayOrRaise

i-walker:
Should I add an Example in the docs 
[noob Issue](https://github.com/arrow-kt/arrow/issues/1357)


modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Async.kt
```diff
@@ -22,227 +22,236 @@ typealias Proc<A> = ((Either<Throwable, A>) -> Unit) -> Unit
...
+     *   //sampleEnd
+     *   println(result)
+     * }
+     * ```
+     */
+    fun <A> defer(ctx: CoroutineContext, f: () -> Kind<F, A>): Kind<F, A> =
+            just(Unit).continueOn(ctx).flatMap { defer(f) }
+
+    /**
+     * Delay a computation on provided [CoroutineContext].
```
pakoito: `Delay a branching computation on the provided [CoroutineContext].`

And you could use this opportunity to add a doc on the original `delayOrRaise` too as  `Delay a branching computation` :D

.bash_profile
```diff
@@ -0,0 +1 @@
+export JAVA_HOME=$(/usr/libexec/java_home -v 1.8.0_201)
```
pakoito: This shouldn't be on the repo hahaha

.gitignore
```diff
@@ -130,3 +130,5 @@ reports/
 .DS_Store
 
 target/
+
+.bash_profile
```
pakoito: Shouldn't bash_profile be on your root anyway?

pakoito:
Thank you for the PR!!

.gitignore
i-walker: I tried putting it into the root, but the settings won't cascade to the actual arrow project. Can you help me, maybe I did it wrong. I have a Macbook Pro 10.14.3

pakoito: You may have needed to restart your terminal, or source the bash profile? I use zsh.

Issue #1347, opened 2019-03-10
Add countdown-based `retry` for ApplicativeError


pakoito:
Something like

```
fun <E, A> Kind<F, A>.retry(AE: ApplicativeError<F, E>, count: Int): Kind<F, A> = AE.run {
  handleErrorWith { withRetry(it, AE, count) }
}

fun <E, A> Kind<F, A>.withRetry(err: E, AE: ApplicativeError<F, E>, retries: Int): Kind<F, A> =
  if (retries == 0) AE.raiseError(err)
  else AE. run { this.handleErrorWith { withRetry(it,AE, retries - 1) } }
```

It needs to be tested in ApplicativeErrorLaws for stackoverflow. If it SOs, then we can move it to MonadError and do another version based off `tailRecM`

RivuChk:
I'd like to pick this up, will add the functions, perform the tests and if required will do the fixes as suggested.

pakoito:
It may be part of MonadSuspend actually, because non-suspended functions are deterministic.

1Jajen1:
This can be closed as with schedules one can do any retry/repeat policy, including count based. Schedule only works on either `Concurrent` or `MonadError` + `Timer`, but as @pakoito noted above, only `MonadDefer` and above make sense. That measn one only needs a `Timer` for anything that is not `Concurrent` and implementing a timer (with or without delay) is rather trivial.

PR #1393, opened 2019-04-03
Add port of zio's schedule

1Jajen1:
I needed this for a project (well a subset of it). Should also solve #1347 

This is incomplete, just putting this here so the code can be reviewed before I put too much work in docs etc.

The schedule can be used as follows:
```kotlin
val schedule = with(IO.monadDefer()) {
  // Do exponential backof while the backoff is below 60 seconds, then continue spaced but at most 100 times. Oh and also randomize the delays by some factor between 0-1
  (exponential(1.seconds).doWhile { it < 60.seconds } or spaced(60.seconds).doWhile { it < 100 }).jittered(this)
}

ioAction.repeat(schedule)
// or
ioAction.retry(schedule)
```
(Or that is how it is going to look once done :wink: )

Does this need a license update, because this is basically ported from zio or is the zio license already in and permits this?

The schedule can work with MonadDefer only, but idk if a run method could/should be implemented on MonadDefer only because delays kinda require Async at least, any ideas/suggestions.

Missing:
- [x] Methods on MonadDefer and/or Async that can execute a given schedule
- [x] Docs
- [x] Tests


raulraja:
@1Jajen1 Scalaz is already in the LICENSE list but you can add an entry that specifically mentions ZIO and what has been derived from it.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/Schedule.kt
```diff
@@ -0,0 +1,354 @@
...
+            f(step.finish.value(), step.delay).map { d -> step.copy(delay = d) }
+          }
+        }
+      }
+    }
+
+  fun delayed(f: (Duration) -> Duration): Schedule<F, State, Input, Output> =
+    modifyDelay { _, duration -> MF().run { just(f(duration)) } }
+
+  fun jittered(MF: MonadDefer<F>, min: Double = 0.0, max: Double = 0.0): Schedule<F, State, Input, Output> =
```
kioba: `Jittered` is hard to test, with ZIO, they had the same issue that the bounds do not control the random generations so testing is hard to execute. In ZIO every jittered function receives an `IO[Nothing, Double]` as a last argument.
ZIO issue: https://github.com/zio/zio/issues/359

in Arrow `jittered` could be modified by replacing `min` and `max` arguments with an `IOOf<Double>`.

1Jajen1: Thanks for the hint. I've changed it to use `Kind<F, Double>` as that allows any monadstack. I'll also add a default that brings this back for any `MonadDefer`

1Jajen1:
Thanks to input on the encoding for the dependent type from @nomisRev this now has a decent api which the compiler can actually work with pretty well.

This change removed `reconsiderM`, `reconsider` and `onDecision` from the api, but I think thats fine.

I also added methods to run a schedule on any `F` that has either `Concurrent` or `MonadError` + `Timer`. 

All that's left now is docs and tests, but I'd like someone to look this over prior to check if this structure is good and/or if there are even better ways to do this.

nomisRev:
@1Jajen1 this is looking great to me! Can I help with anything else in regards to docs or tests?

1Jajen1:
> @1Jajen1 this is looking great to me! Can I help with anything else in regards to docs or tests?

I'll write something up tomorrow.

I've also seen [bows docs](https://bow-swift.io/next/docs/effects/retrying-and-repeating-effects/). I could port that over as a seperate page under the fx tab. Although I'd remove the api description and just link to api docs in that case.

I'll also do the tests then, I am not sure how a test-eq instance for a schedule would look like, so I might have to skip the laws for the few typeclasses that it implements and just do property tests.

And regarding typeclasses: Schedule is an Applicative, a Profunctor, a Monoid and I think it's also Alternative. However there are two Applicative and two Monoid instances (and thus also 2 Alternative instances), one with `and` and one with `or` as a combinator. Any idea if or how we should represent this?

nomisRev:
Our goal is to write all docs as KDoc with examples inline. We can generate webpages based on the KDoc and that way we can keep everything in a single place. This is helping is reduce documentation maintenance overhead. Porting the Bow page you linked to the KDoc and perhaps include some small examples that would be perfect!

Regarding the tests, it seems ZIO uses `Ref` to track what's happening within `IO` and verify the state matches what expected for a certain `Schedule`.

Regarding the instances, so far we've always opted for `and`. See `Boolean` and `Int`, I don't think this would be any different in this case.

1Jajen1:
> Regarding the tests, it seems ZIO uses Ref to track what's happening within IO and verify the state matches what expected for a certain Schedule.

Yes, but I really dislike those tests. The method is good, but zios tests don't really add much confidence for me (about as much as any one-example test will ever add, also afaik they did not cover the entire api of a schedule). The only annoying bit is that writing an eq instance is somewhat hard (but I think I have a decent idea on how to do it...)

> Regarding the instances, so far we've always opted for and. See Boolean and Int, I don't think this would be any different in this case.

Do we offer the other instances somewhere as another option?

1Jajen1:
The docs are now complete I think. Most of it is from https://bow-swift.io/next/docs/effects/retrying-and-repeating-effects/ with ank checked runnable examples.

I have only two things I'd dislike about the code atm:
- Type inference is not great, most of the time when combining schedules you need to manually fill out one or two types
- The tests cover a lot of important aspects, but some parts of the api are not covered, that is usually because other than example based tests I could not come up with good tests (limits of kotlintest and in general some aspects are not easy to test). I don't think this is too bad, because typeclass tests + the ones that are there cover most of the important api.

I don't really have many ideas how to improve type inference, so I think this is ready to go.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/Schedule.kt
```diff
@@ -0,0 +1,1022 @@
...
+ * Common use cases
+ * Once we have building blocks and ways to combine them, let’s see how we can use them to solve some use cases.
+ *
+ * ### Repeating an effect and dealing with its result
+ *
+ * When we repeat an effect, we do it as long as it keeps providing successful results and the scheduling policy tells us to keep recursing. But then, there is a question on what to do with the results provided by each iteration of the repetition.
+ *
+ * There are at least 3 possible things we would like to do:
+ *
+ * - Discard all results; i.e., return `Void`.
```
nomisRev:
```suggestion
 * - Discard all results; i.e., return `Unit`.
```

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/ScheduleTest.kt
```diff
@@ -0,0 +1,286 @@
...
+    tailrec fun go(s: Any?, rem: Int, acc: List<Schedule.Decision<Any?, A>>): List<Schedule.Decision<Any?, A>> =
+      if (rem <= 0) acc
+      else {
+        val res = update(i, s).value()
+        go(res.state, rem - 1, acc + listOf(res))
+      }
+    return go(initialState.value(), n, emptyList())
+  }
+
+  fun refTimer(ref: Ref<ForIO, Duration>): Timer<ForIO> = object : Timer<ForIO> {
```
nomisRev: 👌

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/Schedule.kt
1Jajen1: :see_no_evil:

PR #1395, opened 2019-04-05
Added function effectAnd with explicit continueOn param.

### LINKED ISSUE #1394: CoroutineContext.effect doesn't switch back to caller thread
Maybe this is a self-assumption but I think it would be really helpful to avoid multiple thread switches, specially on "UI-driven" frameworks such as Android.

Quick example, having:

```
private suspend fun printCurrentThread() {
    println(Thread.currentThread().name)
}

val result1 = fx {
    !effect { printCurrentThread() }
    !NonBlocking.effect { printCurrentThread() }
    !effect { printCurrentThread() }
}

fun main() {
    unsafe { runBlocking { result1 } }
}
```

At least from a newbie point of view I'd expect the output to be (when called from the main thread):

```
main
ForkJoinPool-1-worker-1
main
```

But actually is:

```
main
ForkJoinPool-1-worker-1
ForkJoinPool-1-worker-1
```

This is due to the `effect` ext fun behaving similarly to `continueOn`, although I'd expect that **ideally** only the lambda inside is executed in the `NonBlocking` context and then returning to the caller context (`main` in this case).

aballano:
As dicussed in a [Kotlin Slack's conversation](https://kotlinlang.slack.com/archives/C8UK6RTHU/p1554471170039500?thread_ts=1554467490.038700&cid=C8UK6RTHU) here's an alternative extfun for Fx's `effect`,   `effectAnd`, which requires an explicit context in which to continue after executing the effect:

```kotlin
!NonBlocking.effect(continueOn = UI) {
    foo()
}
```

This essentially allows to transform this code:

```kotlin
fx {
    // executed in main 
    !effect { printCurrentThread() }
    // executed in NonBlocking
    !NonBlocking.effect { printCurrentThread() }
    continueOn(Main)
    // executed in main
    !effect { printCurrentThread() }
}
```

into this one:

```kotlin
fx {
    // executed in main thread
    !effect { printCurrentThread() }
    // executes in NonBlocking and switches back to main
    !NonBlocking.effectAnd(continueOn = Main) { printCurrentThread() }
    // executed in main thread
    !effect { printCurrentThread() }
}
```

Not much of a gain at first but this pattern is very often repeated on UI apps so it saves time and makes the code less dev-error-prone. It might be a convenient approach if it's not possible or easy to switch to the previous context implicitly as @pakoito suggested. Btw shall we create a ticket for this if not existing already?

Fixes #1394


raulraja:
I think we should actually change the behavior of ctx.effect to switch back to the context it started on instead of this. This is unrelated to Android even if the Android module contains enhancement syntax for this kind of pattern.

raulraja:
@aballano would something like this work? (untested)
```kotlin
suspend fun <A> CoroutineContext.effect(f: suspend () -> A): Kind<F, A> =
    suspendCoroutine { cont ->
      cont.resume(
        asyncOp { defer(this@effect) { f.effect() } }.continueOn(cont.context)
      )
    }
```
The idea here is that we switch back inmediatelly to the context of the continuation we are on after invoking the effect in the user provided context. This requires no additional api's or deprecation cycles.

pakoito:
>I think we should actually change the behavior of ctx.effect to switch back to the context it started on instead of this.

There's no such thing with coroutines. If the current context is EmptyCoroutineContext, the default, using it won't make it come back to the same thread. It requires installing event loops on arbitrary threads to make it work end-to-end. This is the best intermediate solution.

aballano:
Discussed with @raulraja and @nomisRev we will look for another, simpler, approach.

PR #1397, opened 2019-04-07
added Error message for unpackged sealed classes

i-walker:


i-walker:
I am listening @nomisRev 


> We can fix this in the generator to not include the `package unnamed` in case it's not in a named package.
> 
> If you're interested I can guide you through it but I'd prefer to see it fixed since it's about the same amount of code.

nomisRev:
@pakoito I took this up with @i-walker on Slack and we're going to fix this bug for all current processors :)

modules/meta/arrow-meta/src/main/java/arrow/extensions/PolyTemplateGenerator.kt
```diff
@@ -103,7 +103,7 @@ interface PolyTemplateGenerator : MetaApi {
     val monoidImports = if (info.applicativeRequiresMonoid()) "\nimport arrow.core.extensions.monoid" else ""
     return replace(
       "_imports_applicative_",
-      "import ${applicativePackageName.value.quote()}.just$monoidImports"
+      if (info.instance.packageName.value != "`unnamed package`") "import ${applicativePackageName.value.quote()}.just$monoidImports" else ""
```
nomisRev:
```suggestion
 "import ${applicativePackageName.value.quote()}.just$monoidImports"
```

This is unrelated to including the package names in generated files. This relates to polymorphic documentation written on typeclasses.

i-walker:
> This branch now contains the whole docset, and that's not what we want.

My bad i will delete that

i-walker:
arrow-meta does not support generic optics like this?
```
import arrow.optics.optics

@optics
data class Account<A>(val balance: List<A>, val available: Int) {
  companion object
}
```

PR #1404, opened 2019-04-12
Apply typeclass

### LINKED ISSUE #284: Apply and Divide typeclasses

pedrovgs:
### :pushpin: References
* **Issue:** Closes #284 

Add ``Apply`` typeclass to the project and update ``Applicative`` typeclass implementation to be based on it.

**Disclaimer: I'm creating this PR as a draft because the implementation is not finished. There are some pending tasks related to the extensions and documentation I'd like to review with the team before. Formatting the code properly is also a task I have pending but I can't spend more time on this issue for now.**

### How is it being implemented?

Crating the brand new ``Apply``  typeclass was easy. We only needed to create the new ``Apply`` type and move all the code from ``Applicative`` but the ``just`` related code. The hard part was related to the laws and the usage of simulated higher kinded types.  

These are some points I'd like to review with the team:

* It was weird to see how the laws already described for ``Aplicative`` doesn't match with the one I've found in other implementations like [Cat's Apply](https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/ApplyLaws.scala) or [Scalaz's Apply typeclasses](https://github.com/scalaz/scalaz/blob/series/7.3.x/tests/src/test/scala/scalaz/ApplyTest.scala).
* I've updated the ``AplicativeLaws`` implementation to be divided in ``ApplyLaws``  and ``ApplyCartesianLaws``. This change let us write some of our laws without coupling the implementation to any concrete type. I'm afraid I couldn't implement this for the cartesian laws because of the internal map implemented. **Ideally, we'd like to don't even couple this implementation to the ``Applicative`` instance passed as parameter, but there is no way to lift any instance into ``Kind<F, A>`` w/o an ``Applicative`` instance.** I've reviewed the laws implementation with @delr3ves thinking about how could reduce the coupling but it wasn't possible for this scenario.
* As @raulraja [already told us](https://github.com/arrow-kt/arrow/issues/284#issuecomment-482526136): "import arrow.effects.extensions.io.applicative will no longer be available in that location but instead in:import arrow.effects.extensions.io.apply.[functionName]". ⚠️ This breaking change will not be important now because we are not in version > 1.0.0 but it's important to keep it in mind.

**About the rest of the task**. I have pending to write the documentation and implement the extensions for the types already providing an ``Applicative`` instance. However I have questions about these tasks:

* Should we create a new page for the ``Apply`` typeclass is it ok if we just update the ``Applicative`` page?
* As ``Apply`` is only needed to be able to implement other typeclasses for now, is adding all the extensions worth it?

Please, review the current implementation and my questions about the docs and the extensions. As soon as we consider this implementation correct I'll update the PR with the rest of the code 😃 

### How can it be tested?

:robot: All the code already using the ``ApplicativeLaws`` will now use the ``ApplyLaws``.


pedrovgs:
Finally, I had to rever the ``ApplyLaw`` changes to one closer to the original laws 😞

modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/try.kt
```diff
@@ -87,6 +87,15 @@ interface TryFunctor : Functor<ForTry> {
     fix().map(f)
 }
 
+@extension
+interface TryApply : Apply<ForTry> {
+  override fun <A, B> TryOf<A>.ap(ff: TryOf<(A) -> B>): Try<B> =
+          fix().ap(ff)
```
raulraja: I believe in Arrow we use 2 spaces but we are in the process of adding Ktlint // cc @aballano . You can leave or change this but eventually, it's gonna get reformated.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Apply.kt
```diff
@@ -0,0 +1,264 @@
+@file:Suppress("UNUSED_PARAMETER")
+
+package arrow.typeclasses
+
+import arrow.Kind
+import arrow.core.*
+import java.math.BigDecimal
+
+interface Apply<F> : Functor<F> {
```
pakoito: Cool, I'll never need to open this file anymore.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ApplyLaws.kt
```diff
@@ -0,0 +1,63 @@
...
+import arrow.test.generators.intSmall
+import arrow.typeclasses.Applicative
+import arrow.typeclasses.Apply
+import arrow.typeclasses.Eq
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+
+object ApplyLaws {
+
+    fun <F, A> laws(AP: Applicative<F>, gen: Gen<A>, EQ: Eq<Kind<F, A>>): List<Law> =
```
pakoito: We should keep the laws separated between those that use Apply and those that use Applicative.

Then I realised all of them needed both lol

I'd keep them as ApplicativeLaws tbh.

1Jajen1: But do they? If just is only used to lift into `Kind<F, A>` then it might be better to either use a constructor function or a `Gen<Kind<F, A>>` directly. And all laws that strictly need just belong to applicative anyway. Although I am not too familiar with applicative laws, so idk how many that leaves o.O

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Apply.kt
```diff
@@ -0,0 +1,264 @@
+@file:Suppress("UNUSED_PARAMETER")
+
+package arrow.typeclasses
+
+import arrow.Kind
+import arrow.core.*
+import java.math.BigDecimal
```
1Jajen1: This import seems weird, is it used anywhere?

1Jajen1:
Looks good so far :+1: I'd be in favor of adding the instances as it could be confusing not having them. Although I don't think it is super important, but at least a ticket for adding them should be added

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ApplyLaws.kt
nomisRev: This is an issue we have in more places.

We should indeed be using `Gen<Kind<F, A>>` directly here. There is some low hanging fruit in the test setup.

modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/try.kt
pedrovgs: Yeap, I couldn't find any tool to automatically format the code and I decided to left the formatting issue as is until I get the answers about the docs and the extensions 😃 I'll format the code properly before finishing this PR. Don't worry 😃

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ApplyLaws.kt
pedrovgs: Hi all, I commented one interesting point about the laws in the PR description:

> It was weird to see how the laws already described for Aplicative doesn't match with the one I've found in other implementations like Cat's Apply or Scalaz's Apply typeclasses.

> I've updated the AplicativeLaws implementation to be divided in ApplyLaws and ApplyCartesianLaws. This change let us write some of our laws without coupling the implementation to any concrete type. I'm afraid I couldn't implement this for the cartesian laws because of the internal map implemented. Ideally, we'd like to don't even couple this implementation to the Applicative instance passed as parameter, but there is no way to lift any instance into Kind<F, A> w/o an Applicative instance. I've reviewed the laws implementation with @delr3ves thinking about how could reduce the coupling but it wasn't possible for this scenario.

Finally, as ``FunctorLaws`` need some dependencies I can't fix for now without rewriting most of the laws and this issue was supposed to bee "noob friendly" I decided not to refactor them and just move them to the ``ApplyLaw`` folder.

Are we keep the laws in the ``ApplicativeLaw`` folder and remove this part of the PR? I tried to use Gen<Kind<F, A>>, and a method to lift any value into Kind, but it is not so easy without refactoring the rest of the code laws.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Apply.kt
pedrovgs: You are right, I have pending the imports optimization and the code formatting for this PR.

pedrovgs:
Thanks for your comments guys! Are we all agree if we finish the following task before considering this PR as final?

* Format code and optimize imports.
* Recover ``ApplicativeLaw`` changes.
* Add the missing extensions for Apply.
* Update ``Applicative`` documentation with a small section about Apply.

modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/try.kt
aballano: Still working on it! #1405

pakoito:
>Update Applicative documentation with a small section about Apply.

Why not a new docs page?

pedrovgs:
I asked for this in the PR description @pakoito 

> Should we create a new page for the Apply typeclass is it ok if we just update the Applicative page?

As the usage of ``Apply`` is not so common and most of the people will use ``Applicative`` I thought adding a section inside applicative's page should be enough. You can see how [Cat's documentation](https://typelevel.org/cats/typeclasses/applicative.html) uses this approach. 

Let me know if you'd prefer a new page or just a section inside applicative and I'll update the PR with it.

pedrovgs:
Hi there! As discussed in this draft PR I've updated this PR with the extensions needed, the documentation updated and a new section added to the applicative documentation about the Apply typeclass 😃  So I'm moving this PR to the final stage. I'd like to thank you all for your help and comments. Please, let me know if you think there is anything else you'd like to add or review 😃

modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/try.kt
```diff
@@ -105,6 +106,15 @@ interface TryFunctor : Functor<ForTry> {
...
+
+  override fun <A, B> TryOf<A>.map(f: (A) -> B): Try<B> =
+          fix().map(f)
```
raulraja: format

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ApplicativeLaws.kt
```diff
@@ -12,42 +12,42 @@ import io.kotlintest.properties.forAll
...
   fun <F> laws(A: Applicative<F>, EQ: Eq<Kind<F, Int>>): List<Law> =
-    FunctorLaws.laws(A, EQ) + listOf(
-      Law("Applicative Laws: ap identity") { A.apIdentity(EQ) },
-      Law("Applicative Laws: homomorphism") { A.homomorphism(EQ) },
-      Law("Applicative Laws: interchange") { A.interchange(EQ) },
-      Law("Applicative Laws: map derived") { A.mapDerived(EQ) },
-      Law("Applicative Laws: cartesian builder map") { A.cartesianBuilderMap(EQ) },
-      Law("Applicative Laws: cartesian builder tupled") { A.cartesianBuilderTupled(EQ) }
-    )
+          FunctorLaws.laws(A, EQ) + listOf(
```
raulraja: format

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Applicative.kt
```diff
@@ -3,22 +3,11 @@
 package arrow.typeclasses
 
 import arrow.Kind
```
raulraja: does the file need to be renamed?

raulraja: nevermind

modules/docs/arrow-docs/docs/docs/patterns/glossary/README.md
```diff
@@ -157,7 +157,7 @@ binding {
 ```
 
 ```kotlin:ank
-import arrow.core.extensions.option.applicative.map
+import arrow.core.extensions.option.apply.map
```
raulraja: have we found collision issues with the stdlib `apply` function when requesting the `apply()` instance explicitly from a companion? Seems like this could be potentially confusing to users.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/applicative/README.md
```diff
@@ -98,6 +98,24 @@ Computation happens when `.value()` is invoked.
 Option.applicative().run { Some(1).map2Eval(Eval.later { Some("x") }, { z: Tuple2<Int, String> ->  "${z.a}${z.b}" }).value() }
 ```
 
+### Apply
+
+A closely related type class is Apply which is identical to Applicative, modulo the pure method. Indeed Applicative is a subclass of Apply with the addition of this method.
```
pakoito: `pure` is called `just` in Arrow.

modules/docs/arrow-docs/docs/docs/patterns/glossary/README.md
pedrovgs: I hope not. I know the name can be confusing for some users but as most of the people will consume it as is in this example I guess this won't be an issue.

modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/try.kt
pedrovgs: Fixed!

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ApplicativeLaws.kt
pedrovgs: Fixed! I updated this branch to be aligned with master and used ktlintFormat gradle task 🤔 I thought our ktlint config would fix this, however, I had to fix these files manually.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/applicative/README.md
pedrovgs: Doc updated 😃

pedrovgs:
Thanks for your comments @raulraja @pakoito applied the changes you suggested and updated this PR 😃

raulraja:
Awesome stuff, thanks @pedrovgs ! :clap:

pedrovgs:
Thank you all for your comments and feedback. I'll keep contributing to this awesome project when possible 😃

modules/docs/arrow-docs/docs/docs/patterns/glossary/README.md
nomisRev: @raulraja there is no issue with `apply`. It's called `invoke` in Kotlin. Scala mix-up :D

pedrovgs: @nomisRev I think @raulraja is talking about this one:

```kotlin
package kotlin

/**
 * Calls the specified function [block] with `this` value as its receiver and returns `this` value.
 */
@kotlin.internal.InlineOnly
public inline fun <T> T.apply(block: T.() -> Unit): T {
    contract {
        callsInPlace(block, InvocationKind.EXACTLY_ONCE)
    }
    block()
    return this
}
```

PR #1413, opened 2019-04-18
Add finger tree data structure

### LINKED ISSUE #958: Look into FingerTrees
https://chrispenner.ca/posts/intro-to-finger-trees

juliankotrba:
The implementation is based on the paper _Finger trees: a simple general-purpose data structure_ written by Ralf Hinze and Ross Paterson ([Link](http://www.staff.city.ac.uk/~ross/papers/FingerTree.html)) and [this](http://andrew.gibiansky.com/blog/haskell/finger-trees/) blog post.

There is still a lot of work to do:
- [x] Instances
  - [x] Applicative
  - [x] EQ
  - [x] Monoid
  - [x] Functor
  - [x] Monad
- [x] Tests
- [ ] Documentation
- [ ] Optimizations
- [ ] Random access

closes #958


modules/core/arrow-extras-data/src/main/kotlin/arrow/data/fingertree/FingerTree.kt
```diff
@@ -0,0 +1,57 @@
+package arrow.data.fingertree
+
+import arrow.core.Option
+import arrow.data.fingertree.internal.FingerTreeInternal
+
+class FingerTree<T> internal constructor(internal: FingerTreeInternal<T> = FingerTreeInternal.Empty()) {
```
pakoito: Shouldn't this be `private var fingerTreeInternal: FingerTreeInternal<T>?`

modules/core/arrow-extras-data/src/main/kotlin/arrow/data/fingertree/internal/FingerTreeInternal.kt
```diff
@@ -0,0 +1,107 @@
...
+
+  data class Deep<T>(
+    val prefix: Affix<T>,
+    val deeper: FingerTreeInternal<Node<T>>,
+    val suffix: Affix<T>
+  ) : FingerTreeInternal<T>()
+
+
+  fun prepend(item: T): FingerTreeInternal<T> = when (val tree = this) {
+
```
pakoito: extra line

modules/core/arrow-extras-data/src/main/kotlin/arrow/data/fingertree/internal/FingerTreeInternal.kt
```diff
@@ -0,0 +1,107 @@
...
+  fun viewL(): Option<Tuple2<T, FingerTreeInternal<T>>> = when (this) {
+    is Empty -> Option.empty()
+    is Single -> Option.just(Tuple2(this.a, Empty()))
+    is Deep -> when (this.prefix) {
+      is Affix.One -> {
+
+        this.deeper.viewL().fold(
+          { Option.just(Tuple2(this.prefix.a, suffix.toFingerTree())) },
+          { (node, tree) -> Option.just(Tuple2(this.prefix.a, Deep(Affix.fromList(node.toList()), tree, this.suffix))) }
```
pakoito: This fromList/toList can probably be merged to Affix.fromNode() and matching the number of fingers

modules/core/arrow-extras-data/src/main/kotlin/arrow/data/fingertree/internal/Affix.kt
```diff
@@ -0,0 +1,35 @@
+package arrow.data.fingertree.internal
+
+internal sealed class Affix<A> {
+  data class One<A>(val a: A) : Affix<A>()
+  data class Two<A>(val a: A, val b: A) : Affix<A>()
+  data class Three<A>(val a: A, val b: A, val c: A) : Affix<A>()
+  data class Four<A>(val a: A, val b: A, val c: A, val d: A) : Affix<A>()
+
+  fun append(item: A): Affix<A> = fromList(this.toList() + listOf(item))
```
pakoito: I'm sure this can be made faster by removing the additional intermediate list of items `this.toList() + listOf(item)`

modules/core/arrow-extras-data/src/main/kotlin/arrow/data/fingertree/internal/FingerTreeInternal.kt
```diff
@@ -0,0 +1,107 @@
...
+        )
+      }
+
+      else -> Option.just(
+        Tuple2(this.prefix.toList()[0],
```
pakoito: Instead of creating an additional list in `this.prefix.toList()[0]`, add `head` to Affix that gets the first element for each using when.

modules/core/arrow-extras-data/src/main/kotlin/arrow/data/fingertree/internal/Affix.kt
```diff
@@ -0,0 +1,35 @@
+package arrow.data.fingertree.internal
+
+internal sealed class Affix<A> {
```
pakoito: You can use @AutoFold to generate a `fold` function to use instead of the `when` below.

modules/core/arrow-extras-data/src/main/kotlin/arrow/data/fingertree/internal/Node.kt
```diff
@@ -0,0 +1,22 @@
+package arrow.data.fingertree.internal
+
+internal sealed class Node<T> {
+
+  abstract fun toList(): List<T>
+
+  data class Branch2<T>(val first: T, val second: T) : Node<T>() {
```
pakoito: First and second never get accessed directly outside of the class. Maybe it's faster to just store the list directly on the `fromList` constructor and desambiguate on size.

modules/core/arrow-extras-data/src/main/kotlin/arrow/data/fingertree/FingerTree.kt
juliankotrba: I implemented it that way to make the class `FingerTree` testable. 
Please let me know if you know a better way 😀

modules/core/arrow-extras-data/src/main/kotlin/arrow/data/fingertree/FingerTree.kt
```diff
@@ -0,0 +1,57 @@
...
+
+  private var fingerTreeInternal: FingerTreeInternal<T> = internal
+
+  fun append(item: T) {
+    this.fingerTreeInternal = fingerTreeInternal.append(item)
+  }
+
+  fun prepend(item: T) {
```
raulraja: we can't justify having this api that returns Unit like this after 0.9.1. Is there a version of this data structure that has an immutable only API? otherwise this may not belong in Arrow as it deviates from the rest of the library which exclusively offers pure APIs except for unsafe runners in the effects library. Not saying this is isn't useful just not in `arrow.data`.

raulraja: FTR internal mutability for optimizations is fine but not public functions that return `Unit` and mutate the internal state of data they are part of.

juliankotrba: `FingerTreeInternal` is the immutable only API. I've created the `FingerTree` wrapper class for convenient use of the internal finger tree, but I totally get you point!
So I would say that we just get rid of the wrapper and make the internal finger tree non-internal?

raulraja: @juliankotrba yes, I believe that would be a much better fit in Arrow. It will help to see at a glance a list of functions in the public API to understand where this will best fit and if it can provide extensions for some of type classes as well.

build.gradle
```diff
@@ -35,6 +35,7 @@ buildscript {
         projectReactorVersion = '3.2.6.RELEASE'
         retrofitVersion = '2.5.0'
         rxJavaVersion = '2.2.7'
+        mockkVersion = '1.9.3'
```
pakoito: I don't believe we need to do mock testing, we use property-based testing everywhere with real data.

modules/core/arrow-extras-data/src/test/kotlin/arrow/data/FingerTreeTest.kt
```diff
@@ -0,0 +1,291 @@
...
+    }
+
+    "prepend() should add the item to the first place of the tree prefix when the finger tree has three prefix elements and an empty deeper tree" {
+      val fingerTree = Deep(Three(1, 2, 3), Empty(), One(4))
+
+      fingerTree.prepend(5) shouldBe Deep(Four(5, 1, 2, 3), Empty(), One(4))
+    }
+
+    "prepend() should create a finger tree with the new affix, the result of the prepend on the deeper tree and the same suffix when the prefix has four elements" {
+      val mockDeeperFingerTree = mockk<FingerTree<Node<Int>>>(relaxed = true)
```
pakoito: Please use property-based testing based off KotlinTest for this. Mocking is not the right approach for these as we're testing behavior and not whether something has been called.

juliankotrba:
> Use KotlinTest and Generator instances instead of MockK for testing.

K :thumbsup:

pakoito:
@juliankotrba Ping us on Slack or Gitter if you need guidance for the instances!

modules/core/arrow-extras-data/src/main/kotlin/arrow/data/fingertree/FingerTree.kt
```diff
@@ -111,7 +111,7 @@ sealed class FingerTree<T> {
 
   fun isEmpty() = this is Empty
 
-  fun asList(): List<T> = this.asListHelper(emptyList(), this)
+  fun asList(): List<T> = this.asSequence().toList()
```
pakoito: Why the extra jump?

juliankotrba: Because it is so much faster than my implementation. Unfortunately, I don't know why yet 😅

pakoito: Okay, I see what's there. Keep it!

juliankotrba: K. I think it's so much slower because of the list concat.

modules/core/arrow-extras-data/src/test/kotlin/arrow/data/FingerTreeTest.kt
```diff
@@ -16,6 +16,7 @@ import io.kotlintest.properties.Gen
 import io.kotlintest.properties.forAll
 import io.kotlintest.shouldBe
 import io.kotlintest.specs.StringSpec
+import java.util.*
```
juliankotrba: @pakoito is it ok to have references to java? MPP?

pakoito: For the tests it's okay, as they're JVM-only because of the runner.

juliankotrba: Okay, thanks! 👍

modules/core/arrow-extras-data/src/main/kotlin/arrow/data/fingertree/FingerTree.kt
```diff
@@ -249,6 +245,39 @@ sealed class FingerTree<T> : FingerTreeOf<T> {
 
     fun <A> fromArgs(vararg items: A): FingerTree<A> =
       fromList(items.toList())
+
+    fun <A, B> tailRecM(a: A, f: (A) -> Kind<ForFingerTree, Either<A, B>>): FingerTree<B> =
+      tailRecMHelper(empty(), f, f(a).fix())
+
+    private tailrec fun <A, B> tailRecMHelper(
```
juliankotrba: @pakoito could you also please explain the purpose of `tailRecM()` if you have some time? I tried to copy it from the `ListK` implementation, but I do not get the deeper meaning of it. Also, if the function `f()` returns a `Kind<F, Either<A, B >>`, with the first element being a `Either.Left`, the function will run forever?

pakoito: Sure! So, these data structures aren't stack-safe by default, meaning that if you write a recursive function that uses flatMap and handleErrorWith they'll end up throwing an StackOverflow because the compiler can't do tail-call optimization even when you say the function is `tailrec`.

These implementations use the heap and monadic behavior to prevent that. Hence, `tailrecM`.

juliankotrba: Ah, now I get it. Thanks for this detailed explanation!! 😄

modules/core/arrow-extras-data/src/test/kotlin/arrow/data/FingerTreeTest.kt
```diff
@@ -1,9 +1,6 @@
 package arrow.data
 
-import arrow.core.None
-import arrow.core.Option
-import arrow.core.Some
-import arrow.core.Tuple2
+import arrow.core.*
```
pakoito: ktlint will complain about this

juliankotrba: Fixed, thanks 👍

juliankotrba:
Hey, I would like to give you a status update on this PR. I am currently very busy, so I don't have time to wrap my head around implementing the _random access_. Should we just keep this PR open or should we merge it after I wrote some documentation and create an issue for adding random access to the finger tree data structure?

raulraja:
Thanks @juliankotrba Let's keep the PR open until you have time or someone wants to help with that part, if we merge as is it will be hard later to track the missing functionalities.

sraaphorst:
Out of curiosity, are these frequently used? I'm hardly an expert in functional programming but I've been using it for a few years and have never come across a finger tree until now.

pakoito:
It's a specialized collection, so I'd say it's not down to FP but the domain you work on. The paper lists the main use case as sequences you can traverse both directions.

sraaphorst:
Thanks, @pakoito! Much appreciated.

JorgeCastilloPrz:
In an effort to clean the PR list we've decided to close all the ones that are older than 2 months and consider them obsolete except for specific circumstances that we evaluate per PR. In this case the PR is quite interesting to be honest and we've considered keeping it open, but given that's been open for too long already, we've decided to close it for now. Note that this doesn't mean we're discarding the change, since it can be re opened once the author @juliankotrba has some free time to spend on it. 

Thank you really much for the proposal, very appreciated! and please feel free to get it back to life as soon as you feel ready for it 🙏

PR #1419, opened 2019-04-24
Remove Partial Functions.

diesalbla:
The API of the PartialFunction class is similar to that of the [`PartialFunction`](https://www.scala-lang.org/api/current/scala/PartialFunction.html) class from the Scala standard libary: an interface that extends from the normal `Function` by adding an "isDefinedAt" method, which says if a function is defined for a value of its domain.

The `PartialFunction` class is probably a mistake in the design of the Scala standard library. Its design departs from the ideas of both functional and object-oriented programming.

On the functional side, the preferred approach to FP is to write a programs as a set of functions that, as far as possible, are total: they should terminate without errors on all inputs of a domain, as  specified through its type. A partial function is the negation of this idea.

On the object-oriented side, for a `PartialFunction` to extend from the `Function` interface goes against the Liskov Substitution Principle (LSP). According to thes, in a well-formed inheritance relation any instance of a subclass can replace a reference to any super-class in every context in which the reference to the super-class appears. Therefore, thesub-class can never restrict the pre-condition, or loose the post-condition, of any method in the super-class.  Howver, a `PartialFunction` is intended to allow the `apply` method to fail (error) for some values of the input type. 

The Partial Functions is a redundant construct. One can indicate that a function may not give a value for some types by simply using an Option return type: `(A) -> Option<B>`. If a faster alternative, one without allocating  `Some` objects, one can also use Kotlin's nullable types.

One possible reason why the Scala library defines a `PartialFunction` as a subclass of a `Function` could be to allow writing function literals using brace-cases. It also allows having functions, like `collect`, which can be written as pattern-filters in an `awk`-ward style. 
However, it seems that the culture of the community of Kotlin programmers favours safe and clear code, even if longer, code over syntax candy. So, that benefit from partial functions disappears. 

In conclussion: Partial Functions are not an entirely sound idea in Scala, and the Arrow library for  Kotlin can do without them.


JorgeCastilloPrz:
All you saying here makes sense to me. I'll stay out from the thumbs up since I believe I don't have enough knowledge in this matter. I believe we're not using partial functions much within Arrow internals so it should be easy to remove in case you're not finished. Can you take a look too @raulraja ? thx

pakoito:
This was one approach we took before all the existing machinery around lenses and such was put in place. I'm okay with removing it too.

MarioAriasC:
I'm against altering the API in such a drastic way in just one step.

Mark the instances and functions related as `@Deprecated` pointing that will be deleted in the next "x.x.x" version, and the alternative.

It doesn't need months to be deleted it just a couple of releases one to deprecate everything and another one to finally delete it

modules/optics/arrow-optics/src/main/kotlin/arrow/optics/Optional.kt
```diff
@@ -79,15 +79,6 @@ interface POptional<S, T, A, B> : POptionalOf<S, T, A, B> {
       override fun set(s: S, b: B): T = set(s, b)
     }
 
-    /**
-     * Invoke operator overload to create a [POptional] of type `S` with focus `A`.
-     * Can also be used to construct [Optional]
-     */
-    operator fun <S, A> invoke(partialFunction: PartialFunction<S, A>, set: (S, A) -> S): Optional<S, A> = Optional(
```
MarioAriasC: Don't we need a version with `(S) -> Option<A>` to replace the `PartialFunction<S, A>`?

modules/optics/arrow-optics/src/main/kotlin/arrow/optics/Prism.kt
```diff
@@ -65,15 +65,6 @@ interface PPrism<S, T, A, B> : PPrismOf<S, T, A, B> {
       override fun reverseGet(b: B): T = reverseGet(b)
     }
 
-    /**
-     * Invoke operator overload to create a [PPrism] of type `S` with focus `A` with a [PartialFunction]
-     * Can also be used to construct [Prism]
-     */
-    operator fun <S, A> invoke(partialFunction: PartialFunction<S, A>, reverseGet: (A) -> S): Prism<S, A> = Prism(
```
MarioAriasC: Same here

nomisRev:
Nice PR @diesalbla! 👏 You explained it perfectly, I am in favor of this change.

raulraja:
Also in favor, thanks Diego and happy to see you contributing to arrow 🥰

adam-arold:
One more thing:

```kotlin
abstract class PartialFunction<in A, out B> : (A) -> B {
  abstract fun isDefinedAt(a: A): Boolean
}
```

This won't work in a multiplatform environment because inheriting from functions is not supported on the Javascript target.

pakoito:
I would redo this PR from scratch at this point.

nomisRev:
@pakoito why? We always squash & merge so this history will disappear anyway.

pakoito:
Before you fixed the merged it was messy lol

modules/optics/arrow-optics/src/main/kotlin/arrow/optics/Prism.kt
i-walker: @nomisRev how did you solve this?

nomisRev: https://github.com/arrow-kt/arrow/pull/1419/files#diff-7e03fe423fc60d0d4ff339cb00e547c1R283

PR #1423, opened 2019-04-25
Add deprecation notice to PartialFunction

pakoito:


diesalbla:
@pakoito It should also mark as deprecated every extension function or method in Arrow using `PartialFunction`, with a deprecation notice pointing to the `A -> Option<B>` variants (adding them if necessary).

PR #1405, opened 2019-04-13
Add Ktlint as code formatter

aballano:
### What's done

* Added Ktlint as default code formatter.
* It is already integrated with Travis (see second commit's failures)

### How to use 

For verification is enough by doing `gradle build` or more explicitly `gradle ktlint`.

For auto-correction: `gradle ktlintformat` and it will fix most issues except some that cannot be auto corrected.

### What's missing

There are **2 issues** that cannot be auto corrected and have to be fixed manually **within this PR scope**. The issues are:
* Wildcard imports (*)
* Single classes with different filename (example: [MonoidInvariant in a file named monoid.kt](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/monoid.kt))

Those doesn't seem to be configurable, so I'd rather prefer to hear opinions before changing them myself cc @arrow-kt/maintainers 

#### => The solution would be either ignore this issues file by file or fix them and prescind of wildcard imports and align file names with the class inside.

PR #1424, opened 2019-04-25
Remove several deprecated APIs

pakoito:


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
```diff
@@ -1,15 +1,9 @@
...
 import arrow.Kind
-import arrow.core.Either
-import arrow.core.Left
-import arrow.core.Right
-import arrow.core.Try
-import arrow.core.left
-import arrow.core.recover
-import arrow.core.right
-import arrow.data.k
+import arrow.core.*
```
pablisco: I actually prefer to use start imports, but I thought I ask, what's the project standard?

modules/core/arrow-core-data/src/test/kotlin/arrow/core/TryTest.kt
```diff
@@ -1,5 +1,6 @@
 package arrow.core
 
+import arrow.core.extensions.*
```
aballano: I think ktlint will fail because of this, you can modify your IDEA settings to not put on demand imports in any case

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
aballano: @pablisco After #1405 we had to either ignore all wildcard imports or fix them, we ended up fixing them.

pakoito: I have this setup on my IDE, what's the rule now?

modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/try.kt
```diff
@@ -3,21 +3,13 @@
 package arrow.core.extensions
 
 import arrow.Kind
-import arrow.core.Either
-import arrow.core.Eval
-import arrow.core.ForTry
-import arrow.core.Success
-import arrow.core.Try
+import arrow.core.*
```
aballano: same here

PR #1427, opened 2019-04-25
Concurrent instances for Rx

pakoito:


pakoito:
That last build has also passed, I don't know why it hasn't updated: https://travis-ci.org/arrow-kt/arrow/jobs/543218762

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/observablek.kt
```diff
@@ -149,6 +163,83 @@ interface ObservableKEffect : Effect<ForObservableK>, ObservableKAsync {
...
+          emitter.onNext(Fiber(s.k(), ObservableK {
+            conn.dispose()
+          }))
+        }
+      }.k()
+    }
+
+  override fun <A> cancelableF(k: ((Either<Throwable, A>) -> Unit) -> ObservableKOf<CancelToken<ForObservableK>>): ObservableK<A> =
+    ObservableK.asyncF { kindConnection, function ->
+      k(function).map { kindConnection.push(it) }
```
nomisRev: When should this get popped? 🤔

pakoito: I don't know the semantics of `pop` well, I'm open to suggestions.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
```diff
@@ -326,7 +326,10 @@ interface Concurrent<F> : Async<F> {
    * @see cancelableF for a version that can safely suspend impure callback registration code.
    */
   fun <A> cancelable(k: ((Either<Throwable, A>) -> Unit) -> CancelToken<F>): Kind<F, A> =
-    cancelableF { cb -> delay { k(cb) } }
+    cancelableF { cb ->
+      val token = k(cb)
```
nomisRev: Was the encoding incorrect? Not saying I don't agree this encoding is better.

modules/effects/arrow-effects-rx2-data/src/main/kotlin/arrow/effects/rx2/ObservableK.kt
```diff
@@ -79,29 +78,30 @@ data class ObservableK<A>(val observable: Observable<A>) : ObservableKOf<A>, Obs
...
-            }, { e2 ->
-              emitter.onError(Platform.composeErrors(e, e2))
-            })
-          } else {
-            throw e
+    Observable.create<B> { emitter ->
+      val dispose =
+        handleErrorWith { t -> Observable.fromCallable { emitter.onError(t) }.flatMap { Observable.error<A>(t) }.k() }
+          .concatMap { a ->
+            if (emitter.isDisposed) {
```
nomisRev: This is the case when a cancel signal was received during `acquire`, correct?

modules/effects/arrow-effects-rx2-data/src/main/kotlin/arrow/effects/rx2/ObservableK.kt
```diff
@@ -79,29 +78,30 @@ data class ObservableK<A>(val observable: Observable<A>) : ObservableKOf<A>, Obs
...
+              release(a, ExitCase.Canceled).fix().observable.subscribe({}, emitter::onError)
+              Observable.never<B>().k()
+            } else {
+              defer { use(a) }
+                .value()
+                .doOnError { t: Throwable ->
+                  defer { release(a, ExitCase.Error(t.nonFatalOrThrow())) }.value().subscribe({ emitter.onError(t) }, emitter::onError)
```
nomisRev: Where does this `OrThrow` go? Here it's swallowed by `defer`, which means `release` is not called but it's posted to `emitter::onError` immediately.

Extracting it outside of `defer` will it get swallowed by `create`, no?

I feel like we should drop `nonFatalOrThrow` here because I don't think `RxJava` takes into account fatal errors but I might be mistaken.

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/flowablek.kt
```diff
@@ -160,6 +172,83 @@ interface FlowableKEffect :
...
+  override fun <A> asyncF(k: ProcF<ForFlowableK, A>): FlowableK<A> =
+    FlowableK.asyncF({ _, cb -> k(cb) }, BS())
+
+  override fun <A> asyncF(fa: ConnectedProcF<ForFlowableK, A>): FlowableK<A> =
+    FlowableK.asyncF(fa)
+
+  override fun <A> CoroutineContext.startFiber(kind: Kind<ForFlowableK, A>): Kind<ForFlowableK, Fiber<ForFlowableK, A>> =
+    asScheduler().let { scheduler ->
+      Flowable.create<Fiber<ForFlowableK, A>>({ emitter ->
+        if (!emitter.isCancelled) {
```
nomisRev: Can it already be cancelled at this point? 🤔

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/observablek.kt
```diff
@@ -149,6 +163,83 @@ interface ObservableKEffect : Effect<ForObservableK>, ObservableKAsync {
...
+  override fun <A> asyncF(k: ProcF<ForObservableK, A>): ObservableK<A> =
+    ObservableK.asyncF { _, cb -> k(cb) }
+
+  override fun <A> asyncF(fa: ConnectedProcF<ForObservableK, A>): ObservableK<A> =
+    ObservableK.asyncF(fa)
+
+  override fun <A> CoroutineContext.startFiber(kind: ObservableKOf<A>): ObservableK<Fiber<ForObservableK, A>> =
+    asScheduler().let { scheduler ->
+      Observable.create<Fiber<ForObservableK, A>> { emitter ->
+        if (!emitter.isDisposed) {
```
nomisRev: Can it already be disposed at this point?

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/observablek.kt
```diff
@@ -149,6 +163,83 @@ interface ObservableKEffect : Effect<ForObservableK>, ObservableKAsync {
...
+          val s: ReplaySubject<A> = ReplaySubject.create()
+          val conn: RxDisposable = kind.value().subscribeOn(scheduler).subscribe(s::onNext, s::onError)
```
nomisRev: I'm seeing some `UndeliveredException` here. I tried throwing some synchronization at this, and the `cancel` operation but didn't seem to have any effect. Completely removing `s::onError` also didn't remove the `UndeliveredException` so maybe this is not the culprit.

Another thing I can imagine is an `UndeliveredException` is our `asScheduler` because that seems to throw in `schedule` but haven't gone further down that path.

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/flowablek.kt
```diff
@@ -160,6 +172,83 @@ interface FlowableKEffect :
...
+        emitter.setCancellable { dda.dispose(); ddb.dispose() }
+        val ffa = Fiber(sa.toFlowable(BS()).k(), FlowableK { dda.dispose() })
+        val ffb = Fiber(sb.toFlowable(BS()).k(), FlowableK { ddb.dispose() })
+        sa.subscribe({
+          emitter.onNext(RacePair.First(it, ffb))
+        }, emitter::onError, emitter::onComplete)
+        sb.subscribe({
+          emitter.onNext(RacePair.Second(ffa, it))
+        }, emitter::onError, emitter::onComplete)
+      }, BS()).subscribeOn(scheduler).observeOn(Schedulers.trampoline()).k()
```
nomisRev: What is reason for "observeOn(Schedulers.trampoline())" ?

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/observablek.kt
```diff
@@ -149,6 +163,83 @@ interface ObservableKEffect : Effect<ForObservableK>, ObservableKAsync {
...
+        emitter.setCancellable { dda.dispose(); ddb.dispose() }
+        val ffa = Fiber(sa.k(), ObservableK { dda.dispose() })
+        val ffb = Fiber(sb.k(), ObservableK { ddb.dispose() })
+        sa.subscribe({
+          emitter.onNext(RacePair.First(it, ffb))
+        }, emitter::onError, emitter::onComplete)
+        sb.subscribe({
+          emitter.onNext(RacePair.Second(ffa, it))
+        }, emitter::onError, emitter::onComplete)
+      }.subscribeOn(scheduler).observeOn(Schedulers.trampoline()).k()
```
nomisRev: any reason for `observeOn(Schedulers.trampoline())`?

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/observablek.kt
```diff
@@ -149,6 +163,83 @@ interface ObservableKEffect : Effect<ForObservableK>, ObservableKAsync {
...
+    ObservableK.asyncF { kindConnection, function ->
+      k(function).map { kindConnection.push(it) }
+    }
+
+  override fun <A, B> CoroutineContext.racePair(fa: ObservableKOf<A>, fb: ObservableKOf<B>): ObservableK<RacePair<ForObservableK, A, B>> =
+    asScheduler().let { scheduler ->
+      Observable.create<RacePair<ForObservableK, A, B>> { emitter ->
+        val sa = ReplaySubject.create<A>()
+        val sb = ReplaySubject.create<B>()
+        val dda = fa.value().subscribe(sa::onNext, sa::onError)
```
nomisRev: Any reason we don't complete here but we do in `raceTriple`?

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/observablek.kt
pakoito: Unlikely, but plausible.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
pakoito: No, it was incorrect before and it was causing a bunch of errors.

modules/effects/arrow-effects-rx2-data/src/main/kotlin/arrow/effects/rx2/ObservableK.kt
pakoito: Yep!

modules/effects/arrow-effects-rx2-data/src/main/kotlin/arrow/effects/rx2/ObservableK.kt
pakoito: It does IIRC.

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/observablek.kt
pakoito: Yes, we'd like to continue on the current thread afterwards, yet release it for other purposes.

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/observablek.kt
pakoito: ...a mistake. Having the complete is probably the correct version now that we use concatMap

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/observablek.kt
nomisRev: Okay, so this is for fairness?

modules/effects/arrow-effects-rx2-data/src/main/kotlin/arrow/effects/rx2/ObservableK.kt
nomisRev: So is there still a need for us to interfere there or should we just pass the fatal error to Rx?

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
nomisRev: Oh now I see. damn.

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/observablek.kt
pakoito: And a bit of trial and error.

modules/effects/arrow-effects-rx2-data/src/main/kotlin/arrow/effects/rx2/ObservableK.kt
pakoito: Let's do it anyway.

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/observablek.kt
nomisRev: I'll add a concrete version off cancelable in favor of `asyncF`. I need to do a PR to deprecated asyncF.

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/observablek.kt
nomisRev: broke the build 😭

modules/effects/arrow-effects-rx2-data/src/main/kotlin/arrow/effects/rx2/FlowableK.kt
```diff
@@ -79,29 +79,27 @@ data class FlowableK<A>(val flowable: Flowable<A>) : FlowableKOf<A>, FlowableKKi
...
-        }
-      }, emitter::onError, emitter::onComplete)
-    }, mode))
+    Flowable.create<B>({ emitter ->
+      val dispose =
+        handleErrorWith { e -> Flowable.fromCallable { emitter.onError(e) }.flatMap { Flowable.error<A>(e) }.k() }
+          .value()
+          .concatMap { a ->
+            if (emitter.isCancelled) {
+              release(a, ExitCase.Canceled).value().subscribe({}, emitter::onError)
```
pakoito: This emitter may be the one that's causing the exception (?)

nomisRev: But what cancelation is causing the error? I'll try this out

PR #1433, opened 2019-05-02
add tag to Either, replace pattern match with fold

richard-gibson:


modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -5,30 +5,21 @@ import arrow.core.Either.Left
...
-   */
-  internal abstract val isRight: Boolean
+sealed class Either<out A, out B>(@JvmField @PublishedApi internal val tag: Int) : EitherOf<A, B> {
 
-  /**
-   * Returns `true` if this is a [Left], `false` otherwise.
-   * Used only for performance instead of fold.
-   */
-  internal abstract val isLeft: Boolean
+  fun isLeft(): Boolean = tag == LeftTag
```
pakoito: I'd say deprecate this API.

modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -176,13 +153,20 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
...
       val ev: Either<L, Either<A, B>> = f(a).fix()
-      return when (ev) {
-        is Left -> Left(ev.a)
-        is Right -> {
+      return when (ev.tag) {
+        LeftTag -> {
+          ev as Left<L>
+          Left(ev.a)
+        }
+        else -> {
```
pakoito: No else, it has to be RightTag to get all the perf benefits.

modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -201,13 +185,9 @@ fun <R> Right(right: R): Either<Nothing, R> = Either.right(right)
...
  */
+@Suppress("UNCHECKED_CAST")
 inline fun <A, B, C> EitherOf<A, B>.flatMap(f: (B) -> Either<A, C>): Either<A, C> =
-  fix().let {
-    when (it) {
-      is Right -> f(it.b)
-      is Left -> it
-    }
-  }
+  fix().let { eith -> eith.fold({ eith as Left<A> }, f) }
```
pakoito: Is the cast necessary? If so, and given that it's pervasive, please add a `unsafeRecast` inline function that does this.

nomisRev: This is not `unsafeRecast` this is an optimization to avoid re-wrapping `Left` values on short-circuits. Would you be in favor writing directly using tags here instead of an inlined fold then?

pakoito: I mean that there are several instances of this through the codebase, let's unify them in a single function.

And because you're returning an either from both sides, the cast may not be necessary in this specific instance as Left is defined with Nothing for the right side.

raulraja: The idea is that we can write a faster `fold` in terms of tags which is consistently faster to `when` because of the table switch optimization this fold becomes inline and the rest of the Arrow code base uses `fold` wherever is now using `when` with `is Left` etc.

raulraja:
Ideally, benchmarks also include `Either.fold` based on tags vs `Result` based on when. https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-result/index.html

raulraja:
Since these are not part of the effects benchmarks I propose we have just a generic module `arrow-benchmarks` where we have all benchmarks for now so at least the name is consistent.

modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
pakoito: Yeah, that I understand. I mean literally a function `fun <A> Either<Nothing, Nothing>.recastLeftUnsafe(): Either<A, Nothing> = eith as Left<A>`

raulraja:
I verified in my house with further benchmarks that there is no difference that is noticeable because the kotlin `when` is just doing `instanceof` for simple cases which is cheap. This is more noticeable in Scala where Pattern matching does a lot more work. Unless anyone has different results I think we should close and discard this for now as soon as @nomisRev  verifies it's the same case for Fx.

raulraja:
```
Benchmark             Mode  Cnt          Score        Error  Units
TagsVsWhen.tagsFold  thrpt   20  280262944.752 ? 652391.416  ops/s
TagsVsWhen.whenFold  thrpt   20  263570800.470 ? 733552.433  ops/s
```
they beat each other out in that range so I'm not sure this is significant. What we need to attempt perhaps is to see if we can inline all of our folds or check for ones that have not been inlined and should.

raulraja:
Seems like this optimization may be worth it in cases where there is a larger number of cases https://stackoverflow.com/questions/6705955/why-switch-is-faster-than-if

Issue #1428, opened 2019-04-26
Make Foldable#iterateRight more Kotlin-y


pakoito:
This function looks lifted from a Haskell codebase, with currying and all. Let's make it more like a Kotlin function, maybe an extension function on `Iterable`.

https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Foldable.kt#L202-L208

The function also shouldn't live in Foldable, but be Eval.kt

plastic-karma:
I can take a look at that. @pakoito, do you have any other "make it more like a Kotlin function" things in mind except making it an extension function?

nomisRev:
I'd start with something like `fun <A, B> Iterator<A>.iterateRight(lb: Eval<B>, f: (Eval<B>, Eval<B>) -> Eval<B>): Eval<B>` this, this makes it similar to the signature of `foldRight`.

pakoito:
@plastic-karma like @nomisRev said, it's just pulling the function into another file, and changing the signature. It's a good starter ticket to setup the project :)

plastic-karma:
Sounds good, I’ll make the change.

plastic-karma:
@pakoito, is it okay though to move the function from ``arrow-typeclasses`` to ``arrow-core-data``?

pakoito:
Yep!

raulraja:
Fixed in #1434

PR #1434, opened 2019-05-05
`iterateRight` as extension of `Iterator`

plastic-karma:
fixing https://github.com/arrow-kt/arrow/issues/1428


pakoito:
We're in business, thanks for fixing it!

PR #1436, opened 2019-05-09
Add new recommended constructor for Ref

pakoito:
Making the allocation lazy so every call to `IO` generates a new one, and we prevent eager expensive operations unless needed.

### DIFF
--- a/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/MVar.kt
+++ b/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/MVar.kt
@@ -318,8 +318,8 @@ interface MVar<F, A> {
  *
  * fun main(args: Array<String>) {
  *   //sampleStart
- *   val mvarPartial: MVarPartialOf<ForIO> = MVar(IO.async())
- *   val intVar: IOOf<MVar<ForIO, Int>> = mvarPartial.of(5)
+ *   val mvarPartial: MVarFactory<ForIO> = MVar.factoryUncancelable(IO.async())
+ *   val intVar: IOOf<MVar<ForIO, Int>> = mvarPartial.just(5)
  *   val stringVar: IOOf<MVar<ForIO, String>> = mvarPartial.empty<String>()
  *   //sampleEnd
  * }
@@ -336,8 +336,8 @@ interface MVarFactory<F> {
    *
    * fun main(args: Array<String>) {
    *   //sampleStart
-   *   val mvarPartial: MVarPartialOf<ForIO> = MVar(IO.async())
-   *   val intVar: IOOf<MVar<ForIO, Int>> = mvarPartial.of(5)
+   *   val mvarPartial: MVarFactory<ForIO> = MVar.factoryUncancelable(IO.async())
+   *   val intVar: IOOf<MVar<ForIO, Int>> = mvarPartial.just(5)
    *   //sampleEnd
    * }
    * ```
@@ -353,7 +353,7 @@ interface MVarFactory<F> {
    *
    * fun main(args: Array<String>) {
    *   //sampleStart
-   *   val mvarPartial: MVarPartialOf<ForIO> = MVar(IO.async())
+   *   val mvarPartial: MVarFactory<ForIO> = MVar.factoryUncancelable(IO.async())
    *   val stringVar: IOOf<MVar<ForIO, String>> = mvarPartial.empty<String>()
    *   //sampleEnd
    * }
--- a/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/MVar.kt
+++ b/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/MVar.kt
@@ -26,13 +26,13 @@ interface MVar<F, A> {
    *
    * fun main(args: Array<String>) {
    * //sampleStart
-   * val mvar = MVar(IO.async())
+   * val mvar = MVar.factoryUncancelable(IO.async())
    *
    * mvar.empty<Int>().flatMap { v ->
    *   v.isEmpty()
    * }.unsafeRunSync() == true
    *
-   * mvar.of(10).flatMap { v ->
+   * mvar.just(10).flatMap { v ->
    *   v.isEmpty()
    * }.unsafeRunSync() == false
    * //sampleEnd
@@ -51,9 +51,9 @@ interface MVar<F, A> {
    *
    * fun main(args: Array<String>) {
    * //sampleStart
-   * val mvar = MVar(IO.async())
+   * val mvar = MVar.factoryUncancelable(IO.async())
    *
-   * mvar.of(10).flatMap { v ->
+   * mvar.just(10).flatMap { v ->
    *   v.isNotEmpty()
    * }.unsafeRunSync() == true
    *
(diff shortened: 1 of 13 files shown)


nomisRev:
So @pakoito you're in favor of keeping both constructors then?

```
fun <F, A> of(a: A, MD: MonadDefer<F>): Kind<F, Ref<F, A>>
fun <F, A> of(MD: MonadDefer<F>, f: () -> A): Kind<F, Ref<F, A>>
```

We should to the same for `MVar`. This is not required for `Semaphore` since it only operators on `Long` nor `Promise` unless we want add a constructor to create a completed `Promise`.

pakoito:
@nomisRev the first constructor is the same as `Ref.unsafe` because the operation is eagerly executed and cached. I'm for removing it altogether.

Will do MVar when I get home.

nomisRev:
Not exactly but it requires the user to deal with it, so I agree this is a good API improvement.

```
val ref: IO<Ref<String>> = IO { notAConstant() }.flatMap { a: String -> Ref.of(a, MD) } 
```

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/Ref.kt
```diff
@@ -110,6 +110,13 @@ interface Ref<F, A> {
       unsafe(a, MD)
     }
 
+    /**
+     * Creates an asynchronous, concurrent mutable reference initialized using the supplied function.
+     */
+    fun <F, A> of(MD: MonadDefer<F>, f: () -> A): Kind<F, Ref<F, A>> = MD.delay {
```
raulraja: Why should also add this factory in MonadDefer to create new refs without the argument. That way in fx we can just create refs from the this scope

raulraja: Also for all others that are constructed similarly

nomisRev: We need to do this for MD/Async & Concurrent for cancelation.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/MVar.kt
```diff
@@ -288,18 +288,18 @@ interface MVar<F, A> {
     fun <F, A> uncancelableOf(initial: A, AS: Async<F>): Kind<F, MVar<F, A>> =
       UncancelableMVar(initial, AS)
 
-    operator fun <F> invoke(AS: Async<F>) = object : MVarPartialOf<F> {
+    operator fun <F> invoke(AS: Async<F>) = object : MVarFactory<F> {
```
nomisRev: 👍

nomisRev:
```suggestion
    fun <F> factory(AS: Async<F>) = object : MVarFactory<F> {
```

Consistent with `Ref`

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/MonadDefer.kt
```diff
@@ -36,6 +38,11 @@ interface MonadDefer<F> : MonadThrow<F>, Bracket<F, Throwable> {
   fun <A> delayOrRaise(f: () -> Either<Throwable, A>): Kind<F, A> =
     defer { f().fold({ raiseError<A>(it) }, { just(it) }) }
 
+  /**
+   * Creates a [Ref] to purely manage mutable state, initialized by the function [f]
+   */
+  fun <A> ref(f: () -> A): Kind<F, Ref<F, A>> = Ref(this, f)
```
nomisRev:
```suggestion
  fun <A> Ref(f: () -> A): Kind<F, Ref<F, A>> = Ref(this, f)
```

Should we fake constructors? This a common pattern that they use throughout the Kotlin lib as well.

pakoito: I'd rather not for conflict purposes.

nomisRev: Conflict with? This is only accessible in scoped areas `IO.concurrent().run { }`, `IO.concurrent().fx.concurrent { }`, `IO.fx { }`.

pakoito: With the constructor I mean.

pakoito: Different overrides tho. But still, I don't like it.

nomisRev: > Ref(), Semaphore(5), Promise(), MVar(), MVar(5)

I'd still like to discuss this because I am really in favor of doing them with capital letter but if we all agree to them as regular functions I can live with that but we should be consistent. And/or properly define when to do this and when not, with a consistent API for 1.0 in mind.

* We currently do this for [State](https://github.com/arrow-kt/arrow/blob/c2de52c6f5850c6b67d54366a43203cebb0503a3/modules/core/arrow-extras-data/src/main/kotlin/arrow/data/State.kt#L50) to provide a Kotlin idiomatic `State` constructor over `StateT`. I got stuck improving the API further.
* Kotlinx does this also for [Runnable](https://github.com/Kotlin/kotlinx.coroutines/blob/34287743c5cb2ae6cdd0aab09a4b42403ae82ef7/kotlinx-coroutines-core/common/src/Runnable.common.kt#L21) and I've seen PRs doing it for [Semaphore](https://github.com/Kotlin/kotlinx.coroutines/blob/2cbca1acf3f6be5c68339d1b33050e4396854777/kotlinx-coroutines-core/common/src/sync/Semaphore.kt#L25)

Given that the `fx` DSL is going to be one of our selling point AFAIK, I am in favor of giving it the semantics of a constructor there.
In the DSL you'll be able to work in a way you would with KEEP-87, where all instances can be injected by us. Even including syntax that has a default value for `CoroutineContext`.

```
fx.concurrent {
   val (fiber) = IO { "Hello from dispatchers().default() }.fork()
   val (timeOut) = (fiber race sleep(100.miliseconds))
   val (semaphore) = Semaphore(5)
   val (ref) = Ref(1)
   val (parResult) = parMapN(IO { 1 }, IO { 1 }) { a, b -> a + b }
}
```

pakoito: We do `State` because there's no `State` constructor, it's an alias for `StateT<ForId, S, A>` with the value partially applied.

In this case we have a constructor, and the function that'll only be available in MonadDefer/Concurrent contexts. They have different overloads, so that's a positive. By the time KEEP87 lands, the new function will be obsolete anyway.

I don't have strong feelings either way, if you want to change it go for it.

nomisRev: > By the time KEEP87 lands, the new function will be obsolete anyway.

Which means we can remove the alias in `Concurrent` and people's code will remain the same, right?

pakoito: fair, fair :D PR?

PR #1439, opened 2019-05-15
Add support for suspend functions via the effect constructor in Async. Rename delay to later.

pakoito:
This PR adds the constructor `effect` in the `Async` typeclass, and a copy in the IO datatype.

Screen Shot 2019-05-15 at 4 29 15 PM


modules/docs/arrow-docs/docs/docs/effects/async/README.md
```diff
@@ -87,9 +87,23 @@ Behind the scenes `continueOn()` starts a new coroutine and passes the rest of t
 
 The function `continueOn()` is also available inside [`Monad Comprehensions`]({{ '/docs/patterns/monad_comprehensions' | relative_url }}).
 
+#### effect
+
+Similar to `MonadDefer`'s `invoke`, this constructor takes a single suspended function and optionally the `CoroutineContext` it has to be run on.
```
nomisRev:
```suggestion
Similar to `MonadDefer`'s `delay`, this constructor takes a single suspended function and optionally the `CoroutineContext` it has to be run on.
```

modules/docs/arrow-docs/docs/docs/effects/async/README.md
```diff
@@ -87,9 +87,23 @@ Behind the scenes `continueOn()` starts a new coroutine and passes the rest of t
...
+    // In CommonPool
+    requestSuspend(createUserFromId(123))
+  }
+}
+```
+
 #### invoke with CoroutineContext
 
-Similar to `MonadDefer`'s `invoke`, this constructor it takes a single generator function and the `CoroutineContext` it has to be run on.
+Similar to `MonadDefer`'s `invoke`, this constructor takes a single generator function and the `CoroutineContext` it has to be run on.
```
nomisRev:
```suggestion
Similar to `MonadDefer`'s `delay`, this constructor takes a single generator function and the `CoroutineContext` it has to be run on.
```

nomisRev:
We deprecated and removed `MonadDefer#invoke` in favor of `delay`.

raulraja:
Looks good but let's get rid of the invoke. Also Simon should approve since he has outstanding changes that may clash here

pakoito:
@raulraja which invoke? the IO one to replace with effect? that's a perf hit of async jumps for many IOs out there.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/AsyncLaws.kt
```diff
@@ -56,9 +59,9 @@ object AsyncLaws {
 
   fun <F> Async<F>.asyncConstructor(EQ: Eq<Kind<F, Int>>): Unit =
     forFew(5, Gen.intSmall(), Gen.intSmall()) { threadId1: Int, threadId2: Int ->
-      delay(newSingleThreadContext(threadId1.toString())) { getCurrentThread() }
+      later(newSingleThreadContext(threadId1.toString())) { getCurrentThread() }
```
raulraja: shouldn't these be `effect`?

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/AsyncLaws.kt
```diff
@@ -79,26 +82,46 @@ object AsyncLaws {
...
+        val br = later { promise }.bracketCase(use = { fa }, release = { r, exitCase ->
           when (exitCase) {
             is ExitCase.Completed -> r.complete(b)
             is ExitCase.Error -> r.complete(b)
             else -> just<Unit>(Unit)
           }
         })
 
-        asyncF<Unit> { cb -> delay { cb(Right(Unit)) }.flatMap { br.attempt().`as`(Unit) } }
+        asyncF<Unit> { cb -> later { cb(Right(Unit)) }.flatMap { br.attempt().`as`(Unit) } }
```
raulraja: `effect`?

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/AsyncLaws.kt
```diff
@@ -79,26 +82,46 @@ object AsyncLaws {
...
           .flatMap { promise.get() }
       }.equalUnderTheLaw(just(b), EQ)
     }
   }
 
+  fun <F> Async<F>.effectCanCallSuspend(EQ: Eq<Kind<F, Int>>): Unit =
+    forAll(Gen.int()) { id ->
+      val fs: suspend () -> String = { Thread.currentThread().name }
+
+      effect(newSingleThreadContext(id.toString())) {
```
raulraja: :+1:

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
```diff
@@ -144,7 +144,7 @@ object ConcurrentLaws {
   fun <F> Concurrent<F>.cancelableCancelableFCoherence(EQ: Eq<Kind<F, Int>>): Unit =
     forAll(Gen.either(Gen.throwable(), Gen.int())) { eith ->
       cancelable<Int> { cb -> cb(eith); just<Unit>(Unit) }
-        .equalUnderTheLaw(cancelableF { cb -> delay { cb(eith); just<Unit>(Unit) } }, EQ)
+        .equalUnderTheLaw(cancelableF { cb -> later { cb(eith); just<Unit>(Unit) } }, EQ)
```
raulraja: most usages of later I'm seeing here are effects. Also I don;t understand why there are versions of `later` that take a ctx as argument as a context is always used to perform effects.

modules/docs/arrow-docs/docs/docs/effects/async/README.md
```diff
@@ -87,9 +87,23 @@ Behind the scenes `continueOn()` starts a new coroutine and passes the rest of t
...
 The function `continueOn()` is also available inside [`Monad Comprehensions`]({{ '/docs/patterns/monad_comprehensions' | relative_url }}).
 
+#### effect
+
+Similar to `MonadDefer`'s `later`, this constructor takes a single suspended function and optionally the `CoroutineContext` it has to be run on.
+
+```kotlin
+IO.async().run {
+  // In current thread
+  effect(CommonPool) {
```
raulraja: I would much rather not have this constructor and only have to teach `continueOn` which is more explicit and requires no nesting.

modules/docs/arrow-docs/docs/docs/effects/async/README.md
```diff
@@ -103,7 +117,7 @@ IO.async().run {
 
 #### defer with CoroutineContext
```
raulraja: A context switch is an effect, that is why you can only do it in the presence of a continuation. For that reason, I believe all `CoroutineContext` references can only appear on `effect` and up but not in `later`.`later` signature will only accept pure functions, pure functions should not fail and a context switch or task submission to a pool can result in a https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/RejectedExecutionException.html . As such `continueOn` should be suspended and can only be accepted as an `effect`

raulraja: The comment above is in regards to APIs like https://github.com/arrow-kt/arrow/pull/1439/files#diff-3d061ce877a7468475fa61a0b9fe6b15R227

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -42,9 +45,22 @@ sealed class IO<out A> : IOOf<A> {
 
     fun <A> raiseError(e: Throwable): IO<A> = RaiseError(e)
 
-    operator fun <A> invoke(f: () -> A): IO<A> = defer { Pure(f()) }
+    operator fun <A> invoke(ctx: CoroutineContext, f: suspend () -> A): IO<A> =
+      effect(ctx, f)
 
-    fun <A> defer(f: () -> IOOf<A>): IO<A> = Suspend(f)
+    operator fun <A> invoke(f: suspend () -> A): IO<A> =
```
raulraja: How about something like:
```suggestion
    operator fun <A> invoke(f: suspend IO.Companion.() -> A): IO<A> =
```

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Async.kt
```diff
@@ -137,7 +140,7 @@ interface Async<F> : MonadDefer<F> {
    * }
    * ```
    */
-  fun <A> delay(ctx: CoroutineContext, f: () -> A): Kind<F, A> =
+  fun <A> later(ctx: CoroutineContext, f: () -> A): Kind<F, A> =
```
raulraja: I'm not sure we need these APIs or what use case they attempt to cover.

nomisRev:
@raulraja @pakoito what are we doing with this? I am assuming we still want to refactor/deprecate to `later`?

@pakoito I can take over and clean this up to be merged if you like.

raulraja:
@nomisRev I believe `later` was decided. I don't think we should deprecate it since `delay` is already confusing enough with coroutines and we are not granting any kind of bin compat in this release

build.gradle
```diff
@@ -20,7 +20,6 @@ buildscript {
     ext {
         // Constants
         ankKotlinVersion = '1.3.11'
-        daggerVersion = '2.21'
```
pakoito: 🚎

pakoito:
I'm always up for breaking changes, so go for it. Delay is 0.9 only right?

PR #1442, opened 2019-05-17
IO incomplete `IORunLoop#step`

nomisRev:
Currently `IO.just(1).uncancelable().unsafeRunSync()` hangs, because `step` is a partial function. This adds the missing step.

* Is there a way to enforce exhaustiveness without capturing the value? And/or would the additional bytecode of capturing the value slow down the loop/step?

### DIFF
--- a/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IORunLoop.kt
+++ b/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IORunLoop.kt
@@ -5,6 +5,7 @@ import arrow.core.Either
 import arrow.core.Left
 import arrow.core.NonFatal
 import arrow.core.Right
+import arrow.effects.internal.ArrowInternalException
 import arrow.effects.internal.Platform.ArrayStack
 import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.startCoroutine
@@ -71,7 +72,7 @@ internal object IORunLoop {
         }
         is IO.Async -> {
           // Return case for Async operations
-          return suspendInAsync(currentIO as IO<A>, bFirst, bRest, currentIO.k)
+          return suspendInAsync(currentIO, bFirst, bRest) as IO<A>
         }
         is IO.Bind<*, *> -> {
           if (bFirst != null) {
@@ -87,9 +88,7 @@ internal object IORunLoop {
             bRest.push(bFirst)
           }
           val localCurrent = currentIO
-
           val currentCC = localCurrent.cc
-
           val localCont = currentIO.cont
 
           bFirst = { c: Any? -> IO.just(c) }
@@ -108,8 +107,15 @@ internal object IORunLoop {
           bFirst = currentIO as BindF
           currentIO = currentIO.source
         }
+        is IO.ContextSwitch -> {
+          return suspendInAsync(currentIO, bFirst, bRest) as IO<A>
+        }
         null -> {
           currentIO = IO.RaiseError(NullPointerException("Stepping on null IO"))
+        }
+        else -> {
+          // Since we don't capture the value of `when` kotlin doesn't enforce exhaustiveness
+          currentIO = IO.raiseError(IORunLoopMissingStep)
         }
       }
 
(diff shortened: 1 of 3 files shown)


nomisRev:
Running the benchmarks on the CI might be a little tricky since the currently enabled set takes ~25 minutes on my machine. That would effectively more than double the current build time.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IORunLoop.kt
```diff
@@ -273,6 +271,10 @@ internal object IORunLoop {
         null -> {
           currentIO = IO.RaiseError(NullPointerException("Looping on null IO"))
         }
+        else -> {
+          // Since we don't capture the value of `when` kotlin doesn't enforce exhaustiveness
```
pakoito: This doesn't sound correct.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IORunLoop.kt
```diff
@@ -273,6 +271,10 @@ internal object IORunLoop {
         null -> {
           currentIO = IO.RaiseError(NullPointerException("Looping on null IO"))
```
pakoito: Extract this excepction at the bottom too.

raulraja:
@nomisRev You can enforce exhaustiveness in when statements to enforce them to become expressions with:
```kotlin
sealed class Term {
  object Y : Term()
  object Z : Term()
}

private val <A> A.exhaustive: A // The identity function
    get() = this

when (val x: Term = term()) {
  Y -> println("Y") //Unit return
  Z -> println("Z") //Unit return
}.exhaustive // turns when into an expression of `Unit` return
```
I can't foresee this causing any performance impact since this is most likely optimized given `exhaustive` it's just the identity function. Perhaps it can be forced `inline` as a function but while developing the entire Loop it may be useful.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IORunLoop.kt
raulraja: This is actually the case for all `when` blocks that are not expressions and whose RHS produce `Unit`. In that case, Kotlin does not require exhaustiveness.

raulraja: https://kotlinlang.org/docs/reference/control-flow.html
> If when is used as an __expression__, the else branch is mandatory

pakoito: 👍 

Use the `exhaustive` trick then

nomisRev:
@raulraja @pakoito oh it seems I've missed another trick we leverage in the `when`.

We're actually matching on `Kind<ForIO, Any?>?` which is valid but it doesn't give exhaustiveness. If we want exhaustiveness we must *always* cast. That'll be a downer on performance, since we already saw that casting is quite slow in Kotlin since they also insert null checks.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IORunLoop.kt
nomisRev: Oops... left a general comment instead of a reply here.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IORunLoop.kt
```diff
@@ -55,44 +56,38 @@ internal object IORunLoop {
         }
         is IO.Suspend -> {
           val thunk: () -> IOOf<Any?> = currentIO.thunk
-          currentIO = executeSafe { thunk() }
+          currentIO = executeSafe(thunk)
```
pakoito: perfffffff

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IORunLoop.kt
```diff
@@ -55,44 +56,38 @@ internal object IORunLoop {
...
           // Return case for Async operations
           return suspendAsync(currentIO, bFirst, bRest) as IO<A>
         }
+        is IO.Effect -> {
+          return suspendAsync(currentIO, bFirst, bRest) as IO<A>
+        }
         is IO.Bind<*, *> -> {
           if (bFirst != null) {
-            if (bRest == null) bRest = ArrayStack()
+            if (bRest == null) {
```
pakoito: Cheeky

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IORunLoop.kt
```diff
@@ -108,7 +103,10 @@ internal object IORunLoop {
           currentIO = currentIO.source
         }
         is IO.ContextSwitch -> {
-          return wrapInAsync(currentIO, bFirst, bRest) as IO<A>
```
pakoito: Is wrapInAsync even used anymore?

PR #1443, opened 2019-05-17
IO rewrite ContinueOn impl

nomisRev:
This PR rewrites the current `ContinueOn` impl.
I've added `Effect` already because I rely on it to implement `ContinueOn`.

The reason for this rewrite is so we can keep track of `CoroutineContext` when adding the `Effect` impl, so following test passes [source](https://github.com/arrow-kt/arrow/blob/693c2d2861b6fab0474130b7c7266855d4c8e41f/modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/IOTest.kt#L144)

```
"fx can pass context state across not/bind" {
  fx {
    val ctx = !effect { kotlin.coroutines.coroutineContext }
    !effect { ctx shouldBe EmptyCoroutineContext }
    continueOn(CoroutineName("Simon")) // this is immediately lost and useless.
    val ctx2 = !effect { kotlin.coroutines.coroutineContext }
    !effect { ctx2 shouldBe CoroutineName("Simon") }
  }.unsafeRunSync()
}
```

### DIFF
--- a/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IORunLoop.kt
+++ b/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IORunLoop.kt
@@ -140,10 +140,13 @@ internal object IORunLoop {
     bRest: CallStack?
   ): IO<A> =
     // Hitting an async boundary means we have to stop, however if we had previous `flatMap` operations then we need to resume the loop with the collected stack
-    if (bFirst != null || (bRest != null && bRest.isNotEmpty())) IO.Async { conn, cb ->
-      loop(currentIO, conn, cb as Callback, null, bFirst, bRest)
+    if (bFirst != null || (bRest != null && bRest.isNotEmpty())) {
+      IO.Async { conn, cb ->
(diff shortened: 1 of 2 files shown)


modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IORunLoop.kt
```diff
@@ -136,21 +137,13 @@ internal object IORunLoop {
...
-    when {
-      bFirst != null || (bRest != null && bRest.isNotEmpty()) ->
-        IO.Async { conn, cb ->
-          val rcb = RestartCallback(conn, cb as Callback)
-          rcb.prepare(bFirst, bRest)
-          register(conn, rcb)
-        }
-      else -> currentIO
+    // Hitting an async boundary means we have to stop, however if we had previous `flatMap` operations then we need to resume the loop with the collected stack
+    if (bFirst != null || (bRest != null && bRest.isNotEmpty())) IO.Async { conn, cb ->
```
pakoito: Brackets here for the if and else please.

PR #1435, opened 2019-05-07
Fx on typeclass

nomisRev:
This PR will remove the current `Fx` hierarchy in favor from merging them in the current typeclass hierarchy.

```
[arrow.typeclasses.suspended.monad.Fx]
[arrow.typeclasses.suspended.monaderror.Fx]
[arrow.effects.typeclasses.suspended.monaddefer.Fx]
[arrow.effects.typeclasses.suspended.concurrent.Fx]
```

The corresponding `fx` will be available on the corresponding typeclasses and will be available in the following manner.

```
fun <F> Monad<F>.program(): Kind<F, Unit> = fx.monad { ... } 
fun <F> MonadThrow<F>.program(): Kind<F, Unit> = fx.monadThrow { ... } 
fun <F> Async<F>.program(): Kind<F, Unit> = fx.async { ... } 
fun <F> Concurrent<F>.program(): Kind<F, Unit> = fx.concurrent { ... } 
```

**breaking changes**

This PR will remove the currently deprecated `binding`, and it'll be replaced with `fx` as a typeclass method. Additionally `bindingCancelable` on `MonadDefer` will be removed, and `binding` on `MonadDefer` will be moved to `Async`.

PR #1447, opened 2019-05-19
IO forking

nomisRev:
Currently, we rely on the following snippet to fork starting an `IO`.
This PR removes that boilerplate throughout the code and will share the `Continuation`/`RestartCallback` instance throughout the loop instead of tossing it like before.

```
val a: suspend () -> A = {
      suspendCoroutine { ca: Continuation<A> ->
        IORunLoop.startCancelable(ioA, connA) { either: Either<Throwable, A> ->
          either.fold({ error ->
            ca.resumeWith(Result.failure(error))
          }, { a ->
            ca.resumeWith(Result.success(a))
          })
        }
      }
    }

```

I don't think can optimize this further since we want to guarantee forking in all concurrency operators (in the case the CC is capable of it).


nomisRev:
@pakoito this change changes the threading semantics a little bit. Before we the result in the receiver `CoroutineContext` but now it finishes on the last seen `ctx`. Here `"6"` instead of `"all"`.

```
"parallel mapping is done in the expected CoroutineContext" {
  fun makePar(num: Long) =
    IO(newSingleThreadContext("$num")) {
      // Sleep according to my number
      Thread.sleep(num * 20)
      num
    }
  val result =
    newSingleThreadContext("all").parMapN(
      makePar(6), IO.just(1L), makePar(4), IO.defer { IO.just(2L) }, makePar(5), IO { 3L }) { _, _, _, _, _, _ ->
      Thread.currentThread().name
    }.unsafeRunSync()
  result shouldBe "all"
}
```

modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/IOTest.kt
```diff
@@ -301,11 +299,11 @@ class IOTest : UnitSpec() {
 
       val result =
         newSingleThreadContext("all").parMapN(
-          makePar(6), IO.just(1L), makePar(4), IO.defer { IO.just(2L) }, makePar(5), IO { 3L }) { _, _, _, _, _, _ ->
+          makePar(6), just(1L), makePar(4), IO.defer { just(2L) }, makePar(5), IO { 3L }) { _, _, _, _, _, _ ->
           Thread.currentThread().name
         }.unsafeRunSync()
 
-      result shouldBe "all"
+      result shouldBe "6"
```
pakoito: Why 6 and not any other?

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -123,6 +126,39 @@ sealed class IO<out A> : IOOf<A> {
...
+   *
+   * ```kotlin:ank:playground
+   * import arrow.effects.*
+   * import arrow.effects.extensions.io.async.async
+   * import arrow.effects.extensions.io.monad.binding
+   * import kotlinx.coroutines.Dispatchers
+   *
+   * fun main(args: Array<String>) {
+   *   //sampleStart
+   *   binding {
```
raulraja: should use `fx`

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -123,6 +126,39 @@ sealed class IO<out A> : IOOf<A> {
...
+   *     val promise = Promise.uncancelable<ForIO, Int>(IO.async()).bind()
```
raulraja: `!`

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -123,6 +126,39 @@ sealed class IO<out A> : IOOf<A> {
...
+   *     val fiber = promise.get().fix().startFiber(Dispatchers.Default).bind()
```
raulraja: Should use our dispatchers

modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/MVarTest.kt
```diff
@@ -78,9 +78,9 @@ class MVarTest : UnitSpec() {
         binding {
           val av = mvar.empty<Int>().bind()
 
-          val f1 = av.put(10).startFiber(Dispatchers.Default).bind()
-          val f2 = av.put(20).startFiber(Dispatchers.Default).bind()
-          val f3 = av.put(30).startFiber(Dispatchers.Default).bind()
+          val f1 = av.put(10).fix().startFiber(Dispatchers.Default).bind()
```
raulraja: not sure why we need `fix()` if `startFiber` is for all `Kind<F, A> : Concurrent<F>`

modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/MVarTest.kt
```diff
@@ -78,9 +78,9 @@ class MVarTest : UnitSpec() {
...
+          val f2 = av.put(20).fix().startFiber(Dispatchers.Default).bind()
```
raulraja: Ideally, when we change these we also change them to use `!`

modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/IOTest.kt
nomisRev: 6 always finishes last in this test.

```
fun makePar(num: Long) =
  IO(newSingleThreadContext("$num")) {
    // Sleep according to my number
    Thread.sleep(num * 20)
    num
  }
```

pakoito: Put it on a comment IMO.

nomisRev:
@raulraja I wanted to postpone to do this in 1 all in one go during #1435 fx on type classes PR.

Rewriting felt like a waste since the fx API is going to change, and we can rewrite them with the full power of the DSL (without passing CC / Concurrent everywhere).

PR #1449, opened 2019-05-19
Concurrent - sleep & waitFor

nomisRev:
This PR will add an efficient `sleep` implementation to `Concurrent` and some derived functions.

We need to shift immediately after we get called by `ScheduledExecutorService`, this is currently impossible in `IO` since we don't have Arrow's default dispatcher in the arrow-effects-data module. Given that the `sleep`'s scheduler is internal we can't add an extension in `arrow-effects-io-extensions`.


modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
```diff
@@ -722,6 +724,87 @@ interface Concurrent<F> : Async<F> {
...
+   *     val fallbackWorld = delay { println("Hello from the backup") }
+   *     return world.timeoutTo(fallbackWorld, 1.seconds)
+   *   }
+   *   //sampleEnd
+   *   IO.unsafeRunBlocking(IO.concurrent().timedOutWorld())
+   * }
+   * ```
+   * @see timeout
+   **/
+  fun <A> Kind<F, A>.timeoutTo(default: Kind<F, A>, duration: Duration): Kind<F, A> =
```
raulraja: Is this a better name?
```kotlin
fun <A> Kind<F, A>.waitFor(duration: Duration, otherwise: Kind<F, A>): Kind<F, A>

IO { expensive() }.waitFor(4.seconds, otherwise = IO { cached() })
```

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
```diff
@@ -722,6 +724,87 @@ interface Concurrent<F> : Async<F> {
...
+   *
+   * ```kotlin:ank:playground
+   * import arrow.effects.*
+   * import arrow.effects.extensions.io.concurrent.concurrent
+   * import arrow.effects.typeclasses.seconds
+   *
+   * fun main(args: Array<String>) {
+   *   //sampleStart
+   *   fun <F> Concurrent<F>.timedOutWorld(): Kind<F, Unit> {
+   *     val world = sleep(3.seconds).flatMap { delay { println("Hello World!") } }
```
raulraja: should not use `delay` with println

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/internal/ConcurrentSleep.kt
```diff
@@ -0,0 +1,51 @@
...
+import arrow.effects.typeclasses.Duration
+import arrow.effects.typeclasses.mapUnit
+import java.util.concurrent.Executors
+import java.util.concurrent.ScheduledExecutorService
+import kotlin.coroutines.Continuation
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.startCoroutine
+
+@Suppress("FunctionName")
+fun <F> Concurrent<F>.ConcurrentSleep(duration: Duration): Kind<F, Unit> = cancelable { cb ->
```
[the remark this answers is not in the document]
nomisRev: It should, it'll become `expect/actual` when we do MPP.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
nomisRev: What about replacing both `timeout` and `timeoutTo` with the following signature?

```
fun <A> Kind<F, A>.waitFor(duration: Duration, otherwise: Kind<F, A> = just(TimeoutException)): Kind<F, A>
```

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
nomisRev: `effect` is not yet in `Async`/`Concurrent`.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
raulraja: looks great

pakoito: Not a fan of defaults in interfaces. They cannot be overriden.

PR #1451, opened 2019-05-20
IO effect in runloop

nomisRev:


modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/FxSyntax.kt
```diff
@@ -40,39 +36,46 @@ interface FxSyntax<F> : Concurrent<F>, BindSyntax<F> {
...
-    { ka, kb -> suspend { this(ka.bind(), kb.bind()) }.effect() }
+  fun <A, B> (suspend (A) -> B).effect(): (Kind<F, A>) -> Kind<F, B> = { fa ->
+    fa.flatMap { a ->
+      effect { this(a) }
+    }
+  }
 
-  fun <A, B, C, D> (suspend (A, B, C) -> D).effect(): (Kind<F, A>, Kind<F, B>, Kind<F, C>) -> Kind<F, D> =
-    { ka, kb, kc -> suspend { this(ka.bind(), kb.bind(), kc.bind()) }.effect() }
+  fun <A, B, C> (suspend (A, B) -> C).effect(): (Kind<F, A>, Kind<F, B>) -> Kind<F, C> = { ka, kb ->
```
raulraja: We need to check if we need these as they may be too esoteric for the public api

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Async.kt
```diff
@@ -146,6 +149,33 @@ interface Async<F> : MonadDefer<F> {
...
+   *   fun <F> Async<F>.invokeOnDefaultDispatcher(): Kind<F, String> =
+   *     _effect_(Dispatchers.Default, { getThreadSuspended() })
+   *
+   *   val result = _extensionFactory_.invokeOnDefaultDispatcher()
+   *   //sampleEnd
+   *   println(result)
+   * }
+   * ```
+   */
+  fun <A> effect(ctx: CoroutineContext = EmptyCoroutineContext, f: suspend () -> A): Kind<F, A> =
```
pakoito: I wouldn't add defaults in interfaces, if I did on my PR it was a mistake.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/FxSyntax.kt
```diff
@@ -40,39 +36,46 @@ interface FxSyntax<F> : Concurrent<F>, BindSyntax<F> {
...
-
-  fun <A> (suspend () -> A).effect(unit: Unit = Unit): Kind<F, A> = effect(this)
+  fun <A> (suspend () -> A).effect(unit: Unit = Unit): Kind<F, A> = effect(f = this)
 
-  fun <A, B> (suspend (A) -> B).effect(): (Kind<F, A>) -> Kind<F, B> =
-    { suspend { this(it.bind()) }.effect() }
-
-  fun <A, B, C> (suspend (A, B) -> C).effect(): (Kind<F, A>, Kind<F, B>) -> Kind<F, C> =
-    { ka, kb -> suspend { this(ka.bind(), kb.bind()) }.effect() }
+  fun <A, B> (suspend (A) -> B).effect(): (Kind<F, A>) -> Kind<F, B> = { fa ->
```
pakoito: Currying isn't really a thing in Kotlin. Pass Kind<F, A> as the first argument.

nomisRev:
@pakoito no default params is that a rule we want to enforce in our codebase type classes?

i.e.
```
fun <A> Kind<F, A>.fork(ctx: CoroutineContext = dispatchers().default())
fun <A> Kind<F, A>.waitFor(duration: Duration, default: Kind<F, A> = raiseError(TimeoutException)): Kind<F, A>`
```

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/FxSyntax.kt
nomisRev: I think we need to evaluate all the functions in this file. I'd like to get rid of `BindSyntax` and `FxSyntax` completely. Being in scope of `Concurrent` should give you the APIs as `fx.concurrent { }` IMO.

pakoito:
Yes because implementors cannot replace that parameter, so it enforces semantics that may not be what you'd like.

raulraja:
@pakoito can this be merged?

nomisRev:
> I'd also make effect with CoroutineContext = EmptyCoroutineContext parameter the invoke constructor for IO.

I am in favor of doing a 1:1 copy of the type class API for educational reasons and it'll make it easier for people to transition to tagless later on.

Let's address in a separate PR if necessary. I purposely left renaming `delay` and deprecating stuff out this PR since it's already addressed in #1439.

PR #1453, opened 2019-05-24
Add .tupled()/.untupled() to function syntax

glamm:
Changes:

* Adds .tupled() to function syntax, which converts a n-arity
  function to a function taking a single n-tuple argument.
* Adds .untupled() which does the reverse (converts a function
  taking a single n-tuple argument into a n-arity function.

Motivation:

Without this change, applicative instance methods must take
a lambda, even if the method signature appears to match, since
the instance methods pass all arguments as a single tuple:

```
fun concat(a: String, b: String) = a + b

// yields Some("foobar")
// lambda is only used to destructure the Tuple2
Option.applicative().map("foo".some(), "bar".some(), { (a, b) -> concat(a, b) })
```

With this change, a modified function reference can be used instead:

```
fun concat(a: String, b: String) = a + b

// yields Some("foobar")
// no need to use lambda since we can just modify the function reference
Option.applicative().map("foo".some(), "bar".some(), ::concat.tupled())
```


pakoito:
Looks good to me, although the build has failed and I cannot tell why. Let me retry it and we'll see.

Issue #1137, opened 2018-11-18
IOTest flaky 


LordRaydenMK:
When running `./gradlew clean test` two test for `IOTest` fail (I attached the results). After running multiple times, sometimes only one of those tests fail.

The tests are:
IOTest. parallel execution makes all IOs start at the same time
IOTest. parallel execution preserves order for synchronous IOs

When running test from the `IOTest` class ony (from IntelliJ) they run just fine.

I am using: IntelliJ IDEA 2018.2.6 (Community Edition)
Build #IC-182.5107.16, built on November 6, 2018
JRE: 1.8.0_152-release-1248-b22 amd64
JVM: OpenJDK 64-Bit Server VM by JetBrains s.r.o
Linux 4.4.0-139-generic

Let me know if you need more information.

[test results.zip](https://github.com/arrow-kt/arrow/files/2592976/test.results.zip)

pakoito:
It is the concurrency all starts at the same level one?

LordRaydenMK:
I updated the issue with the name of the tests. Maybe someone else will figure it out.

pakoito:
If you check the test, it uses threads per execution context and one to start them all. It'd be good if you could please run the tests with `-i`, or use a try/catch to print the complete exception so we can understand it better, given that you have consistent repro :D

LordRaydenMK:
I executed `./gradlew clean test -i > tests.txt` and here is the result.
[tests.txt](https://github.com/arrow-kt/arrow/files/2593250/tests.txt)

Then I tied running `./gradlew clean test -i --stacktrace > tests2.txt` and the result is: 
[tests2.txt](https://github.com/arrow-kt/arrow/files/2593260/tests2.txt)

```
* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':arrow-effects-instances:test'.
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.executeActions(ExecuteActionsTaskExecuter.java:110)
        at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter.execute(ExecuteActionsTaskExecuter.java:77)
        at org.gradle.api.internal.tasks.execution.OutputDirectoryCreatingTaskExecuter.execute(OutputDirectoryCreatingTaskExecuter.java:51)
        at org.gradle.api.internal.tasks.execution.SkipUpToDateTaskExecuter.execute(SkipUpToDateTaskExecuter.java:59)
[...]

PR #1456, opened 2019-05-24
Enable project parallelism for Gradle

mkobit:
-----

I saw a pretty significant speed up locally from this, time to see see if it applies to CI as well.


pakoito:
Does this affect tests? We have to avoid parallelism in tests.

pakoito:
See #1137

mkobit:
It should not impact tests in each individual project. If the tests in an individual project run one-at-time (or whatever order they do today), then it will continue to do that. This change should make independent projects build in parallel, so hopefully get multiple test tasks of those projects running at the same time. However, surprisingly to me, the build at https://travis-ci.org/arrow-kt/arrow/builds/536896884 was `29 min 13 sec` which doesn't seem much faster than I would have expected.

I did actually see that issue in `IOTest` earlier before this change, so I'll also follow up on https://github.com/arrow-kt/arrow/issues/1137.

aballano:
It seems that it decreased the CI check from 33-34min to 31 (~8.8%) 👍 Let me re-run the check some more times to check that there's no flakiness due to the changes and also to measure the time a bit more 👌

pakoito:
https://travis-ci.org/arrow-kt/arrow/jobs/536896885#L2946-L2947

nomisRev:
@pakoito given that it's green and I could not reproduce locally should we merge?

pakoito:
go for it

PR #1459, opened 2019-05-26
Improve racing APIs

nomisRev:
Currently the race results are modeled as nested `Either`s, which means that if you try to call `fold` it'll use the instance method instead of the `fold` extension function.

The only solution AFAIK is defining actual models with a `fold` method.

One quirk is that you cannot ignore multi-param lambdas and it become mandatory to do
`_, _, _  -> raiseError(..)` but only affects `racePair` and `raceTriple`.

### DIFF
--- a/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/RaceModels.kt
+++ b/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/RaceModels.kt
@@ -1,0 +1,202 @@
+package arrow.effects
+
+import arrow.core.Either
+import arrow.core.Tuple2
+import arrow.effects.typeclasses.Fiber
+
+/** Alias for `Either` structure to provide consistent signature for race methods. */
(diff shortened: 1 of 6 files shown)


modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/RaceModels.kt
```diff
@@ -0,0 +1,212 @@
...
+
+import arrow.core.Either
+import arrow.effects.typeclasses.Fiber
+
+/** Alias for `Either` structure to provide consistent signature for race methods. */
+sealed class RacePair<F, A, B> {
+  data class First<F, A, B>(val winner: A, val fiberB: Fiber<F, B>) : RacePair<F, A, B>()
+  data class Second<F, A, B>(val fiberA: Fiber<F, A>, val winner: B) : RacePair<F, A, B>()
+
+  fun <C> fold(
```
pakoito: inline

PR #1464, opened 2019-05-27
Introduce Timer

nomisRev:
After some more thought I propose to introduce `Timer` as a separate typeclass. Atm you can get it for free from `Concurrent`, or you can provide a custom instance (see RxJava/Reactor).

Reasoning:
 * Data types will be able to provide `Timer` (`sleep`) without providing `Concurrent`.
 * Sleeping becomes more explicit in the code, and it offers an alternative to overwriting schedulers to deal with `sleep` in tests.

Additionally `Timer` can be derived from `Concurrent` so we can provide a 0 overhead solution in `Concurrent` and thus for the `fx` DSL with the possibility for users to very easily replace the `Timer` instance for testing.


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/TimerLaws.kt
```diff
@@ -0,0 +1,56 @@
...
+      Law("Timer Laws: sleep should last specified time") { AS.sleepShouldLastSpecifiedTime(T, Clock(AS), EQ) },
+      Law("Timer Laws: negative sleep should be immediate") { AS.negativeSleepShouldBeImmediate(T, EQ) }
+    )
+
+  fun <F> Async<F>.sleepShouldLastSpecifiedTime(
+    T: Timer<F>,
+    C: Clock<F>,
+    EQ: Eq<Kind<F, Boolean>>
+  ) {
+    val lhs = binding {
```
pakoito: I'd test it multiple times to prevent issues. `forFew(10` or something like that.

PR #1465, opened 2019-05-28
Fx type class hierarchy and a la carte binding strategies

raulraja:
This PR is a joint effort with @nomisRev 

- Removes the current Fx hierarchy in favor of merging it in the current type class hierarchy.

`fx` is now part of the Effect type class hierarchy exposed at different levels (Monad, MonadError, Concurrent, etc.) and accessible via companions to avoid naming collisions. Each data type exports only its most powerful effect by default over its companion but you can always get to less powerful ones via type class instance access.

```kotlin
import arrow.effects.extensions.fx

IO.fx { //Receiver is `Concurrent`

}
```
```kotlin
import arrow.typeclasses.Monad
import arrow.effects.typeclasses.Concurrent

fun <F> Monad<F>.example() = fxMonad {  //Receiver is MonadSyntax /* Monad + BindSyntax */
}

fun <F> Concurrent<F>.example() = fxConcurrent {  //Receiver is ConcurrentSyntax /* Concurrent + BindSyntax */
}
```

- Type class instances that do not require construction arguments or include abstract dependencies are now cached as a `val` at the package level as part of the `ExtensionGenerator` codegen.

```kotlin
@PublishedApi()
internal val functor_singleton: OptionFunctor = object : arrow.core.extensions.OptionFunctor {}
inline fun Companion.functor(): OptionFunctor = functor_singleton
```


JorgeCastilloPrz:
Is this in progress or can be reviewed? Maybe we could start using labels to be explicit on that, or drafts or something

nomisRev:
@JorgeCastilloPrz this can be reviewed. Currently fixing build.

modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/function0.kt
```diff
@@ -92,6 +94,16 @@ interface Function0Monad : Monad<ForFunction0> {
 
   override fun <A, B> Function0Of<Either<A, B>>.select(f: Kind<ForFunction0, (A) -> B>): Kind<ForFunction0, B> =
     fix().fun0Select(f)
+
+  override suspend fun <A> MonadContinuation<ForFunction0, *>.bindStrategy(fa: Function0Of<A>): BindingStrategy<ForFunction0, A> =
+    BindingStrategy.Strict(fa.fix()())
+
+  override val fx: MonadFx<ForFunction0>
+    get() = object : MonadFx<ForFunction0> {
```
pakoito: This will create a new instance on every call. Is that what we want? I believe we should be catching it instead, use something like by lazy or having the object on the companion.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
```diff
@@ -31,24 +25,13 @@ object MonadDeferLaws {
...
-      Law("MonadDefer bind: binding failure") { SC.asyncBindError(EQERR) },
-      Law("MonadDefer bind: unsafe binding") { SC.asyncBindUnsafe(EQ) },
-      Law("MonadDefer bind: unsafe binding failure") { SC.asyncBindUnsafeError(EQERR) },
-      Law("MonadDefer bind: binding in parallel") { SC.asyncParallelBind(EQ) },
-      Law("MonadDefer bind: binding cancellation before flatMap") { SC.asyncCancellationBefore(EQ) },
-      Law("MonadDefer bind: binding cancellation after flatMap") { SC.asyncCancellationAfter(EQ) },
-      Law("MonadDefer bind: bindingInContext cancellation before flatMap") { SC.inContextCancellationBefore(EQ) },
-      Law("MonadDefer bind: bindingInContext cancellation after flatMap") { SC.inContextCancellationAfter(EQ) },
-      Law("MonadDefer bind: bindingInContext throw equivalent to raiseError") { SC.inContextErrorThrow(EQERR) },
-      Law("MonadDefer bind: monad comprehensions binding in other threads equivalence") { SC.monadComprehensionsBindInContextEquivalent(EQ) },
```
pakoito: Where did these go?

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadError.kt
```diff
@@ -128,20 +128,36 @@ interface MonadThrow<F> : MonadError<F, Throwable> {
...
-    val continuation = MonadErrorContinuation<F, B>(this)
-    val wrapReturn: suspend MonadErrorContinuation<F, *>.() -> Kind<F, B> = { just(c()) }
-    wrapReturn.startCoroutine(continuation, continuation)
-    return continuation.returnedMonad()
-  }
+  fun <B> bindingCatch(c: suspend MonadThrowContinuation<F, *>.() -> B): Kind<F, B> =
+    fx.monadThrow(c)
 
+  @Deprecated(
+    "`bindingCatch` is getting renamed to `fx` for consistency with the Arrow Fx system. Use the Fx extensions for comprehensions",
```
pakoito: `binding` is getting renamed...

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadError.kt
```diff
@@ -128,20 +128,36 @@ interface MonadThrow<F> : MonadError<F, Throwable> {
...
    */
+  override val fx: MonadThrowFx<F>
+    get() = object : MonadThrowFx<F> {
+      override val ME: MonadThrow<F> = this@MonadThrow
+    }
+
   @Deprecated(
     "`bindingCatch` is getting renamed to `fx` for consistency with the Arrow Fx system. Use the Fx extensions for comprehensions",
-    ReplaceWith("fx")
+    ReplaceWith("fx.monadThrow")
```
pakoito: "fx.monadThrow(c)" is the right syntax for IJ to do the right replacement.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadError.kt
```diff
@@ -128,20 +128,36 @@ interface MonadThrow<F> : MonadError<F, Throwable> {
...
-    val wrapReturn: suspend MonadErrorContinuation<F, *>.() -> Kind<F, B> = { just(c()) }
-    wrapReturn.startCoroutine(continuation, continuation)
-    return continuation.returnedMonad()
-  }
+  fun <B> bindingCatch(c: suspend MonadThrowContinuation<F, *>.() -> B): Kind<F, B> =
+    fx.monadThrow(c)
 
+  @Deprecated(
+    "`bindingCatch` is getting renamed to `fx` for consistency with the Arrow Fx system. Use the Fx extensions for comprehensions",
+    ReplaceWith("fx.monad")
```
pakoito: Same as above for the replacement syntax

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -123,6 +126,12 @@ sealed class IO<out A> : IOOf<A> {
     /* For parMap, look into IOParallel */
   }
 
+  suspend fun suspended(): A = suspendCoroutine { cont ->
```
pakoito: `await` fits better here

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadError.kt
```diff
@@ -128,20 +128,36 @@ interface MonadThrow<F> : MonadError<F, Throwable> {
...
+    fx.monad(c)
 
   fun <A> Throwable.raiseNonFatal(): Kind<F, A> =
     if (NonFatal(this)) raiseError(this) else throw this
 }
+
+interface MonadThrowFx<F> : MonadFx<F> {
+  val ME: MonadThrow<F>
+  override val M: Monad<F> get() = ME
+  fun <A> monadThrow(c: suspend MonadThrowContinuation<F, *>.() -> A): Kind<F, A> {
```
pakoito: This is a great chance to replace `MonadThrowContinuation` with a new set of interfaces that don't expose internal. Same for the other continuations.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/Concurrent.kt
```diff
@@ -777,15 +791,15 @@ interface Concurrent<F> : Async<F> {
...
+    val continuation = ConcurrentContinuation<F, B>(this)
+    val wrapReturn: suspend ConcurrentContinuation<F, *>.() -> Kind<F, B> = { just(c()) }
     wrapReturn.startCoroutine(continuation, continuation)
-    return continuation.returnedMonad() toT continuation.disposable()
+    return continuation.returnedMonad()
   }
 
   override fun <B> binding(c: suspend MonadContinuation<F, *>.() -> B): Kind<F, B> =
-    bindingCancellable { c() }.a
+    bindingConcurrent { c() }
```
pakoito: `bindingConcurrent(c)`

modules/effects/arrow-effects-extensions/src/main/kotlin/arrow/effects/extensions/io.kt
```diff
@@ -84,6 +86,10 @@ interface IOMonad : Monad<ForIO> {
 
   override fun <A> just(a: A): IO<A> =
     IO.just(a)
+
+  override suspend fun <A> MonadContinuation<ForIO, *>.bindStrategy(fa: Kind<ForIO, A>): BindingStrategy<ForIO, A> = BindingStrategy.Suspend {
+    fa.fix().suspended()
+  }
```
pakoito: `BindingStrategy.Suspend(fa.fix()::await)` after renaming suspended to `await`

pakoito: Isn't this strategy forcing the IO to be run to completion on every suspension point? I thought we wanted to do the `effect` approach instead.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
raulraja: These were there for all the `bind` overloads that were removed from `MonadDefer`. They are gone.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
raulraja: Wouldn't it be confused with `async/await`?

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadDeferLaws.kt
nomisRev: These functions were all invalid. They defined `Async` ops in `MonadDefer`.

The global cancelation is removed all together since it breaks RT.

nomisRev: I.e. `bindIn(ctx) {..}` becomes `effect(ctx) {..}.bind()` etc

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
nomisRev: Originally we had it as `val suspended` but that felt less ergonomic to use since you can just do `io::suspended` instead but not the other way around

modules/meta/arrow-meta/src/main/java/arrow/meta/encoder/jvm/StringTypeExtensions.kt
```diff
@@ -13,6 +13,9 @@ fun String.asPlatform(): String =
     .replaceFirst("kotlin.collections.", "java.util.")
     .replaceFirst("kotlin.", "java.lang.")
 
+/**
+ * TODO this is horrible is there a canonical way to obtain a kotlin type given a fqn java type name?
+ */
 fun String.asKotlin(): String =
```
pakoito: internal fun?

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadContinuations.kt
```diff
@@ -24,8 +24,12 @@ open class MonadContinuation<F, A>(M: Monad<F>, override val context: CoroutineC
     returnedMonad = value
   }
 
+  @Suppress("UNCHECKED_CAST")
   override fun resumeWithException(exception: Throwable) {
```
pakoito: Is this respected by inheritors?

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
pakoito: Maybe it collides with `await`, although it's semantically the same. In any case I'd like a verb as a name.

modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/function0.kt
nomisRev: Interfaces `by lazy` won't work here. For `* -> *` we can cache this on the companion, however.

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
nomisRev: I liked `val suspended: suspend () -> A` but that requires excessive wrapping, which I wanted to avoid this way.

The semantics with `Deferred#await` are not the same (well..), in Kotlinx they always talk about `await`ing eager async operations. That's why I am not a big fan of `await` but I agree a verb would be better.

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadError.kt
nomisRev: In the draft PR you suggest doing this in a separate PR, I'd prefer to give this a shot in a follow-up PR.

modules/effects/arrow-effects-extensions/src/main/kotlin/arrow/effects/extensions/io.kt
nomisRev: This is the definition of the `!` operator. I'm not sure what you mean with "I thought we wanted to do the effect approach instead."

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadContinuations.kt
nomisRev: `MonadThrow` overwrites it to capture the `Throwable` instead of throwing it.

```
  @Suppress("UNCHECKED_CAST")
  override fun resumeWithException(exception: Throwable) {
    returnedMonad = when (exception) {
      is ContinuationShortcircuitThrowable -> exception.exit as Kind<F, A>
      else -> raiseError(exception)
    }
  }
```

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadError.kt
```diff
@@ -160,4 +160,7 @@ interface MonadThrowFx<F> : MonadFx<F> {
     wrapReturn.startCoroutine(continuation, continuation)
     return continuation.returnedMonad()
   }
+
```
pakoito: There's a thing about this, whether we try/catch the block in Monad or not, and how this can change that behavior.

modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/either.kt
```diff
@@ -13,23 +13,7 @@ import arrow.core.Right
...
-import arrow.typeclasses.Hash
-import arrow.typeclasses.Monad
-import arrow.typeclasses.MonadContinuation
-import arrow.typeclasses.MonadError
-import arrow.typeclasses.Monoid
-import arrow.typeclasses.Semigroup
-import arrow.typeclasses.SemigroupK
-import arrow.typeclasses.Show
-import arrow.typeclasses.Traverse
+import arrow.typeclasses.*
```
pakoito: inbf4 detekt

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadContinuations.kt
```diff
@@ -71,6 +71,7 @@ open class MonadContinuation<F, A>(M: Monad<F>, override val context: CoroutineC
       is BindingStrategy.Strict -> strategy.a
       is BindingStrategy.ContinuationShortCircuit -> suspendCoroutineUninterceptedOrReturn { c ->
```
pakoito: Isn't there a simpler API for this? Without `orReturn`

nomisRev: It's used to short circuit the continuation. Could we keep track of it ourselves with a simple bool?

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadError.kt
nomisRev: Right. The problem I was trying to solve here was that if you overwrite `fx` in `Monad` and you overwrite `fx` in `MonadThrow` then the `fx.monad` from `Monad` gets lots.

This api seems dangerous. Should we just got for `fx`, `fxCatch`, `fxAsync` and `fxConcurrent`?

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadContinuations.kt
pakoito: I mean there should be another easier short in the API like `suspendCoroutineUnintercepted`

nomisRev: Nope, such an api doesn't exist there. You're supposed to return `COROUTINE_SUSPENDED`. Actually, what is the difference between the 2 examples below? 🤔 The docs didn't make me any wiser, and I haven't dug into the code yet.

```
suspendCoroutineUninterceptedOrReturn { c ->
  c.resumeWithException(throw RuntimeException("Boom!"))
  COROUTINE_SUSPENDED
}
```

and

```
suspendCoroutine { c ->
  c.resumeWithException(throw RuntimeException("Boom!"))
}
```

pakoito: Beats me. Keep it as it is I guess. There's no interception in Unintercepted, that much we know...probably.

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/observablek.kt
```diff
@@ -180,7 +180,7 @@ fun ObservableK.Companion.monadErrorSwitch(): ObservableKMonadError = object : O
 }
 
 // TODO ObservableK does not yet have a Concurrent instance
-fun <A> ObservableK.Companion.fx(c: suspend AsyncContinuation<ForObservableK, *>.() -> A): ObservableK<A> =
+fun <A> ObservableK.Companion.fx(c: suspend AsyncContext<ForObservableK>.() -> A): ObservableK<A> =
```
pakoito: ...I'd use MonadSyntax, AsyncSyntax, MonadThrowSyntax

Context is too loaded with Android terminology :P

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadError.kt
raulraja: I think every type should expose a single fx if that is an issue but multiple names is confusing

pakoito: My point was not to override it.

modules/effects/arrow-effects-rx2-extensions/src/main/kotlin/arrow/effects/rx2/extensions/observablek.kt
nomisRev: Fair enough :D

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/MonadError.kt
nomisRev: We can't override because of the receiver param?

PR #1466, opened 2019-05-28
Generalizes `firstOption` and promotes it to `Foldable`.

raulraja:
- [x] Generalizes `firstOption` making it part of the `Foldable` algebra.
- [x] Removes the need to pass the `Either.monad()` constrain by implementing `get` in terms of `foldLeft` and using `Either`'s API directly.
- [x] Deprecates all usages of `firstOption` in `arrow-syntax`

### DIFF
--- a/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Reducible.kt
+++ b/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Reducible.kt
@@ -2,7 +2,6 @@ package arrow.typeclasses
 
 import arrow.Kind
 import arrow.core.Eval
-import arrow.core.ForEither
 import arrow.core.None
 import arrow.core.Option
 import arrow.core.Some
--- a/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Reducible.kt
+++ b/modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Reducible.kt
@@ -143,11 +143,11 @@ interface NonEmptyReducible<F, G> : Reducible<F> {
     return 1 + tail.size(MN)
   }
 
-  override fun <A> Kind<F, A>.get(M: Monad<Kind<ForEither, A>>, idx: Long): Option<A> =
+  override fun <A> Kind<F, A>.get(idx: Long): Option<A> =
     if (idx == 0L)
       Some(this.split().a)
     else
-      FG().run { split().b.get(M, idx - 1L) }
+      FG().run { split().b.get(idx - 1L) }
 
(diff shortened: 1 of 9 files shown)


MarioAriasC:
It needs tests

raulraja:
@raulraja I came up with these two ad-hoc laws for these utilities. Let me know if you can think of any others.
```kotlin
fun <F> Foldable<F>.firstOptionReturnsNoneIfEmpty(cf: (Int) -> Kind<F, Int>) =
  forAll(Gen.int().map(cf)) { fa: Kind<F, Int> ->
    if (fa.isEmpty()) fa.firstOption().isEmpty()
    else fa.firstOption().isDefined()
  }

fun <F> Foldable<F>.firstOptionReturnsNoneIfPredicateFails(cf: (Int) -> Kind<F, Int>) =
  forAll(Gen.int().map(cf)) { fa: Kind<F, Int> ->
    fa.firstOption { false }.isEmpty()
}
```

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Reducible.kt
```diff
@@ -143,11 +143,11 @@ interface NonEmptyReducible<F, G> : Reducible<F> {
     return 1 + tail.size(MN)
   }
 
-  override fun <A> Kind<F, A>.get(M: Monad<Kind<ForEither, A>>, idx: Long): Option<A> =
```
pakoito: Neat.

nomisRev:
These operations already exist in Arrow-optics based on `Monoid`/`Option`. Defined 
 [firstOptionMonoid](https://github.com/arrow-kt/arrow/blob/e93c27d42340744d139d484596009b4548daf3dc/modules/optics/arrow-optics/src/main/kotlin/arrow/optics/predef.kt#L18) and [lastOptionMonoid](https://github.com/arrow-kt/arrow/blob/e93c27d42340744d139d484596009b4548daf3dc/modules/optics/arrow-optics/src/main/kotlin/arrow/optics/predef.kt#L26), you can use this in combination with `fun <A> Kind<F, A>.fold(MN: Monoid<A>): A` from `Foldable`.

Given that these are probably not very popular `Monoid` instance, are these instances we want to lift up from optics and expose?

raulraja:
@nomisRev most likely since we want to aim for a single API unless they bring value to Optics as they stand today

nomisRev:
@raulraja it's used in `Optics` to define `Fold#headOption` and `Fold#lastOption` based on `Fold#fold`. So lifting it up to `Foldable` and using them to define the same operators there would make sense and would mean the instances get re-used for `Foldable` & it's Optic version `Fold`.

raulraja:
@nomisRev sounds good

PR #1472, opened 2019-06-04
Deprecate Monad.*effect* methods without suspensions

glamm:
* Deprecated `Monad.effectM`, `Monad.forEffect`, and `Monad.forEffectEval` in
favor of `Monad.flatTap`, `Monad.productL`, and `Monad.productLEval`,
respectively.  Rationale:
    * Methods named with `effect` have semantic meaning in suspension
functions in Arrow;
    * The names match the corresponding method names in `cats`.


modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Monad.kt
```diff
@@ -48,15 +48,36 @@ interface Monad<F> : Selective<F> {
...
     flatMap { fb.value() }
 
+  @Deprecated(
+    "effectM is being renamed to flatTap",
+    ReplaceWith("flatTap")
+  )
   fun <A, B> Kind<F, A>.effectM(f: (A) -> Kind<F, B>): Kind<F, A> =
+    flatTap(f)
+
+  fun <A, B> Kind<F, A>.flatTap(f: (A) -> Kind<F, B>): Kind<F, A> =
```
pakoito: Where does `flatTap` come from?

nomisRev: It seems they have moved to the this naming model in Cats [FlatMap](https://github.com/typelevel/cats/blob/f59a9b0d250c90bca85e6d3a5fdfcc076be453cb/core/src/main/scala/cats/FlatMap.scala#L150).

pakoito: I don't like it but I guess it'll pass >.<

nomisRev: Can't find a similar signature on Hoogle. I remember seeing `flatTap` in other libs as well but I'm drawing a blank right now.

raulraja: the rationale here is that `effect` is associated to suspend so it should not be used in places where there is no suspension or it will become confusing to users

modules/core/arrow-typeclasses/src/main/kotlin/arrow/typeclasses/Monad.kt
```diff
@@ -48,15 +48,36 @@ interface Monad<F> : Selective<F> {
...
     flatMap { a -> f(a).map { a } }
 
-  fun <A, B> Kind<F, A>.forEffect(fb: Kind<F, B>): Kind<F, A> =
+  fun <A, B> Kind<F, A>.productL(fb: Kind<F, B>): Kind<F, A> =
     flatMap { a -> fb.map { a } }
 
-  fun <A, B> Kind<F, A>.forEffectEval(fb: Eval<Kind<F, B>>): Kind<F, A> =
+  @Deprecated(
+    "forEffect is being renamed to productL",
+    ReplaceWith("productL")
```
raulraja: is this the proper syntax for replacements?

glamm: It's consistent with `@Deprecated`/`ReplaceWith` as used by elsewhere in Arrow.

pakoito: Which is wrong lol

glamm: As near as I can tell from the Kotlin docs the right syntax is `ReplaceWith("productL(fb)")`, does that look right?

pakoito: yeah!

raulraja:
thanks @glamm @glammr1 !

PR #1474, opened 2019-06-04
Move the Documentation for `Id` from README.md to the Id.kt

maruiz81:
The Playground has been fixed. And also the existing links to id from other pages.


maruiz81:
Thanks guys it was a pleasure!

Issue #1414, opened 2019-04-18
parTraverse isn't always executing in parallel


pakoito:
In this line https://github.com/arrow-kt/arrow/blob/ce539f2c807683eb937190dfcdb76ddd2db75cad/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/suspended/FxSyntax.kt#L35 the `fa.map(f)` happens eagerly if the Kind isn't suspended.

pakoito:
I've checked, it turns out NonBlocking uses the same thread for small payloads and it may look sequential

```
arrow.effects.IOTest > bla STANDARD_OUT
    ForkJoinPool-1-worker-1
    ForkJoinPool-1-worker-1
    ForkJoinPool-1-worker-1
    ForkJoinPool-1-worker-1
    ForkJoinPool-1-worker-2
    ForkJoinPool-1-worker-3
    ForkJoinPool-1-worker-3
    ForkJoinPool-1-worker-1
    ForkJoinPool-1-worker-1
    ForkJoinPool-1-worker-1
    ForkJoinPool-1-worker-3
    ForkJoinPool-1-worker-4
    ForkJoinPool-1-worker-4
```

alexd6631:
Hello @pakoito,

I have made some tests with Arrow-Fx and tried to run some workload in parallel with parSequence/parTraverse but, in my case, it does not seem to run in parallel. 
The sample workload is implemented using coroutine delay, as I think it is representative of an non-blocking HTTP call. Same behavior is observed by switching to regular blocking Thread.sleep() calls.

In the code below, I also made a quick implementation of parSequence which starts a set of fibers and  join the results, achieving parallelism (~1sec to execute them all). However, Arrow parSequence seems to execute the tasks in a sequential, taking ~5secs to execute them all.

BTW, thank you for the great work on Arrow and Fx, it is great to have a principled effect system while having a lightweight and naturak Kotlin syntax.

```
package org.alexd.testfx

import arrow.effects.IO
import arrow.effects.extensions.io.fx.fx
import arrow.effects.fix
import kotlinx.coroutines.delay
import kotlin.coroutines.CoroutineContext

suspend fun putStrLn(s: String) { println(s) }

suspend fun getTime() = System.currentTimeMillis()

suspend fun work(name: String): String {
    println("Starting $name from ${Thread.currentThread()}")
    delay(1000)
    println("Done $name")
    return name
}

[...]

PR #450, opened 2017-11-07
Fix table + social meta

israelperezglez:
@calvellido could you review this PR please? Thanks

PR #405, opened 2017-10-13
Rx2 traverse and foldable instances

nomisRev:
closes #291

PR #1477, opened 2019-06-08
Fix FxSyntax parallelism issue

alexd6631:
This fix a issue with current FxSyntax.parTraverse implementation, where effects are run in sequential fashion, instead of being run in parallel. I left a comment in #1414 providing a way to reproduce the issue. The github issue was already closed, but the issue seems to be still present in 0.9.1-SNAPSHOT

It appears that in parTraverse (in case of IO), we were traversing a sequence of startFiber() immediatly "flatMapped" with join(), IO Applicative used by traverse (while representing mapping independant computations) does not perform IO in parallel (it is probably implemented in terms of flatMap).

This implementation perform the traversal twice, one for starting all fibers, one for joining them all, achieving real parallelism.

### DIFF
--- a/modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/EffectsSuspendDSLTests.kt
+++ b/modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/EffectsSuspendDSLTests.kt
@@ -336,7 +336,7 @@ class EffectsSuspendDSLTests : UnitSpec() {
 
     "List.parTraverse should run effects in parallel" {
       fxTest {
-        fx {
+        IO.fx {
           val timeline = mutableListOf<String>()
           val tasks = (1..3).map {
             effect {
--- a/modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/EffectsSuspendDSLTests.kt
+++ b/modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/EffectsSuspendDSLTests.kt
@@ -22,6 +22,7 @@ import io.kotlintest.runner.junit4.KotlinTestRunner
 import io.kotlintest.shouldBe
 import io.kotlintest.shouldThrow
 import kotlinx.coroutines.ObsoleteCoroutinesApi
+import kotlinx.coroutines.delay
 import kotlinx.coroutines.newSingleThreadContext
 import org.junit.runner.RunWith
 import java.util.concurrent.atomic.AtomicReference
@@ -331,6 +332,24 @@ class EffectsSuspendDSLTests : UnitSpec() {
           }
         }
       } shouldBe false
+    }
+
+    "List.parTraverse should run effects in parallel" {
+      fxTest {
+        fx {
+          val timeline = mutableListOf<String>()
+          val tasks = (1..3).map {
+            effect {
+              timeline.add("start $it")
+              delay(100L * it)
+              timeline.add("end $it")
+              it
+            }
+          }
+          !newSingleThreadContext("A").parTraverse(tasks, ::identity)
+          timeline
+        }
+      } shouldBe listOf("start 3", "start 2", "start 1", "end 1", "end 2", "end 3")
     }
 
     "FX supports polymorphism" {
--- a/modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/EffectsSuspendDSLTests.kt
+++ b/modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/EffectsSuspendDSLTests.kt
@@ -349,7 +349,7 @@ class EffectsSuspendDSLTests : UnitSpec() {
           !newSingleThreadContext("A").parTraverse(tasks, ::identity)
           timeline
         }
-      } shouldBe listOf("start 3", "start 2", "start 1", "end 1", "end 2", "end 3")
+      } shouldBe listOf("start 1", "start 2", "start 3", "end 1", "end 2", "end 3")
     }
 
     "FX supports polymorphism" {
--- a/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/FxSyntax.kt
+++ b/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/FxSyntax.kt
@@ -4,7 +4,6 @@ import arrow.Kind
 import arrow.core.Either
 import arrow.core.OptionOf
 import arrow.core.TryOf
-import arrow.core.extensions.id.applicative.just
 import arrow.core.identity
 import arrow.data.extensions.list.traverse.sequence
 import arrow.data.extensions.listk.traverse.traverse
@@ -32,7 +31,7 @@ interface FxSyntax<F> : Concurrent<F>, BindSyntax<F> {
       release = { fibers, exit -> when (exit) {
         ExitCase.Canceled -> fibers.traverse(this@FxSyntax) { it.cancel() }.map { Unit }
         else -> Unit.just()
-      }}
+      } }
     ).map { it.fix() }
 
   fun <A> CoroutineContext.parSequence(effects: Iterable<Kind<F, A>>): Kind<F, List<A>> =
--- a/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/FxSyntax.kt
+++ b/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/FxSyntax.kt
@@ -4,6 +4,7 @@ import arrow.Kind
 import arrow.core.Either
 import arrow.core.OptionOf
 import arrow.core.TryOf
+import arrow.core.extensions.id.applicative.just
 import arrow.core.identity
 import arrow.data.extensions.list.traverse.sequence
 import arrow.data.extensions.listk.traverse.traverse
@@ -26,9 +27,12 @@ interface FxSyntax<F> : Concurrent<F>, BindSyntax<F> {
   ): Kind<F, List<B>> =
     effects.fold(emptyList<Kind<F, Fiber<F, B>>>()) { acc, fa ->
       acc + startFiber(fa.map(f))
-    }.sequence(this@FxSyntax).bracket(
-      { fibers -> fibers.traverse(this@FxSyntax) { it.cancel() }.map { Unit } },
-      { fibers -> fibers.traverse(this@FxSyntax) { it.join() } }
+    }.sequence(this@FxSyntax).bracketCase(
+      use = { fibers -> fibers.traverse(this@FxSyntax) { it.join() } },
+      release = { fibers, exit -> when (exit) {
+        ExitCase.Canceled -> fibers.traverse(this@FxSyntax) { it.cancel() }.map { Unit }
+        else -> Unit.just()
+      }}
     ).map { it.fix() }
 
   fun <A> CoroutineContext.parSequence(effects: Iterable<Kind<F, A>>): Kind<F, List<A>> =
--- a/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/FxSyntax.kt
+++ b/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/FxSyntax.kt
@@ -5,7 +5,8 @@ import arrow.core.Either
 import arrow.core.OptionOf
 import arrow.core.TryOf
 import arrow.core.identity
-import arrow.data.extensions.list.traverse.traverse
+import arrow.data.extensions.list.traverse.sequence
+import arrow.data.extensions.listk.traverse.traverse
 import arrow.data.fix
 import arrow.typeclasses.ApplicativeError
 import arrow.typeclasses.Monad
@@ -25,8 +26,8 @@ interface FxSyntax<F> : Concurrent<F>, BindSyntax<F> {
   ): Kind<F, List<B>> =
     effects.fold(emptyList<Kind<F, Fiber<F, B>>>()) { acc, fa ->
       acc + startFiber(fa.map(f))
-    }.traverse(this@FxSyntax) { kind ->
-      kind.flatMap { it.join() }
+    }.sequence(this@FxSyntax).flatMap { fibers ->
+      fibers.traverse(this@FxSyntax) { it.join() }
     }.map { it.fix() }
 
   fun <A> CoroutineContext.parSequence(effects: Iterable<Kind<F, A>>): Kind<F, List<A>> =
--- a/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/FxSyntax.kt
+++ b/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/FxSyntax.kt
@@ -5,11 +5,14 @@ import arrow.core.Either
 import arrow.core.OptionOf
 import arrow.core.TryOf
 import arrow.core.identity
-import arrow.data.extensions.list.traverse.sequence
+import arrow.data.ListK
 import arrow.data.extensions.listk.traverse.traverse
 import arrow.data.fix
+import arrow.data.k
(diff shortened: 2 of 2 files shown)


pakoito:
First PR and a very ambitious one. Thank :D Simon will validate this, or if you could add a positive/negative test it'd be great.

alexd6631:
@pakoito Thanks for your feedback ! 
The tricky part of writing an automated test of parallelism/sequential behavior, is to avoid dependency to time (to avoid flaky tests, or slowing the tests with too conservative delays). Maybe I can come up with something using latches of some sorts.

I don't know how it could fit into Fx/IO but maybe having "test schedulers" à la Rx/Reactor would be practical for testing this kind of stuff.

In the meantime, there is a small snippet in #1414 that exhibit this behavior.

alexd6631:
I pushed a quick test to assert parallel behavior (I am not too fan about the delay(100), maybe it should be increased to avoid flaky tests on a overloaded CI ?), and tested it to fail with previous implementation. 

Doing so I remarked that tasks are started in a reverse order (this was previously the case), and wondered if this is the expected behavior, since parMapN variants are tested to preserve order for synchronous IO in IOTests.kt ?

modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/EffectsSuspendDSLTests.kt
```diff
@@ -332,6 +333,24 @@ class EffectsSuspendDSLTests : UnitSpec() {
...
+          val tasks = (0 until 3).map {
+            effect {
+              timeline.add("start $it")
+              delay(100)
+              timeline.add("end $it")
+              it
+            }
+          }
+          !newSingleThreadContext("A").parTraverse(tasks, ::identity)
+          timeline.toList()
```
pakoito: This is unnecessary as it doesn't modify the list in place.

modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/EffectsSuspendDSLTests.kt
```diff
@@ -332,6 +333,24 @@ class EffectsSuspendDSLTests : UnitSpec() {
...
     }
 
+    "List.parTraverse should run effects in parallel" {
+      fxTest {
+        fx {
+          val timeline = mutableListOf<String>()
+          val tasks = (0 until 3).map {
+            effect {
+              timeline.add("start $it")
+              delay(100)
```
pakoito: I'd multiply the delay by the position (or the inverse), so you know deterministically who's supposed to end first.

pakoito:
>Doing so I remarked that tasks are started in a reverse order (this was previously the case), and wondered if this is the expected behavio

The implementation of applicative is like that, it's not a concern.

nomisRev:
@alex6631 Hi, I've been looking into this since this, since the current implementation is not what we want, it's concrete to `Iterable`.It's should be polymorphic to ` G` with an alias.

The implementation you provided also doesn't correctly take into consideration cancelation.

I have some ideas for tests/laws using `Promise` as latches.

If you are interested on collaborating we can discuss it further once I have something working locally.

alexd6631:
Hi @nomisRev, I amended the tests after @pakoito comments.

Indeed, I did not wanted to change much to the actual implementation, and did not take special consideration to cancellation, assuming it would be somehow done by canceling the chain of fiber.join() flatmapped together by the applicative.

Feel free to come back later to me to discuss a better implementation.

alexd6631:
@nomisRev 
As of curiosity, I did implement some support for cancellation. This is done using bracket instead of flatMap after fibers creation.  At first I thought there could be a race condition if the whole effect is suspended too early, creating the fibers but never calling the release effect. 
But from my understanding of bracket, I think it is safe to do so, since fibers creation is not cancellable, and bracket wrapping the previous IO, the cancellation of all fibers should be always called. Feel free to correct me if I'm wrong.

This is not yet polymorphic regarding the container, but I could work on it (and rebase on top of the new hierarchy), if you are still interested in this pull request.

I also noticed, that the cancellation system does not seems, by default, cancel "natively" the underlying coroutine. (In my tests the coroutine delay call was not cancelled despite the fibers being cancelled). 
Correct me if am missing something, but I think it is something that should be at least documented.

nomisRev:
@alexd6631 the new implementation with `bracket` looks correct to me, and your understanding of it is correct.

I always love contributions so I don't want to stop you from figuring it out! If we can figure it out in this PR then awesome! I am myself still trying to figure out how to support correct cancellation semantics in a polymorphic way.

> I also noticed, that the cancellation system does not seems, by default, cancel "natively" the underlying coroutine.

AFAIK, there is no support for cancelling a coroutine in the Kotlin std. Cancellation to delay only works within Kotlinx IIRC, and using `delay` outside of kotlinx will result in the same behavior in or outside of Arrow. In the next release (already in master) Arrow will include our its own `Timer#sleep`, which is cancelable in a tagless fashion. But again there is no concept of cancelation on a language level, only on a library level atm.

If you have ideas on how to improve this then I'd love to hear them.

alexd6631:
Hi,

While testing the implementation with cancellation, I noticed that in the case of completed fibers, the systematic cancellation done by the bracket release block, prevent the result from being emitted (the downstream IO seems to be cancelled). I don't know if this is the expected behavior (we should avoid to cancel a completed fiber), or some side effects in the IO run loop.

As a workaround, I switched to bracketCase to cancel only in case of explicit cancellation from downstream. This is not fully satisfying, since if a fiber errors, we want other fiber to be cancelled, and it is not the case with this implementation.

Regarding a polymorphic version, I pushed some code in the fix-partraverse-parallelism-poly branch.
It uses Traverse as a functor to store the fibers in the same "container" type than the input, to later traverse them (joining/cancelling all the fibers). Do you think it would be problematic for some containers implementing Traverse ?

Regarding coroutine, you can control cancellation by providing custom CoroutineContext, one possible way is to use Job, here is a quick version of a coroutine-cancellable effect version :

 ```kotlin
import arrow.Kind
import arrow.effects.typeclasses.suspended.concurrent.Fx
import kotlinx.coroutines.Job
import kotlinx.coroutines.withContext

fun <F, A> Fx<F>.cancellableEffect(block: suspend () -> A): Kind<F, A> = fx {
  !effect { Job().also { it.start() } }.bracket(
    {j -> effect { j.cancel() }},
    {j -> effect {
      withContext(j) { block() }
    }}
  )
}
```

However, I am not too sure, from the API standpoint, if this should be done explicitly or automatically for every effect or fiber, but with Job some bridging can be done with IO.

nomisRev:
I am not sure I follow the first bit but "This is not fully satisfying, since if a fiber errors, we want other fiber to be cancelled, and it is not the case with this implementation." is not correct. The other fibers should be cancelled in case of cancelation or error. 

> Do you think it would be problematic for some containers implementing Traverse?
That could be problematic if you have nested cancelable effect types.

Including kotlinx.coroutines-core into arrow-effects is not an option. It has a very different goal as Arrow, and we don't want to expose its API from arrow-effects. Additionally, we try to keep all modules small, isolated and fine-grained.

alexd6631:
> I am not sure I follow the first part but "This is not fully satisfying, since if a fiber errors, we want other fiber to be cancelled, and it is not the case with this implementation." is not correct. The other fibers should be cancelled in case of cancelation or error.
> 

I think we are both saying the same thing, fibers *should* be cancelled in case of cancellation or errors, but the implementation in my latest commit, only handling the cancellation case is not fully  satisfying. I will need to look more into the IO loop implementation to understand the behavior described in my previous comment.

> > Do you think it would be problematic for some containers implementing Traverse?
> 
> Yes, that could be problematic if you have nested cancelable effect types. Although that'll most likely depend on the impl of `Traverse`. Also, nesting effect types are less than ideal.
> 
I think the sample implementation would work ok for Traverse backed by in-memory containers (Option, List, Tree, ...). Do you think it is good enough to limit to them (maybe by requiring an other typeclass) or is there use cases to support other traverse with effect (I personally would avoid nesting effects, since it would find hard to reason about what is happening when) ? 

> Including kotlinx.coroutines-core into arrow-effects is not an option. It has a very different goal as Arrow, and we don't want to expose its API from arrow-effects. Additionally, we try to keep all modules small, isolated and fine-grained.

I am not sure it is related to my commit (since the dependency to coroutines-core was already there)
or some more general reflexion for the future of arrow-effects ?

alexd6631:
Ok, I got the meaning of your latest remark, it was about coroutines cancellation, and Job being in kotlinx.coroutines and not in kotlin.coroutine.

I understand the need to keep the library dependencies to a minimum. Personnaly I would use arrow-fx to wrap and compose side effects in coroutines (for instance httprequests coming from a 3rd party) in a more principled fashion, so I would likely already have kotlinx.core in my dependencies it would not be a problem for me.

Maybe this can be provided as an extension to Fx in a separate module.
Another approach, which I am not sure is possible, would be to implement a custom CoroutineContext (ie reimplement Job equivalent in arrow-effects)

nomisRev:
It's for sure not enough to limit to in-memory containers as there would be no way to differentiate from it in a polymorphic way, unless we introduce typeclasses/constraints like you said which I don't think is the way to go. This problem is solved in [Cats-effects](https://github.com/typelevel/cats-effect) so we can solve it in a similar manner. `Concurrent` implies `Parallel` in `Cats-effects` that's why we decided to incorporate it directly there. I am going to dive deeper into next week, so I could give you more details then.

Concerning the cancellation and kotlinx. I think we should provide some integrations, this will be useful for anyone using kotlinx today. I haven't very deeply looked into it, some of the cancelation stuff seems wired into the dispatchers etc. So it would require some dedicated time to figure it out.

On another note I've been meaning to look into why `CancelableCoroutine` and `suspendCancelableCoroutine` are in kotlinx and not in the std because this would solve that issue in a more generic way for everyone.

alexd6631:
I guess that having a Parallel typeclass would allow to have applicative instances that really perform work in parallel (and in case of IO supports cancellation at the applicative level), so the parTraverse implementation would not need to take care of fibers/cancellation, just to compose parallel work together like a regular traverse.

https://github.com/typelevel/cats/blob/a4ef42468562022fcc63bafdaa196fb5e5426e4b/core/src/main/scala/cats/Parallel.scala#L129

It is true, that is a much better and generic approach that the provided implementation.
I don't know the impact of modifying the hierarchy and including it in arrow-fx, I guess this is what you are thinking/working on now.

So feel free to either drop this PR or use it as an "interim" version that will be removed when Parallel is available (the current version on master does not parallelize neither handle cancellation for IO, so it is better than nothing)

nomisRev:
@alexd6631 I am not sure we need `Parallel` in Arrow, and we definitely don't need it as a constraint for `Concurrent`. The resulting API in Cats-effects is a little quirky and AFAIK they're looking for an alternative [cats-par](https://github.com/ChristopherDavenport/cats-par), [Cats-effects #405](https://github.com/typelevel/cats-effect/issues/405), [Cats-effects #450](https://github.com/typelevel/cats-effect/issues/450).

Some background on [Parallel](https://gitter.im/typelevel/cats?at=5cf6771dfaac6439343b5173) in Cats and why it's there.

alexd6631:
@nomisRev Thanks for the links.

I don't think we should copy Parallel in its current encoding (especially the need for 2 types, which is alleviated by cats-par), but I think that having some way to have an alternative Applicative (no pun intended) used for parallel/cancellable computation looks promising to solve the current issue. 

I don't have a precise idea for the encoding, maybe this applicative could be provided by the Fx instance (or its hierarchiy) of a given type, and used by parallel methods. 

I don't have enough background on Arrow/Cats-Effects so there may be a better approch, or an approach that does not require Parallel.

Looking forward to hear from you.

alexd6631:
After reading your links and pondering on the subject, I think I understand better your point of view in your last comment. Since Concurrent can implement parMap solely in term of race or fibers, there is no need for Parallel to implement such operations. 

So I couldn't resist to try the following implementation, since Concurrent offers parMapN (given a CoroutineContext) it is possible to implement a "parallel" applicative on top of it, and traverse with it through any Traverse. In theory this should create a nested parMap chain, which should support cancellation (a bit like high arity parMapN are implemented on top of 2-3 versions). The implementation works for nominal case, but does not play well with cancellation. If you have time, I would be curious to have your opinion on why this approach would not work, maybe you have tried similar code on your side ?

Also, despite Concurrent can implement parallel computation on its own, I personally see some benefits in having some kind of typeclasses for parallel computation : 

- If there is a native parallel operation available for the underlying effect type, there would be an overhead to recreate it on top of race/fibers
- I may be mistaken here, but the implementation with fibers would work with single-valued effects but I am not sure it would work with multi-valued effects like Observable/Flux 
- It would make straightforward to implement the instance on top of Rx/Reactor zip family of operators.
- It would make the implementation of parTraverse straightforward.

If you have time, I'd like to hear your rationale/counter-arguments on this points.
Thanks

nomisRev:
@alexd6631 I will get back to you later this week when I had some time to investigate. I will keep you in the loop of any developments :)

nomisRev:
@alexd6631 Hi alex, this is what I came up [#1483](https://github.com/arrow-kt/arrow/pull/1483).

Like you already mentioned, and what was mentioned in the reference I shared we can derive a parallel`Applicative` from `parMapN`. I had to rewrite `parMapN` however to take some cancellation edge cases into account properly.

> If there is a native parallel operation available for the underlying effect type, there would be an overhead to recreate it on top of race/fibers

This is a non-issue with typeclasses as you can just provide an overwrite for the instance. As you can see in #1483, we overload the `parMapN` overload with the faster concrete version of `IO`.

> I may be mistaken here, but the implementation with fibers would work with single-valued effects but I am not sure it would work with multi-valued effects like Observable/Flux.

Those fibers support multi-value. There currently is a PR open ready to be merged [here](https://github.com/arrow-kt/arrow/pull/1427).

> It would make straightforward to implement the instance on top of Rx/Reactor zip family of operators.

Feedback on #1427 is welcome :) And the implementations for Reactor is still missing if you would like to give it a try that'd be awesome!

> It would make the implementation of parTraverse straightforward.

I agree, I like this encoding better than `Parallel` to be honest. We can still iterate on this further later if necessary.

If you have any thoughts on the implementation in #1483 I'd love to hear them.

alexd6631:
Hello @nomisRev,

Thanks for your feedback and your explanations.

I think we agreed on the fundamental idea to be able to provide some kind of parallel applicative with a default implementation for any Concurrent and have the possibility to override this implementation with a native/optimized version. (interestingly, having to implement parMapN, reminds me of Semigroupal encoding (https://github.com/typelevel/cats/blob/master/core/src/main/scala/cats/Semigroupal.scala), which is  in some way a non-currified version of Applicative)  

I'll take a look at the code more in depth this weekend. 
One thing I noticed is that the old "parTraverse" seems to be in FxSyntax, but I guess this is the WIP and will be removed later ?

I will also try to play with polymorphic Fx program, and see how it works with a different runtime. (If time allow me, I will give a try at implementing a Reactor runtime, which should be very similar to Rx).

I plan to write a technical blog article about Arrow-fx and a general introduction to effects system in the upcoming week. I'll be very happy to have your feedback before publishing it.

nomisRev:
Hey @alexd6631,

> I plan to write a technical blog article about Arrow-fx and a general introduction to effects system in the upcoming week. I'll be very happy to have your feedback before publishing it.

That's awesome! I would love to collaborate with you on that, I can guarantee that other Arrow maintainers are also excited about this and would love to help out 2 if you like.

> One thing I noticed is that the old "parTraverse" seems to be in FxSyntax, but I guess this is the WIP and will be removed later?

The `parTraverse` that ended up there was a bit ad-hoc also given that it was not fully polymorphic so that would be replaced by a more powerful concrete version yes.

It's indeed very similar to `Semigroupal` and we also need to overwrite `product` in `Applicative` to achieve a parallel instance. [reference](https://github.com/arrow-kt/arrow/pull/1483/files#diff-cf4adba06de0c8e637d114e3ba561b1e).

Looking forward to your contributions and if there is anything we hang out on the Kotlin slack in #arrow or #arrow-contributors.

PR #1478, opened 2019-06-08
Favor Either and IO instead of Try

pakoito:
This diff stabilizes a proposal that's been floating for a while. `Try` is a defective `Either` with a nice constructor. Let's promote using `Either` or `IO` instead.

### DIFF
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -189,6 +189,13 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
     }
 
     fun <L, R> cond(test: Boolean, ifTrue: () -> R, ifFalse: () -> L): Either<L, R> = if (test) right(ifTrue()) else left(ifFalse())
+
+    fun <R> catch(f: () -> R): Either<Throwable, R> =
+      try {
+        f().right()
+      } catch (t: Throwable) {
+        t.nonFatalOrThrow().left()
+      }
   }
 }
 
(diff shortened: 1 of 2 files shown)


modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -189,6 +189,16 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
     }
 
     fun <L, R> cond(test: Boolean, ifTrue: () -> R, ifFalse: () -> L): Either<L, R> = if (test) right(ifTrue()) else left(ifFalse())
+
+    fun <R> catch(f: () -> R): Either<Throwable, R> =
```
raulraja: This should be `suspend fun <R> catch...`

modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -189,6 +189,16 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
...
+      catch(::identity, f)
+
+    fun <L, R> catch(fe: (Throwable) -> L, f: () -> R): Either<L, R> =
```
raulraja: suspend

modules/core/arrow-core-data/src/main/kotlin/arrow/core/Try.kt
```diff
@@ -16,8 +16,16 @@ sealed class Try<out A> : TryOf<A> {
 
   companion object {
 
+    @Deprecated(
+      "Try promotes eager execution and works like Either, so it's better if you work with Either or a suspended type like IO",
+      ReplaceWith("Either.just(a)")
+    )
     fun <A> just(a: A): Try<A> = Success(a)
```
raulraja: This constructor is fine if we use suspend for `Try`'s effects

Issue #1475, opened 2019-06-05
["Request"] convenient Option.mapNotNull


StefanSchrass:
What version are you currently using?
0.9.0

What would you like to see?
I hope I am not missing out on something, but I would like to have a convenient `Option.mapNotNull`.

What I find myself doing often is 
```
Option.fromNullable(foo)
    .flatMap { Option.fromNullable(it.bar) }
    .flatMap { Option.fromNullable(it.baz) }
```
And what I would love having is something similar to
```
Option.fromNullable(foo)
    .mapNotNull { it.bar }
    .mapNotNull { it.baz }
```
So it kind of prevents `Some(null)` from happening and smartcasts from `T?` to `T`.

raulraja:
@StefanSchrass I feel your pain but I think the best solution is the other way around. We need to add `map`, `flatMap` and essentially any available combinator in Option as extension functions for nullable types for folks that would rather use the classic FP combinators over direct syntax:
```kotlin
fun <A, B> A?.fold(ifEmpty: () -> B, ifFull: (A) -> B): B = 
  this?.let(ifFull) ?: ifEmpty()
fun <A, B> A?.map(f: (A) -> B): B? = 
  this?.let(f)
fun <A, B> A?.flatMap(f: (A) -> B?): B? = 
  this?.let(f)
```
These are just stylistic because as you can observe their implementation body is simpler than the actual function signature. Kotlin has chosen to model 2 effects hardcoded in its type system. Option <-> ? and Effects <-> suspend

If you still want to pursue this I'll help review. Also if you have a more elaborate example that illustrates the pain more than foobar we can maybe suggest a better encoding for your problem.

nomisRev:
@raulraja I think the API @StefanSchrass is looking for is `fun <A, B> Option<A>.flatMap(f: (A) -> B?): Option<B> = flatMap { a -> Option.fromNullable(f(a)) }`, not an `API` for nullable types.

raulraja:
@nomisRev yes but I don't think we should add those to Option so I'm proposing an alternative so that he does not have to put those in Option in the first place if the only thing he is doing is to call `map` on them. That is why I asked for a more elaborate example since the `foo` `bar` does not make sense if all this code is doing is call `map` or `flatMap`.

StefanSchrass:
@raulraja yes indeed my motivation is a stylisitc one and I pretty much had in mind what @nomisRev wrote.

[...]

PR #1479, opened 2019-06-09
fixes ["Request"] convenient Option.mapNotNull #1475

StefanSchrass:
As discussed in https://github.com/arrow-kt/arrow/issues/1475 this is a proposed fix to the issue.
A convenience method was added `mapNotNull` with KDoc and a corresponding UnitTest.

### DIFF
--- a/modules/core/arrow-core-data/src/test/kotlin/arrow/core/OptionTest.kt
+++ b/modules/core/arrow-core-data/src/test/kotlin/arrow/core/OptionTest.kt
@@ -16,13 +16,7 @@ import arrow.mtl.extensions.option.traverseFilter.traverseFilter
 import arrow.syntax.collections.firstOption
 import arrow.test.UnitSpec
 import arrow.test.generators.option
-import arrow.test.laws.FunctorFilterLaws
-import arrow.test.laws.HashLaws
-import arrow.test.laws.MonadFilterLaws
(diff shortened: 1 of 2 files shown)


modules/core/arrow-core-data/src/test/kotlin/arrow/core/OptionTest.kt
```diff
@@ -16,13 +16,7 @@ import arrow.mtl.extensions.option.traverseFilter.traverseFilter
...
 import arrow.test.UnitSpec
 import arrow.test.generators.option
-import arrow.test.laws.FunctorFilterLaws
-import arrow.test.laws.HashLaws
-import arrow.test.laws.MonadFilterLaws
-import arrow.test.laws.MonoidLaws
-import arrow.test.laws.MonoidalLaws
-import arrow.test.laws.ShowLaws
-import arrow.test.laws.TraverseFilterLaws
+import arrow.test.laws.*
```
pakoito: Ktlint will flag this as an error. You need to not keep the *

StefanSchrass: Told my IDE to not "optimize imports"

StefanSchrass:
Glad that you find it worth adding! My pleasure!

PR #1480, opened 2019-06-11
Ank not fail fast

nomisRev:
Given that we have quite some docs already, and we're doing some refactorings it's becoming increasingly hard to maintain docs. Luckily Ank is there to save us!

To make the task of updating the docs less annoying I thought it'd be better if we collected all incorrect docs instead of failing fast on the first one. Saves a bunch of build time.

Given that there is not feature roadmap for Ank I just did the simplest thing possible.

### DIFF
--- a/modules/ank/arrow-ank/src/main/kotlin/arrow/ank/ank.kt
+++ b/modules/ank/arrow-ank/src/main/kotlin/arrow/ank/ank.kt
@@ -14,9 +14,7 @@ import arrow.data.combine
 import arrow.data.extensions.list.foldable.reduceLeftOption
 import arrow.data.extensions.list.semigroup.List.semigroup
 import arrow.data.extensions.nonemptylist.semigroup.semigroup
-import arrow.data.fix
 import arrow.data.validNel
-import arrow.effects.typeclasses.Async
 import arrow.effects.typeclasses.Concurrent
 import java.nio.file.Path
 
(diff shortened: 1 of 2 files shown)


modules/ank/arrow-ank/src/main/kotlin/arrow/ank/ank.kt
```diff
@@ -24,9 +40,9 @@ fun <F> Concurrent<F>.ank(source: Path, target: Path, compilerArgs: List<String>
     !effect { printConsole("Current heap used: ${(heapSize - Runtime.getRuntime().freeMemory()).humanBytes()}") }
     !effect { printConsole("Starting ank with Heap Size: ${heapSize.humanBytes()}, Max Heap Size: ${heapMaxSize.humanBytes()}") }
     val path = !effect { createTargetDirectory(source, target) }
-    val paths = !!effect {
+    val validatedPaths = !effect {
       path.ankFiles().map { (index, p) ->
-        suspend {
+        effect(suspend {
```
pakoito: is suspend even needed anymore? `effect {`

pakoito:
JFC yes! One of the best features ever.

modules/ank/arrow-ank/src/main/kotlin/arrow/ank/ank.kt
nomisRev: Thanks!

PR #1482, opened 2019-06-11
Bump to Kotlin 1.3.31

ogolberg:
See https://youtrack.jetbrains.com/issue/KT-31664

### DIFF
--- a/modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/EffectsSuspendDSLTests.kt
+++ b/modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/EffectsSuspendDSLTests.kt
@@ -280,7 +280,7 @@ class EffectsSuspendDSLTests : UnitSpec() {
             suspend { 1 },
             suspend { 2 },
             suspend { 3 }
-          ).traverse(::effectIdentity)
+          ).traverse { it }
         }
       } shouldBe listOf(1, 2, 3)
     }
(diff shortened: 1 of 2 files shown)


modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/EffectsSuspendDSLTests.kt
```diff
@@ -281,7 +281,7 @@ class EffectsSuspendDSLTests : UnitSpec() {
             suspend { 1 },
             suspend { 2 },
             suspend { 3 }
-          ).traverse(::effectIdentity)
```
pakoito: `::identity`?

ogolberg: `identity` doesn't work because it's not `suspend`

`effectIdentity` was giving me Screen Shot 2019-06-12 at 8 43 02 AM

modules/effects/arrow-effects-data/src/test/kotlin/arrow/effects/EffectsSuspendDSLTests.kt
```diff
@@ -281,7 +281,7 @@ class EffectsSuspendDSLTests : UnitSpec() {
...
+          ).traverse { it }
```
nomisRev:
```suggestion
          ).sequence()
```

pakoito:
That small nit was okay. Let's merge! Thanks :D

PR #1483, opened 2019-06-11
Concurrent parApplicative/parTraverse&co

nomisRev:
This PR will add a `parApplicative` factory to `Concurrent` and enables combinators such as `parTraverse`.

- [ ] Add laws


modules/docs/arrow-docs/src/main/kotlin/test.kt
```diff
@@ -0,0 +1,59 @@
+import arrow.data.ListK
```
raulraja: Do we need this file? 🔥

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/internal/ConcurrentParMap3.kt
```diff
@@ -0,0 +1,92 @@
+package arrow.effects.internal
+
+import arrow.Kind
+import arrow.effects.typeclasses.Concurrent
+import arrow.effects.typeclasses.ExitCase
+import kotlin.coroutines.CoroutineContext
+
+fun <F, A, B, C, D> Concurrent<F>.parMap3(ctx: CoroutineContext, fa: Kind<F, A>, fb: Kind<F, B>, fc: Kind<F, C>, f: (A, B, C) -> D): Kind<F, D> = ctx.run {
+  tupled(startFiber(fa), startFiber(fb), startFiber(fc)).bracketCase(use = { (fiberA, fiberB, fiberC) ->
```
raulraja: If you look at it sideways is even more tripier

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
```diff
@@ -72,7 +75,10 @@ object ConcurrentLaws {
       Law("Concurrent Laws: race cancels both") { CF.raceCancelCancelsBoth(EQ, ctx) },
       Law("Concurrent Laws: race is cancellable by participants") { CF.raceCanBeCancelledByParticipants(EQ, ctx) },
       Law("Concurrent Laws: parallel map cancels both") { CF.parMapCancelCancelsBoth(EQ, ctx) },
-      Law("Concurrent Laws: action concurrent with pure value is just action") { CF.actionConcurrentWithPureValueIsJustAction(EQ, ctx) }
+      Law("Concurrent Laws: action concurrent with pure value is just action") { CF.actionConcurrentWithPureValueIsJustAction(EQ, ctx) },
+      Law("Concurrent Laws: parTraverse can traverse effectfull computations") { CF.parTraverseCanTraverseEffectfullComputations(EQ) },
```
pakoito: Typo: efectful

alexd6631:
Hello @nomisRev,

Here some feedback on this PR.

As mentioned in the previous PR, I think the parTraverse implementation should be removed/replaced in FxSyntax.
https://github.com/arrow-kt/arrow/blob/2b158ba6f6d73dca906bd61f2d9e594aa772d614/modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/typeclasses/FxSyntax.kt#L22

It may be convenient to keep this specialized version in FxSyntax (to avoid having to wrap with ListK and passing a Traverse instance manually each time in user code) but implemented on top of the polymorphic version.

```kotlin
  fun <A, B> CoroutineContext.parTraverse(
    effects: Iterable<Kind<F, A>>,
    f: (A) -> B
  ): Kind<F, List<B>> =
    effects.toList().k().parTraverse(this, ListK.traverse()) { it.map(f) }.map { it.fix() }
```

I made some quick tests of parallel work and parallelism/error/cancellation seems to work as expected 👍 

One thing I noticed with parTraverseN, is that in a scenario where I have a lot of parallel tasks (50) with a low parallelism (2), when I simulate my first task failing after a few hundred mills, it takes a few seconds after the failure to cancel the whole chain and propagate the error (Sometimes I even get a UserCancellationException instead of the initial error). 
If I have enough parallelism the whole chain will be cancelled as expected after a few milliseconds.
I can post a snippet if the above description is not clear enough.

Alex

nomisRev:
Hi @alexd6631,

Thanks for the feedback. I think we agree on the initial design of `parTraverse` :)
I've added the overload for `Iterable`, we can refine later and hopefully, we can remove all the sugar at some later point.

> One thing I noticed with parTraverseN, is that in a scenario where I have a lot of parallel tasks (50) with a low parallelism (2), when I simulate my first task failing after a few hundred mills, it takes a few seconds after the failure to cancel the whole chain and propagate the error (Sometimes I even get a UserCancellationException instead of the initial error).
If I have enough parallelism the whole chain will be cancelled as expected after a few milliseconds.
I can post a snippet if the above description is not clear enough.

I'd love some input on this, if you can share some snippets that'd be awesome. Designing laws/tests for these operators is a little tricky so all input/help is very welcome :) Especially if you've encountered some edge cases.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/AsyncLaws.kt
```diff
@@ -19,6 +19,10 @@ import kotlinx.coroutines.newSingleThreadContext
 
 object AsyncLaws {
 
+  // Using newSingleThreadContext in forAll will create 1001 Thread per usage in every forAll.
+  private val one = newSingleThreadContext("1")
```
pakoito: 🙈

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -47,7 +47,7 @@ typealias IOProcF<A> = ((Either<Throwable, A>) -> Unit) -> IOOf<Unit>
 @Suppress("StringLiteralDuplication")
 sealed class IO<out A> : IOOf<A> {
 
-  companion object {
+  companion object : IOParMap2, IOParMap3, IORacePair, IORaceTriple {
```
pakoito: Nice trick

pakoito: Can we put this some other places lol

nomisRev: We need KEEP-87 :D

Issue #1484, opened 2019-06-13
Async#mapEffect


pakoito:
```
fun <A, B> Kind<F, A>.mapEffect(f: suspend (A) -> B): Kind<F, B> =
  flatMap { a -> effect { f(a) } }
```

* [ ] Add function to Async.kt
* [ ] Add dokka on top of the function (see other functions on the file)
* [ ] Add examples below https://arrow-kt.io/docs/effects/async/#continueon

FireZenk:
Let me do this one

PR #1485, opened 2019-06-13
Add effectMap to Async

FireZenk:
#1484

- [x] Add function to Async.kt
- [x] Add dokka on top of the function (see other functions on the file)
- [x] Add examples below https://arrow-kt.io/docs/effects/async/#continueon


FireZenk:
@nomisRev  @JorgeCastilloPrz

modules/docs/arrow-docs/docs/docs/effects/async/README.md
```diff
@@ -87,6 +87,18 @@ Behind the scenes `continueOn()` starts a new coroutine and passes the rest of t
...
+
+Helper function that provides an easy way to construct a suspend effect
+
+```kotlin
+val userCreation = IO.async().run {
+  just(id)
+    .mapEffect { id -> createUserFromId(id) }
+}
+userCreation.request(it)
+```
```
nomisRev: This snippet is invalid because the type of `userCreation` is `Kind<ForIO, User>`.

pakoito:
I suggested the snippet lol Any fix is good.

modules/docs/arrow-docs/docs/docs/effects/async/README.md
FireZenk: Let's remove it then as suggested.

pakoito:
New errors:

```
  //sampleStart
  suspend fun logAndIncrease(s: String): Int {
     println(s)
     return s.toInt() + 1
  }
  val result = ObservableK.async().mapEffect { s -> logAndIncrease(s) }
  //sampleEnd
  println(result)
}
```
error: unresolved reference: ObservableK
  val result = ObservableK.async().mapEffect { s -> logAndIncrease(s) }
               ^
error: cannot infer a type for this parameter. Please specify it explicitly.
  val result = ObservableK.async().mapEffect { s -> logAndIncrease(s) }
                                               ^
error: suspend function 'logAndIncrease' should be called only from a coroutine or another suspend function
  val result = ObservableK.async().mapEffect { s -> logAndIncrease(s) }
```

You have to add the relevant imports (check other examples), and hint `{ s: String ->`

**Comment by pakoito** [comment]
@FireZenk a refactor has landed, the class is now in `arrow-fx` so please move the code there :D

**Comment by pakoito** [comment]
Almost there! Add `* _imports_` like the prior snippets have!

**Comment by pakoito** [comment]
Bit closer:

```
import arrow.fx.rx2.extensions.observablek.applicative.*
```

Add the `async` instance here on top of the `applicative` one, that's what's missing.

**Comment by pakoito** [comment]
image

**Comment by pakoito** [comment]
The code is now correct, except there's a problem in codegen for suspend functions.

```
fun <A, B> Kind<ForIO, A>.effectMap(f: suspend A.() -> B): IO<B> = arrow.fx.IO.async().run {
  this@effectMap.effectMap<A, B>(f) as arrow.fx.IO<B>
}
```

nomisRev:
@pakoito @FireZenk I'll take a look into the processor to see if we can fix this so we can include it in the release :)

nomisRev:
I debugged this for a couple of hours and ran out of time. Having a hard time freeing up time to look into this. Created a [ticket](https://github.com/arrow-kt/arrow/issues/1554).

raulraja:
I'll look into this

raulraja:
Merging this into a local fix branch

PR #1486, opened 2019-06-15
Deprecate concurrent async

nomisRev:
This PR removes follow two methods from the `Concurrent` API.
 - `fun <A> async(f: (KindConnection<F>, (Either<Throwable, A>) -> Unit) -> Unit): Kind<F, A>` 
- `fun <A> asyncF(f: (KindConnection<F>, (Either<Throwable, A>) -> Unit) -> Kind<F, Unit>): Kind<F, A>`

These APIs are removed because of free control over the `push`/`pop` mechanism to control the cancel stack in a public API proofed too dangerous.
Not correctly pushing/poping the `CancelToken`s on the stack can corrupt the rest of the program.

This also reduces the API surface and leaves `cancelable`/`cancelableF` as the only builder that supports async cancelable operation, and `async`/`asyncF` (Async typeclass) are the builders with no cancelation support.


modules/aql/arrow-query-language/src/test/kotlin/arrow/aql/tests/IOExample.kt
```diff
@@ -54,10 +54,10 @@ class DefaultRepository(
 ) : Repository {
 
   override fun allTasksByUser(user: User): IO<List<Task>> =
-    localDS.allTasksByUser(user).handleErrorWith {
-      when (it) {
+    localDS.allTasksByUser(user).handleErrorWith { e: Throwable ->
```
pakoito: 👍

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
```diff
@@ -127,7 +154,7 @@ sealed class IO<out A> : IOOf<A> {
         }
       }
 
-    val never: IO<Nothing> = async { _, _ -> Unit }
+    val never: IO<Nothing> = async { Unit }
```
pakoito: Is the `Unit` necessary?

modules/effects/arrow-effects-rx2-data/src/main/kotlin/arrow/effects/rx2/FlowableK.kt
```diff
@@ -190,41 +186,75 @@ data class FlowableK<A>(val flowable: Flowable<A>) : FlowableKOf<A>, FlowableKKi
...
 
-        val dispose = fa(conn) { either: Either<Throwable, A> ->
+    fun <A> cancelableF(fa: ((Either<Throwable, A>) -> Unit) -> FlowableKOf<CancelToken<ForFlowableK>>, mode: BackpressureStrategy = BackpressureStrategy.BUFFER): FlowableK<A> =
+      Flowable.create({ emitter: FlowableEmitter<A> ->
+        val cb = { either: Either<Throwable, A> ->
           either.fold({
-            emitter.onError(it)
-          }, {
-            emitter.onNext(it)
+            emitter.tryOnError(it).let { Unit }
```
pakoito: Is the `Unit` necessary?

modules/effects/arrow-effects-rx2-data/src/main/kotlin/arrow/effects/rx2/FlowableK.kt
```diff
@@ -190,41 +186,75 @@ data class FlowableK<A>(val flowable: Flowable<A>) : FlowableKOf<A>, FlowableKKi
...
+        }, { e -> emitter.tryOnError(e) })
 
         emitter.setCancellable {
-          dispose.dispose()
-          conn.cancel().value().subscribe({}, {})
+          disp.dispose()
+          val token = cancelOrToken.getAndSet(Left(Unit))
+          token?.fold({}, {
+            it.value().subscribe({}, { e -> emitter.tryOnError(e) })
+          })
```
pakoito: I didn't validate these, I trust they'll be okay :D

modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/IO.kt
nomisRev: Nope, good catch. :+1:

modules/effects/arrow-effects-rx2-data/src/main/kotlin/arrow/effects/rx2/FlowableK.kt
nomisRev: Yes, `tryOnError` checks if the emitter is disposed first and returns a boolean. `onNext` and `onComplete` already checks isDisposed first but return void instead.

PR #1487, opened 2019-06-15
Adding Bitraverse Typeclass

maruiz81:
Implement the Bitraverse interface in the arrow-core-typeclasses


modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/either.kt
```diff
@@ -149,16 +152,34 @@ interface EitherFoldable<L> : Foldable<EitherPartialOf<L>> {
...
+    fix().fold({ f(c, it) }, { g(c, it) })
+
+  override fun <A, B, C> EitherOf<A, B>.bifoldRight(c: Eval<C>, f: (A, Eval<C>) -> Eval<C>, g: (B, Eval<C>) -> Eval<C>): Eval<C> =
+    fix().fold({ f(it, c) }, { g(it, c) })
+}
+
 fun <G, A, B, C> EitherOf<A, B>.traverse(GA: Applicative<G>, f: (B) -> Kind<G, C>): Kind<G, Either<A, C>> =
   fix().fold({ GA.just(Either.Left(it)) }, { GA.run { f(it).map { Either.Right(it) } } })
 
+fun <G, A, B, C, D> EitherOf<A, B>.bitraverse(AP: Applicative<G>, f: (A) -> Kind<G, C>, g: (B) -> Kind<G, D>):
```
pakoito: This function is unnecessary at this level, even if it's possible to implement.

modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/tuple.kt
```diff
@@ -120,10 +123,22 @@ interface Tuple2Foldable<F> : Foldable<Tuple2PartialOf<F>> {
...
+
+  override fun <A, B, C> Tuple2Of<A, B>.bifoldRight(c: Eval<C>, f: (A, Eval<C>) -> Eval<C>, g: (B, Eval<C>) -> Eval<C>): Eval<C> =
+    fix().let { f(it.a, g(it.b, c)) }
+}
+
 fun <F, G, A, B> Tuple2Of<F, A>.traverse(GA: Applicative<G>, f: (A) -> Kind<G, B>): Kind<G, Tuple2<F, B>> = GA.run {
   fix().let { f(it.b).map(it.a::toT) }
 }
 
+fun <G, A, B, C, D> Tuple2Of<A, B>.bitraverse(GA: Applicative<G>, f: (A) -> Kind<G, C>, g: (B) -> Kind<G, D>): Kind<G, Tuple2<C, D>> = GA.run {
```
pakoito: Same as with Either

pakoito:
Ignore the failing build for now, that's kind of my fault -.- Add the fixes and we'll assert and merge once we're sure the problem is fixed.

modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/either.kt
maruiz81: Do you mean to move inside of EitherBitraverse?

pakoito: Yes. Bifoldable will mostly be used as an interface, so it's not expected for this to be here. There's only some cases where we've done it in Composed instances.

maruiz81: yes, I totally agree with you

modules/core/arrow-core-data/src/test/kotlin/arrow/core/IorTest.kt
```diff
@@ -44,7 +48,9 @@ class IorTest : UnitSpec() {
       ShowLaws.laws(Ior.show(), EQ) { Right(it) },
       MonadLaws.laws(Ior.monad(Int.semigroup()), Eq.any()),
       TraverseLaws.laws(Ior.traverse(), Ior.applicative(Int.semigroup()), ::Right, Eq.any()),
-      HashLaws.laws(Ior.hash(Hash.any(), Int.hash()), Ior.eq(Eq.any(), Int.eq())) { Right(it) }
+      HashLaws.laws(Ior.hash(Hash.any(), Int.hash()), Ior.eq(Eq.any(), Int.eq())) { Right(it) },
+      BifoldableLaws.laws(Ior.bifoldable(), { Right(it) }, Eq.any()),
+      BitraverseLaws.laws(Ior.bitraverse(), { Right(it) }, Eq.any())
```
pakoito: Bitraverse Laws imply that the datatype also passes the Bifoldable Laws, so these two lines are redundant.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BitraverseLaws.kt
```diff
@@ -0,0 +1,28 @@
...
+import arrow.test.generators.functionAToB
+import arrow.test.generators.intSmall
+import arrow.typeclasses.Bitraverse
+import arrow.typeclasses.Eq
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+
+object BitraverseLaws {
+  fun <F> laws(BT: Bitraverse<F>, cf: (Int) -> Kind2<F, Int, Int>, EQ: Eq<Kind2<F, Int, Int>>): List<Law> =
+    listOf(Law("Bitraverse Laws: Identity") { BT.identityBitraverse(BT, cf, EQ) })
```
pakoito: As Bitraverse implies Bifoldable, these tests are: `BifoldableLaws + listOf(...)`

See Foldable and Traverse Laws.

pakoito:
Just one last point in the review and we're ready to go!

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BitraverseLaws.kt
maruiz81: ok, I will take a look.

modules/core/arrow-core-data/src/test/kotlin/arrow/core/IorTest.kt
maruiz81: Make sense.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Bitraverse.kt
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * The type class `Bitraverse` defines the behaviour of two separetes `Traverse` over a data type.
+ *
+ * Every instance of `Bitraverse<F>` must contains the next functions:
+ *
+ * ## Bitraverse
+ *
+ * `Bitraverse` perfoms a`Traverse` over both side of the Data type which is `Bifoldable`.
+ *
+ * ```kotlin
```
raulraja: these should be ank playgrounds similar to how it's done in Functor or Id https://github.com/arrow-kt/arrow/blob/38703b5a7b9477892be5093109d5421b85ac2dac/modules/core/arrow-core-data/src/main/kotlin/arrow/core/Id.kt#L13

modules/core/arrow-core-data/src/test/kotlin/arrow/core/ValidatedTest.kt
```diff
@@ -42,7 +46,9 @@ class ValidatedTest : UnitSpec() {
       SemigroupKLaws.laws(
         Validated.semigroupK(String.semigroup()),
         Validated.applicative(String.semigroup()),
-        Eq.any())
+        Eq.any()),
+      BifoldableLaws.laws(Validated.bifoldable(), { Valid(it) }, Eq.any()),
```
raulraja: duplicated law check

modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/either.kt
```diff
@@ -149,6 +151,15 @@ interface EitherFoldable<L> : Foldable<EitherPartialOf<L>> {
     fix().foldRight(lb, f)
 }
 
+@extension
+interface EitherBifoldable : Bifoldable<ForEither> {
+  override fun <A, B, C> EitherOf<A, B>.bifoldLeft(c: C, f: (C, A) -> C, g: (C, B) -> C): C =
+    fix().fold({ f(c, it) }, { g(c, it) })
```
raulraja: These implementations should delegate to methods in `Either` so thatthe data type also includes `bifoldLeft` and `bifoldRight`

modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/ior.kt
```diff
@@ -96,6 +98,26 @@ interface IorTraverse<L> : Traverse<IorPartialOf<L>>, IorFoldable<L> {
     fix().traverse(AP, f)
 }
 
+@extension
+interface IorBifoldable : Bifoldable<ForIor> {
+  override fun <A, B, C> IorOf<A, B>.bifoldLeft(c: C, f: (C, A) -> C, g: (C, B) -> C): C =
+    fix().fold({ f(c, it) }, { g(c, it) }, { a, b -> g(f(c, a), b) })
```
raulraja: Same here, move all instances implementations when possible to the data type. Future versions of arrow won't export the syntax as individual extension functions and we are interested in the data type always including the relevant methods that it wants to expose directly without the type class indirection were possible.

modules/core/arrow-core-extensions/src/main/kotlin/arrow/core/extensions/either.kt
maruiz81: ok, but what about Validated and Tuple, should I move for them too?

pakoito:
@raulraja ??

maruiz81:
Thank you! It was tricky at the beginning but I think now I understand better some concepts. Thank you guys for the lot of help you gave me. My idea is to keep on contributing, so we will be in touch here.

PR #1495, opened 2019-06-23
refactor task for effects

i-walker:
This PR collapse the extension modules from effects into each designated data module:

After this is done. 
We rename the module and artifact names to:
- [x] arrow-effects-data -> arrow-fx
- [x] create arrow-effects-mtl 
- [x] arrow-effects-mtl -> arrow-fx-mtl
- [x] arrow-effects-reactor-data -> arrow-fx-reactor
- [x] arrow-effects-rx2-data -> arrow-fx-rx2


modules/effects/arrow-effects-data/src/main/kotlin/arrow/effects/extensions/io.kt
```diff
@@ -238,3 +243,33 @@ interface IOUnsafeRun : UnsafeRun<ForIO> {
   override suspend fun <A> unsafe.runNonBlocking(fa: () -> Kind<ForIO, A>, cb: (Either<Throwable, A>) -> Unit) =
     fa().fix().unsafeRunAsync(cb)
 }
+
+@extension
+interface IODispatchers : Dispatchers<ForIO> {
+  override fun default(): CoroutineContext =
+    IODispatchers.CommonPool
```
pakoito: This forces effects-extensions to be JVM-only! I would not go ahead with this refactor as it is.

pakoito:
```
Effects
io.arrow-kt:arrow-effects:$arrow_version
VVVV
io.arrow-kt:arrow-effects-data:$arrow_version
io.arrow-kt:arrow-effects-typeclasses:$arrow_version
io.arrow-kt:arrow-effects-extensions-core:$arrow_version
io.arrow-kt:arrow-effects-extensions-aql:$arrow_version
io.arrow-kt:arrow-effects-extensions-mtl:$arrow_version
io.arrow-kt:arrow-effects-extensions-effects:$arrow_version
```

That's what we got in the planning about effects

i-walker:
The old list in google is not up-to-date with the current plans: 
But, sure I forgot
`arrow-fx-mtl`

pakoito:
What changed?

raulraja:
@pakoito in order to match the marketing, logo etc we are being consistent across the board and renaming the `effects` package and library name to be just `fx` and the artifact `arrow-fx`. This is more something I proposed to Imran based on what I discussed with marketing at 47 but if anyone disagrees we are still on time to discuss. I personally prefer cohesive names across marketing materials, logos etc in line with the code and library. Also `arrow.fx.IO` looks more badass

pakoito:
I'm okay with that, I meant is the package structure different than what I posted above?

raulraja:
Yes, there are no individual modules except for the integrations because arrow-fx includes IO and the instances. We discussed the same for optics so users get all they need without fiddling with artifacts. We would have to add more modules when we do the expect for the dispatchers multiplatform.

pakoito:
So wait, we had it like that and then the splits planned:

```
Effects
io.arrow-kt:arrow-fx:$arrow_version
VVVV
io.arrow-kt:arrow-fx-data:$arrow_version   <-- includes typeclasses now
io.arrow-kt:arrow-fx-extensions-core:$arrow_version
io.arrow-kt:arrow-fx-extensions-aql:$arrow_version
io.arrow-kt:arrow-fx-extensions-mtl:$arrow_version
io.arrow-kt:arrow-fx-extensions-effects:$arrow_version
```

i-walker:
The plan is to include the regular extension of a specific module within the boundaries of the toplevel module:
```
E.g.:Effects
io.arrow-kt:arrow-fx:$arrow_version <- includes arrow-fx-extensions-fx
VVVV
io.arrow-kt:arrow-fx-data:$arrow_version   <-- includes typeclasses now
io.arrow-kt:arrow-fx-extensions-core:$arrow_version
io.arrow-kt:arrow-fx-extensions-aql:$arrow_version
io.arrow-kt:arrow-fx-extensions-mtl:$arrow_version
``` 
The core of each module is essentially encapsulates the mainly the data types and typeclasses, meaning core is unnecessary:
```
E.g.:Effects
io.arrow-kt:arrow-fx:$arrow_version <- includes arrow-fx-extensions-fx
VVVV
io.arrow-kt:arrow-fx-data:$arrow_version   <-- includes typeclasses now
io.arrow-kt:arrow-fx-extensions-aql:$arrow_version
io.arrow-kt:arrow-fx-extensions-mtl:$arrow_version
```
Because of marketing, specific for arrow-fx, it makes sense to collapse datatypes and extensions, because people endup using both.
```
E.g.:Effects <- now Fx
io.arrow-kt:arrow-fx:$arrow_version <- includes arrow-fx-extensions-fx,  typeclasses and data types
VVVV
io.arrow-kt:arrow-fx-extensions-aql:$arrow_version
io.arrow-kt:arrow-fx-extensions-mtl:$arrow_version
```
If there are any AQL instances in Fx it makes sense to create an artifact for it, but I did not find any. @pakoito are there any I may have missed?

raulraja:
I discussed this offline with @pakoito and the current proposed layout matches what we need for type classes and auto extensions with compiler plugins. Also splitting instances would not allow us to form coherent exportable instances so the layout in the PR is ok for the time being.

PR #1497, opened 2019-06-26
Replace IO suspend with defer

Octogonapus:
This PR fixes the `IO.suspend` docs so they are up to date with the latest version (`0.9.0`). I think `suspend` was renamed to `defer` (my artifacts don't have a `suspend`) so the docs should be updated.


pakoito:
Where do you use Arrow, if you don't mind me asking? Company, side-projects?

Octogonapus:
Mostly for robotics research that lives here https://github.com/CommonWealthRobotics

PR #1500, opened 2019-07-01
Add linting documentation documentation

### LINKED ISSUE #1385: Redundant suspend modifiers
When you use ArrowFx you are expected to `suspend` your effectful functions so you can then wrap them on `effect {}` inside the `fx {}` block. The usage of `suspend` on them is equivalent to wrapping them on `IO` in a sense, since it acts as a safety lock that ensures you'd never be able to call it from a non suspended block (which we provide by using `effect {}`).

One "issue" we have is that default IntellIJ Kotlin linting rules mark `suspend` modifiers that do not call suspended functions inside as redundant, and prompt you do remove them. 

Screenshot 2019-03-22 at 19 05 40

This is not super bad, but in our use case we'd ideally avoid this since removing those is not really an option, given that you shouldn't use `fx` without `suspend`.

This is not a linter rule but an Idea inspection:

Screenshot 2019-03-28 at 10 25 54

Screen Shot 2019-03-28 at 10 27 06 AM

### Action points 

Best thing we can do about this is add a section in README and Quick Start docs page warning about IDEA linting and how you should disable this rule, providing guiding images like the ones attached here and [a link to the official docs about this](https://www.jetbrains.com/help/idea/intention-actions.html).

### LINKED ISSUE #1386: Unit return types
When you're writing code in an FP manner, you want to keep all effects under control. **Functions that return `Unit` are effectful by definition**, since otherwise they'd do nothing. On their bodies, they always access the external world, as in printing, logging, network queries, database reads / writes, modifying the value of class properties... etc. 

### Action points 
  
* [x] The redundant Unit return type is an IntellIJ IDEA Inspection. Unit return types are never redundant from an FP perspective, since they make function "referentially transparent". The function states what's it doing by looking at it's input and output types, so best thing we can do is to give feedback to Arrow users about how to disable it in README and "Quick start" docs section as also suggested for #1385. [Also include a link to the official intention action docs](https://www.jetbrains.com/help/idea/intention-actions.html). 

Screenshot 2019-03-28 at 12 07 22

* [ ] ~Apart from the IntellIJ IDEA Inspection, Ktlint also complaints about Unit return types regardless of where they're used. That's also something we could suggest users how to disable on their ktlint configurations, since from an FP perspective it's always important to declare return types regardless of what they are, so it's explicit that the function has side effects inside.~

Tylos:
### References
* Fixes #1385
* Fixes #1386 

### Description

Add Linting documentation as a new section explaining `redundant suspend modifier` and `redundant unit return type` IDEA warnings and how they can be disabled

The section is added to the sidebar menu and referenced in the project README

### 📸 

image
deployed in localhost


modules/docs/arrow-docs/docs/docs/quickstart/linting/README.md
```diff
@@ -0,0 +1,33 @@
...
+permalink: /docs/quickstart/linting/
+---
+
+## Arrow Linting
+
+Some linters might complain about some code practices that are common when working with functional programming. In this section
+
+### Working with ArrowFx
+
+When working with [ArrowFx]() side effects are expressed as `suspend` functions.
```
JorgeCastilloPrz: You're missing the link here!

modules/docs/arrow-docs/docs/docs/quickstart/linting/README.md
```diff
@@ -0,0 +1,33 @@
...
+## Arrow Linting
+
+Some linters might complain about some code practices that are common when working with functional programming. In this section
+
+### Working with ArrowFx
+
+When working with [ArrowFx]() side effects are expressed as `suspend` functions. 
+
+However, IDEA will show *redundant suspend modifiers* warning if a function does 
+not invoke other suspended functions. For ArrowFx we want to be explicit about impure functions, so this warning
```
JorgeCastilloPrz: do you have a random line break here?

modules/docs/arrow-docs/docs/docs/quickstart/linting/README.md
```diff
@@ -0,0 +1,33 @@
...
+is not useful
+
+You can disable this warning in Preferences > Editor > Inspections. Then search for the option Kotlin > Redundant constructs > Redundant suspend modifier
+
+![gif](/img/linting/linting_suspend_modifier.gif)
+
+### Working with purity
+
+In functional programming being explicit about the return type of functions is important.
+Based on the return type of a function we can try to identify whether it is a pure or an
```
JorgeCastilloPrz: Random line break?

modules/docs/arrow-docs/docs/docs/quickstart/linting/README.md
Tylos: Yup, fixing

modules/docs/arrow-docs/docs/docs/quickstart/linting/README.md
Tylos: Yep! fixing

PR #1509, opened 2019-07-05
moved FunctorFilter to core

i-walker:
FunctorFilter needs to be moved to the core typeclasses, given it’s gonna be needed in AQL and IO when IO is able to autoflatten nested Option and Either effects.


raulraja:
The actual instances for core data types like Try etc should also be relocated alongside the core instances. Does AQL need a dependency anymore on MTL now that FunctorFilter is in core? If not it should be removed and just depend on core

pakoito:
@i-walker We should also move MonadFilter!

PR #1511, opened 2019-07-15
move TraverseFilter, MonadCombine and MonadFilter

i-walker:
also cleans up some import aliases 

follow up from: https://github.com/arrow-kt/arrow/pull/1509#issuecomment-509267111


modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/TraverseFilter.kt
```diff
@@ -1,30 +1,40 @@
...
-import arrow.typeclasses.FunctorFilter
-import arrow.typeclasses.Traverse
 
 /**
- * ank_macro_hierarchy(arrow.mtl.typeclasses.TraverseFilter)
+ * ank_macro_hierarchy(arrow.typeclasses.TraverseFilter)
  */
 interface TraverseFilter<F> : Traverse<F>, FunctorFilter<F> {
 
+  private object IdApplicative : Applicative<ForId> {
```
pakoito: _squint_ why?

pakoito: Oh, right, because it's moved too.

modules/core/arrow-core-data/src/test/kotlin/arrow/core/extensions/ComposedInstancesTest.kt
```diff
@@ -108,7 +107,7 @@ class ComposedInstancesTest : UnitSpec() {
       FunctorLaws.laws(ComposedFunctor(Option.functor(), NonEmptyList.functor()), cf, EQ_OPTION_NEL),
       ApplicativeLaws.laws(ComposedApplicative(Option.applicative(), NonEmptyList.applicative()), EQ_OPTION_NEL),
       FoldableLaws.laws(ComposedFoldable(Option.foldable(), NonEmptyList.foldable()), cf, Eq.any()),
-      TraverseLaws.laws(ComposedTraverse(Option.traverse(), NonEmptyList.traverse(), NonEmptyList.applicative()), ComposedFunctor.invoke(Option.functor(), NonEmptyList.functor()), cf, EQ_OPTION_NEL),
+      TraverseLaws.laws(ComposedTraverse(arrow.core.extensions.traverse(), arrow.core.extensions.traverse(), NonEmptyList.applicative()), ComposedFunctor.invoke(Option.functor(), NonEmptyList.functor()), cf, EQ_OPTION_NEL),
```
pakoito: This doesn't look quite right.

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/const.kt
```diff
@@ -70,7 +70,7 @@ interface ConstApply<A> : Apply<ConstPartialOf<A>> {
   override fun <T, U> ConstOf<A, T>.map(f: (T) -> U): Const<A, U> = fix().retag()
 
   override fun <T, U> ConstOf<A, T>.ap(ff: ConstOf<A, (T) -> U>): Const<A, U> =
-    constAp(MA(), ff)
+    ap(MA(), ff)
```
pakoito: Good catch!

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/either.kt
```diff
@@ -90,7 +85,7 @@ interface EitherApply<L> : Apply<EitherPartialOf<L>>, EitherFunctor<L> {
   override fun <A, B> EitherOf<L, A>.map(f: (A) -> B): Either<L, B> = fix().map(f)
 
   override fun <A, B> EitherOf<L, A>.ap(ff: EitherOf<L, (A) -> B>): Either<L, B> =
-    fix().eitherAp(ff)
+    fix().ap(ff)
```
pakoito: This is going to cause a stackOverflow because you'd think it calls its member function but no. It calls itself instead.

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/either.kt
```diff
@@ -135,7 +130,7 @@ interface EitherApplicativeError<L> : ApplicativeError<EitherPartialOf<L>, L>, E
   override fun <A> raiseError(e: L): Either<L, A> = Left(e)
 
   override fun <A> EitherOf<L, A>.handleErrorWith(f: (L) -> EitherOf<L, A>): Either<L, A> =
-    fix().eitherHandleErrorWith(f)
+    fix().handleErrorWith(f)
```
pakoito: Same here, and in the others. Some of these are defined as extfuns already, so it picks up this one instead unless you alias it.

i-walker: thank you I went crazy to find the problem

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/TraverseFilter.kt
i-walker: exactly

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/either.kt
pakoito: There's only a couple like constAp where this doens't apply.

i-walker:
Isn't this a recursive call, too. 
We're not using map2 internally, that's why I checked it. 
https://github.com/arrow-kt/arrow/blob/652b9c2cc5266c104a48dad172d451e654bff0fc/modules/mtl/arrow-mtl/src/main/kotlin/arrow/mtl/extensions/option.kt#L41

pakoito:
Yes, make sure to test them somehow!!

PR #1513, opened 2019-07-16
FunctorFilter cleanup

i-walker:
- [X] added the missing FunctorFilter Laws from Cats https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/FunctorFilterLaws.scala
- [X] removed FunctorFilter instance in NonEmptyList
- [X] implemented missing mapFilter / map2 functions
- [X] moved standalone FunctorFilter instances to their extension files
- [X] rewrote ForSequenceK.toList to an instance method
- [X] renamed mapFilter to filterMap


nomisRev:
Could we add a meaningful law for this and add it here? https://github.com/arrow-kt/arrow/blob/cb579c88e710916d158d08ba9594671d2eeb472d/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/FunctorFilterLaws.kt

modules/core/arrow-core-data/src/test/kotlin/arrow/core/SequenceKTest.kt
```diff
@@ -35,22 +35,22 @@ class SequenceKTest : UnitSpec() {
...
     val eq: Eq<Kind<ForSequenceK, Int>> = object : Eq<Kind<ForSequenceK, Int>> {
       override fun Kind<ForSequenceK, Int>.eqv(b: Kind<ForSequenceK, Int>): Boolean =
-        toList() == b.toList()
+        fix().toList() == b.fix().toList()
     }
 
     val associativeSemigroupalEq: Eq<Kind<ForSequenceK, Tuple2<Int, Tuple2<Int, Int>>>> = object : Eq<Kind<ForSequenceK, Tuple2<Int, Tuple2<Int, Int>>>> {
       override fun Kind<ForSequenceK, Tuple2<Int, Tuple2<Int, Int>>>.eqv(b: Kind<ForSequenceK, Tuple2<Int, Tuple2<Int, Int>>>): Boolean =
-        this.toList() == b.toList()
+        this.fix().toList() == b.fix().toList()
```
pakoito: `this` isn't necessary here

pakoito:
Why is Try not a FF?

raulraja:
@pakoito what would be Try's empty identity?

pakoito:
🤔 

None is not

modules/core/arrow-core-data/src/main/kotlin/arrow/core/Try.kt
```diff
@@ -85,7 +85,7 @@ sealed class Try<out A> : TryOf<A> {
   fun filter(p: Predicate<A>): Try<A> =
     flatMap { if (p(it)) Success(it) else Failure(TryException.PredicateException("Predicate does not hold for $it")) }
 
-  fun <B> mapFilter(f: (A) -> Option<B>): Try<B> =
+  fun <B> filterMap(f: (A) -> Option<B>): Try<B> =
```
pakoito: This function needs to be deleted if we're not providing the typeclass.

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/try.kt
```diff
@@ -190,6 +192,16 @@ interface TryTraverse : Traverse<ForTry> {
     fix().foldRight(lb, f)
 }
 
+@extension
+interface TryFunctorFilter : FunctorFilter<ForTry> {
```
pakoito: Oh, nevermind. It's here :D

i-walker: but you're right we need to delete this

PR #1516, opened 2019-07-17
Adds Bimonad Laws

i-walker:
- [x] write the Laws based on Cats https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/BimonadLaws.scala
- [x] add BimonadLaws to every Test that uses this typeclass

I am also inclined to write Doc's for Bimonad and move them to KDocs


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BimonadLaws.kt
```diff
@@ -0,0 +1,34 @@
...
+               EQ1: Eq<Int>,
+               f2: (Int) -> Kind<F, Kind<F, Int>>,
+               EQ2: Eq<Kind<F, Kind<F, Int>>>): List<Law> =
+    listOf(
+      Law("Bimonad Laws: Extract Identity") { BF.extractIsIdentity(f1, EQ1) },
+      Law("Bimonad Laws: CoflatMap Composition") { BF.coflatMapComposition(f1, EQ2) },
+      Law("Bimonad Laws: Extract FlatMap") { BF.extractFlatMap(f2, EQ1) }
+    )
+
+  fun <F> Bimonad<F>.extractIsIdentity(f: (Int) -> Kind<F, Int>, EQ: Eq<Int>): Unit =
```
pakoito: f isn't necessary as it's `just`, the constructor

i-walker:
I did not find any other instance, If you (@pakoito) agree on the current set up we can merge, after build :)

modules/core/arrow-core-data/src/test/kotlin/arrow/core/EvalTest.kt
```diff
@@ -18,16 +20,21 @@ import org.junit.runner.RunWith
...
   init {
 
-      testLaws(
-        MonadLaws.laws(Eval.monad(), EQ),
-        ComonadLaws.laws(Eval.comonad(), ::Now, EQ)
-      )
+    testLaws(
+      MonadLaws.laws(Eval.monad(), EQ1),
+      ComonadLaws.laws(Eval.comonad(), ::Now, EQ1),
+      BimonadLaws.laws(Eval.bimonad(), { Now(Now(it)) }, Eq.any(), EQ2)
```
pakoito: Isn't bimonad Comonad + Monad already?

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BimonadLaws.kt
```diff
@@ -8,27 +8,35 @@ import io.kotlintest.properties.forAll
 
 object BimonadLaws {
   fun <F> laws(BF: Bimonad<F>,
-               f1: (Int) -> Kind<F, Int>,
+               f: (Int) -> Kind<F, Kind<F, Int>>,
```
pakoito: This is still possible to make simply with Monad. It's `just(just(it))`

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BimonadLaws.kt
```diff
@@ -8,27 +8,35 @@ import io.kotlintest.properties.forAll
...
                EQ1: Eq<Int>,
-               f2: (Int) -> Kind<F, Kind<F, Int>>,
                EQ2: Eq<Kind<F, Kind<F, Int>>>): List<Law> =
     listOf(
```
pakoito: This should be MonadLaws + ComonadLaws + listOf

i-walker: Currently, with your suggestion, BimonadLaws would be unnecessary because in every datatype we already have that exact setting. What I wanted to introduce here is that the laws work in conjunction, and not solely in parallel, as before. 

That's why the `coflatMap` composition with `just`.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BimonadLaws.kt
i-walker: Makes sense

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BimonadLaws.kt
pakoito: >BimonadLaws would be unnecessary because in every datatype we already have that exact setting

It should replace it, and extend it with the additional laws below!

pakoito: For example, something that's Async and Comonad may not need the AsyncLaws check removed, but the Comonad one can be replaced with Bimonad.

i-walker: agree

PR #1518, opened 2019-07-18
delete FF extension for Try

i-walker:


i-walker:
Not sure why but bintray says that the push is successfull, github does not give the same answer. 
I am going to merge this, anyways.

PR #1519, opened 2019-07-18
Add persistent map

adam-arold:
In this PR I

- Added a new module, `arrow-persistent-data-structures` to Arrow
- Augmented `Foldable` to include the `remove` operation
- And added a preliminary implementation for `PersistentMap` which wraps a ported `HashArrayMappedTrie` implementation


modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
```diff
@@ -12,6 +12,7 @@ import arrow.core.Some
 import arrow.core.flatMap
 import arrow.core.identity
 import arrow.core.right
+import me.eugeniomarletti.kotlin.metadata.shadow.load.kotlin.hasVoidReturnType
```
pakoito: ????

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
```diff
@@ -71,6 +72,31 @@ interface Foldable<F> {
       }
     }
 
+  /**
+   * Transforms this [Foldable] into a new one which doesn't contain the given [element].
+   * Will return an identical [Foldable] if it doesn't contain [element] in the first place.
+   * Example:
+   * ```
```
pakoito: Check other snippets for ank documentation so it gets generated and is runnable :D

modules/persistent/arrow-persistent-data-structures/src/main/kotlin/arrow/persistent/data/PersistentMap.kt
```diff
@@ -0,0 +1,27 @@
...
+
+  operator fun get(key: K): Option<V> = map[key]
+
+  fun containsKey(key: K): Boolean = map.containsKey(key)
+
+  fun put(key: K, value: V): PersistentMap<K, V> {
+    return PersistentMap(map.put(key, value))
+  }
+
+  fun remove(key: K): PersistentMap<K, V> {
```
pakoito: Expression bodies please 🙈

modules/persistent/arrow-persistent-data-structures/build.gradle
```diff
@@ -0,0 +1,17 @@
+apply plugin: 'kotlin-kapt'
+
+dependencies {
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
+    compile project(':arrow-annotations')
+    compile project(':arrow-core')
+    compile project(':arrow-core-data')
```
i-walker:
```suggestion
```
you only need to depend on arrow-core, it will pull arrow-core-data automatically.

modules/persistent/arrow-persistent-data-structures/gradle.properties
```diff
@@ -0,0 +1,4 @@
+# Maven publishing configuration
+POM_NAME=Arrow Persistent Data Structures
+POM_ARTIFACT_ID=arrow-persistent-data-structures
```
i-walker: Can we look for a cool artifact ID, maybe shorter than this :)

modules/persistent/arrow-persistent-data-structures/src/main/kotlin/arrow/persistent/data/PersistentMapK.kt
```diff
@@ -0,0 +1,73 @@
...
+data class PersistentMapK<K, A>(
+  private val map: HashArrayMappedTrie<K, A> = HashArrayMappedTrie.empty())
+  : PersistentMapKOf<K, A> {
+
+  val isEmpty: Boolean = map.isEmpty
+  val size: Int = map.size
+
+  fun <B> map(f: (A) -> B): PersistentMapK<K, B> {
+    var result = HashArrayMappedTrie.empty<K, B>()
+    map.iterator().forEach { (a, b) ->
```
i-walker: Maybe a flatMap would work here, rather than an imperative implementation.

modules/persistent/arrow-persistent-data-structures/src/main/kotlin/arrow/persistent/data/PersistentMapK.kt
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  fun remove(key: K): PersistentMapK<K, A> {
+    return PersistentMapK(map.remove(key))
+  }
+
+  companion object
+}
+
+class ForPersistentMapK private constructor() {
```
i-walker: The Higherkinded boilerplate is always at the top of each file

modules/persistent/arrow-persistent-data-structures/gradle.properties
adam-arold: Do you have a suggestion? `arrow-persistent-data` maybe?

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
```diff
@@ -71,6 +71,44 @@ interface Foldable<F> {
...
     }
 
+  /**
+   * Transforms this [Foldable] into a new one which doesn't contain the given [element].
+   * Will return an identical [Foldable] if it doesn't contain [element] in the first place.
+   * Example:
+   *
+   * {: data-executable='true'}
+   *
+   * ```kotlin:ank
```
nomisRev:
```
-   * {: data-executable='true'}
-   *
-   * ```kotlin:ank
```

modules/persistent/arrow-persistent-data-structures/build.gradle
```diff
@@ -0,0 +1,16 @@
+apply plugin: 'kotlin-kapt'
+
+dependencies {
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
```
nomisRev:
```suggestion
```

Already included in `arrow-core`

modules/persistent/arrow-persistent-data-structures/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+    compile project(':arrow-annotations')
+    compile project(':arrow-core')
+    kapt project(':arrow-meta')
+    kaptTest project(':arrow-meta')
+    compileOnly project(':arrow-meta')
```
nomisRev:
```suggestion
```

This shouldn't be necessary, it doesn't affect the binary but it clouds your IDEA suggestions.

modules/persistent/arrow-persistent-data-structures/src/main/kotlin/arrow/persistent/extensions/persistentmap.kt
```diff
@@ -0,0 +1,53 @@
...
+  }
+}
+
+@extension
+interface PersistentMapKMonoid<K, A> : Monoid<PersistentMapK<K, A>>, PersistentMapKSemigroup<K, A> {
+  override fun empty(): PersistentMapK<K, A> = PersistentMapK()
+}
+
+@extension
+interface PersistentMapKEq<K, A> : Eq<PersistentMapK<K, A>> {
```
nomisRev: TIP: Here you'd want to have a constraint on `Eq<K>` and `Eq<A>` to implement `Eq<PersistentMapK<K, A>>`. This is done by having abstract functions in the interface.

```
interface PersistentMapKEq<K, A> : Eq<PersistentMapK<K, A>> {
  fun EQK(): Eq<K>
  fun EQA(): Eq<A>
```

modules/persistent/arrow-persistent-data-structures/gradle.properties
adam-arold: @i-walker ping

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
```diff
@@ -71,6 +71,44 @@ interface Foldable<F> {
...
+   *
+   * @param element the element to remove
+   * @param monoidK the [Monoid] which produces the empty result [Foldable]
+   * @param AF the [Applicative] which will be used to lift the values in this [Foldable]
+   */
+  fun <A> Kind<F, A>.remove(
+    element: A,
+    monoidK: MonoidK<F>,
+    AF: Applicative<F>
+  ): Kind<F, A> {
```
i-walker: Expression body?

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
```diff
@@ -71,6 +71,44 @@ interface Foldable<F> {
...
+   *  //sampleEnd
+   *  println(result)
+   * }
+   * ```
+   *
+   * @param element the element to remove
+   * @param monoidK the [Monoid] which produces the empty result [Foldable]
+   * @param AF the [Applicative] which will be used to lift the values in this [Foldable]
+   */
+  fun <A> Kind<F, A>.remove(
```
i-walker: I am not sure if this method makes sense at all. I checked in with Haskell and Cat's Library and none of them implement a remove function. 
Nonetheless, this remove function is isomorphic to a simple fold function, where you preserve the Datatype which implements Foldable and just exclude the element you don't want. This makes more sense if you keep it in the Datatype as a function for convenience, but I object having it in this type class.

modules/persistent/arrow-persistent-data-structures/src/main/kotlin/arrow/persistent/extensions/persistentmap.kt
i-walker: We can stay with `arrow-persistent-data` and if you define any extension instance name that module `arrow-persistent`.

modules/persistent/arrow-persistent-data-structures/src/main/kotlin/arrow/persistent/extensions/persistentmap.kt
```diff
@@ -0,0 +1,53 @@
+package arrow.persistent.extensions
```
i-walker: All these extensions need to be in another module `arrow-persistent`, you need to compile `arrow-persistent-data` and check what you need in your tests. 

If you write TestLaws, they need to be in `arrow-persistent-data` - Just testCompile what you need.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt
nomisRev: I do not object on having it here and I'd like to see it in `FunctorFilter` too with a simpler signature.

Just wondering if there is a type that has an instance for `Foldable` and `MonoidK` but no instance for `FunctorFilter` 🤔 cc\ @raulraja @pakoito

i-walker: I am not saying that we should not have a remove function, but it should not be a typeclass-level function. We can have them for datatypes and refer to them with a fold. 
I don't see the point of a simpler API, if we add this function.

adam-arold: @i-walker please discuss this in the relevant place where there is a very long conversation which also explains why this is here: [link](https://kotlinlang.slack.com/archives/C8UK6RTHU/p1563102430078300).

modules/persistent/arrow-persistent-data-structures/src/main/kotlin/arrow/persistent/extensions/persistentmap.kt
adam-arold: Can you elaborate a bit further? If I understand correctly we need to rename `arrow-persistent-data-structures` to `arrow-persistent-data`, but I'm not sure what you are trying to say here. Do we need 2 different modules?

adam-arold: We've discussed this with @nomisRev on Slack and we agreed that it is fine as it stands now (since I inlined the `@higherkind` boilerplate and the whole thing will be rendered unnecessary after KEEP-87.

modules/persistent/arrow-persistent-data-structures/gradle.properties
kylegoetz: arrow-persist? arrow-eternal? :D

purrgrammer:
Maybe it makes sense for Arrow to integrate with immutable persistent collections from https://github.com/Kotlin/kotlinx.collections.immutable instead of shipping its own implementation? They already have implementations of map, set and list over there.

pakoito:
I wish we could. It looks like someone's side-project and hasn't seen development in 4 months. Nowhere it's been promoted or stablished as "done". The best we could aim for IMO is to fork it and be the stewards.

@adam-arold wanted to do something around it that was unified for all Persistent collections.

In any case I agree, we shouldn't ship our own collections straight from the main repo.

purrgrammer:
FTR a nice overview of the state of the art for immutable & persistent collections on the JVM: https://github.com/lacuna/bifurcan/blob/master/doc/comparison.md#the-libraries-and-versions

adam-arold:
@purrgrammer I went through this list and ended up using `kotlinx.collections.immutable` at the end because it is the only "official" solution which is also MPP. I think it would make sense if Arrow's implementation would just wrap it (or we could also vendor it since it is Apache 2.0). What do you think @pakoito ?

PhBastiani:
@adam-arold ```kotlinx.collections.immutable``` is right way to play with immutable collections in a MPP context. But, from memory, this lib does not provide any solution for persistent collections.

@pakoito I think that https://github.com/arrow-kt/arrow-collections could be the right place to propose one or several arrow solution / wrappers

adam-arold:
@PhBastiani it does provide persistent collections. In fact, I've been using that lib in other projects ever since it became MPP. I'd definitely go MPP since Arrow is becoming MPP pretty soon.

pakoito:
Do you believe we should ask Jetbrains about investment/vendoring? Same as Spek, they may feel okay with leaving ownership to someone that can put the time.

adam-arold:
In my opinion, adding a simple dependency to `kotlinx.collections.immutable` and just wrapping it in an Arrow-style abstraction is enough and cost-efficient. I only mentioned vendoring, because someone said that adding dependencies like this is not an option. Although I don't understand the underlying reasons.

rachelcarmena:

PR #1527, opened 2019-07-22
Delete inline fun in DataTypes

i-walker:
I left out those in Fx, MTL, Free and meta (mostly the Processor for CodeGen).
If those need to go to just leave a comment


modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -46,7 +46,7 @@ sealed class Either<out A, out B> : EitherOf<A, B> {
    * @param ifRight the function to apply if this is a [Right]
    * @return the results of applying the function
    */
-  inline fun <C> fold(ifLeft: (A) -> C, ifRight: (B) -> C): C = when (this) {
+  fun <C> fold(ifLeft: (A) -> C, ifRight: (B) -> C): C = when (this) {
```
pakoito: I would keep it here just so people can call `suspend` because the implementation doesn't prevent it.

modules/core/arrow-core-data/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -79,16 +79,16 @@ sealed class Option<out A> : OptionOf<A> {
...
 
-  inline fun <B, R> map2(fb: Kind<ForOption, B>, f: (Tuple2<A, B>) -> R): Option<R> =
+  fun <B, R> map2(fb: Kind<ForOption, B>, f: (Tuple2<A, B>) -> R): Option<R> =
     flatMap { a: A -> fb.fix().map { b -> f(a toT b) } }
 
   fun <B> filterMap(f: (A) -> Option<B>): Option<B> =
     flatMap { a -> f(a).fold({ empty<B>() }, { just(it) }) }
 
-  inline fun <R> fold(ifEmpty: () -> R, ifSome: (A) -> R): R = when (this) {
+  fun <R> fold(ifEmpty: () -> R, ifSome: (A) -> R): R = when (this) {
```
pakoito: Same as Either

modules/core/arrow-core-data/src/main/kotlin/arrow/core/Try.kt
```diff
@@ -98,7 +98,7 @@ sealed class Try<out A> : TryOf<A> {
   /**
    * Applies `ifFailure` if this is a `Failure` or `ifSuccess` if this is a `Success`.
    */
-  inline fun <B> fold(ifFailure: (Throwable) -> B, ifSuccess: (A) -> B): B =
+  fun <B> fold(ifFailure: (Throwable) -> B, ifSuccess: (A) -> B): B =
```
pakoito: Same as Either and Option

pakoito:
IMO we should keep inline in `fold` because the implementation is a when, so people don't have to decide between fold and when depending on whether they're inside a suspen fun.

pakoito:
@JorgeCastilloPrz @nomisRev @raulraja opinion?

raulraja:
I'm fine with inline fold since the purity plugin will still enforce suspend calls won't happen there even if today is abused to let those thru

nomisRev:
I am absolutely in favor of that!

JorgeCastilloPrz:
Agree, It's good to leave it open to any lambdas including suspended ones imho.

modules/ank/arrow-ank/src/main/kotlin/arrow/ank/interpreter.kt
```diff
@@ -130,8 +130,8 @@ val interpreter: AnkOps = object : AnkOps {
     val result: Tuple3<SnippetParserState, Sequence<String>, Sequence<Snippet>> =
       content
         .fold(
-          Tuple3(SnippetParserState.Searching, emptySequence(), emptySequence())
-        ) { (state, lines, snippets), line ->
+          Tuple3(SnippetParserState.Searching as SnippetParserState, emptySequence(), emptySequence())
```
raulraja: Is Inference not working as expected when the inline modifier is removed? If so that looks like a compiler bug

modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
raulraja: We can keep it inline here and in all autofolds because the purity plugin will ensure these do not actually accept suspend but users looking for fair optimization use and not just a hack over suspend can still benefit from inlining

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/EitherT.kt
```diff
@@ -64,7 +64,7 @@ data class EitherT<F, A, B>(private val value: Kind<F, Either<A, B>>) : EitherTO
 
   fun value(): Kind<F, Either<A, B>> = value
 
-  inline fun <C> fold(FF: Functor<F>, crossinline l: (A) -> C, crossinline r: (B) -> C): Kind<F, C> = FF.run {
+  fun <C> fold(FF: Functor<F>, l: (A) -> C, r: (B) -> C): Kind<F, C> = FF.run {
```
raulraja: If other folds are preserved inlined so should this one

modules/ank/arrow-ank/src/main/kotlin/arrow/ank/interpreter.kt
i-walker: That is true

PR #1531, opened 2019-07-24
@Deprecation cleanUp

i-walker:
- Open Issue on Validated datatype


modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Monad.kt
```diff
@@ -58,36 +58,15 @@ interface Monad<F> : Selective<F> {
   fun <A, B> Kind<F, A>.followedByEval(fb: Eval<Kind<F, B>>): Kind<F, B> =
     flatMap { fb.value() }
 
-  @Deprecated(
-    "effectM is being renamed to flatTap",
```
pakoito: I believe this hasn't even been deployed yet.

pakoito:
I'd check that all the deleted @deprecated methods were already marked in 0.9.0 for deprecation (go to the 0.9.0 tag on the repository), otherwise we've added them during the 0.10.0 dev cycle.

One thing, @JorgeCastilloPrz  suggested adding a deprecation notice atop Try, and also change the text of the existing in new ones to be more explicit that it'll be removed in the not-so-distant future.

JorgeCastilloPrz:
I was recently told by an old friend that sometimes we've removed stuff that was being widely used (sometimes abused like Try) without proper Deprecation notes, so that got them into a bit of trouble. I believe that was with some older versions, but still I believe it's something to always keep in mind. So If we're removing Try in 1 or 2 versions and given how it's massively used as it seems for what I've seen around in the Android community, I suggest to be as explicit as possible about the fact that is being removed pretty soon in the deprecation message starting now.

modules/core/arrow-syntax/src/main/kotlin/arrow/syntax/collections/array.kt
```diff
@@ -1,59 +0,0 @@
-package arrow.syntax.collections
-
-import arrow.core.Option
-import arrow.core.Predicate
-import arrow.core.toOption
-
-@Deprecated(message = "`firstOption` is now part of the Foldable interface and generalized to all foldable data types")
```
i-walker: Should I let them deprecated or as here (deleted), because they are in Foldable and I don't see a problem deleting these. @pakoito @raulraja

modules/free/arrow-free-data/src/main/kotlin/arrow/free/StackSafeMonadContinuation.kt
```diff
@@ -49,18 +49,6 @@ open class StackSafeMonadContinuation<F, A>(M: Monad<F>, override val context: C
   }
 }
 
-@Deprecated(
```
i-walker: This function doesn't exist in 0.9.0 stable, do we agree on just deleting it. It originated in "Next".

modules/core/arrow-syntax/src/main/kotlin/arrow/syntax/collections/array.kt
pakoito: fair

modules/free/arrow-free-data/src/main/kotlin/arrow/free/StackSafeMonadContinuation.kt
pakoito: I'm surprise it didn't exist. Maybe as a member function?

i-walker: Maybe.., but deleting it is fine right?

raulraja: I think we agreed in deleting this as part of the binding rework and decide to remove it because it was hard to refactor. We are gonna support this natively with plugins for all comprehensions that are flagged as stack safe without the Free runtime using suspension and a small loop.

modules/docs/arrow-docs/docs/docs/arrow/data/eithert/README.md
```diff
@@ -166,18 +172,18 @@ import arrow.fx.rx2.extensions.fx
 
 fun getCountryCode(personId: Int): ObservableK<Either<BizError, String>> =
   ObservableK.fx {
-    val person = findPerson(personId).bind()
-    val address = person.fold (
+    val person = !findPerson(personId)
```
JorgeCastilloPrz: are we removing `.bind()` ? I actually liked it more than `!`, never been super convinced about overriding that operator tbh.

i-walker: No we're allowing different styles. We still have other areas in the docs where we use bind, 
But I don't mind changing it back to `.bind()`

modules/core/arrow-core-data/src/main/kotlin/arrow/core/Try.kt
```diff
@@ -5,25 +5,23 @@ import arrow.higherkind
...
 
-/**
- * The `Try` type represents a computation that may either result in an exception, or return a
- * successfully computed value.
- *
- * Port of https://github.com/scala/scala/blob/v2.12.1/src/library/scala/util/Try.scala
- */
+@Deprecated(
+  "Try will be deleted soon as it promotes eager execution of effects, so it’s better if you work with Either’s suspend constructors or a an effect handler like IO",
+  ReplaceWith("Either<Nothing, A>")
```
pakoito: `Either<Throwable, A>`

modules/core/arrow-syntax/src/main/kotlin/arrow/syntax/collections/iterable.kt
```diff
@@ -1,18 +1,11 @@
...
-import arrow.core.toOption
-
-@Deprecated(message = "`firstOption` is now part of the Foldable interface and generalized to all foldable data types")
-fun <T> Iterable<T>.firstOption(): Option<T> = firstOrNull().toOption()
-
-@Deprecated(message = "`firstOption` is now part of the Foldable interface and generalized to all foldable data types")
-fun <T> Iterable<T>.firstOption(predicate: Predicate<T>): Option<T> = firstOrNull(predicate).toOption()
 
 @Deprecated(message = "`collect` is now part of the FunctorFilter interface and generalized to all filterable data types")
+@Suppress("UNCHECKED_CAST")
```
pakoito: Is the cast necessay? What's the inferred type?

i-walker:
```
f.andThen { optionB -> optionB.getOrElse { g(value) } } as (A) -> Option<B>
```
This Cast can't be type checked

modules/core/arrow-core-data/src/main/kotlin/arrow/core/Try.kt
i-walker: Thanks :)

modules/docs/arrow-docs/docs/docs/arrow/data/eithert/README.md
JorgeCastilloPrz: I guess it's a longer discussion we need to have outside of this PR. We should probable agree on one version and go with it for everything (including docs) cc @pakoito @nomisRev @raulraja. As said, not blocker for this PR. For now I guess you can keep it as is.

modules/core/arrow-syntax/src/main/kotlin/arrow/syntax/collections/iterable.kt
pakoito: Why not? If you break it down and ask the IDE to fill in the types, what is the suggested result? There is nothing in the definition that cannot be checked.

i-walker: the type checker expects (A) -> Any?

pakoito: That's telling us that the implementation is incorrect. You can see that `optionB.getOrElse { }` returns `Any?` because if it's present then it returns `B`, whereas if it isn't it returns `Option<B>` and that's never flattened. `optionB.fold({ Some(it) }, { g(value) })`

pakoito: `f(value).map { listOf(it) }.getOrElse { emptyList() }` can also be simplified to a single `fold`

modules/ank/arrow-ank/src/main/kotlin/arrow/ank/interpreter.kt
```diff
@@ -223,8 +223,7 @@ val interpreter: AnkOps = object : AnkOps {
           if (snippet.isPlayground)
             snippetContent.replace(snippet.fence, "{: data-executable='true'}\n\n```${snippet.lang}\n${snippet.code}\n```")
           else
-            snippetContent.replace(snippet.fence, "```${snippet.lang}\n${snippet.code}\n```")
-        },
+            snippetContent.replace(snippet.fence, "```${snippet.lang}\n${snippet.code}\n```") },
```
pakoito: This change makes no sense. Either fence the `else` clause with `{` or keep the other one on a new line

pakoito: A nit, not a blocker

modules/optics/arrow-optics/src/main/kotlin/arrow/optics/std/try.kt
```diff
@@ -2,17 +2,14 @@ package arrow.optics
...
 import arrow.core.Right
 import arrow.core.Success
 import arrow.core.Try
-import arrow.core.Invalid
 import arrow.core.Valid
 import arrow.core.Validated
 
-/**
- * [PPrism] to focus into an [arrow.core.Try.Success]
- */
```
pakoito: Missing deprecation notice. Not a blocker.

i-walker: I didnt know any replacements, because Try is getting deleted anyways

PR #1532, opened 2019-07-27
cleanUp on "stable" webpage

i-walker:
fixes some webpage problems users have and deletes references to SnapShot builds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,6 @@
 <img height="100" src="https://avatars2.githubusercontent.com/u/29458023?v=4&amp;s=200" width="100">
 
-[![Maven Central](https://maven-badges.herokuapp.com/maven-central/io.arrow-kt/arrow-core/badge.svg)](https://maven-badges.herokuapp.com/maven-central/io.arrow-kt/arrow-core)
-[![Build Status](https://travis-ci.org/arrow-kt/arrow.svg?branch=master)](https://travis-ci.org/arrow-kt/arrow/)
+[![Maven Central](https://img.shields.io/maven-central/v/io.arrow-kt/arrow-docs.svg?label=Maven%20Central)](https://search.maven.org/search?q=g:%22io.arrow-kt%22%20AND%20a:%22arrow-docs%22)
 [![Kotlin version badge](https://img.shields.io/badge/kotlin-1.3-blue.svg)](https://kotlinlang.org/docs/reference/whatsnew13.html)
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](http://www.apache.org/licenses/LICENSE-2.0)
 [![codecov](https://codecov.io/gh/arrow-kt/arrow/branch/master/graph/badge.svg)](https://codecov.io/gh/arrow-kt/arrow)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,7 +1,6 @@
 <img height="100" src="https://avatars2.githubusercontent.com/u/29458023?v=4&amp;s=200" width="100">
 
-[![Maven Central](https://maven-badges.herokuapp.com/maven-central/io.arrow-kt/arrow-core/badge.svg)](https://maven-badges.herokuapp.com/maven-central/io.arrow-kt/arrow-core)
-[![Build Status](https://travis-ci.org/arrow-kt/arrow.svg?branch=master)](https://travis-ci.org/arrow-kt/arrow/)
+[![Maven Central](https://img.shields.io/maven-central/v/io.arrow-kt/arrow-docs.svg?label=Maven%20Central)](https://search.maven.org/search?q=g:%22io.arrow-kt%22%20AND%20a:%22arrow-docs%22)
```
i-walker: This is one mavenCentral Badge with a 0.9.0 label

modules/docs/arrow-docs/docs/docs/README.md
```diff
@@ -90,7 +56,7 @@ dependencies {
     compile "io.arrow-kt:arrow-extras-data:$arrow_version"
     compile "io.arrow-kt:arrow-extras-extensions:$arrow_version"
     kapt    "io.arrow-kt:arrow-meta:$arrow_version"
-
+
```
pakoito: Unnecessary whitespace

pakoito:
Why aren't reviews necessary to merge here?

README.md
pakoito: The badge updates with the version in central. Is it necessary?

i-walker:
We're still working on a way to block it, but that's just for now..

README.md
i-walker: Yes, it is as of today it only displays the 0.8.2 maven central,  which is odd considering that 0.9.0 is the current stable version.

raulraja:
That is because the referenced artifact is wrong. We discussed before that
arrow-core does not exist in 0.9.0. this badge should point to
arrow-core-data currently and back to arrow-core in 0.10.0

i-walker:
Does this branch need any changes or do you approve?

modules/docs/arrow-docs/docs/docs/README.md
```diff
@@ -111,6 +76,7 @@ dependencies {
     compile "io.arrow-kt:arrow-query-language:$arrow_version" //optional
     compile "io.arrow-kt:arrow-integration-retrofit-adapter:$arrow_version" //optional
 }
+
```
pakoito: Additional line

modules/docs/arrow-docs/docs/docs/README.md
```diff
@@ -90,7 +56,6 @@ dependencies {
     compile "io.arrow-kt:arrow-extras-data:$arrow_version"
     compile "io.arrow-kt:arrow-extras-extensions:$arrow_version"
     kapt    "io.arrow-kt:arrow-meta:$arrow_version"
-
```
pakoito: Missing line between mandatory and optional

PR #1534, opened 2019-07-29
Add Traverse Docs

### LINKED ISSUE #333: `Traverse` docs

### LINKED ISSUE #498: traverse sample code
It would be really useful for beginners to have an example of how to traverse, for example a `Try`
from: Try<Sequence<List<Try<Int>>>>  to Try<Sequence<Try<Int>>>

I have tried to find an example in docs or kdoc without success :(

Thanks!

i-walker:
resolves #333 and resolves #498 
- [X] Add finished Docs to KDocs
- [x] finish several playground examples
- [x] finish text sections


modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+    .sequence_(Option.applicative())
+  //sampleEnd
+  println("someUnit= $someUnit")
+  println("noneUnit= $noneUnit")
+}
+```
+
+### Traversables are Foldable
+
+The `Foldable` type class abstracts over “things that can be folded over” similar to how `Traverse` abstracts over “things that can be traversed.” It turns out `Traverse` is strictly more powerful than `Foldable` - that is, `foldLeft` and `foldRight` can be implemented in terms of `traverse` by picking the right `Applicative`. However, arrow's `Traverse` does not implement `foldLeft` and `foldRight` as the actual implementation tends to be ineffecient, as we one can implement the `Foldable` `fold`'s.
```
i-walker: This is just a guess, but I am guessing that the Foldable fold is more efficient, because traverse can blowUp the stack if it isn't defined by fold. 
https://github.com/typelevel/cats/issues/1473

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
 {:.intermediate}
 intermediate
 
-TODO. Meanwhile you can find a short description in the [intro to typeclasses]({{ '/docs/typeclasses/intro/' | relative_url }}).
+In functional programming it is very common to encode "effects" as data types - common effects include `Option` for possibly missing values, `Either` and `Validated` for possible errors, and [`Promise`]({{ '/docs/effects/promise/' | relative_url }}) for asynchronous computations, which has the same purpose as a `Future`, only pure and lazy.
```
pakoito: "effects" is a word we use for other meaning. Use "behaviors" instead.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+
+Abstracting away the `G` (still imagining `F` to be `List`), `traverse` says given a collection of data, and a function that takes a piece of data and returns an effectful value, it will traverse the collection, applying the function and aggregating the effectful values (in a `List`) as it goes.
+
+In the most general form, `Kind<F, A>` is some sort of context `F` which may contain a value `A` (or several). While `List` tends to be among the most general cases, there also exist `Traverse` instances for `Option`, `Either`, and `Validated` (among others).
+
+#### Why not Foldable
+
+
+
+#### Choose your effect
```
i-walker: @pakoito Should it be 
```
### Choose your data type
```
Or something else

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
pakoito: Not resolved IMO.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
pakoito: ### Choose your implementation ?

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+```
+
+Note the return type `List<Promise<ForIO, Profile>>`. This makes sense given the type signatures, but seems unwieldy. We now have a list of asynchronous values, and to work with those values we must then use the combinators on `Promise` for every single one. It would be nicer instead if we could get the aggregate result in a single `Promise`, say a `Promise<ForIO, List<Profile>>`.
+
+There exists a much more generalized form that would allow us to parse a `List<String>` or validate credentials for a `List<User>`.
+
+Enter `Traverse`.
+
+### The Typeclass
+At center stage of Traverse is the traverse method.
```
pakoito: `traverse` method. Between backticks

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+import arrow.typeclasses.Applicative
+import arrow.typeclasses.Foldable
+import arrow.typeclasses.Functor
+
+interface Traverse<F> : Functor<F>, Foldable<F> {
+  fun <G, A, B> Kind<F, A>.traverse(AP: Applicative<G>, f: (A) -> Kind<G, B>): 
+}
+```
+
+**Essentially, one uses `Traverse`, when there is a function `(A) -> Kind<G, B>` and `G` is an `Applicative` instance e.g.: `(User) -> Promise<ForIO, Profile>` and you want to `lift` the previous function to work on `Kind<F, A>` where `F` is a `Traverse` instance e.g.: `List<User>` and return `Kind<G, Kind<F, B>>` e.g.: `Promise<ForIO, List<Profile>>`**
```
pakoito: This paragraph is too dense. I would split it.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+
+In our above example, `F` is `List`, and `G` is `Option`, `Either`, or `Promise`. For the profile example, traverse says given a `List<User>` and a function `(User) -> Promise<ForIO, Profile>`, it can give you a `Promise<ForIO, List<Profile>>`.
+
+Abstracting away the `G` (still imagining `F` to be `List`), `traverse` says given a collection of data, and a function that takes a piece of data and returns an effectful value, it will traverse the collection, applying the function and aggregating the effectful values (in a `List`) as it goes.
+
+In the most general form, `Kind<F, A>` is some sort of context `F` which may contain a value `A` (or several). While `List` tends to be among the most general cases, there also exist `Traverse` instances for `Option`, `Either`, and `Validated` (among others).
+
+#### Why not Foldable
+
```
pakoito: Empty section?

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+
+
+#### Choose your effect
+
+The type signature of `Traverse` appears highly abstract, and indeed it is - what `traverse` does as it walks the `Kind<F, A>` depends on the effect of the function. Let's see some examples where `F` is taken to be `List`.
```
pakoito: `, and indeed it is` --> `, although it's easier if you think about collections of executable operations`

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+
+```kotlin:ank
+import arrow.core.ValidatedNel
+import arrow.core.invalidNel
+import arrow.core.toT
+import arrow.core.validNel
+import arrow.core.Either
```
pakoito: This one is carried from a prior snippet.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+  //sampleEnd
+  println(validatedList)
+}
+```
+
+Notice that in the `Either` case, should any string fail to parse the entire `traversal` is considered a failure. Moreover, once it hits its first bad parse, it will not attempt to parse any others down the line (similar behavior would be found with using `Option` as the effect). Contrast this with `Validated` where even if one bad parse is hit, it will continue trying to parse the others, accumulating any and all errors as it goes. The behavior of traversal is closely tied with the `Applicative` behavior of the data type.
+
+Going back to our `Promise` example, we can get an `Applicative` instance for `IO`, by converting `Promise` to `IO` that runs each `Promise` concurrently. Then when we traverse a `List<A>` with an `(A) -> Promise<ForIO, B>`, we can imagine the traversal as a scatter-gather. Each `A` creates a concurrent computation that will produce a `B` (the scatter), and as the `Promise`s complete they will be gathered back into a `List`.
+
+{TODO: finish example with Promise}
```
pakoito: TODO

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+fun validateLogin(c: Credentials): Either<SecurityError, Unit> = TODO()
+
+fun userInfo(u: User): Promise<ForIO, Profile> = TODO()
+
+data class Admin(val name: String, val crd: Credentials): User
+
+fun ListK<User>.processLogin()
+```
+
+Evidently, `Traverse` is not limited to `List` or `Nel`, it provides an abstraction over 'things that can be traversed over', like a Binary tree, hence the name `Traverse`.
```
pakoito: `like a Binary tree, a Sequence, or a Stream`

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+
+#### Playing with `Reader`
+
+Another interesting effect we can use is Reader. Recall that a `Reader<D, A>` is a type alias for `Kleisli<ForId, D, A>` which is a wrapper around `(D) -> A`.
```
pakoito: A bit more context. Reader is an `injector for a parameter` or something similar.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+import arrow.core.Id
+import arrow.core.ForId
+import arrow.core.ListK
+import arrow.core.extensions.id.applicative.applicative
+import arrow.mtl.KleisliPartialOf
+import arrow.mtl.extensions.kleisli.applicative.applicative
+import arrow.mtl.fix
+
+fun processTopics(topics: ListK<Topic>): Job<ListK<Result>> =
+  topics.traverse<KleisliPartialOf<ForId, Context>, Result>(Job.applicative<ForId, Context>(Id.applicative())) {
```
pakoito: I'd extract the instance `Job.applicative<ForId, Context>(Id.applicative())` to a field here.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+    processTopic(it)
+  }.fix()
+```
+
+Note the nice return type - `Job<List<Result>>`. We now have one aggregate `Job` that when run, will go through each topic and run the topic-specific job, collecting results as it goes. We say "when run" because a `Job` is some function that requires a Context before producing the value we want.
+
+One example of a "context" can be found in the [Spark](http://spark.apache.org/) project. In Spark, information needed to run a Spark job (where the master node is, memory allocated, etc.) resides in a `SparkContext`. Going back to the above example, we can see how one may define topic-specific Spark jobs `(type Job<A> = Reader<SparkContext, A>)` and then run several Spark jobs on a collection of topics via `traverse`. We then get back a `Job<List<Result>>`, which is equivalent to (`SparkContext) -> List<Result>`. When finally passed a `SparkContext`, we can run the job and get our results back.
```
pakoito: I'd also add an example of Android's Context.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+
+Moreover, the fact that our aggregate job is not tied to any specific `SparkContext` allows us to pass in a `SparkContext` pointing to a production cluster, or (using the exact same job) pass in a test `SparkContext` that just runs locally across threads. This makes testing our large job nice and easy.
+
+Finally, this encoding ensures that all the jobs for each topic run on the exact same cluster. At no point do we manually pass in or thread a `SparkContext` through - that is taken care for us by the (applicative) effect of `Reader` and therefore by `traverse`.
+
+### Sequencing
```
pakoito: TBH given the number of requests we get I'd start with sequencing, and then explaining that traverse is a specialization.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+  val emptyList =
+    listOf(1.some(), none(), 3.some())
+      .sequence(Option.applicative())
+  //sampleEnd
+  println("optionList = $optionList")
+  println("emptyList = $emptyList")
+}
+```
+
+-- Add Text
```
pakoito: TODO

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+
+### Traversables are Functors
+
+{TODO: finish this section}
```
pakoito: TODO

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+
+### Traversing for effect
```
pakoito: sequence_ is in foldable, isn't it?

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,366 @@ redirect_from:
...
+```
+
+### Traversables are Foldable
+
+The `Foldable` type class abstracts over “things that can be folded over” similar to how `Traverse` abstracts over “things that can be traversed.” It turns out `Traverse` is strictly more powerful than `Foldable` - that is, `foldLeft` and `foldRight` can be implemented in terms of `traverse` by picking the right `Applicative`. However, arrow's `Traverse` does not implement `foldLeft` and `foldRight` as the actual implementation tends to be ineffecient, as we one can implement the `Foldable` `fold`'s.
+
+For brevity and demonstration purposes we’ll implement an isomorphic `foldMap` method in terms of `traverse` by using `arrow.typeclasses.Const`. You can then implement `foldRight` in terms of `foldMap`, and `foldLeft` can then be implemented in terms of `foldRight`, though the resulting implementations may be slow.
+
+```kotlin:ank
+// TODO: finish this section
```
pakoito: TODO

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
i-walker: I don't have a better idea, I would go for your suggestion :)

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
i-walker: yes

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
i-walker: I deleted this section because the next paragraph explains the same concept, just with more concrete Datatypes. I am shifting away from this style of writing.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
i-walker: I purpose to add a more tagless final approach, so Users are not inclined to use MTL.
What do you think @pakoito ?

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,491 @@ redirect_from:
...
+  val promise: IO<Promise<ForIO, Profile>> =
+    Promise.uncancelable<ForIO, Profile>(IO.async()).fix()
+
+  fun userInfo(u: User): IO<Profile> =
+    promise.flatMap { p -> p.complete(DummyProfile(u)).flatMap { p.get() } }
+
+  fun ListK<User>.processLogin() =
+    traverse(IO.applicative()) { userInfo(it) }
+
+  val res = listOf(DummyUser("Micheal"), DummyUser("Juan"), DummyUser("T'Challa")).k()
```
i-walker: As a todo I will turn this into an Fx style

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,491 @@ redirect_from:
...
+import arrow.typeclasses.Applicative
+import arrow.typeclasses.Foldable
+import arrow.typeclasses.Functor
+
+interface Traverse<F> : Functor<F>, Foldable<F> {
+  fun <G, A, B> Kind<F, A>.traverse(AP: Applicative<G>, f: (A) -> Kind<G, B>): Kind<G, Kind<F, B>>
+}
+```
+
+In our above example, `F` is `List`, and `G` is `Option`, `Either`, or `Promise`. For the profile example, traverse says given a `List<User>` and a function `(User) -> Promise<ForIO, Profile>`, it can give you a `Promise<ForIO, List<Profile>>`.
```
rcd27: So `F` could be named a `container`, am I right?

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,491 @@ redirect_from:
...
+```kotlin:ank
+import arrow.core.Either
+import arrow.core.Option
+import arrow.fx.ForIO
+import arrow.fx.Promise
+
+interface Profile
+interface User
+
+fun userInfo(u: User): Promise<ForIO, Profile> = TODO()
```
adam-arold: This is just some nitpicking but `userInfo` would be better suitable for a field name. What do you think about `loadUserInfo`, or something similar?

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,491 @@ redirect_from:
...
+```
+
+Each function asks only for the data it actually needs; in the case of `userInfo`, a single `User`. We certainly could write one that takes a `List<User>` and fetch profile for all of them, would be a bit strange. If we just wanted to fetch the profile of just one user, we would either have to wrap it in a `List` or write a separate function that takes in a single user anyways. More fundamentally, functional programming is about building lots of small, independent pieces and composing them to make larger and larger pieces - does this hold true in this case?
```
adam-arold:
```suggestion
Each function asks only for the data it actually needs; in the case of `userInfo`, a single `User`. We could certainly write one which takes a `List<User>` and fetches a profile for all of them, but it would be a bit strange. If we wanted to fetch the profile of just one user, we would either have to wrap it in a `List` or write a separate function that takes in a single user anyways. More fundamentally, functional programming is about building lots of small, independent pieces and composing them to make larger and larger pieces - does this hold true in this case?
```

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
adam-arold: I'd love to have readable type parameters instead of `G`, `A`, `B` which convey no information in themselves.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,491 @@ redirect_from:
...
+  
+  // val optionMapKBoilered = map.foldLeft(Some(emptyMap())) { acc: Option<MapK<String, String>>, i: Int ->
+  //   acc.fold({ emptyMap() }, { /*Some logic to retrieve the key of a value, transform it and add it to the accumulated Map*/ })
+  // }
+  //sampleEnd
+  println(optionMapK)
+}
+```
+
+Both values try to attain the same thing, but what `Foldable` lacks is that it solely drills down to it's `A` here `Int` and does not preserve it's shape `F` - `MapK`. This is where `Traverse` shines, whenever you care about the Output `B` from `(A) -> B` and the existing shape of `F` you would use `traverse`.
```
adam-arold: This is a great comparison! These nuggets of information are the ones which I'm missing from the rest of the docs! Great work! 👍

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,491 @@ redirect_from:
...
+
+Additionally, your able to wrap your context `F` within a `G`. That is, may among others, the reason why `Traverse` is strictly more powerful than `Foldable`.
```
adam-arold:
```suggestion
Additionally, your are able to wrap your context `F` within a `G`. That is, among others, the reason why `Traverse` is strictly more powerful than `Foldable`.
```

adam-arold: I'd add some counterexamples to justify the existence of `Foldable`. If `Traverse` is a strict upgrade why would I use `Foldable`?

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,491 @@ redirect_from:
...
+fun main() {
+  //sampleStart
+  val validatedList = listOf("1", "22w", "3", "33s").k()
+    .traverse(ValidatedNel.applicative(Nel.semigroup<NumberFormatException>()), ::parseIntValidated).fix()
+  //sampleEnd
+  println(validatedList)
+}
+```
+
+Notice that in the `Either` case, should any string fail to parse the entire `traversal` is considered a failure. Moreover, once it hits its first bad parse, it will not attempt to parse any others down the line (similar behavior would be found with using `Option`). Contrast this with `Validated` where even if one bad parse is hit, it will continue trying to parse the others, accumulating any and all errors as it goes. The behavior of traversal is closely tied with the `Applicative` behavior of the data type.
```
adam-arold:
```suggestion
Notice that in the `Either` case, should any string fail to parse the entire `traverse` is considered a failure. Moreover, once it hits its first bad parse, it will not attempt to parse any others down the line (similar behavior would be found with using `Option`). Contrast this with `Validated` where even if one bad parse is hit, it will continue trying to parse the others, accumulating any and all errors as it goes. The behavior of traversal is closely tied with the `Applicative` behavior of the data type.
```

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,491 @@ redirect_from:
...
+
+Going back to our `Promise` example, we can get an `Applicative` instance for `IO`, by converting `Promise` to `IO` that runs each `Promise` concurrently. Then when we traverse a `List<A>` with an `(A) -> Promise<ForIO, B>`, we can imagine the traversal as a scatter-gather. Each `A` creates a concurrent computation that will produce a `B` (the scatter), and as the `Promise`s complete they will be gathered back into a `List`.
```
adam-arold:
```suggestion
Going back to our `Promise` example, we can get an `Applicative` instance for `IO`, by converting `Promise` to `IO` that runs each `Promise` concurrently. Then when we traverse a `List<A>` with an `(A) -> Promise<ForIO, B>`, we can imagine the traversal as a divide & conquer. Each `A` creates a concurrent computation that will produce a `B` (divide), and as the `Promise`s complete they will be gathered back into a `List` (conquer).
```

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
i-walker: They have to be polymorphic :) @adam-arold
Yes @rcd27 , `F` can be seen as a "container", to be precise it is a parameter for type constructors, e.g.: containers like List, Either..

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
i-walker: there is a section down below which explains that

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
adam-arold: @i-walker what I meant is that you could use `CONTAINER` instead of `T` in the example above. The point is that the generic type parameter would convey what it stands for not just an arbitrary letter like `F`.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
i-walker: thx

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,495 @@ redirect_from:
...
+  
+  // val optionMapKBoilered = map.foldLeft(Some(emptyMap())) { acc: Option<MapK<String, String>>, i: Int ->
+  //   acc.fold({ emptyMap() }, { /*Some logic to retrieve the key of a value, transform it and add it to the accumulated Map*/ })
+  // }
+  //sampleEnd
+  println(optionMapK)
+}
+```
+
+Both values try to attain the same thing, but what `Foldable` lacks is that it solely drills down to it's `A` here `Int` and does not preserve it's shape `F` - `MapK`. This is where `Traverse` shines, whenever you care about the Output `B` from `(A) -> B` and the existing shape of `F` you would use `traverse`.
```
AdrianRaFo:
```suggestion
Both values try to attain the same thing, but what `Foldable` lacks is that it solely drills down to it's `A` here `Int` and does not preserve it shapes `F` - `MapK`. This is where `Traverse` shines, whenever you care about the Output `B` from `(A) -> B` and the existing shape of `F` you would use `traverse`.
```

i-walker:
```suggestion
Both values try to attain the same thing, but what `Foldable` lacks is that it solely drills down to it's `A` here `Int` and does not preserve its shape `F` - `MapK`. This is where `Traverse` shines, whenever you care about the Output `B` from `(A) -> B` and the existing shape of `F` you would use `traverse`.
```
Thx, I meant `its` instead of `it's`

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,495 @@ redirect_from:
 {:.intermediate}
 intermediate
 
-TODO. Meanwhile you can find a short description in the [intro to typeclasses]({{ '/docs/typeclasses/intro/' | relative_url }}).
+In functional programming it is very common to encode "behaviors" as data types - common behaviors include `Option` for possibly missing values, `Either` and `Validated` for possible errors, and [`Promise`]({{ '/docs/effects/promise/' | relative_url }}) for asynchronous computations, which has the same purpose as a `Future`, only pure and lazy.
```
JorgeCastilloPrz: I think we should always add links to all data types / typeclasses we mention first time in an article. People doesn't read docs in order so they'll get to this page sometimes even knowing about other essential stuff. (Meaning to add links for Option, Either, Validated here, not just for Promise)

i-walker:
Eventually, all docs need to go to KDocs. That's Why I start writing them at the regular place, where users expect to see Docs and copy them to KDocs. Once, we fully integrate KDocs in Arrow, I'll be glad to delete the legacy docs and then we can start promoting KDocs.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,495 @@ redirect_from:
...
+  //sampleEnd
+  println(optionMapK)
+}
+```
+
+Both values try to attain the same thing, but what `Foldable` lacks is that it solely drills down to it's `A` here `Int` and does not preserve its shape `F` - `MapK`. This is where `Traverse` shines, whenever you care about the Output `B` from `(A) -> B` and the existing shape of `F` you would use `traverse`.
+
+Additionally, you're able to wrap your context `F` within a `G`. That is, may among others, the reason why `Traverse` is strictly more powerful than `Foldable`.
+
+You can also misuse it's powers - where a `map` is more considerable:
```
JorgeCastilloPrz: What do you mean with this?

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,495 @@ redirect_from:
...
+fun main() {
+  //sampleStart
+  val validatedList = listOf("1", "22w", "3", "33s").k()
+    .traverse(ValidatedNel.applicative(Nel.semigroup<NumberFormatException>()), ::parseIntValidated).fix()
+  //sampleEnd
+  println(validatedList)
+}
+```
+
+Notice that in the `Either` case, should any string fail to parse the entire `traverse` is considered a failure. Moreover, once it hits its first bad parse, it will not attempt to parse any others down the line (similar behavior would be found with using `Option`). Contrast this with `Validated` where even if one bad parse is hit, it will continue trying to parse the others, accumulating any and all errors as it goes. The behavior of `traverse` is closely tied with the `Applicative` behavior of the data type.
```
JorgeCastilloPrz: I like these examples 👍

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
JorgeCastilloPrz: Exactly. The most usual problem is the one about "reversing the types" which is what sequence resolves. You can start with sequence, then in the end unveil its equivalent to traverse with identity, then explain Traverse and how it's even more powerful being able to use other transformation functions.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,495 @@ redirect_from:
...
+
+Abstracting away the `G` (still imagining `F` to be `List`), `traverse` says given a collection of data, and a function that takes a piece of data and returns an value `B` wrapped in a container `G`, it will traverse the collection, applying the function and aggregating the values (in a `List`) as it goes.
+
+In the most general form, `Kind<F, A>` is some sort of context `F` which may contain a value `A` (or several). While `List` tends to be among the most general cases, there also exist `Traverse` instances for `Option`, `Either`, and `Validated` (among others).
+
+Essentially, one uses `Traverse`, when there is a function `(A) -> Kind<G, B>` and `G` is an `Applicative` instance, thus `G` provides `just` and `map` functions e.g.: just as `Promise<ForIO, B>` provides and you want to `lift` the previous function to work on `Kind<F, A>` where `F` is a `Traverse` instance - and you still care about `A` (the `Profiles`) `(List<User>)` and return `Kind<G, Kind<F, B>>` e.g.: `Promise<ForIO, List<Profile>>`
```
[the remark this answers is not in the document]
i-walker: I choose to remove it. It really creates more overhead then it helps. I added it to stress a more generalized use case for Traverse. I hope it's still derivable from the other explanations.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,495 @@ redirect_from:
...
+import arrow.core.fix
+import arrow.core.k
+
+fun main(){
+  //sampleStart
+  val map: MapK<String, Int> = mapOf("one" to 1, "two" to 2).k()
+
+  val optionMapK: Option<MapK<String, String>> = map.traverse(Option.applicative()) { Some("$it") }.fix()
+  
+  // val optionMapKBoilered = map.foldLeft(Some(emptyMap())) { acc: Option<MapK<String, String>>, i: Int ->
```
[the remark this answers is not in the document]
i-walker: What do you mean exactly, I didn't want to implement the fold version, because the result is just a nasty way to ducktape a new `MapK`

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
i-walker: Here I wanted to give a counter-example, how not to use `Traverse`. Here it results in an inefficient implementation because we `wrap` and `unwrap` the `Int`

i-walker: I think I had to add that exact sentence from above to the markdown, haha

i-walker:
I excuse my delay in this PR. I will work on it after the 23th Dec.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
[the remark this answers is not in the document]
i-walker: Done

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/traverse/README.md
```diff
@@ -11,16 +11,495 @@ redirect_from:
...
+import arrow.fx.ForIO
+import arrow.fx.Promise
+
+interface Profile
+interface User
+
+fun userInfo(u: User): Promise<ForIO, Profile> = TODO()
+```
+
+Each function asks only for the data it actually needs; in the case of `userInfo`, a single `User`. Certainly, we could write one that takes a `List<User>` and fetches profile for all of them,but it would be a bit strange. If we just wanted to fetch the profile of just one user, we would either have to wrap it in a `List` or write a separate function that takes in a single user anyways. More fundamentally, functional programming is about building lots of small, independent pieces and composing them to make larger and larger pieces - does this hold true in this case?
```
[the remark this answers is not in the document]
i-walker: Makes sense. I edited it

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ * }
+ * ```
+ *
+ * Both methodologies try to attain the same thing, but what [Foldable] lacks is that it solely drills down to it's `A` here `Int` and does not preserve it's shape `F` - `MapK`. Resulting in a tiresome implementation, where we need to come up with an algorithm to resolve a key of a given value in the `Map` - let alone that this algorithm is not universal over any given `Map`.
+ *
+ * This is where [Traverse] shines, whenever you care about the Output `B` from `(A) -> B` and the existing shape of `F` you may use [traverse].
+ *
+ * Additionally, you're able to wrap your context `F` within a `G`. That is, may among others, the reason why [Traverse] is strictly more powerful than [Foldable].
+ *
+ * However, for certain computations [traverse] may also be simplified to a `map`:
```
1Jajen1: I don't think anyone would do this in the wild, probably better to go the other way around and argue that `traverse` over the identity functor also defines map. Not sure tho

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ * fun main() {
+ *   //sampleStart
+ *   val validatedList = listOf("1", "22w", "3", "33s").k()
+ *     .traverse(ValidatedNel.applicative(Nel.semigroup<NumberFormatException>()), ::parseIntValidated).fix()
+ *   //sampleEnd
+ *   println(validatedList)
+ * }
+ * ```
+ *
+ * Notice that in the `Either` case, should any string fail to parse the entire [traverse] is considered a failure. Moreover, once it hits its first bad parse, it will not attempt to parse any others down the line (similar behavior would be found with using `Option`). Contrast this with `Validated` where even if one bad parse is hit, it will continue trying to parse the others, accumulating any and all errors as it goes. The behavior of [traverse] is closely tied with the [Applicative] behavior of the data type, where computations are runed in isolation.
```
1Jajen1:
```suggestion
 * Notice that in the `Either` case, should any string fail to parse the entire [traverse] is considered a failure. Moreover, once it hits its first bad parse, it will not attempt to parse any others down the line (similar behavior would be found with using `Option`). Contrast this with `Validated` where even if one bad parse is hit, it will continue trying to parse the others, accumulating any and all errors as it goes. The behavior of [traverse] is closely tied with the [Applicative] behavior of the data type, where computations are run in isolation.
```

This a good place to introduce three other concepts as well: 
`traverse` order (which depends on `Foldable` but is generally always left to right for most of arrows datatypes).
And also laziness, `traverse` (with an applicative that supports it, by implementing `lazyAp`) will not perform unnecessary steps and work on infinite structures.
And lasty the evaluation order of your applicative matters: For all arrow types it is left to right (listOf(1,2,3,4).traverse { IO { println(it); it } } puts out 1,2,3,4 and returns the list in the same order ) but you can easily define a type `Backwards` that executes the effects in reverse order (by flipping argument order of ap). Backwards would still return the list in same order but would print 4,3,2,1.

Not sure which of these concepts are important enough to be included, your choice :), but this could be a good place for it

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ *
+ * Going back to our `IO` example from the beginning with concurrency in mind, we can get an [Applicative] instance for `IO`, by using `parApplicative`. Then when we traverse a `List<A>` with its [Traverse] instance `ListK.traverse()` and a function`(A) -> IO<B>`, we can imagine the traversal as a scatter-gather. Each `A` creates a concurrent computation that will produce a `B` (the scatter), and as the `IO` operations completes they will be gathered back into a `List`.
```
1Jajen1: It is worth mentioning that this will not short circuit the same way as most other applicatives because `parApplicative` does not implement `lazyAp` and creates all `IO`'s upfront and executes all at the same time. Short circuit behaviour is then done by `raceN` and sorts where the first computation to fail with cancel all others, but they still may run. With a normal applicative an effect is only run if the previous ones were successful. Also `parApplicative` breaks monad-applicative-consistency laws, which is by design ^^

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ *
+ * Ultimately, we use `parTraverse` that runs each operation in parallel:
```
1Jajen1: This sounds weird ^-^
```suggestion
 * Ultimately, we have the `parTraverse` as a shorthand for `IO` which calls `traverse` using the `parApplicative`:
```

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ * fun main() {
+ *   unsafe { runBlocking { program() } }
+ * }
+ * ```
+ *
+ * Evidently, [Traverse] is not limited to [List] or [Nel], it provides an abstraction over 'things that can be traversed over', like a Binary tree, a [SequenceK], or a `Stream`, hence the name [Traverse].
+ *
+ * #### Playing with `Reader`
+ *
+ * Another interesting data type we can use is `Reader`. Recall that a `Reader<D, A>` is a type alias for `Kleisli<ForId, D, A>` which is a wrapper around `(D) -> A`.
```
1Jajen1:
```suggestion
 * Another interesting data type we can use is `Reader`. Recall that a `Reader<D, A>` is a type alias for `Kleisli<ForId, D, A>` which is a wrapper around `(D) -> Id<A>`.
```
`Kleisli<ForId, D, A>` is just that, you could and probably should add that that is isomorphic to `(D) -> A` tho.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ * interface Result
+ *
+ * typealias Job<A> = Reader<Context, A>
+ *
+ * fun processTopic(t: Topic): Job<Result> = TODO()
+ * ```
+ *
+ * We can imagine we have a data pipeline that processes a bunch of data, each piece of data being categorized by a topic. Given a specific topic, we produce a `Job` that processes that topic. (Note that since a `Job` is just a `Reader`/`Kleisli`, one could write many small `Jobs` and compose them together into one `Job` that is used/returned by `processTopic`.)
+ *
+ * Corresponding to bunch of topics, a `List<Topic>` if you will. Since `Reader` has an [Applicative] instance, we can [traverse] over this list with `processTopic`.
```
1Jajen1:
```suggestion
 * Corresponding to bunch of topics, a `List<Topic>` if you will. Since `Reader` has an [Applicative] instance, we can [traverse] over this list.
```
The method name is an implementation detail and not related to `traverse`. Mentioning it there may cause confusion as to what that method actually is I think

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ *
+ * val JobForContext = Job.applicative<ForId, Context>(Id.applicative())
+ *
+ * fun processTopics(topics: ListK<Topic>): Job<ListK<Result>> =
+ *   topics.traverse<KleisliPartialOf<ForId, Context>, Result>(JobForContext) {
+ *     processTopic(it)
+ *   }.fix()
+ * ```
+ *
+ * Note the nice return type - `Job<List<Result>>`. We now have one aggregate `Job` that when run, will go through each topic and run the topic-specific job, collecting results as it goes. We say "when run" because a `Job` is some function that requires a Context before producing the value we want.
```
1Jajen1:
```suggestion
 * Note the nice return type - `Job<List<Result>>`. We now have one aggregate `Job` that, when run, will go through each topic and run the topic-specific job, collecting results as it goes. We say "when run" because a `Job` is some function that requires a Context before producing the value we want.
```

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ *
+ * Moreover, the fact that our aggregate job is not tied to any specific `SparkContext` allows us to pass in a `SparkContext` pointing to a production cluster, or (using the exact same job) pass in a test `SparkContext` that just runs locally across threads. This makes testing our large job nice and easy.
+ *
+ * Finally, this encoding ensures that all the jobs for each topic run on the exact same cluster. At no point do we manually pass in or thread a `SparkContext` through - that is taken care for us by the (applicative) behavior of `Reader` and therefore by [traverse].
+ *
+ * ### Traversables are Functors
+ *
+ * As it turns out every [Traverse] is a lawful [Functor]. By carefully picking the `G` to use in [traverse] we can implement `map`.
```
1Jajen1: This is a duplicate, but I like this section much more ^^

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ *
+ * object IdApplicative : Applicative<ForId> {
+ *   override fun <A, B> IdOf<A>.ap(ff: IdOf<(A) -> B>): Id<B> =
+ *     fix().ap(ff)
+ *
+ *   override fun <A, B> IdOf<A>.map(f: (A) -> B): Id<B> =
+ *     fix().map(f)
+ *
+ *   override fun <A> just(a: A): Id<A> =
+ *     Id.just(a)
```
1Jajen1: While technically correct it may be better to show the actual implementation instead of some delegate. These methods are easy enough anyway. Also skip the map implementation, it is only needed because kotlin can't choose between the superclass Functor one and the Applicative default impl, but a reader doesn't care about that.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ * }
+ * ```
+ *
+ * We can implement [map] by wrapping and unwrapping [Id] as necessary.
+ *
+ * ### Theory Wrap-up
+ *
+ * Unsurprisingly, [Foldable] and [Traverse] act on multiple elements and reduce them into a single value - in cat theory - `Catamorphisms`.
```
1Jajen1:
```suggestion
 * [Foldable] and [Traverse] act on multiple elements and reduce them into a single value - in cat theory - `Catamorphisms`.
```
Gives of the wrong tone imo ^^

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ *
+ * In contrast, `homomorphisms` such as `Monoid` and [Applicative] preserve their structure, hence adding two `List<Int>` will yield a `List<Int>` or an [Applicative] example, where `List<Int>.map(::toString)` results in a `List<String>`.
```
1Jajen1:
```suggestion
 * In contrast, `homomorphisms` such as `Monoid` and [Applicative] preserve their structure, hence adding two elements of type `List<Int>` will yield a `List<Int>` or an [Applicative] example, where `List<Int>.map(::toString)` results in a `List<String>`.
```
The second example is a functor rather than applicative example, not that it matters, just inaccurate

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ *
+ * Given just `(User) -> IO<Profile>`, what should we do if we want to fetch profiles for a `List<User>`? We could try familiar combinators like map.
+ *
+ * ```kotlin:ank
+ * fun profilesFor(users: List<User>): List<IO<Profile>> = users.map(::userInfo)
+ * ```
+ *
+ * Note the return type `List<IO<Profile>>`. This makes sense given the type signatures, but seems unwieldy. We now have a list of asynchronous values, and to work with those values we must then use the combinators on `IO` for every single one. It would be nicer instead if we could get the aggregate result in a single `IO`, say a `IO<List<Profile>>`.
+ *
+ * ### Sequencing
```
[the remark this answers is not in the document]
i-walker: @pakoito and @JorgeCastilloPrz suggested to start first with Sequence and then explain `traverse`. in that sense, I reverted the former structure. Maybe we can add something tangible in the Text, to add guide the reader towards a redline. If you have something in mind, feel free to comment it :)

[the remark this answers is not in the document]
1Jajen1: It is a good idea to start with `sequence`, but not by explaining it as `traverse` + `id`. I'd say don't mention `traverse` at all up until you introduce `traverse` and then quickly mention how `sequence` is `traverse(AP, ::identity)`

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
i-walker: I'll delete that section. Down below I already explain how one can define map with traverse using `Id`

i-walker: In fact I will add it down below to that example as you said :) Thanks so much @1Jajen1

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
i-walker: Done :)

i-walker:
Thanks @1Jajen1 for the review. I toke another spin on it for a clearer redline. Comment here if you have something in mind :)

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
i-walker: The intention here is to further highlight the power and versatility of `Traverse`, through different angles.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ *
+ * In general this holds:
+ *
+ * ```kotlin
+ * map(::f).sequence(AP) == traverse(AP, ::f)
+ * ```
+ *
+ * where AP stands for an `Applicative<G>`, which is in the prior snippet `Applicative<ForOption>`.
+ *
+ * There exists a much more generalized and powerful form that would allow us to parse a `List<String>` or validate credentials for a `List<User>`.
```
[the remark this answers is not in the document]
1Jajen1: The way the introduction is written is better now, but now this doesn't fit because it was already introduced :see_no_evil:

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ *   val validatedList = listOf("1", "22w", "3", "33s").k()
+ *     .traverse(ValidatedNel.applicative(Nel.semigroup<NumberFormatException>()), ::parseIntValidated).fix()
+ *   //sampleEnd
+ *   println(validatedList)
+ * }
+ * ```
+ *
+ * Notice that in the [Either] case, should any string fail to parse the entire [traverse] is considered a failure. Moreover, once it hits its first bad parse, it will not attempt to parse any others down the line (similar behavior would be found with using `Option`). Contrast this with `Validated` where even if one bad parse is hit, it will continue trying to parse the others, accumulating any and all errors as it goes. The behavior of [traverse] is closely tied with the [Applicative] behavior of the data type, where computations are run in isolation.
+ *
+ * Going back to our `IO` example from the beginning with concurrency in mind, we can get an [Applicative] instance for `IO`, that will not short-circuit, by using `parApplicative`.
```
1Jajen1:
```suggestion
 * Going back to our `IO` example from the beginning with concurrency in mind, we can get an [Applicative] instance for `IO`, that traverses concurrently, by using `parApplicative`. This is in contrast to usual `IO.applicative()` which executes in sequence.
```
Don't think short-circuit is whats important at this point. I would however stretch the differences between this and the default applicative for IO

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ *
+ * It is worth mentioning that `parApplicative` does not implement `lazyAp` in contrast to `IO.applicative()`. `parApplicative` creates all IO's upfront and executes them in parallel. More importantly, `parApplicative` breaks monad-applicative-consistency laws by design. Consequently, `IO.applicative()` will only run an effect, when the previous one is successful.
```
1Jajen1:
```suggestion
 * It is worth mentioning that `parApplicative` does not implement `lazyAp` in contrast to `IO.applicative()`, which means `parApplicative` creates all IO's upfront and executes them in parallel. If one fails, the resulting `IO` fails, but the effects may still have already run. As a consequence `parApplicative` breaks monad-applicative-consistency laws which is by design. On the flipside, `IO.applicative()` will only run an effect, when the previous one is successful.
```
I think this is clearer. Having implemented `lazyAp` I know what it does and what implications it holds, but a first time reader won't. This shows the significance of it in contrast to the normal applicative.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ *
+ * Then when we traverse a `List<A>` with its [Traverse] instance `ListK.traverse()` and a function`(A) -> IO<B>`, we can imagine the traversal as a scatter-gather. Each `A` creates a concurrent computation that will produce a `B` (the scatter), and as the `IO` operations completes they will be gathered back into a `List`.
```
1Jajen1:
```suggestion
 * Then, when we traverse a `List<A>` with its [Traverse] instance `ListK.traverse()` and a function`(A) -> IO<B>`, we can imagine the traversal as a scatter-gather. Each `A` creates a concurrent computation that will produce a `B` (the scatter), and as the `IO` operations completes they will be gathered back into a `List`.
```

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ *     .processLogin().bind()
+ *   println(list)
+ * }
+ * //sampleEnd
+ * fun main() {
+ *   unsafe { runBlocking { program() } }
+ * }
+ * ```
+ *
+ * Evidently, [Traverse] is not limited to [List] or [Nel], it provides an abstraction over 'things that can be traversed over', like a Binary tree, [SequenceK], or a `Stream`, hence the name [Traverse].
```
1Jajen1:
```suggestion
 * [Traverse] is not limited to [List] or [Nel], it provides an abstraction over 'things that can be traversed over', like a Binary tree, [SequenceK], or a `Stream`, hence the name [Traverse].
```
"Evidently" makes no sense if all we've done so far is traversing List or List-like structures

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,519 @@ import arrow.core.Id
...
+ * ```kotlin:ank
+ * import arrow.core.ForId
+ * import arrow.core.Id
+ * import arrow.core.IdOf
+ * import arrow.core.fix
+ * import arrow.typeclasses.Applicative
+ *
+ * interface IdApplicative : Applicative<ForId> {
+ *   override fun <A, B> IdOf<A>.ap(ff: IdOf<(A) -> B>): Id<B> =
+ *     ff.fix().flatMap { f -> map(f) }.fix()
```
1Jajen1:
```suggestion
 *     Id(ff.value().invoke(this.value()))
```
I know that the flatMap one is the one used in Id, and it is correct, however it doesn't show what is actually happening. Either do the above or include the flatMap definition somewhere.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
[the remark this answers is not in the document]
i-walker: I think that is one bargain we have to take when introducing sequence first. It's inconclusive if we introduce sequence without explaining it and still delaying `traverse`. For what its worth I suggest a different order.  Where we start with Sequence then Traverse and then Sequence Effects. But I am not sure if that changes anything.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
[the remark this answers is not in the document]
i-walker: What was already introduced? The sections before address Sequencing result types and this is for effects.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
i-walker: This is explained in a few lines down below

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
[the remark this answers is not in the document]
1Jajen1: Both traverse the method and Traverse the typeclass have already been used and named before, this makes it sound like something new is coming which is just not the case.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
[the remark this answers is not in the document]
1Jajen1: Starting with sequence is a really good starting point because most who are stuck and need traverse are sitting on a type `Kind<F, Kind<G, A>>` already and just need `sequence` to solve their problem. Then transition into the more general form `traverse`, show that `traverse(::identity) == sequence` and `map(f).sequence() == traverse(f)`. After that the rest of the tutorial around sequencing for effects, foldable, etc. That is how I would structure it.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
1Jajen1: I know, the point was not that it is not explained or wrong, but that it is not the important part. It is also not correct, it will short-circuit, but it can run effects from later elements before.
`listOf(1,2,3,4).parTraverse(IO { println(it); throw Error() })` can print anything from 1 to 4 or even all of it depending on who wins the race, but the result will be the first error encountered. So it short circuits as soon as it encounters an error, but because the effects start concurrently it can execute the effects of other items before doing so. The result is the same, except for side-effects

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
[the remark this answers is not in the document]
i-walker: I see. Let me rephrase that

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
i-walker: I phrased it a bit differently but it catches your point

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
i-walker: Makes sense. I think the part with `short-circuit`, especially in regards to side-effects could be a great addition to the parTraverse Docs.
I hope I could nail your point.

i-walker:
🎉 @1Jajen1 awesome Feedback. The order as you suggested reads a lot better. What do you think?

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,526 @@ import arrow.core.Id
...
+ * fun writeToStore(data: Data): IO<Unit> = TODO("")
+ * //sampleStart
+ * fun writeManyToStore(data: ListK<Data>): IO<ListK<Unit>> =
+ *   data.traverse(IO.applicative()) { writeToStore(it) }.fix()
+ * //sampleEnd
+ * ```
+ *
+ * We end up with a `IO<ListK<Unit>>`! A `ListK<Unit>` is not of any use to us, and communicates the same amount of information as a single [Unit] does.
+ *
+ * Traversing solely for the sake of the effects (ignoring any values that may be produced, [Unit] or otherwise) is common, so [Foldable] (superclass of [Traverse]) provides [traverse_] and [sequence_] methods that do the same thing as [traverse] and [sequence] but ignores any value produced along the way, returning [Unit] at the end.
```
1Jajen1:
```suggestion
 * Traversing solely for the sake of the effects (ignoring any values that may be produced, [Unit] or otherwise) is common, so [Foldable] (superclass of [Traverse]) provides [traverse_] and [sequence_] methods that do the same thing as [traverse] and [sequence] but ignore any value produced along the way, returning [Unit] at the end.
```

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,526 @@ import arrow.core.Id
...
+ *   val emptyList: Option<List<Int>> =
+ *     listOf(1.some(), none(), 3.some())
+ *       .traverse(Option.applicative(), ::identity).map { it.fix() }
+ *   //sampleEnd
+ *   println("optionList = $optionList")
+ *   println("emptyList = $emptyList")
+ * }
+ * ```
+ *
+ * Interestingly enough, [traverse] is a much more generalized and powerful form of [sequence] that would allow us to parse a `List<String>` or validate credentials for a `List<User>`.
```
1Jajen1:
```suggestion
 * [traverse] is a much more generalized and powerful form of [sequence] that would allow us to parse a `List<String>` or validate credentials for a `List<User>`.
```
I'm no native english speaker, but leading sentences like this seems weird. Also begs the question: "interestingly enough" for whom? Also why "enough". Imo just drop it, it adds no value to the text.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,526 @@ import arrow.core.Id
...
+ *
+ *   // val optionMapKBoilered = map.foldLeft(Some(emptyMap())) { acc: Option<MapK<String, String>>, i: Int ->
+ *   //   acc.fold({ emptyMap() }, { /*Some logic to retrieve the key of a value, transform it and add it to the accumulated Map*/ })
+ *   // }
+ *   //sampleEnd
+ *   println(optionMapK)
+ * }
+ * ```
+ *
+ * Both methodologies try to attain the same thing, but what [Foldable] lacks is that it solely drills down to it's `A` here `Int` and does not preserve it's shape `F` - `MapK`. Resulting in a tiresome implementation, where we need to come up with an algorithm to resolve a key of a given value in the `Map` - let alone that this algorithm is not universal over any given `Map`.
```
1Jajen1:
```suggestion
 * Both methodologies try to attain the same thing, but what [Foldable] lacks is that it solely drills down to its `A` here `Int` and does not preserve it's shape `F` - `MapK`. Resulting in a tiresome implementation, where we need to come up with an algorithm to resolve a key of a given value in the `Map` - let alone that this algorithm is not universal over any given `Map`.
```

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,526 @@ import arrow.core.Id
...
+ *
+ * This is where [Traverse] shines, whenever you care about the Output `B` from `(A) -> B` and the existing shape of `F` you may use [traverse].
+ *
+ * Additionally, you're able to wrap your context `F` within a `G`. That is one reason, may among others, why [Traverse] is strictly more powerful than [Foldable].
+ *
+ * ### Traversables are Foldable
+ *
+ * The [Foldable] type class abstracts over “things that can be folded over” similar to how [Traverse] abstracts over “things that can be traversed.” It turns out [Traverse] is strictly more powerful than [Foldable] - that is, [foldLeft] and [foldRight] can be implemented in terms of [traverse] by picking the right [Applicative]. However, arrow's [Traverse] does not implement [foldLeft] and [foldRight] with [traverse] as the actual implementation tends to be inefficient.
```
1Jajen1:
```suggestion
 * The [Foldable] type class abstracts over “things that can be folded over” similar to how [Traverse] abstracts over “things that can be traversed.” [Traverse] is strictly more powerful than [Foldable] - that is, [foldLeft] and [foldRight] can be implemented in terms of [traverse] by picking the right [Applicative]. However, arrow's [Traverse] does not implement [foldLeft] and [foldRight] with [traverse] as the actual implementation tends to be inefficient.
```

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,526 @@ import arrow.core.Id
...
+ * fun parseIntEither(s: String): Either<NumberFormatException, Int> =
+ *   if (s.matches(Regex("-?[0-9]+"))) Either.Right(s.toInt())
+ *   else Either.Left(NumberFormatException("$s is not a valid integer."))
+ *
+ * fun parseIntValidated(s: String): ValidatedNel<NumberFormatException, Int> =
+ *   if (s.matches(Regex("-?[0-9]+"))) s.toInt().validNel()
+ *   else NumberFormatException("$s is not a valid integer.").invalidNel()
+ * ```
+ *
+ * We can now [traverse] structures that contain strings parsing them into integers and either with a `fail-fast` strategy using [Either] or accumulating failures with [ValidatedNel].
```
1Jajen1:
```suggestion
 * We can now [traverse] structures that contain strings and parse them into integers with either with a `fail-fast` strategy using [Either] or accumulating failures with [ValidatedNel].
```

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,526 @@ import arrow.core.Id
...
+ *
+ * Going back to our `IO` example from the beginning with concurrency in mind, we can get an [Applicative] instance for `IO`, that traverses concurrently, by using `parApplicative`. Contrary to `IO.applicative()`, which runs in sequence.
+ *
+ * It is worth mentioning that `parApplicative` does not implement `lazyAp` in contrast to `IO.applicative()`, which means `parApplicative` creates all IO's upfront and executes them in parallel. More importantly, `parApplicative` breaks monad-applicative-consistency laws by design. The aforementioned law holds in the case of `IO.applicative()`, where an effect only runs , when the previous one is successful - hence the monadic nature.
```
1Jajen1:
```suggestion
 * It is worth mentioning that `parApplicative` does not implement `lazyAp` in contrast to `IO.applicative()`, which means `parApplicative` creates all IO's upfront and executes them in parallel. Consequently `parApplicative` breaks monad-applicative-consistency laws by design. The aforementioned law holds in the case of `IO.applicative()`, where an effect only runs , when the previous one is successful - hence the monadic nature.
```

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,526 @@ import arrow.core.Id
...
+ *
+ * Continuing with the example, we traverse a `List<A>` with its [Traverse] instance `ListK.traverse()` and a function`(A) -> IO<B>`, we can imagine the traversal as a scatter-gather. Each `A` creates a concurrent computation that will produce a `B` (the scatter), and as the `IO` operations completes they will be gathered back into a `List`.
+ *
+ * Ultimately, we utilize `parTraverse` that calls traverse with `parApplicative` in an concurrent environment [F] - in the following example `IO`:
```
1Jajen1:
```suggestion
 * There exists a utility method `parTraverse` that calls traverse with `parApplicative` in an concurrent environment [F] - in the following example `IO`:
```

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
```diff
@@ -5,11 +5,526 @@ import arrow.core.Id
...
+ *
+ * interface Traverse<F> : Functor<F>, Foldable<F> {
+ *   fun <G, A, B> Kind<F, A>.traverse(AP: Applicative<G>, f: (A) -> Kind<G, B>): Kind<G, Kind<F, B>>
+ *
+ *   override fun <A, B> Kind<F, A>.map(f: (A) -> B): Kind<F, B> =
+ *     traverse(Id.applicative()) { Id(f(it)) }.value()
+ * }
+ * ```
+ *
+ * Both have an `Kind<F, A>` receiver and a similar `f` parameter. [traverse] expects the return type of `f` to be `Kind<G, B>` whereas `map` just wants `B`. Similarly the return type of [traverse] is `Kind<G, Kind<F, B>>` whereas for `map` it's just `Kind<F, B>`. This suggests we need to pick a `G` such that `Kind<G, A>` communicates exactly as much information as `A`. We can conjure one up by simply wrapping an `A` in `arrow.core.Id`.
```
1Jajen1:
```suggestion
 * Both have a `Kind<F, A>` receiver and a similar `f` parameter. [traverse] expects the return type of `f` to be `Kind<G, B>` whereas `map` just wants `B`. Similarly the return type of [traverse] is `Kind<G, Kind<F, B>>` whereas for `map` it's just `Kind<F, B>`. This suggests we need to pick a `G` such that `Kind<G, A>` communicates exactly as much information as `A`. We can conjure one up by simply wrapping an `A` in `arrow.core.Id`.
```

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Traverse.kt
i-walker: It's really just a stylistic choice of my writing. We will have a proper language review from fellow Arrow Maintainers soon. So I'd leave this as is once it goes through language review there will be more edits to this :)

PR #1536, opened 2019-07-31
Fix AndThen#contramap doc

MakeNowJust:
`AndThen#contramap` seems not `andThen` alias. Correctly it is `compose` alias.


pakoito:
Thank yoyu for catching this error!

PR #1541, opened 2019-08-07
Improve Try#filter documentation

juliankotrba:
I was just reading the documentation of `Try` and IMHO I think the `Try#filter` example can be slightly improved.
The documentation says "convert successful computations in failures" but the result of the code example is `Failure(exception=Line_1$AuthorizationException)`

Bildschirmfoto 2019-08-07 um 14 37 33

This means the computation was never successful and I would also expect to see a `TryException.PredicateException` in the code example.

With this change the example in the documentation should return a `Failure(exception=PredicateException(message=Predicate does not hold for 0))`.

Image source: [https://arrow-kt.io/docs/arrow/core/try/](https://arrow-kt.io/docs/arrow/core/try/) (07.08.2019)

### DIFF
--- a/modules/docs/arrow-docs/docs/docs/arrow/core/try/README.md
+++ b/modules/docs/arrow-docs/docs/docs/arrow/core/try/README.md
@@ -93,8 +93,8 @@ lotteryTry.getOrElse { emptyList() }
 If you want to perform a check on a possible success, you can use `filter` to convert successful computations in failures if conditions aren't met:
 
 ```kotlin:ank
-lotteryTry.filter {
-    it.size < 4
+Try.just(0).filter {
+    it > 1
 }
 ```
 


i-walker:
In the previous snippet the `lotteryTry` variable already encapsulates the Error, thus further computations end in `Failure`. 
Try is short-circuit and eager.  
Does this make sense? @juliankotrba 
Either way, we're deprecating Try after 0.10.0. Check out `Either.catch { }`

modules/docs/arrow-docs/docs/docs/arrow/core/try/README.md
```diff
@@ -93,8 +93,8 @@ lotteryTry.getOrElse { emptyList() }
 If you want to perform a check on a possible success, you can use `filter` to convert successful computations in failures if conditions aren't met:
 
 ```kotlin:ank
```
i-walker: You can turn this into a `kotlin:ank:playground`
And add costume `message` to your Exceptions to get your `PredicateException(message=Predicate does not hold for $exceptionVariable)`

juliankotrba:
Thank you for your answer @i-walker , but as I already said the documentation of the `filter` function says that it converts successful computations into failures. That's why I think the code example should also represent a successful computation which results, after filtering, in a failure. But anyway, since this will be deprecated we  can close this PR 😄  .

Issue #1196, opened 2018-12-13
["Request"] Add support to JPMS/Java >= 9


gortiz:
What version are you currently using?
Any, including 0.8.1

What would you like to see?
It is not possible to use Arrow on an application/library running on the module path of Java >= 9. There are some minor problems (like not having module-info.java nor automatic name on the manifest) but the main one is related with arrow-annotations. It depends on `kindedj`, but also includes in its own sources [this file](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-annotations/src/main/java/io/kindedj/hkjs.kt), which is under the package io.kindedj. Therefore when adding arrow-annotations on the module-path, there package io.kindedj is declared in two jars, which is forbidden.

I didn't checked the sources of io.kindedj, but if `io.kindedj.Hk` is public, the solution could be as easy as to move hkjs.kt to another package.

pakoito:
I thought Kotlin didn't allow Java 9. This seems like an easy fix as that file is just type aliases :)

gortiz:
Kotlin does not compile to Java 9 bytecode, but Kotlin programs run fine on the module path. Jetbrain has even modified Kotlin libraries to include a module name, so transition will be easier.

[This plugin](https://github.com/beryx/badass-jar-plugin/blob/master/README.adoc) makes very easy to support both Java 8 and Java >= 9 on libraries and applications and has Kotlin examples.

Pyeroh:
I replied on this issue because I think it's relevant, considering the topic.

Despite #1546, when I compile with the latest snapshot (0.10.0-20190818.134003-13) of arrow-core-data, I have this error :  
`java.lang.module.ResolutionException: Modules arrow.core.data and arrow.annotations export package arrow to module X`
Whenever I rebuild, the module changes, or it shows that error :
`java.lang.module.ResolutionException: Module arrow.annotations contains package arrow, module arrow.core.data exports package arrow to arrow.annotations`

I can't find any module-info, but I believe these errors occurs because, in the module descriptor of the library, the whole package arrow is exported. I looked in arrow-core-data and arrow-annotations, this export is meaningful, but not compatible with the module system.

cacsar:
Upgrading from 0.11.0 to 0.13.2 took care of the ResolutionException for my code.

PR #1545, opened 2019-08-11
Moved hkjs.kt to package arrow in order to enable java9+ support

Cottand:
Moved hkjs.kt out of io.kindedj in order to not declare io.kindedj in more than one jar, see #1196


pakoito:
Please check what's failing on the travis build so we can fix it before approving this PR :D

Thanks for your effort!

Cottand:
Oh sorry about that

pakoito:
????

pakoito:
There was no need to delete it, the build was failing, it just needed some fix! :D

Issue #1501, opened 2019-07-02
Add FunctorFilter/MonadFilter instances for Rx and Reactor


pakoito:
https://twitter.com/sla_shalafi/status/1145988260004585472

plattysoft:
What we did was:

```
fun <T> Observable<Option<T>>.flatten(): Observable<T> {
    return flatMap {
        it.map {
            Observable.just<T>(it)
        }.getOrElse {
            Observable.empty<T>()
        }
    }
}
```

I'm happy to send a PR with it if you are happy with it.

pakoito:
What we'd need is for `filterMap` to be implemented, and flattenOption can be derived from there.

One thing I'd suggest is using `fold` instead of `map` then `getOrElse`. Same results, one fewer intermediate object created.

pakoito:
If you do a PR, make sure to add it to Flowable, Maybe and Flux too, as all of them are capable of having this pattern.

plattysoft:
fold sounds better indeed, cleaner, more concise.

I'll probably send a PR later this week.

lenguyenthanh:
Is anyone working on this issue? If the answer is no, I'm happy to try. I have free time this weekend.

raulraja:
`aql` could also have the rx and reactor integration modules since these instances will enable most of the AQL operators.

plattysoft:
Oh, I completely forgot about this... I might have some time tomorrow, otherwise take it.

plattysoft:
I didn't do it exactly as you intentded @pakoito , this is what made sense in my head, but I'm happy to iterate and discuss it: https://github.com/arrow-kt/arrow/pull/1550

pakoito:
Flatten is already taken as an extension function. Let's break it down:

```
fun <T, A, B> Flowable<A>.mapFilter(f: (A) -> Option<B>) =
  flatMap { a ->
    f(a).fold({ Flowable.empty() }, { b -> Flowable.just(b) })
  }
```

So now we have a function that works for every filter function, and also maps all in one go. Great. Let's specialize it!

```
fun <T, A> Flowable<Option<A>>.filterNone() =
  mapFilter { it } // mapFilter(::identity) to prevent an additional allocation
```

It turned out to be super easy to implement, and we've gotten functionality that's generic, allowing us to implement the FunctorFilter typeclass on the way. 2x1!

lenguyenthanh:
[...]

PR #1550, opened 2019-08-16
added flatten extension function to Rx Observable, Flowable and Maybe

plattysoft:
Initial implementation suggested for https://github.com/arrow-kt/arrow/issues/1501

Added flatten extension function to Observable, Flowable and Maybe.

If this implementation is ok and on the right place, I'll proceed to add some tests.


pakoito:
See the comment here! https://github.com/arrow-kt/arrow/issues/1501#issuecomment-522175483

nomisRev:
I added a new issue in relation to this #1871.
@plattysoft can we close this PR and open a new one for #1871 instead?

plattysoft:
I recall this was already fixed on another branch, sure, let's close this.

PR #1551, opened 2019-08-18
Revisit Apply

i-walker:
This PR adds:
- [ ] Apply Doc's as well as for KDocs
- [x] Semigroupal instances for DT who have an Apply instance 
- [x] Apply instances
- [ ] ApplyLaw Tests for DT
- [X] ApplyLaws from cat's [ApplyLaws](https://github.com/typelevel/cats/blob/master/laws/src/main/scala/cats/laws/ApplyLaws.scala)

The initial draft also introduces the compose function for Apply, which is derived from cats [ComposedApply](https://github.com/typelevel/cats/blob/241a7b9ba1f3baff8010ead2140d969b8253ed05/core/src/main/scala/cats/Apply.scala#L210), interpreted in a RecieverFunction. 

Furthermore, the Semigroupal instances are substantial for the ApplyLaws.

As an addition Composed extension instances are now in `arrow-mtl`, which was forgotten from the refactor task in https://github.com/arrow-kt/arrow/pull/1549


modules/docs/arrow-docs/docs/docs/arrow/typeclasses/apply/README.md
```diff
@@ -0,0 +1,38 @@
...
+layout: docs
+title: Apply
+permalink: /docs/arrow/typeclasses/apply/
+redirect_from:
+  - /docs/typeclasses/apply/
+---
+
+## Apply
+
+`Apply` extends the `Functor` type class (which features the familiar `map` function) with a new function `ap`. The `ap` function is similar to `map` in that we are transforming a value in a context (a context being the `F` in `Kind<F, A>`;
```
i-walker: I am currently working on the code and after that is done and build I finish the docs. 

I am aware that the text is incorrect and this will be corrected.

modules/core/arrow-core-data/src/test/kotlin/arrow/core/MapKTest.kt
```diff
@@ -35,13 +38,24 @@ class MapKTest : UnitSpec() {
...
 
     testLaws(
       ShowLaws.laws(MapK.show(), EQ_TC) { mapOf(it.toString() to it).k() },
       TraverseLaws.laws(MapK.traverse(), MapK.functor(), { a: Int -> mapOf("key" to a).k() }, EQ),
       MonoidLaws.laws(MapK.monoid<String, Int>(Int.semigroup()), Gen.mapK(Gen.string(), Gen.int()), EQ),
+      /*ApplyLaws.laws(MapK.apply(), MapK.semigroupal(), EQ,
+        { a: Int -> mapOf("key" to a).k() },
+        { a: Int -> mapOf("key" to { i: Int -> a + i }).k() },
+        { b -> b.fix().map { it.reverse() } },
+        EQ2)*/
```
AdrianRaFo: Leave an explanation, uncomment it or delete it

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/either.kt
```diff
@@ -94,7 +97,7 @@ interface EitherApply<L> : Apply<EitherPartialOf<L>>, EitherFunctor<L> {
 }
 
 @extension
-interface EitherApplicative<L> : Applicative<EitherPartialOf<L>>, EitherFunctor<L> {
+interface EitherApplicative<L> : Applicative<EitherPartialOf<L>>, EitherFunctor<L>, EitherApply<L> {
```
pakoito: If it has `EitherApply` than it doesn't need `EitherFunctor`. Neither here nor in the other extensions.

i-walker: `Apply` acts as an Extension to our current Typeclassopedia in `Arrow`
I discussed it in the `contributor` channel, whether to drop `Apply`, because it lacks in infrastructure in `Arrow`. After discussing and further research I am trying to give it a place in Arrow. 
Screenshot 2019-08-26 at 12 21 24

i-walker: That's why we will keep both `<Datatype>Apply` and `<Data>`

pakoito: `interface Apply<F> : Functor<F>`

My point is that you already have a functor in the hierarchy of these classes via Apply, there's not need to have both.

i-walker: My Bad, we figured that the picture from above is wrong...

i-walker: Let me finish up with ApplyLaws and move forward

i-walker:
Closing this. It is way too outdated

PR #1552, opened 2019-08-20
Docs: add note about fold

rachelcarmena:
**Super tiny** PR to add a note in order to differentiate:
* Optics `Fold`
* Collection aggregate operation `fold`


pakoito:
Works for me. Thanks for spotting this one. It'll go live in next.arrow-kt.io

PR #1557, opened 2019-08-24
Fix one conflict that reached master

pakoito:


nomisRev:
Derp in a comment. Thanks @pakoito!

PR #1573, opened 2019-09-01
Rename functions called run to avoid collision with stdlib

pakoito:
Title


modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/Kleisli.kt
```diff
@@ -15,8 +15,11 @@ import arrow.typeclasses.Monad
  */
 typealias KleisliFun<F, D, A> = (D) -> Kind<F, A>
 
+@Deprecated("Kleisli#run can be easily mistaken with the standard library extension function", ReplaceWith("runK(d)"))
 fun <F, D, A> KleisliOf<F, D, A>.run(d: D): Kind<F, A> = fix().run(d)
 
+fun <F, D, A> KleisliOf<F, D, A>.runK(d: D): Kind<F, A> = fix().run(d)
```
raulraja: can we make this `fun operator invoke`?

modules/streams/arrow-streams/src/test/kotlin/arrow/streams/internal/FreeCTest.kt
```diff
@@ -236,7 +236,7 @@ class FreeCTest : UnitSpec() {
     "Running a FlatMapped value using Either" {
       forAll(Gen.string()) { s ->
         FreeC.FlatMapped(FreeC.just("")) { FreeC.just<EitherPartialOf<Throwable>, String>(s) }
-          .run(Either.monadError()) == Right(Some(s))
+          .runFC(Either.monadError()) == Right(Some(s))
```
raulraja: we could also use invoke in these cases for Free. I think this is inline that both Free and Kleisli represent functions that you can apply to evaluate the structure.

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/Kleisli.kt
pakoito: I'd rather we didn't, at least for this first version. Automatic replacement can cause syntactic errors.

pakoito:
https://travis-ci.org/arrow-kt/arrow/jobs/579540609#L41885

rachelcarmena:

PR #1574, opened 2019-09-02
Adds blog section

AntonioMateoGomez:
This PR adds blog section, for the moment it will be apart


pakoito:
Same as the other diff, I can merge when you say go for it

PR #1575, opened 2019-09-02
remove codevoc

ffgiraldez:

Talking with @JorgeCastilloPrz and @i-walker we don't see the needs of coverage reports until a stable version is released, codecov was disabled on #1108 but still executed as CI stage. 

This PR remove that stage on CI, badge on README.MD and codecov configuration file

### :memo: Notes
I still keep the codecov integration for the Arrow-kt organization, but without any repo access


JorgeCastilloPrz:
Thanks!

Issue #1576, opened 2019-09-03
Browser home


israelperezglez:
We have to create an html mockup for the browser in the home page

PR #1578, opened 2019-09-03
Home code runable

AntonioMateoGomez:
This PR adds the main "home code block" behaviour but still needs to add:

- [ ] Proper code for each library

- [ ] Give browser style to the code block #1576 

- [ ] Control responsive mode and readjust the top position of the `fadeInDown` in the code block div.


pakoito:
I removed that check that went nowhere :)

PR #1580, opened 2019-09-04
Make home code editable and runable

AntonioMateoGomez:
This PR makes the home code snippet runable and editable through arrowplayground.
We still need to include the real code snippet for each library


arrow-version/_code/core-home-code.md
```diff
@@ -1,7 +1,9 @@
 ---
 library: core
 ---
-```kotlin:ank:playground
+
+{: data-executable="true"}
+```kotlin
 import arrow.*
 import arrow.core.*
```
pakoito: Given that we're here and it's an easy change, could you change this snippet to call `magic` with a value?

Also, given that we're trying to convince people that there isn't any undecipherable magic involved, I'd rename that variable to `chain` or `combination`.

AntonioMateoGomez: Ok @pakoito, I will try it. Keep in mind that we are still not including the definitive code fragment for each library.

AntonioMateoGomez:
This has been solved thorugh this commit 
https://github.com/arrow-kt/arrow/commit/78d3cd52a6061e89e24941609b2bf0e37382ed5d

PR #1593, opened 2019-09-11
Mobile style and hover control

AntonioMateoGomez:
In this PR we set the home section style in mobile view.


arrow-version/js/main.js
```diff
@@ -65,59 +98,69 @@ function loadEvent() {
...
   // incubator elements
   const incubatorPlayHover = document.getElementById('incubator');
 
   incubatorPlayHover.addEventListener('mouseenter', () => {
-    arrayArrowFeatures.map(obj => obj.classList.remove('active'));
-    incubatorPlayHover.classList.add('active');
-    checkActiveFeature(arrayArrowFeatures);
-    incubatorHoverAnimation.play();
-    arrowBaseAnimation.stop();
+    if (screenSize() > 992) {
```
pakoito: Why 992?

pakoito: Also, why not make it a constant in the file?

AntonioMateoGomez: 992 is the break point we´ve set to hide the code console and the main logo, so that´s the point in which it has no sense to keep the main hover behaviour. At that point the mobil mode comes into play.
Anyway is a good idea make it a constant of course.
Thanks @pakoito !

PR #1600, opened 2019-09-13
Either right if null extension

Neuman968:
Added rightIfNull Either extension function for kotlin.Any? type. Provides inverse functionality of rightIfNotNull.

### DIFF
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -383,6 +383,20 @@ fun <A, B> B?.rightIfNotNull(default: () -> A): Either<A, B> = when (this) {
 }
 
 /**
+ * Returns [Either.Right] if the value of type Any? is null, otherwise the specified A value wrapped into an
+ * [Either.Left].
+ *
+ * Example:
+ * ```
+ * null.rightIfNull { "left" }    // Right(b="Unit")
+ * "value".rightIfNull { "left" ) // Left(a="left")
+ * ```
+ */
+fun <A> Any?.rightIfNull(default: () -> A): Either<A, Unit> = when (this) {
+  null -> Either.right(Unit)
+  else -> Either.left(default())
+}
+/**
  * Applies the given function `f` if this is a [Left], otherwise returns this if this is a [Right].
  * This is like `flatMap` for the exception.
  */
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -388,12 +388,12 @@ fun <A, B> B?.rightIfNotNull(default: () -> A): Either<A, B> = when (this) {
  *
  * Example:
  * ```
- * null.rightIfNull { "left" }    // Right(b="Unit")
+ * null.rightIfNull { "left" }    // Right(b="Nothing?")
  * "value".rightIfNull { "left" } // Left(a="left")
  * ```
  */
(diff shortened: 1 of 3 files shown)


modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -382,6 +382,20 @@ fun <A, B> B?.rightIfNotNull(default: () -> A): Either<A, B> = when (this) {
...
 }
 
+/**
+ * Returns [Either.Right] if the value of type Any? is null, otherwise the specified A value wrapped into an
+ * [Either.Left].
+ *
+ * Example:
+ * ```
+ * null.rightIfNull { "left" }    // Right(b="Unit")
+ * "value".rightIfNull { "left" ) // Left(a="left")
```
pakoito: There's a mismatched `{`

pakoito:
Nice, very complete PR. Thank you :D

I would need an example of where `Unit` in the right is the right approach. Wouldn't you want `Nothing?` instead?

Neuman968:
@pakoito  Thank you! This is my first public PR so forgive me if I miss any formalities. 
I did consider using `Nothing?` but from my understanding Nothing cannot be instantiated so the code

```kotlin
fun <A> Any?.rightIfNull(default: () -> A): Either<A, Nothing?> = when (this) {
  null -> Either.right(Nothing())
  else -> Either.left(default())
}
```
Doesn't compile. Maybe I am missing something. I have a few projects that use this library and found a few places where I used this rightIfNull functionality so I thought it would be a good first contribution. 

Thanks for such an awesome project! and Thank you so much for your time :)

modules/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
Neuman968: Thank you! Fixed.

Neuman968:
On second thought, It could be something like 

```kotlin
fun <A> Any?.rightIfNull(default: () -> A): Either<A, Nothing?> = when (this) {
  null -> Either.right(null)
  else -> Either.left(default())
}
```
I updated the code and README.

modules/docs/arrow-docs/docs/docs/arrow/core/either/README.md
```diff
@@ -317,6 +317,21 @@ Example:
...
 ```
 
+The inverse of `rightIfNotNull`, `rightIfNull`.
+If the value is null it will be transformed to the specified `Either.right` and the type will be `Nothing?`.
+If the value is not null than it will be transformed to the specified `Either.Left`.
+
+Example:
+
+```kotlin:ank
+// Results in Either.Left
```
pakoito: These aren't necessary, ank will display the result inlined below this snippet

modules/docs/arrow-docs/docs/docs/arrow/core/either/README.md
```diff
@@ -317,6 +317,21 @@ Example:
...
+"value".rightIfNull { "left" }
+```
+
+```kotlin:ank
+// Results in Either.Right
```
pakoito: Same here

pakoito:
You're doing okay, no formalities are necessary :D As soon as things are fixed and CI passes we're good to go.

modules/docs/arrow-docs/docs/docs/arrow/core/either/README.md
Neuman968: Thanks! should be corrected now :)

pakoito:
Passing and merged. Thank you for the contribution!

Neuman968:
Thank you!

PR #1603, opened 2019-09-16
Implement FunctorFilter instances for Rx

lenguyenthanh:
Here is my pr for #1501.

For now I just implemented FunctorFilter  instance for ObserableK because I want to have some feedback before implementing FunctorFilter for other types.

### DIFF
--- a/modules/fx/arrow-fx-rx2/src/main/kotlin/arrow/fx/rx2/extensions/observablek.kt
+++ b/modules/fx/arrow-fx-rx2/src/main/kotlin/arrow/fx/rx2/extensions/observablek.kt
@@ -3,6 +3,7 @@ package arrow.fx.rx2.extensions
 import arrow.Kind
 import arrow.core.Either
 import arrow.core.Eval
+import arrow.core.Option
 import arrow.core.Tuple2
 import arrow.fx.CancelToken
 import arrow.fx.RacePair
@@ -37,6 +38,7 @@ import arrow.typeclasses.Applicative
 import arrow.typeclasses.ApplicativeError
 import arrow.typeclasses.Foldable
 import arrow.typeclasses.Functor
+import arrow.typeclasses.FunctorFilter
 import arrow.typeclasses.Monad
 import arrow.typeclasses.MonadError
 import arrow.typeclasses.MonadThrow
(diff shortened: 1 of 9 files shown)


modules/fx/arrow-fx-rx2/src/main/kotlin/arrow/fx/rx2/extensions/observablek.kt
```diff
@@ -281,3 +283,12 @@ interface ObservableKTimer : Timer<ForObservableK> {
     ObservableK(io.reactivex.Observable.timer(duration.nanoseconds, TimeUnit.NANOSECONDS)
       .map { Unit })
 }
+
+@extension
+interface ObservableKFunctorFilter : FunctorFilter<ForObservableK> {
+  override fun <A, B> Kind<ForObservableK, A>.filterMap(f: (A) -> Option<B>): ObservableK<B> =
+    fix().filterMap(f)
```
JorgeCastilloPrz: Is `filterMap` availabel for `ObservableK`?

JorgeCastilloPrz:
It looks good to me.

lenguyenthanh:
Hey @JorgeCastilloPrz, do I need to add some tests in some where? To avoid what happen in my first commit? I thought if I passed FunctorFilter

JorgeCastilloPrz:
For now at least you'd need to cover `ObservableK` with the `FunctorFilterLaws` ([here they are](https://github.com/arrow-kt/arrow/blob/84d52075e33252ffb0b9a77e75658e27e1492be5/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/FunctorFilterLaws.kt)). You can add that line [here](https://github.com/arrow-kt/arrow/blob/5686f6f06fd02e704f5c2360da3c8d19a2e71017/modules/fx/arrow-fx-rx2/src/test/kotlin/arrow/fx/ObservableKTests.kt#L59).

lenguyenthanh:
So I implemented 3 `FunctorFilter` instances for `MaybeK`, `ObservableK`  and `FlowableK`.

I'm not sure how we should  implement `filterMap` for `SingleK`. Here is my draft:

```
  fun <B> filterMap(f: (A) -> Option<B>): SingleK<B> =
    single.flatMap { a ->
      f(a).fold({ Single.error<B>(NullPointerException()) }, { b -> Single.just(b) })
    }.k()
```

Any idea about this?

raulraja:
Hi @lenguyenthanh, In order to filter you need an empty value for when there is not a match for the predicate. What is the empty value of Single? I don't think this instance is possible if there isn't such a thing. Using an exception is not a valid implementation we can add to Arrow if it's ever gonna surface to the user in any shape.

lenguyenthanh:
Hi @raulraja, I agree with that. So I think this PR is in a good shape now. Do you see anything I need to improve? Or otherwise I'll rebase, so you can merge to master.

JorgeCastilloPrz:
Single doesn't emit empty values, it just emits one value or an error. It doesn't support FunctorFilter.

lenguyenthanh:
Hey guys, thanks for reviewing and comments. I think I'll write some docs in other pr (I'll also create an issue for docs).

I did squash commits. So I think this is good to merge now ;)

Thanh

pakoito:
Thank you for the contribution! The squash is done automatically by github, so there was no need to do it manually hahahha

lenguyenthanh:
This is my first PR for Arrow, I hope that I can contribute more.

P/S I use gitlab too much these days 😭

pakoito:
Glad to have you around, I can see you're active on the channels too 👏

PR #1607, opened 2019-09-18
Menu redesign

### LINKED ISSUE #1567: New doc menu redesign
Once the final menu is defined we must implement it and modify the differents files to make it.

AntonioMateoGomez:
This PR organizes the new sidebar menu redesign since now it is divided into `core`, `optics`, `fx`, `meta` and `incubator`. It also modifies the sidebar menu doc version button position and improves the sidebar layout reducing the five needed `.html` includes into one.
A couple of points to consider:

- API Docs links are not working since dokka is not configured in the module where we are implementing the new site. **It will be tested in an easier way when we would make the integration in the right module**. This is going to be the next step.

- Arrow Meta documentation is not implemented yet since it is still in process.

Fixes #1567


AntonioMateoGomez:
> I think that including the markdown files as part of this feature branch will be more problematic than helpful (lots of conflicts, outdated content, etc.). So, I'd start by removing the new ones being added in this PR, then, I either remove the rest of them from the target branch, or I'd keep them but having in mind that the content source will always be in `master`.

Ok, markdown files in this PR are just to visualize how will each library menu be, **we have to mind this in the integration process** to do it as you say @calvellido , the content source will be what´s is in master.

Issue #1525, opened 2019-07-22
Upgrade ktlint and disable no-unit-return rule


Tylos:
### References

Related to #1502 

### Description

Upgrade ktlint plugin in build.gradle or set ktlint version manually to 0.34.0 or above. This will allow to globally disable no unit returns errors in arrow, allowing to add explicit unit return types

Tylos:
Seems like version  `0.34.0` has some regression bugs, I am keeping an eye to see when there is a stable version we could include

raulraja:
Closing as this is no longer relevant and looks like we are now in 0.39.0 for KtLint. Thanks, @Tylos !

PR #1610, opened 2019-09-19
Upgrade ktlint and disable no-unit-return rule

Tylos:
### References

#1525

### Description

Upgrade ktlint to the latest version where rules can be globally disabled via `.editorconfig` and include `no-unit-return` so functions can declare `Unit` return type explicitly

### Implementation

Bump ktlint gradle plugin to latest version, from 7.3.0 to 8.2.0. This would set ktlint to use version 0.34.0. This version shipped with some bugs, so have forced ktlint to latest minor where they are fixed by setting the library version through KtlintExtension in `build.gradle`. 

This effectively bumps ktlint from 0.31.0 to 0.34.2, allowing `no-unit-return` to be disabled. Apart from it, the following rules have required attention in the codebase

#### import-ordering

Added on version `0.34.0` to the [standard rule set](https://github.com/pinterest/ktlint/commit/adb4d4d20a63ad0ab5c67e426429c273bbdec2ad#diff-ccd5a7f365eb13a0c3aea71168cf3bc6R15). It changes imports to follow alphabetical order. This will conflict with Intellij order and the reason why it has been **ignored**

```
./gradlew ktlintFormat

(...)
230 files changed, 536 insertions(+), 537 deletions(-)
```

#### curly-spacing

Already existed in the standard ruleset but must have some fixes/corrections in previous versions. Fixed the violations found in the project and **kept** the rule

```
./gradlew ktlintFormat

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Bifunctor.kt     | 3 +--
modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Contravariant.kt | 3 +--
modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Foldable.kt      | 3 +--
modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Functor.kt       | 3 +--
modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/function0.kt      | 6 ++----
5 files changed, 6 insertions(+), 12 deletions(-)
```

#### parameter-list-wrapping

Also existed in the standard ruleset but its implementation has changed in previous versions. Fixed the violations found in the project and **kept** the rule
```
./gradlew ktlintFormat

 modules/meta/arrow-meta/src/main/java/arrow/meta/ast/ast.kt | 18 ++++++++++++------
 1 file changed, 12 insertions(+), 6 deletions(-)
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,3 +3,4 @@
 indent_size=2
 insert_final_newline=true
 max_line_length=off
+disabled_rules=import-ordering
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,4 +3,4 @@
 indent_size=2
 insert_final_newline=true
 max_line_length=off
-disabled_rules=import-ordering
+disabled_rules=import-ordering,no-unit-return
--- a/build.gradle
+++ b/build.gradle
@@ -21,6 +21,7 @@ buildscript {
         // Constants
         ankKotlinVersion = '1.3.11'
         dokka_version = '0.9.18'
+        ktlint_version = "0.34.2"
         ktlint_gradle_version = '8.2.0'
         gradleVersionsPluginVersion = '0.21.0'
         jUnitVersion = '4.12'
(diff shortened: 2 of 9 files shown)


pakoito:
>The job exceeded the maximum log length, and has been terminated.

lol

nomisRev:
That's been going on since the RxJava instances. I haven't found a way to disable those logs tho.

PR #1613, opened 2019-09-20
Added Boolean.rightIfTrue extension method. 

Neuman968:
## Changes
Added Boolean Extension method rightIfTrue.

Functionally the same as using Either.cond but simplified syntax. 

Example

```kotlin
( 1 == 2).rightIfTrue({"Success" }, { "Failure" })
```
Or

```kotlin
( 1 == 1).rightIfTrue({ "Failure" })
```
results in Either.Right(null)


Neuman968:
Since this functionality already exists I totally get it if this isn't something that should be added, just thought it might be useful. Thanks :)

pakoito:
Yes, these could live outside of the lib :D

Neuman968:
Ok No Problem, This uses/replicates Either.cond
Thanks though!

PR #1622, opened 2019-09-24
Make Rx family covariant

### LINKED ISSUE #1609: Make ObservableK and the family covariant
We found that when working with Observable and Either, mapping into `Either.left` would cause an error because the other side can't be hinted and Observable is invariant, so you need to create a local variable to give the correct type to Nothing.

We can fix this in ObservableK and suppress the variance issues.

cc @Laimiux @lenguyenthanh

lenguyenthanh:
Here is my first try to fix #1609.
I only made `ObservableK` covariant this time, after making sure everything fine with `ObservableK` I'll continue to work on other types of Rx family.

P/S: I had to remove `ObservableKKindedJ<A>` from inherited list because `io.kindedj.Hk<ForObservableK, A>` is invariant.

### DIFF
--- a/modules/fx/arrow-fx-rx2/src/main/kotlin/arrow/fx/rx2/FlowableK.kt
+++ b/modules/fx/arrow-fx-rx2/src/main/kotlin/arrow/fx/rx2/FlowableK.kt
@@ -24,7 +24,6 @@ class ForFlowableK private constructor() {
   companion object
 }
 typealias FlowableKOf<A> = arrow.Kind<ForFlowableK, A>
-typealias FlowableKKindedJ<A> = io.kindedj.Hk<ForFlowableK, A>
 
 @Suppress("UNCHECKED_CAST", "NOTHING_TO_INLINE")
 inline fun <A> FlowableKOf<A>.fix(): FlowableK<A> =
@@ -32,9 +31,9 @@ inline fun <A> FlowableKOf<A>.fix(): FlowableK<A> =
 
 fun <A> Flowable<A>.k(): FlowableK<A> = FlowableK(this)
 
-fun <A> FlowableKOf<A>.value(): Flowable<A> = fix().flowable
+fun <A> FlowableKOf<A>.value(): Flowable<A> = fix().flowable as Flowable<A>
 
-data class FlowableK<A>(val flowable: Flowable<A>) : FlowableKOf<A>, FlowableKKindedJ<A> {
+data class FlowableK<out A>(val flowable: Flowable<out A>) : FlowableKOf<A> {
 
   fun <B> map(f: (A) -> B): FlowableK<B> =
     flowable.map(f).k()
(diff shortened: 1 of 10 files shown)


modules/fx/arrow-fx-rx2/src/main/kotlin/arrow/fx/rx2/ObservableK.kt
```diff
@@ -23,18 +23,18 @@ class ForObservableK private constructor() {
...
 @Suppress("UNCHECKED_CAST", "NOTHING_TO_INLINE")
 inline fun <A> ObservableKOf<A>.fix(): ObservableK<A> =
   this as ObservableK<A>
 
 fun <A> Observable<A>.k(): ObservableK<A> = ObservableK(this)
 
+@Suppress("UNCHECKED_CAST")
 fun <A> ObservableKOf<A>.value(): Observable<A> =
-  fix().observable
+  fix().observable as Observable<A>
```
pakoito: Why is this necesssary? What's the complaint?

pakoito:
So far so good

modules/fx/arrow-fx-rx2/src/main/kotlin/arrow/fx/rx2/ObservableK.kt
lenguyenthanh: Screenshot 2019-09-24 at 22 23 48
It seems it cannot auto cast `Observable<out A>` to `Observable<A>`

pakoito: see if there's any of the @annotations that may fix it :D

lenguyenthanh: where should I look for annotations? I checked here: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-annotation.html, but I didn't find any thing useful.

pakoito: JvmSuppressWildcards or JvmWildcard

lenguyenthanh: Both of them dont' work :(

lenguyenthanh: do I need to do anything to fix this? @pakoito

modules/fx/arrow-fx-rx2/src/main/kotlin/arrow/fx/rx2/FlowableK.kt
```diff
@@ -24,17 +24,17 @@ class ForFlowableK private constructor() {
   companion object
 }
 typealias FlowableKOf<A> = arrow.Kind<ForFlowableK, A>
-typealias FlowableKKindedJ<A> = io.kindedj.Hk<ForFlowableK, A>
```
JorgeCastilloPrz: Are we removing KindedJ support? @pakoito

pakoito: No, covariance needs a compatibility function that's on the `arrow-kindedj` module.

modules/fx/arrow-fx-rx2/src/main/kotlin/arrow/fx/rx2/ObservableK.kt
pakoito: No, leave it as is.

I'd suggest testing a bit just in case.

lenguyenthanh:
Any thing I can do to make this go forward?

pakoito:
I'll click the green button as soon as all the checks pass, and set a reminder for it!

lenguyenthanh:
No hurry, I'm just checking ;)

pakoito:
Neat :D If you wanna round it up, try the same with Project Reactor!

lenguyenthanh:
Will do it soon!

Issue #1617, opened 2019-09-24
Monad transformer threading issue


ps-feng:
Hi there,

I'm not sure if I did something wrong, but it seems monad transformers (at least `EitherT`) are not respecting the specified `CoroutineContext`. This is a simplified version of a larger piece of code where I found the error:
```kotlin
fun eitherTIOMonad(): EitherTMonad<ForIO, String> = EitherT.monad(IO.monad())

...

val eitherTIOM: IO<Either<String, String>> = eitherTIOMonad().fx.monad {
    println("Hi") // <-- runs on main thread, why?
    "Hola"
}.value().fix()

IO.fx {
    continueOn(Dispatchers.IO)
    eitherTIOM.bind() // <-- this line gets executed on IO thread
}.unsafeRunAsync { ... }
```
What I was playing with was basically just encapsulating some IO mixed with `Either` within an `EitherT` and I noticed that the first instruction within the `EitherT` monad was always executed on the caller thread. This code is simplified but with further `EitherT`s bound inside, the rest do get executed on IO worker threads.

Is this a bug or I did something wrong?

Cheers.

PS. I can work around this by doing `continueOn` within the `EitherT` but it doesn't seem like the correct thing to do as we're pushing threading contexts within that Monad, which should ideally be agnostic.
```kotlin
fun <A, B> MonadSyntax<EitherTPartialOf<ForIO, A>>.liftIO(fa: Kind<ForIO, B>): EitherT<ForIO, A, B> {
    return run { EitherT.liftF(IO.functor(), fa) }
}

val eitherTIOM: IO<Either<String, String>> = eitherTIOMonad().fx.monad {
    liftIO(IO.unit.continueOn(Dispatchers.IO)).bind()
    println("Hi") // <-- now runs on a worker thread
    "Hola"
}.value().fix()
```

pakoito:
can you please do `!effect { println("Hi") }` and see what happens?

ps-feng:
I unable to do that because I'm inside an `EitherT`.

pakoito:
printing to screen is a side-effect. Inlining a side-effect in an `fx` block or a `map` block or somewhere like that is allowed by the language but you're bound to find edge cases like these.

Wrapping side-effects assures correct behavior.

pakoito:
The good news is that @raulraja is working on a purity checker that'd flag these :)

ps-feng:
Yeah, not the best example I guess. 

[...]

PR #1624, opened 2019-09-25
Fix IO.fx eager evaluation

ps-feng:
The previous implementation was launching a coroutine straight away and running the provided code until the first suspension point.

This was causing the issue described in https://github.com/arrow-kt/arrow/issues/1617, which can be summarized as
```kotlin
val someIO = IO.fx {
    someFun() // <-- called on caller's thread because it's evaluated straight away
}

IO.fx {
    continueOn(Dispatchers.IO)
    someIO.bind()
}.unsafeRunAsync { }
```


pakoito:
Could you please see if there are other `fx` functions with the same issue, maybe in the Rx or Reactor packages?

ps-feng:
The issue is indeed quite extended, I'll see if I can fix them all in this PR.

ps-feng:
This is trickier than I expected. Actually the initial problem I reported isn't solved because most of the monads' `fx` extension use `MonadFx.monad`, which launches the coroutine immediately.
```kotlin
interface MonadFx<F> {
  val M: Monad<F>
  fun <A> monad(c: suspend MonadSyntax<F>.() -> A): Kind<F, A> {
    val continuation = MonadContinuation<F, A>(M)
    val wrapReturn: suspend MonadContinuation<F, *>.() -> Kind<F, A> = { just(c()) }
  }
}
```
That's fine for most monads, but not so much for `IO` or transformers that contain `IO`, so my original case is still not fixed by the change suggested in this PR because it's using precisely `MonadFx.monad`:
```kotlin
EitherT.monad(IO.monad()).fx.monad {
    whatever() // <-- eagerly evaluated, then suspends
    someEitherT.bind()
}
```
Any ideas?

pakoito:
`fx.monad` is correct because those don't allow suspension. The ones that do are supposed to be overriden with that version.

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/IOTest.kt
```diff
@@ -282,6 +282,17 @@ class IOTest : UnitSpec() {
         .unsafeRunSync()
     }
 
+    "fx should defer evaluation until run" {
```
JorgeCastilloPrz: 👍

JorgeCastilloPrz:
Shouldn't an `EitherT` for IO allow suspension @pakoito? I was under the impression that it should given it's an `IO` empowered with `Either` capabilities, but still an `IO`.

pakoito:
Ooooooh, that's a tricky one. @raulraja do you know how it is in Scala?

raulraja:
Nine of this matter faster October when the comprehension plugin is out since binding and suspension will be unrelated. Binding it's gonna be translated to flatMap and the only way to inject suspension at compile time is via effect or the IO constructor which does not eagerly run. All the fx interfaces and typeclass will be gone since fx is gonna be powered by synthetic codegen via compiler plugins. As for the current fix whatever works I'm fine with but we can't change signatures now to change them back in October. It would force us into a deprecation cycle we don't want to through before meta and compiler plugins take over the FP style

ps-feng:
If that's the case, should we just close this ticket and wait for the plugin?

JorgeCastilloPrz:
The fix doesn't seem to change signatures, but deprecation is needed for breaking changes not just on signatures. Since when we've got this bug @nomisRev? If it's been there since long ago the fix could also be considered a breaking change given it drastically changes the method behaviour. In the other hand, If it's been there since 0.10.0 only then I say we fix it fast since it seems to be a quite important bug from my point of view.

pakoito:
Let's merge with this. We should fix Rx and Reactor too, if they have the same problem.

ps-feng:
I'm not familiar with coroutines but I believe they have the same problem.
```kotlin
fun <A> MonoK.Companion.fx(c: suspend AsyncSyntax<ForMonoK>.() -> A): MonoK<A> =
  MonoK.async().fx.async(c).fix()

interface AsyncFx<F> : MonadThrowFx<F> {
  val async: Async<F>
  override val ME: MonadThrow<F> get() = async
  fun <A> async(c: suspend AsyncSyntax<F>.() -> A): Kind<F, A> {
    val continuation = AsyncContinuation<F, A>(async)
    val wrapReturn: suspend AsyncSyntax<F>.() -> Kind<F, A> = { just(c()) }
    wrapReturn.startCoroutine(continuation, continuation) // <-- will execute until next bind I think
  }
}
```

pakoito:
Yep, so just wrap in `defer` for now.

pakoito:
Actually, wait, it's

fun <A> async(c: suspend AsyncSyntax<F>.() -> A): Kind<F, A>

what should be wrapped.

ps-feng:
I think it makes no difference but I'll check tomorrow.

ps-feng:
I did verify that the same issue was occuring in Rx and Reactor, so I created a separate PR for them: https://github.com/arrow-kt/arrow/pull/1663

pakoito:
We should have merged this already, damn. I'll check the build.

PR #1627, opened 2019-09-29
Make coproducts covariant

jereksel:


pakoito:
EZPZ. What are you using them for?

jereksel:
Unfortunately, I don't use them personally, but I thought this could be useful for others.

Issue #1614, opened 2019-09-20
Links in Library Quickstart ToC are broken


pakoito:
https://github.com/arrow-kt/arrow/blob/master/modules/docs/arrow-docs/docs/docs/quickstart/libraries/README.md#table-of-content

They need to be relative links, i.e. `({{ '/docs/patterns/glossary/#instances-and-extensions-interfaces' | relative_url }})`

i-walker:
I got this. Thanks, @pakoito for reporting.

PR #1629, opened 2019-09-30
Link quickfix

i-walker:
Regarding https://github.com/arrow-kt/arrow/issues/1614


raulraja:
thanks @i-walker !

pakoito:
Thank you for the fix <3

PR #1647, opened 2019-09-30
Compatibility with incremental annotation processing

jereksel:
Makes Arrow annotation processors compatible with Gradle's incremental compilation - for Gradle >= 5.4.0.

Docs: https://docs.gradle.org/current/userguide/java_plugin.html#sec:incremental_annotation_processing

TODO:
- [ ] Move kdoc generation to separate annotation processor
- [ ] Move coproduct somewhere else - propably separate module due to number of classes


jereksel:
https://github.com/arrow-kt/arrow/issues/1651#issuecomment-537001070

pakoito:
I would keep it tbh. Was there anything needed it to fix it?

PR #1655, opened 2019-10-01
Adding some, many and alt functions to Alternative

### LINKED ISSUE #1410: Add functions alt, some and many to Alternative type class
**Useful links**
* http://hackage.haskell.org/package/base-4.12.0.0/docs/Control-Applicative.html#g:2
* https://blog.jle.im/entry/free-alternative-regexp.html

Cotel:
Close #1410

Should we change `alt` into something else? It is supposed to be Haskell's `<|>`. I've seen some places where it is called `orElse` but it can be misleading.

I've also added a law for testing if `alt` is associative. I suppose `empty` laws are already covered with MonoidK. I took these laws from [here](https://en.wikibooks.org/wiki/Haskell/Alternative_and_MonadPlus).

I'll write some docs in another PR for completing Hacktoberfest 😋


modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
```diff
@@ -228,6 +229,9 @@ interface ListKMonadCombine : MonadCombine<ForListK> {
 
   override fun <A> Kind<ForListK, A>.combineK(y: Kind<ForListK, A>): ListK<A> =
     fix().listCombineK(y)
+
+  override fun <A> Kind<ForListK, A>.alt(b: Kind<ForListK, A>): Kind<ForListK, A> =
```
pakoito: The return can be ListK<A> due to covariance.

pakoito:
I like `orElse`, that way we typeclass it.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Alternative.kt
```diff
@@ -1,6 +1,16 @@
...
 /**
  * ank_macro_hierarchy(arrow.typeclasses.Alternative)
  */
-interface Alternative<F> : Applicative<F>, MonoidK<F>
+interface Alternative<F> : Applicative<F>, MonoidK<F> {
+  fun <A> Kind<F, A>.some(): Kind<F, ListK<A>> = map(this, many()) { (v, vs) -> (listOf(v) + vs).k() }
+
+  fun <A> Kind<F, A>.many(): Kind<F, ListK<A>> = some() alt just(emptyList<A>().k())
+
+  infix fun <A> Kind<F, A>.alt(b: Kind<F, A>): Kind<F, A>
```
pakoito: Please provide at least one non-infix alternative just in case :P

Cotel: Should I leave alt as the infix version and provide an orElse non-infix function? I wish we could use operators and name it <|> 😅

pakoito: ye, works for me

pakoito: One nit. Could you please swap things around so `alt` is implemented in terms of `orElse` instead?

Cotel: Sure 👍

PR #1656, opened 2019-10-01
Using Eq typeclass NonZero class within validation module

franciscodr:
Some people are getting this compilation error in the validation module when building the arrow project:
```
e: ~/arrow/modules/validation/arrow-validation/build/generated/source/kaptKotlin/main/extension/arrow/validation/refinedTypes/numeric/either/nonZero/EitherNonZero.kt: (18, 47): Type argument is not within its bounds: should be subtype of 'Number'
e: ~/arrow/modules/validation/arrow-validation/build/generated/source/kaptKotlin/main/extension/arrow/validation/refinedTypes/numeric/either/nonZero/EitherNonZero.kt: (18, 55): Type mismatch: inferred type is <no name provided> but EitherNonZero<Any?> was expected
```
It seems the issue is caused by the `isNonZero` function that compares a value of the generic type `A` with the literal int `0`.

This pull request proposes using the typeclass `Eq` provided by arrow to do the comparison.

### DIFF
--- a/modules/validation/arrow-validation/src/main/kotlin/arrow/validation/refinedTypes/numeric/NonZero.kt
+++ b/modules/validation/arrow-validation/src/main/kotlin/arrow/validation/refinedTypes/numeric/NonZero.kt
@@ -32,11 +32,12 @@ interface NonZero<F, A : Number> : Refinement<F, A> {
    * {: data-executable='true'}
    *
    * ```kotlin:ank
+   * import arrow.core.extensions.eq
    * import arrow.validation.refinedTypes.numeric.validated.nonZero.nonZero
    *
    * fun main(args: Array<String>) {
    *   //sampleStart
-   *   val result = 0.nonZero()
+   *   val result = 0.nonZero(Int.eq())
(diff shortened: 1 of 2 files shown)


JorgeCastilloPrz:
Do we know the exact cause? the generic type is strongly typed to be descendant of `Number` so it shouldn't fail.

raulraja:
thanks @franciscodr !

franciscodr:
> Do we know the exact cause? the generic type is strongly typed to be descendant of `Number` so it shouldn't fail.

To be honest, I don't know but, in some cases, the code generated for `NonZero.kt` contains these lines:
```
/**
 * cached extension
 */
@PublishedApi()
internal val nonZero_singleton: EitherNonZero<Any?> = object : EitherNonZero<kotlin.Number> {}
```
which causes the compilation error.

However, if I build the arrow project in a different machine, the code is generated correctly:
```
@PublishedApi()
internal val nonZero_singleton: EitherNonZero<Number> = object : EitherNonZero<kotlin.Number> {}
```
Any insight on this?

PR #1661, opened 2019-10-03
Update some dependencies

LordRaydenMK:
Gradle, Kotlin, Coroutines and Retrofit to latest versions.

Note: I didn't update Retorfit, latest version has support for Coroutines and the retrofit integration is no longer required, right?

If yes, I can create a PR to remove it. It's also the only module in the `integrations` module. So do we remove that completely?


aballano:
The CI is failing in Decoder.kt: (211, 5) -> `ClassName(packageName = pckg.value, simpleName = simpleName)`, seems that the constructor changed so the `simpleName` param name is now different. @LordRaydenMK could you please change that one to see if it passes the checks?

LordRaydenMK:
Fixed that, but looks like there are more breaking changes with KotlinPoet. The latest version does `Any?` in a few places where `Any` was expected.

So I will remove KotlinPoet from this PR and update that separately (or maybe together with kotlintest).

LordRaydenMK:
Looks like updating ALL the things wasn't the best idea.... will try it out incrementally and reopen later.

LordRaydenMK:
Looks like the Gradle update breaks things:

https://github.com/kotlintest/kotlintest/issues/645
https://github.com/gradle/gradle/issues/6453

PR #1662, opened 2019-10-03
Adding some docs for Alternative

Cotel:
Docs taken from [here](https://en.wikibooks.org/wiki/Haskell/Alternative_and_MonadPlus) and [here](https://wiki.haskell.org/Typeclassopedia#Failure_and_choice:_Alternative.2C_MonadPlus.2C_ArrowPlus).

Two things that are weird for me.

One is we don't offer `Option.alternative()` and I have to use `Option.monadCombine()` which is subclass of Alternative but it can confuse readers.

The second thing is that one of the sources I've used for writing these docs says that `<|>` combines computations but the other says it is a choice between alternatives. And it seems depending on the instance it is both of those things. Maybe `orElse` is not a good name enough or we should try to explain it better in the docs 🤔.

### DIFF
--- a/modules/docs/arrow-docs/docs/docs/arrow/typeclasses/alternative/README.md
+++ b/modules/docs/arrow-docs/docs/docs/arrow/typeclasses/alternative/README.md
@@ -11,6 +11,60 @@ redirect_from:
 {:.intermediate}
 intermediate
 
+We use [`Option`]({{ '/docs/arrow/core/option' }}) to indicate a computation can fail somehow (that is, it can have either zero results or one result), and we use lists for computations that can have many possible results (ranging from zero to arbitrarily many results). In both of these cases, one useful operation is combining all possible results from multiple computations into a single computation. The `Alternative` type class captures this combination.
+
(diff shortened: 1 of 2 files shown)


modules/docs/arrow-docs/docs/docs/arrow/typeclasses/alternative/README.md
```diff
@@ -11,6 +11,60 @@ redirect_from:
...
+import arrow.core.extensions.option.monadCombine.monadCombine
+
+Option.monadCombine().run {
+    val x = Option.just(1)
+    val y = Option.empty()
+    x alt y
+}
+```
+
+### Laws
```
pakoito: I'd add `some` and `many` but not as a hard requirement.

Also I've just realized `some` may clash with Option's constructor. Oh well.

Cotel: Yup, didn't realize earlier. Anyway, you need to be in the scope of `Alternative` to have access to this `some` and it will only appear as a function for the object. `some` as an Option constructor won't be suggested by the IDE in that situation 🤔.

Cotel:
I've realized defining `some` and `many` with recursion calling each other results in StackOverflow.  I read that there are specialized implementation for Option and List as it wouldn't be performant even if it worked. I'll write these implementations and docs for them in another PR (spamming Hacktober 😆)

PR #1663, opened 2019-10-03
Fix fx eager evaluation in Reactor and RxJava

ps-feng:
Follows up on https://github.com/arrow-kt/arrow/pull/1624


pakoito:
Thanks for following up! Nice 🙇‍♂️

pakoito:
Both fixes are in now!

PR #1664, opened 2019-10-04
Fixing Alternative's some and many and adding docs for them

Cotel:
I had to change `some` and `many` result type from `Kind<F, ListK<A>>` into `Kind<F, SequenceK<A>>` because these functions will repeat the current computation until it fails. But for example `Option.just(1)` won't fail never so the list will be infinite so we need to work with lazy collections here.

Anyway, the default implementation for these functions provided in Alternative will probably result in stackoverflow if users create their own instances. People from Purescript or Cats don't have these functions implemented in their Alternative typeclasses 🤔


Cotel:
> Why `A: Any`? We allow nulls in polymorphic.

It is for `generateSequence`. It has a constraint that forces `A : Any` 🤷‍♂

pakoito:
There has to be a generator that doesn't require it, even if it's just implementing the interface directly. Are they using null to signal termination or what?

Cotel:
Indeed https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.sequences/generate-sequence.html

> Returns a sequence which invokes the function to calculate the next value on each iteration until the function returns null.

Cotel:
I cannot find any other generator function. Could it be a good idea to make our own generator for `SequenceK` like `fun <A> generateSequenceK(f: () -> Option<A>): SequenceK<A>` ?

pakoito:
Have you tried implementing the `sequence` interface?

Cotel:
Where should I try to implement it? At the alternative typeclass?

pakoito:
I mean inline, in the places where you're writing `generateSequence`

Cotel:
I'm not sure if I understood you but if I mark functions using `generateSequence` as inline nothing changes. It still complains about A not being child of Any 🤔.

pakoito:
I meant using the Sequence class directly, with the Sequence constructor: https://bit.ly/31YZtmW There are a few examples of possible implementations on that same file.

But after reading them and seeing how they behave, I'd rather the implementation was changed to `Iterable` maybe.

I don't know which one would work.

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/sequenceK.kt
```diff
@@ -247,13 +247,21 @@ interface SequenceKMonadCombine : MonadCombine<ForSequenceK> {
...
 
-  override fun <A : Any> Kind<ForSequenceK, A>.many(): SequenceK<SequenceK<A>> =
+      override fun next(): A = it
+    } }.k() }.k()
+
+  override fun <A> Kind<ForSequenceK, A>.many(): SequenceK<SequenceK<A>> =
     if (this.fix().isEmpty()) sequenceOf(emptySequence<A>().k()).k()
-    else map { generateSequence { it }.k() }.k()
+    else map { Sequence { object : Iterator<A> {
+      override fun hasNext(): Boolean = true
```
pakoito: Wait, this passes the tests? lol

I'd expect that you'd need to do some indexed map or something, and use that to check instead.

pakoito: In retrospect, yeah, this makes sense lol

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
```diff
@@ -234,13 +234,21 @@ interface ListKMonadCombine : MonadCombine<ForListK> {
   override fun <A> Kind<ForListK, A>.orElse(b: Kind<ForListK, A>): ListK<A> =
     fix() + b.fix()
 
-  override fun <A : Any> Kind<ForListK, A>.some(): ListK<SequenceK<A>> =
+  override fun <A> Kind<ForListK, A>.some(): ListK<SequenceK<A>> =
     if (this.fix().isEmpty()) ListK.empty()
-    else map { generateSequence { it }.k() }.k()
+    else map { Sequence { object : Iterator<A> {
```
pakoito: Isn't this equivalent to `sequenceOf(it)`?

Cotel: Nope. `sequenceOf(it)` generates a sequence of only one element. This way we are generating an infinite sequence of one repeated element.

pakoito:
Awesome, going merge!

PR #1667, opened 2019-10-04
[META] Testing plugin: initial version

rachelcarmena:


modules/meta/arrow-meta-prototype/testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/Compilation.kt
```diff
@@ -0,0 +1,65 @@
...
+
+data class CompilationData(
+  val sourceFileName: String,
+  val sourceContent: String,
+  val generatedFileContent: String,
+  val generatedClasses: ArrayList<String>,
+  val compilationResult: CompilationResult
+)
+
+fun testCompilation(compilationData: CompilationData) {
```
raulraja: Even for functions that effectful and return `Unit` I consider good practice for all public API's to always specify what they return so there is no ambiguity in the reader as to what type of inference came up with.

modules/meta/arrow-meta-prototype/testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/Compilation.kt
```diff
@@ -0,0 +1,65 @@
...
+  val kotlinSource = SourceFile.kotlin(compilationData.sourceFileName, compilationData.sourceContent)
+
+  val result = KotlinCompilation().apply {
+    sources = listOf(kotlinSource)
+    classpaths = listOf(
+      // TODO
+      // classpathOf("arrow-annotations:x.x.x")
```
raulraja: if we activate the mode in which the plugin includes what's on the test classpath we can just add these dependencies to the build file itself

modules/meta/arrow-meta-prototype/testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/Compilation.kt
```diff
@@ -0,0 +1,65 @@
...
+    )
+    pluginClasspaths = listOf(
+      classpathOf("compiler-plugin")
+//      File("../compiler-plugin/build/libs/compiler-plugin.jar")
+    )
+  }.compile()
+
+  assertThat(result.exitCode).isEqualTo(exitCodeFrom(compilationData.compilationResult))
+
+  if (!result.exitCode.equals(KotlinCompilation.ExitCode.OK)) return
```
raulraja: Early returns like this are great for performance but I think in this case is better to take the bottom of the function remaining and splitting in a new function that is conditionally applied based on the value of the exit code.

modules/meta/arrow-meta-prototype/testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/Compilation.kt
```diff
@@ -0,0 +1,65 @@
...
+
+  val actualGeneratedFileContent = Paths.get(result.outputDirectory.parent, "sources", "${compilationData.sourceFileName}.meta").toFile().readText()
+  assertThat(actualGeneratedFileContent).isEqualTo(compilationData.generatedFileContent)
+
+  val actualGeneratedClasses = result.generatedFiles.map { it.name }.filter { it.endsWith(".class") }
+  assertThat(actualGeneratedClasses).isEqualTo(compilationData.generatedClasses.map { "$it.class" })
+}
+
+fun contentFromResource(fromClass: Class<Any>, resourceName: String): String = fromClass.getResource("/$resourceName").readText()
```
raulraja: We frequently follow the style:
```
fun declaration(): RetType =
  expressionBody
```
To easily tell apart the function signature from the actual implementation which gets confusing if its all in one line.

modules/meta/arrow-meta-prototype/testing-plugin/src/test/kotlin/arrow/meta/plugin/testing/CompilationTest.kt
```diff
@@ -0,0 +1,35 @@
...
+
+  @Test
+  fun `metadebug consideration works as expected`() {
+
+    testCompilation(
+      CompilationData(
+        sourceFileName = "Example.kt",
+        sourceContent = contentFromResource(javaClass, "Example.kt.source"),
+        generatedFileContent = contentFromResource(javaClass, "Example.kt.meta"),
+        generatedClasses = arrayListOf("ExampleKt", "ForId2", "Id2", "ForId2\$Companion"),
```
raulraja: 😍

modules/meta/arrow-meta-prototype/testing-plugin/src/test/resources/Example.kt.meta
```diff
@@ -0,0 +1,17 @@
+import arrow.higherkind
```
raulraja: This file should not be part of VCS as it's rewritten in each compilation by meta

modules/meta/arrow-meta-prototype/testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/Compilation.kt
rachelcarmena: I love that guideline!

modules/meta/arrow-meta-prototype/testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/Compilation.kt
rachelcarmena: Ok! I'm waiting for the new release of Arrow, because I didn't find `higherkind`in the current release. When it's ready, I was thinking about:

* `build.gradle`
```
dependencies {
    testRuntime "io.arrow-kt:arrow-annotations:0.10.1"
    ...
```
* `Compilation.kt`
```
classpaths = listOf(classpathOf("arrow-annotations:0.10.1"))
```

modules/meta/arrow-meta-prototype/testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/Compilation.kt
rachelcarmena: I love that idea! Thanks!

modules/meta/arrow-meta-prototype/testing-plugin/src/main/kotlin/arrow/meta/plugin/testing/Compilation.kt
rachelcarmena: Absolutely agree. Thanks!

modules/meta/arrow-meta-prototype/testing-plugin/src/test/resources/Example.kt.meta
rachelcarmena: It's just used for the comparison as the **expected** result.

The **actual** files are generated in the `/tmp` directory by default:

```
/tmp/Kotlin-Compilation7741532578741152250
├── classes
│   ├── ExampleKt.class
│   ├── ForId2.class
│   ├── ForId2$Companion.class
│   ├── Id2.class
│   └── META-INF
│       └── main.kotlin_module
├── kapt
│   ├── incrementalData
│   ├── kotlinGenerated
│   ├── sources
│   └── stubs
└── sources
    ├── Example.kt
    └── Example.kt.meta
```

modules/meta/arrow-meta-prototype/testing-plugin/src/test/kotlin/arrow/meta/plugin/testing/CompilationTest.kt
rachelcarmena: I love you like it! I tried to wrap the external library and to hide all the complexity.

rachelcarmena:
I uploaded a [new commit with all the improvements](https://github.com/arrow-kt/arrow/pull/1667/commits/fdcdb142308c3602f38bd75e5ffae30a5199d19e).

Thanks for the review @raulraja !!

PR #1682, opened 2019-10-09
Add easy constructor for FunctionK

pakoito:
```
FunctionK { op: TryOf<A> ->
  op.fix().toEither()
}
```


pakoito:
Derp there are no higher order generics on lambdas, I forgot. This cannot be built as is :D

PR #1684, opened 2019-10-09
removes Russian translations

i-walker:
None of the regular maintainers can assure that these docs are up-to-date. We are planning on supporting multiple translations at a later point in time, but we don't have regular contributors in that area, yet.


pakoito:
I would add a description of why we're removing it, given that someone put effort on it :)

tl;dr none of the regular maintainers can assure it's up-to-date

rachelcarmena:
And I would move the files to a folder like `translations` with a README.md with that description (instead of removing them), mentioning the authors and encouraging in case someone wants to keep them updated.

rcd27:
Russian native speaker here, guys. Just ping me when you are ready.

i-walker:
We will definitely love your contribution. After we set up the process :)

PR #1685, opened 2019-10-10
Adding some docs for MonadCombine

Cotel:
Typeclass Hierarchy for MonadCombine is not being displayed currently at the website and I don't know how to fix that. I suspect it is because docs are in `docs/arrow/mtl/typeclasses` instead of being at `/docs/arrow/typeclasses` 🤔 (every other typeclass in mtl folder is not displaying its hierarchy)


pakoito:
IN!

PR #1691, opened 2019-10-10
KIO moved to jitpack.io

colomboe:
Sorry I've updated the repository for the KIO dependency (the old one still works at the moment).


pakoito:
Nice, will update in a second :D

pakoito:
Mergiiiiing

PR #1697, opened 2019-10-11
Clean up and improve fragment provider update handling

jansorg:
CI isn't passing, but that's already the case with `rr-meta-prototype-integration`.

# Changes
- register MetaPlugin's plugins for each project when project is opened. Before this change it was done when the fragment provider was called for the first time. Now this work is done when the project is loaded (by using a `ProjectComponent`)
- scope cache per project. Before a global cache was shared between projects. Changes in one project clear the cache which then releases cached data needed for another project. The other project refreshes, too, and the cache could end up with incomplete content
- init cache once when project is opened and not when the fragment provider was called for the first time
- cache refresh is now handling dumb mode correctly, i.e.  `IndexNotReadyException` shouldn't be thrown anymore while refreshing the cache
- split vfs event handling from the actual fragment provider. Keeping this separate makes it clearer when a refresh is happening
- extract DiagnosticSuppressor from the fragment provider, don't keep two instances per project of the  provider doing different jobs. Before `MetaSyntheticPackageFragmentProvider` was instantiated once for `<packageFragmentProviderExtension>` and once for `<diagnosticSuppressor>`. Each used distinct parts of the class, but init code was called at least twice
- changed printing to stdout to proper IntelliJ logging. This way the level of detail is configurable at runtime and the output will be properly recorded in the idea.log file

### DIFF
--- a/modules/meta/arrow-meta-prototype/README.md
+++ b/modules/meta/arrow-meta-prototype/README.md
@@ -5,3 +5,22 @@
 ```
 ./gradlew clean gradle-plugin:publishArrowPluginMarkerMavenPublicationToMavenLocal gradle-plugin:publishPluginMavenPublicationToMavenLocal idea-plugin:runIde -Dorg.gradle.debug=true -Dkotlin.compiler.execution.strategy="in-process"
 ```
+
+## Logging
+Logging inside of an IntelliJ plugin is done via 
+```
+val LOG = com.intellij.openapi.diagnostic.Logger.getInstance("#arrow.yourLogChannel")
+// LOG.error, LOG.warn, LOG.info, LOG.debug, LOG.trace are available 
+```
+
+You can see the messages in IntelliJ's log file. The log file is available at `Help->Show Log in ...`. 
+debug and trace messages are not shown by default. To enable these, enter the full channel ID at 
+`Help->Debug Log Settings`, for example
+```
+#arrow.trace
+```
+
+To enable trace message, append `:trace` to a channel ID. For example:
+```
+#arrow.trace:trace
+```
--- a/modules/meta/arrow-meta-prototype/README.md
+++ b/modules/meta/arrow-meta-prototype/README.md
@@ -13,8 +13,9 @@ val LOG = com.intellij.openapi.diagnostic.Logger.getInstance("#arrow.yourLogChan
 // LOG.error, LOG.warn, LOG.info, LOG.debug, LOG.trace are available 
 ```
 
-You can see the messages in IntelliJ's log file. The log file is available at `Help->Show Log in ...`. 
-debug and trace messages are not shown by default. To enable these, enter the full channel ID at 
+You can now see the messages in IntelliJ's log file. The log file is available at `Help->Show Log in ...`.
+ 
+`debug` and `trace` messages are not shown by default. To show them, enter the full channel ID at 
 `Help->Debug Log Settings`, for example
 ```
 #arrow.trace
--- a/modules/meta/arrow-meta-prototype/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/MetaPluginRegistrarComponent.kt
+++ b/modules/meta/arrow-meta-prototype/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/MetaPluginRegistrarComponent.kt
@@ -17,14 +17,13 @@ class MetaPluginRegistrarComponent(val project: Project) : ProjectComponent {
   override fun getComponentName(): String = "arrow.meta.registrar"
 
   override fun initComponent() {
-    LOG.warn("initComponent()")
+    LOG.info("initComponent()")
 
     val start = System.currentTimeMillis()
-    // fixme use CompilerConfiguration.EMPTY ?
     val configuration = CompilerConfiguration()
     metaPlugin.registerMetaComponents(project, configuration)
 
-    LOG.warn("initComponent() took ${System.currentTimeMillis() - start}ms")
+    LOG.info("initComponent() took ${System.currentTimeMillis() - start}ms")
   }
 
   override fun disposeComponent() {
--- a/modules/meta/arrow-meta-prototype/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/MetaPluginRegistrarComponent.kt
+++ b/modules/meta/arrow-meta-prototype/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/MetaPluginRegistrarComponent.kt
@@ -27,12 +27,10 @@ class MetaPluginRegistrarComponent(val project: Project) : ProjectComponent {
   }
 
   override fun disposeComponent() {
-    // fixme: make sure that all registered components are disposed
+    // TODO: make sure that all registered components are disposed
   }
 
-  override fun projectClosed() {
-  }
+  override fun projectClosed() {}
 
-  override fun projectOpened() {
-  }
+  override fun projectOpened() {}
 }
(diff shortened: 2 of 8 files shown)


modules/meta/arrow-meta-prototype/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/MetaPluginRegistrarComponent.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    val start = System.currentTimeMillis()
+    val configuration = CompilerConfiguration()
+    metaPlugin.registerMetaComponents(project, configuration)
+
+    LOG.info("initComponent() took ${System.currentTimeMillis() - start}ms")
+  }
+
+  override fun disposeComponent() {
+    // fixme: make sure that all registered components are disposed
```
i-walker:
```suggestion
    // TODO: make sure that all registered components are disposed
```

modules/meta/arrow-meta-prototype/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/MetaPluginRegistrarComponent.kt
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  override fun projectClosed() {
+  }
+
+  override fun projectOpened() {
+  }
```
i-walker:
```suggestion
```

modules/meta/arrow-meta-prototype/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/MetaPluginRegistrarComponent.kt
```diff
@@ -0,0 +1,38 @@
...
+
+    LOG.info("initComponent() took ${System.currentTimeMillis() - start}ms")
+  }
+
+  override fun disposeComponent() {
+    // fixme: make sure that all registered components are disposed
+  }
+
+  override fun projectClosed() {
+  }
```
i-walker:
```suggestion
```

modules/meta/arrow-meta-prototype/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/MetaPluginRegistrarComponent.kt
```diff
@@ -0,0 +1,38 @@
...
+    metaPlugin.registerMetaComponents(project, configuration)
+
+    LOG.info("initComponent() took ${System.currentTimeMillis() - start}ms")
+  }
+
+  override fun disposeComponent() {
+    // fixme: make sure that all registered components are disposed
+  }
+
+  override fun projectClosed() {
```
i-walker:
```suggestion
  override fun projectClosed() {}
```

i-walker:
Add explicit ReturnTypes everywhere

modules/meta/arrow-meta-prototype/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/phases/resolve/DiagnosticSuppressor.kt
```diff
@@ -0,0 +1,58 @@
...
+import org.jetbrains.kotlin.diagnostics.DiagnosticWithParameters2
+import org.jetbrains.kotlin.diagnostics.Errors
+import org.jetbrains.kotlin.psi.KtElement
+import org.jetbrains.kotlin.psi.KtParameter
+import org.jetbrains.kotlin.psi.KtWhenExpression
+import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.kotlin.types.checker.KotlinTypeChecker
+import org.jetbrains.kotlin.utils.addToStdlib.safeAs
+
+open class DiagnosticSuppressor : org.jetbrains.kotlin.resolve.diagnostics.DiagnosticSuppressor {
```
i-walker: Is there a reason why this has to be an open class?

modules/meta/arrow-meta-prototype/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/phases/resolve/package.kt
```diff
@@ -0,0 +1,8 @@
+package arrow.meta.plugin.idea.phases.resolve
+
+import com.intellij.openapi.diagnostic.Logger
+
+/**
+ * @author jansorg
```
i-walker: @raulraja should we add Authors?

jansorg: This slipped through, I'll remove it

modules/meta/arrow-meta-prototype/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/phases/resolve/DiagnosticSuppressor.kt
jansorg: No reason

modules/meta/arrow-meta-prototype/README.md
```diff
@@ -5,3 +5,23 @@
 ```
 ./gradlew clean gradle-plugin:publishArrowPluginMarkerMavenPublicationToMavenLocal gradle-plugin:publishPluginMavenPublicationToMavenLocal idea-plugin:runIde -Dorg.gradle.debug=true -Dkotlin.compiler.execution.strategy="in-process"
 ```
+
```
i-walker: Can you add how users enable the Logger so we can collect the Data?

Please Move this part to a new ReadMe in the dir of IdeSyntax

modules/meta/arrow-meta-prototype/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/phases/resolve/MetaSyntheticPackageFragmentProvider.kt
```diff
@@ -184,13 +117,6 @@ class MetaSyntheticPackageFragmentProvider :
...
     lookupTracker: LookupTracker
   ): PackageFragmentProvider? {
-    project.registerIdeStack {
-      initializeFileListener(project)
-      computeCache(project)
-    }
-//    println("MetaSyntheticPackageFragmentProvider.getPackageFragmentProvider:, cache:\n ${descriptorCache.toList().joinToString {
-//      "${it.first} : ${it.second.size}"
-//    }}")
     return DescriptorCachePackageFragmentProvider(module)
```
i-walker: turn this into an Expression body

modules/meta/arrow-meta-prototype/idea-plugin/src/main/kotlin/arrow/meta/plugin/idea/phases/resolve/MetaSyntheticPackageFragmentProvider.kt
```diff
@@ -205,7 +131,6 @@ class MetaSyntheticPackageFragmentProvider :
   }
 
   inner class BuildCachePackageFragmentDescriptor(module: ModuleDescriptor, fqName: FqName) : PackageFragmentDescriptorImpl(module, fqName) {
```
i-walker: does this has to be an inner class or can we move this to an object and pass it implicitly

raulraja:
Thanks so much @jansorg . This is great. @i-walker let's fix small stuff ourselves so @jansorg can focus on the stability of the plugin. @jansorg test nexts sounds great. Thanks for your time and effort making this stable.

PR #1702, opened 2019-10-12
Implemented AQL's Max and Min 

maruiz81:
Added implementation for Max and Min in AQL

### DIFF
--- a/modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Max.kt
+++ b/modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Max.kt
@@ -1,0 +1,26 @@
+package arrow.aql
+
+import arrow.core.ForId
+import arrow.core.Id
+import arrow.core.identity
+import arrow.core.value
+import arrow.typeclasses.Foldable
+import kotlin.math.max
+
+interface Max<F> {
+
+  fun foldable(): Foldable<F>
+
+  infix fun <A, Z> Query<F, A, Z>.max(f: A.() -> Long): Query<ForId, Long, Long> =
+    foldable().run {
+      Query(
+        select = ::identity,
+        from = Id(from.foldLeft(Long.MIN_VALUE) { acc, a -> max(f(a), acc) })
+      )
+    }
+
+  fun Query<ForId, Long, Long>.value(): Long =
+    foldable().run {
+      this@value.from.value()
+    }
+}
--- a/modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Max.kt
+++ b/modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Max.kt
@@ -20,7 +20,7 @@ interface Max<F> {
       from = Id(foldable().run {
         from.foldLeft(None) { acc: Option<X>, a: A ->
           acc.fold({ Some(f(a)) },
-            { Some(if (ord.run { it > (f(a)) }) it else f(a)) })
+            { Some(if (ord.run { it > f(a) }) it else f(a)) })
         }
       }))
 
--- a/modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Max.kt
+++ b/modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Max.kt
@@ -4,20 +4,25 @@ import arrow.core.ForId
 import arrow.core.Id
 import arrow.core.identity
 import arrow.core.value
+import arrow.core.None
+import arrow.core.Some
+import arrow.core.Option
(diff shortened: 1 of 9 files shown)


modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Max.kt
```diff
@@ -0,0 +1,26 @@
...
+import arrow.core.identity
+import arrow.core.value
+import arrow.typeclasses.Foldable
+import kotlin.math.max
+
+interface Max<F> {
+
+  fun foldable(): Foldable<F>
+
+  infix fun <A, Z> Query<F, A, Z>.max(f: A.() -> Long): Query<ForId, Long, Long> =
```
pakoito: Instead of specializing for `Long`, require an `Order` instance: https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Order.kt

That way it works for everything and future-proofs it :D

modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Max.kt
```diff
@@ -0,0 +1,26 @@
...
+    foldable().run {
+      Query(
+        select = ::identity,
+        from = Id(from.foldLeft(Long.MIN_VALUE) { acc, a -> max(f(a), acc) })
+      )
+    }
+
+  fun Query<ForId, Long, Long>.value(): Long =
```
pakoito: There's a chance you may end up returning `Option<A>` where `A` is the type of the `Order`. This is because you won't have a min/max value to default to when the container is empty.

Either way, returning Long.MIN_VALUE as the min from an empty array seems incorrect.

pakoito:
Good start, some tips on the design to generalize it!

modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Max.kt
maruiz81: Make sense, I am gonna do some changes ;)

modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Max.kt
maruiz81: You are totally right, I'll come back with your suggested changes.

modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Max.kt
```diff
@@ -4,20 +4,25 @@ import arrow.core.ForId
...
-        from = Id(from.foldLeft(Long.MIN_VALUE) { acc, a -> max(f(a), acc) })
-      )
-    }
+  fun <A, X, Z> Query<F, A, Z>.max(ord: Order<X>, f: A.() -> X): Query<ForId, Option<X>, Option<X>> =
+    Query(
+      select = ::identity,
+      from = Id(foldable().run {
+        from.foldLeft(None) { acc: Option<X>, a: A ->
+          acc.fold({ Some(f(a)) },
+            { Some(if (ord.run { it > (f(a)) }) it else f(a)) })
```
pakoito: Nit: `(f(a))` doesn't need the additional parenthesis.

maruiz81: ups, fixed!

maruiz81:
Ohh! yes, I was too fast. I have fixed

modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Min.kt
```diff
@@ -0,0 +1,31 @@
...
+import arrow.core.Some
+import arrow.core.None
+import arrow.typeclasses.Foldable
+import arrow.typeclasses.Order
+
+interface Min<F> {
+
+  fun foldable(): Foldable<F>
+
+  fun <A, Y, Z> Query<F, A, Y>.min(ord: Order<Z>, f: A.() -> Z): Query<ForId, Option<Z>, Option<Z>> =
```
pakoito: I'm still trying to figure out what happens with the `Y` lol

pakoito: I guess we take the result of the existing one and ignore it, so `Any` or `*` would work the same.

pakoito:
It just clicked. So the return doesn't have to be the value we've mapped to, rather the original value. That way you can pick the largest User by proxy of their age, and not just the age.

PR #1710, opened 2019-10-15
Replace usages of java.util.concurrent.atomic with Kotlin's atomic-fu

aballano:
This fixes an crash when Arrow is run on Android with API < 24.

The crash is happening because Arrow uses AtomicReference, where some functions were added on Android in a [latter API version](https://developer.android.com/reference/java/util/concurrent/atomic/AtomicReference#updateAndGet(java.util.function.UnaryOperator%3CV%3E))

cc @arrow-kt/maintainers


modules/core/arrow-core-data/src/main/kotlin/arrow/unsafe.kt
```diff
@@ -1,18 +1,19 @@
...
+import kotlinx.atomicfu.AtomicRef
+import kotlinx.atomicfu.atomic
 import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.EmptyCoroutineContext
 import kotlin.coroutines.RestrictsSuspension
 import kotlin.coroutines.startCoroutine
 
 private class UnsafeContinuation<A>(
-  val result: AtomicReference<A> = AtomicReference()
+  val result: AtomicRef<A?> = atomic(null)
```
JorgeCastilloPrz: Instead of nullable type can't we lateinit?

aballano: I don't think so, if you take a closer look at Java's AtomicRefernece you'll see that they don't initialize the value, which means it will be null until initialized, and it will throw when we get it because is nullable and we expect A.

So this is the most straightforward conversion afaik unless we want to handle that case specifically

aballano:
@pakoito Unfortunately is not, seems that atomic fu cannot be used in local variables without being wrapped

pakoito:
Go have a rest, tomorrow is a big daaaay.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/Concurrent.kt
```diff
@@ -274,15 +274,15 @@ interface Concurrent<F> : Async<F> {
...
       val cb1 = { r: Either<Throwable, A> ->
         try {
           cb(r)
         } finally {
           if (!state.compareAndSet(null, mapUnit)) {
-            val cb2 = state.get()
+            val cb2 = state.value
             state.lazySet(null)
-            cb2(rightUnit)
+            cb2!!(rightUnit)
```
pakoito: Nit: assert null at the dereference, not on the use

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/Atomic.kt
```diff
@@ -0,0 +1,23 @@
+package arrow.core.extensions
+
+import kotlinx.atomicfu.atomic
+import kotlinx.atomicfu.updateAndGet
+
+// TODO should this be hidden from the outside?
```
pakoito: If internal works, great, otherwise I'd ignore it.

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/Atomic.kt
```diff
@@ -0,0 +1,23 @@
...
+    }
+    get() = atomicRef.value
+
+  fun getAndSet(a: A) = atomicRef.getAndSet(a)
+
+  fun updateAndGet(function: (A) -> A) = atomicRef.updateAndGet(function)
+
+  fun compareAndSet(expect: A, update: A) = atomicRef.compareAndSet(expect, update)
+
+  fun lazySet(a: A) = atomicRef.lazySet(a)
```
pakoito: Maybe equals, hash, and a sensible toString delegated to the atomicref?

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
```diff
@@ -105,13 +105,13 @@ object BracketLaws {
 
   fun <F> Bracket<F, Throwable>.bracketMustRunReleaseTask(EQ: Eq<Kind<F, Int>>): Unit =
     forAll(Gen.int(), Gen.int().applicativeError(this)) { i, fa ->
-      val msg: AtomicReference<Int> = AtomicReference(0)
+      val msg = Atomic(0)
```
pakoito: Nit: type hint here

modules/fx/arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/MonoK.kt
```diff
@@ -85,11 +85,11 @@ data class MonoK<out A>(val mono: Mono<out A>) : MonoKOf<A> {
    */
   fun <B> bracketCase(use: (A) -> MonoKOf<B>, release: (A, ExitCase<Throwable>) -> MonoKOf<Unit>): MonoK<B> =
     MonoK(Mono.create<B> { sink ->
-      val isCanceled = AtomicBoolean(false)
-      sink.onCancel { isCanceled.set(true) }
+      val isCanceled = Atomic(false)
```
pakoito: Knowing JBs I'm surprised there isn't a specialised boolean/int.

aballano: There is, I'll create wrappers and replace the refs ;)

aballano:
Verified on a local project that the fix works when it was crashing before ✅

PR #1713, opened 2019-10-15
[wip] Test support for the Arrow IntelliJ plugin

jansorg:
# Changes:
- base class `KotlinIdeBaseTest` which offers helpers to simplify testing Arrow Kotlin code. This will be expanded with more functionality as this PR progresses.
- Test `IconProviderTest` to show how to use the new class
- Run `gradle clean build` in `modules/meta/arrow-meta-prototype/idea-plugin` to run the IntelliJ tests in a headlessly and/or on a CI server.

# To do
## Integration of Arrow metadata in tests
Tests like `IconProviderTest` need Arrow metadata to work.
This metadata is generated by the Arrow Meta compiler plugins. 

These plugins are executed by the Gradle build. The resulting metadata is stored in `*.class` files in folders named `build`. This data must now be made available for the unit tests of Arrow meta to simplify the implementation and to enable CI integration.

I can think of two ways how to integrate this with our gradle tests:
1. Generate the metadata at runtime. The advantage is that this is a bit more universal. Downside is that this is much more complex to implement, prone to errors and probably slower because metadata generation is slow and it's not possible to cache this data.
2. Generate the metadata at compile time, i.e. as part of the main gradle build and before tests are executed. Advantages are that it's probably faster than (1), that the data could be cached and that build issues are easier to debug. Disadvantage is, that this setup only allows to use kotlin files on disk and not Kotlin code defined at runtime and/or by string values.
At this time I assume that this is the better solution to add test support to the Arrow IntelliJ plugin.

### How (2) could work
1.  Gradle needs to be setup to generate Arrow Meta metadata for code used in the unit tests
2. IntelliJ's SDK tests create a temporary IntelliJ project at runtime. It's stored at a temp location. We have to modify the setup of this temp project to integrate the metadata resolving. 
  1. either copy the `build` folder containing the class file with metadata into the `build` folder of the temp project
  2. or add a new source folder to include the original location of the `build` folder in the temp project. (1) is probably best, because the original data can not be modified between tests
3. Add tests to make sure that the basic resolve functionality is working before more specific tests are added

### Requirements implement (2)
1. Decide where to store source code files for the tests.  `src/test/kotlin-test` might be a good location. This should be separate from the Kotlin code which defines the test cases. It's probably also a good idea to keep a package structure in this folder which allows to identify the test case using it, e.g. `src/test/kotlin-test/icon-provider/higherKind.kt`
2. Configure the Gradle build to generate metadata for the folder defined in (1). The IDE test cases are executed in phase `test`, the metadata must be available before this is called.
3. `gradle clean` should remove the generated metadata


modules/meta/arrow-meta-prototype/idea-plugin/src/test/kotlin/arrow/meta/IconProviderTest.kt
```diff
@@ -0,0 +1,50 @@
...
+      // fixme: this is not yet working, because
+      //   - the kotlin class was created on the fly and the Kotlin compiler plugin has
+      //   not yet been run on this source file
+      //   Possible fixes:
+      //   - integrate the Kotlin compiler and the Arrow compiler plugin somehow at runtime
+      //   - don't create a file on-the-fly, but store it on disk. Generate the arrow bytecode and add
+      //     the generated bytecode data to the vcs.
+      //     Then make use of it in the tests, i.e. load the source from disk
+      val psiType = (psi as KtProperty).type()
+      Assert.assertTrue("psi element must resolve to a valid type", psiType !is UnresolvedType)
```
raulraja: This looks great and it's the kind of test we were looking for

nomisRev:
Moved to https://github.com/arrow-kt/arrow-meta

PR #1705, opened 2019-10-13
Test with semigroup and monoid laws

devesh-shetty:
This PR addresses https://github.com/arrow-kt/arrow/issues/460

The following tests have been updated to use semigroup or monoid laws:
- [x] ListKTest
- [x] SequenceKTest
- [x] SetKTest
- [x] TupleTest
- [x] StringTest
- [x] AndMonoid
- [x] ~all extensions in numbers.kt~ all tests already covered in `NumberSemiringTest`

PR #1719, opened 2019-10-17
Use kotlin-collections import alias to not uselessly self-call in Set…

glamm:
…K.combineAll


modules/core/arrow-core/src/test/kotlin/arrow/core/extensions/setkTest.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import arrow.core.SetK
+import arrow.core.extensions.setk.monoid.monoid
+import arrow.core.k
+import arrow.test.UnitSpec
+import io.kotlintest.shouldBe
+
+class setkTests: UnitSpec() {
+  init {
+    "SetK combineAll leads to StackOverflow #1717" {
```
pakoito: Love the error-linked test

pakoito: Could you please move it with the other tests in SetKTests in core?

pakoito:
In the end @devesh-shetty was already working on a PR that added missing tests such as this one, and it caught the error even if he didn't know about these. It's merged already: https://github.com/arrow-kt/arrow/pull/1705, so should we close this now?

pakoito:
Bob said it's okay to merge!

devesh-shetty:
> Bob said it's okay to merge!

But isn't the fix already in? 🤔

pakoito:
*close

PR #1722, opened 2019-10-19
Max and Min Shouldn't modify the Query value

maruiz81:
We need to use a value to define which row is max or min, but not necessarily this value has to be returned.

### DIFF
--- a/modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Max.kt
+++ b/modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Max.kt
@@ -14,17 +14,17 @@ interface Max<F> {
 
   fun foldable(): Foldable<F>
 
-  fun <A, Y, Z> Query<F, A, Y>.max(ord: Order<Z>, f: A.() -> Z): Query<ForId, Option<Z>, Option<Z>> =
+  fun <A, Y, Z> Query<F, A, Y>.max(ord: Order<Z>, f: A.() -> Z): Query<ForId, Option<Y>, Option<Y>> =
     Query(
       select = ::identity,
       from = Id(foldable().run {
-        from.foldLeft(None) { acc: Option<Z>, a: A ->
-          acc.fold({ Some(f(a)) },
-            { Some(if (ord.run { it > f(a) }) it else f(a)) })
-        }
+        from.foldLeft(None) { acc: Option<A>, a: A ->
+          acc.fold({ Some(a) },
+            { Some(if (ord.run { f(it) > f(a) }) it else a) })
+        }.map(select)
       }))
 
-  fun <Z> Query<ForId, Option<Z>, Option<Z>>.value(): Option<Z> =
+  fun <A, Y> Query<ForId, Option<Y>, Option<Y>>.value(): Option<Y> =
     foldable().run {
       this@value.from.value()
     }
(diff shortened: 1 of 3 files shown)


modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Max.kt
```diff
@@ -14,17 +14,17 @@ interface Max<F> {
...
-            { Some(if (ord.run { it > f(a) }) it else f(a)) })
-        }
+        from.foldLeft(None) { acc: Option<A>, a: A ->
+          acc.fold({ Some(a) },
+            { Some(if (ord.run { f(it) > f(a) }) it else a) })
+        }.map(select)
       }))
 
-  fun <Z> Query<ForId, Option<Z>, Option<Z>>.value(): Option<Z> =
+  fun <A, Y> Query<ForId, Option<Y>, Option<Y>>.value(): Option<Y> =
```
pakoito: is foldable() necessary here?

pakoito: A is never used btw

modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Min.kt
```diff
@@ -14,17 +14,17 @@ interface Min<F> {
...
       select = ::identity,
       from = Id(foldable().run {
-        from.foldLeft(None) { acc: Option<Z>, a: A ->
-          acc.fold({ Some(f(a)) },
-            { Some(if (ord.run { it < f(a) }) it else f(a)) })
-        }
+        from.foldLeft(None) { acc: Option<A>, a: A ->
+          acc.fold({ Some(a) },
+            { Some(if (ord.run { f(it) < f(a) }) it else a) })
+        }.map(select)
```
pakoito: map identity? doesn't that return the same value?

modules/aql/arrow-aql/src/test/kotlin/arrow/aql/tests/AQLTests.kt
```diff
@@ -137,7 +143,13 @@ class AQLTests : UnitSpec() {
...
       listOf(john, jane, jack, chris).query {
         selectAll().min(Long.order()) { age.toLong() }
-      }.value() shouldBe Some(30L)
+      }.value() shouldBe Some(john)
+    }
+
+    "AQL is able to `min` and return the select type" {
+      listOf(john, jane, jack, chris).query {
+        select { this.age }.min(Int.order()) { age }
+      }.value() shouldBe Some(30)
```
pakoito: Should be User, not age, right?

pakoito: Oh, the select first.

modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Max.kt
maruiz81: yes, it's not used I can remove it

modules/aql/arrow-aql/src/main/kotlin/arrow/aql/Min.kt
maruiz81: I don't follow you. I map it to select to return `Option<Y>` instead of `Option<X>`

modules/aql/arrow-aql/src/test/kotlin/arrow/aql/tests/AQLTests.kt
maruiz81: 👍

PR #1730, opened 2019-10-22
Fix typos in Higher Kind docs

zsmb13:


pakoito:
Waiting for the statuses to pass!

PR #1736, opened 2019-10-24
Typeclasses: small fix and initial test

rachelcarmena:
* Small fix to find package descriptors in bytecode
* Initial test for typeclasses


modules/meta/arrow-meta-prototype/compiler-plugin/src/main/kotlin/arrow/meta/plugins/typeclasses/TypeClassesPlugin.kt
```diff
@@ -299,7 +300,11 @@ private fun PackageFragmentDescriptor.findExtensionProof(extensionType: KotlinTy
     }
 
 private fun ClassifierDescriptor?.packageFragmentDescriptor(): PackageFragmentDescriptor? =
-  this?.findPsi()?.containingFile.safeAs<KtFile>()?.let { file -> this?.module?.findPackageFragmentForFile(file) }
+  if (this?.source is KotlinJvmBinarySourceElement) {
+    this.containingDeclaration as PackageFragmentDescriptor
```
sraaphorst: Why do we cast if we don't use the results of the cast? (Sorry if that's a newb question. I'm a total newb to this project although I am loving reading these PRs and my FP knowledge has increased dramatically over what I was slogging through in Scala. The code and concepts here are so much clearer.)

rachelcarmena: Happy to hear about it, @sraaphorst! I think all the questions are good questions, don't worry. There is an implicit return. If you look at the signature, the body starts after `=`, so the result of the cast is the result of the function.

PR #1744, opened 2019-10-25
Implement suspend catch for ApplicativeError

### LINKED ISSUE #1732: Add suspend version of catch to `ApplicativeError`
```
suspend fun <A> effectCatch(fe: (Throwable) -> E, f: suspend () -> A): Kind<F, A> = 
  try { f() } catch (t: Throwable) { raiseError(fe(t)) }

suspend fun <A> Applicative<F, Throwable>.effectCatch(f: suspend () -> A): Kind<F, A> = 
  try { f() } catch (t: Throwable) { raiseError(t) }
```

lenguyenthanh:
Close #1732

### DIFF
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ApplicativeErrorLaws.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ApplicativeErrorLaws.kt
@@ -13,6 +13,7 @@ import arrow.typeclasses.ApplicativeError
 import arrow.typeclasses.Eq
 import io.kotlintest.properties.Gen
 import io.kotlintest.properties.forAll
+import kotlinx.coroutines.runBlocking
 
 object ApplicativeErrorLaws {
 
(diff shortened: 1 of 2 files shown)


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ApplicativeErrorLaws.kt
```diff
@@ -13,6 +13,7 @@ import arrow.typeclasses.ApplicativeError
 import arrow.typeclasses.Eq
 import io.kotlintest.properties.Gen
 import io.kotlintest.properties.forAll
+import kotlinx.coroutines.runBlocking
```
pakoito: Could you please use `IO.effect { }.unsafeRunSync()` instead? :D Dogfood our own lib hahah

pakoito:
90% there and tests passing :D

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ApplicativeErrorLaws.kt
lenguyenthanh: Fixed that with the new commit. Good to know this tip.

lenguyenthanh:
I definitely need to do that. I'm quite bad at writing though.

pakoito:
Great chance to practice without pressure then ;)

PR #1748, opened 2019-10-25
Fix typo and replace compile with implementation in various md files 

aballano:
Also added `**/.kotlintest/` folder to gitignore to avoid unnecessary tracked files

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,3 +132,4 @@ reports/
 target/
 
 .bash_profile
+**/.kotlintest/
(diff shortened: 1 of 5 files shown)


pakoito:
So what this means is that we can get a reexport for the whole lib just having a single `build.gradle` with all as `api`?

aballano:
> So what this means is that we can get a reexport for the whole lib just having a single `build.gradle` with all as `api`?

AFAIK yes, so if you want A -> C and you have A -> B, B -> C then you only need to declare api of B -> C so A gets it without explicitly declaring it 👍

PR #1750, opened 2019-10-26
Add sequence birecursive instance and fix corecursive tests

1Jajen1:
I forgot this instance, also corecursive equality of unfolds by == instead of using Eq, that only fails with Sequence, but should be fixed by using equalUnderTheLaw anyway. (The same is not necessary for Recursive because we are only comparing ints there)


1Jajen1:
I also moved elgot and coelgot, it was a mistake to have them in recursive/corecursive.

modules/recursion/arrow-recursion-data/src/test/kotlin/arrow/recursion/SequenceKBirecursive.kt
```diff
@@ -0,0 +1,56 @@
+package arrow.recursion
```
pakoito: Add Test to the name of this file, if possible :D

1Jajen1:
> LGTM, some nit on file names

Should be done :+1:

pakoito:
All ABOOOOOARD!

PR #1751, opened 2019-10-26
Check submodules for Android API compatibility

### LINKED ISSUE #1715: Implement Android check to prevent regressions
Arrow 0.10.1 is crashing because Arrow uses AtomicReference, from which it uses some functions that were added to Android in a [latter API version](https://developer.android.com/reference/java/util/concurrent/atomic/AtomicReference#updateAndGet(java.util.function.UnaryOperator%3CV%3E))

_See https://github.com/arrow-kt/arrow/pull/1710_

junron:
Oracle JDK 7 was added to .travis.yml.

This makes Travis test on JDK 7 as well as 8.
I'm not sure if there is anything else that needs to be done to target JDK 7?

Closes #1715


junron:
According to [Travis documentation](https://docs.travis-ci.com/user/reference/trusty/#jvm-clojure-groovy-java-scala-images),

>  Oracle JDK 7 is not provided because it reached End of Life in April 2015.

Maybe we could use OpenJDK 7 instead?

aballano:
> According to [Travis documentation](https://docs.travis-ci.com/user/reference/trusty/#jvm-clojure-groovy-java-scala-images),
> 
> > Oracle JDK 7 is not provided because it reached End of Life in April 2015.
> 
> Maybe we could use OpenJDK 7 instead?

Hi @junron, first of all thanks for the PR! 

I just [updated the ticket](https://github.com/arrow-kt/arrow/issues/1715#issuecomment-546697022) to mention gradle plugin that we could greatly benefit from, the only tricky part is that it needs to be configured in a specific way. Would you like to try that out? Thanks again!

junron:
I'll look at it tomorrow and get back to you on progress. Thanks for the advice!

.travis.yml
```diff
@@ -4,6 +4,7 @@ dist: trusty
 
 jdk:
 - oraclejdk8
+- oraclejdk7
```
rachelcarmena: Thanks @junron! This PR will be very useful! Please, it's not necessary to change this file, because Travis CI is only used for creating releases now.

junron:
There seems to be an error:
`Could not generate a proxy class for class ru.vyarus.gradle.plugin.animalsniffer.AnimalSniffer.` when running `gradlew check`.
I'm not sure what's causing it.

aballano:
@junron I believe it's due to the project not using Gradle 5, while that comes you can use plugin version 1.4.6

aballano:
@junron Maybe we need to add the `gradlew check` command to bitrise or github actions so the CI checks for that cc @rachelcarmena

junron:
Oh right, I'll add it soon

build.gradle
```diff
@@ -60,8 +62,12 @@ buildscript {
         classpath "gradle.plugin.io.morethan.jmhreport:gradle-jmh-report:0.9.0"
         classpath "org.jlleitschuh.gradle:ktlint-gradle:$ktlint_gradle_version"
         classpath "org.jetbrains.kotlinx:atomicfu-gradle-plugin:$atomicfuVersion"
+        classpath 'ru.vyarus:gradle-animalsniffer-plugin:1.4.6'
     }
 }
+dependencies {
```
aballano: i'm not sure that this is being applied to all submodules, i think you need to move it below inside `allProjects` together with `apply plugin: 'ru.vyarus.animalsniffer'`

build.gradle
```diff
@@ -16,6 +16,8 @@
 
 import com.github.benmanes.gradle.versions.reporter.PlainTextReporter
 
+apply plugin: 'java'
```
aballano: I think this shouldn't be needed

junron: When removed gradle build errors with `Could not find method signature()`. According to documentation `plugin must be declared after java or groovy plugin, otherwise nothing will be registered.`, so I have decided to leave it in.

build.gradle
junron: Ok, moved

junron:
Currently the following subrojects are failing the test:

- arrow-kio-benchmarks
- arrow-meta
- arrow-ank-gradle

I believe I should whitelist them?

bitrise.yml
```diff
@@ -26,6 +26,13 @@ workflows:
             #!/usr/bin/env bash
             set -ex
             ./gradlew :arrow-docs:runAnk
+    - script:
+        title: check
+        inputs:
+          - content: |-
+              #!/usr/bin/env bash
+              set -ex
+              ./gradlew check
```
aballano: I'd suggest adding either bitrise or github actions, but not both. @rachelcarmena Do you know which one should we use? I'd personally go for github actions as it seems it runs faster :)

build.gradle
```diff
@@ -82,6 +82,22 @@ allprojects {
     }
 }
 
+configure(allprojects
+        - project("arrow-meta")
+        - project("arrow-ank-gradle")
+        - project("arrow-ank")
+        - project("arrow-kio-benchmarks")
+        - project("arrow-fx")
+        - project("arrow-fx-reactor")
```
aballano: I believe this both fx projects shouldn't be whitelisted as people shall use them. If the check fails here we could fix them in a separate PR if necessary.

bitrise.yml
junron: Ok, I will revert the bitrise config file

rachelcarmena: Thanks for the mention @aballano. It would be just necessary to update GitHub Actions. because we have the build check there:
- Travis CI is only used to publish snapshots/releases
- Bitrise is only used to generate the documentation

Thanks!

aballano:
> Currently the following subprojects are failing the test:
> 
> * arrow-kio-benchmarks
> * arrow-meta
> * arrow-ank-gradle
> 
> Additional subprojects excluded after further testing:
> 
> * arrow-ank
> * arrow-fx-reactor
> * arrow-fx
> * arrow-integrations-retrofit-adapter
> 
> I believe I should whitelist them?

Sorry didn't see this comment, ank shouldn't be necessary, but the other 3: fx, fx-reactor and retrofit integration I guess should be excluded, yes

junron:
The builds are failing after removing arrow-ank and others from the excluded list. Maybe I got the signature version wrong?

bitrise.yml
junron: Reverted

aballano:
@junron Sorry I expressed myself incorrectly, I just updated my comment above to try to be more clear, essentially the modules that shouldn't be used on Android can be explicitly excluded, the rest should be checked. If you have extra questions feel free to ping me in Slack as well!

aballano:
@junron I think this is working now! 🎉 

This means that this pieces won't work on Androi API 14

[Undefined reference] arrow.fx.(IODispatchers.kt:15)
  >> void java.util.concurrent.ForkJoinPool.<init>()

[Undefined reference] arrow.fx.internal.(Utils.kt:182)
  >> void Throwable.addSuppressed(Throwable)

Now the question is if we want to support this API level and fix those or directly jump to API 21.  @JorgeCastilloPrz @nomisRev You can check detailed versions and the cumulative distribution below 👇 

<details>

image
</details>

nomisRev:
@aballano so we currently support API 15+? Or did I miss understand?

That should be sufficient, no? Since Google recommends min sdk 21. I think we need to support lower than 21, or at least do a Twitter poll of current users?

build.gradle
```diff
@@ -82,6 +82,19 @@ allprojects {
...
+configure(allprojects
+        - project("arrow-kio-benchmarks")
+        - project("arrow-meta")
+        - project("arrow-ank-gradle")
+        - project("arrow-ank")
+) {
+    apply plugin: 'java'
+    apply plugin: 'ru.vyarus.animalsniffer'
+    dependencies {
+        signature 'net.sf.androidscents.signature:android-api-level-14:4.0_r4@signature'
```
aballano:
```suggestion
        signature 'net.sf.androidscents.signature:android-api-level-21:5.0.1_r2@signature'
```

@junron Would you mind trying API 21 and see if everything is ok?

aballano:
Build is failing due to arrow-fx-reactor, after reading their Readme it seems they don't officially support Android, although it should work on API 26. So I guess we could exclude this one as well @junron, sorry for the back and forth 🙏

junron:
I'll do some more local testing to confirm, before pushing my commits

junron:
|Java API|Used in|Android API level|
|:-----:|:-----:|:-----:|
|[`getTypeName` on `java.lang.reflect.Type`](https://developer.android.com/reference/java/lang/reflect/Type#getTypeName())|`arrow-integrations-retrofit-adapter` in `CallKindAdapterFactory.kt`|28

Looks like `arrow-integrations-retrofit-adapter` may not work either

aballano:
@junron Feel free to exclude it as well and I'll take care of checking that with the other maintainers, at least having the check in place will be very useful 👍

pakoito:
> Execution failed for task ':arrow-integrations-retrofit-adapter:animalsnifferMain'.
> 1 AnimalSniffer violations were found in 1 files. See the report at: file:////home/runner/work/arrow/arrow/modules/integrations/arrow-integrations-retrofit-adapter/build/reports/animalsniffer/main.text

This is not very useful without output in CI

aballano:
> > Execution failed for task ':arrow-integrations-retrofit-adapter:animalsnifferMain'.
> > 1 AnimalSniffer violations were found in 1 files. See the report at: file:////home/runner/work/arrow/arrow/modules/integrations/arrow-integrations-retrofit-adapter/build/reports/animalsniffer/main.text
> 
> This is not very useful without output in CI

Agree, I'd say we merge and iterate later on and improve it 💪

aballano:
Thank you @junron!

PR #1753, opened 2019-10-28
Slight tweaks to atomics

aballano:
* Removed Atomic toString functions crashing on Android API 21 [as discussed here](https://github.com/arrow-kt/arrow/issues/1715#issuecomment-546730951)
* Added couple of utility functions for AtomicInt

### DIFF
--- a/modules/core/arrow-core/src/main/kotlin/arrow/core/internal/atomic.kt
+++ b/modules/core/arrow-core/src/main/kotlin/arrow/core/internal/atomic.kt
@@ -22,6 +22,8 @@ class AtomicRefW<A>(a: A) {
   fun compareAndSet(expect: A, update: A) = atomicRef.compareAndSet(expect, update)
 
   fun lazySet(a: A) = atomicRef.lazySet(a)
+
+  override fun toString(): String = value.toString()
 }
 
 class AtomicBooleanW(a: Boolean) {
@@ -40,6 +42,8 @@ class AtomicBooleanW(a: Boolean) {
   fun compareAndSet(expect: Boolean, update: Boolean) = atomicRef.compareAndSet(expect, update)
 
   fun lazySet(a: Boolean) = atomicRef.lazySet(a)
+
+  override fun toString(): String = value.toString()
 }
 
 class AtomicIntW(a: Int) {
@@ -70,4 +74,6 @@ class AtomicIntW(a: Int) {
   fun compareAndSet(expect: Int, update: Int) = atomicRef.compareAndSet(expect, update)
 
   fun lazySet(a: Int) = atomicRef.lazySet(a)
+
+  override fun toString(): String = value.toString()
 }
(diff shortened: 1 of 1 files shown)


modules/core/arrow-core/src/main/kotlin/arrow/core/internal/atomic.kt
```diff
@@ -22,8 +22,6 @@ class AtomicRefW<A>(a: A) {
   fun compareAndSet(expect: A, update: A) = atomicRef.compareAndSet(expect, update)
 
   fun lazySet(a: A) = atomicRef.lazySet(a)
-
-  override fun toString(): String = atomicRef.toString()
```
JorgeCastilloPrz: Why is this one not working for JDK 7? could we print `atomicRef.value` instead of removing?

modules/core/arrow-core/src/main/kotlin/arrow/core/internal/atomic.kt
```diff
@@ -42,8 +40,6 @@ class AtomicBooleanW(a: Boolean) {
   fun compareAndSet(expect: Boolean, update: Boolean) = atomicRef.compareAndSet(expect, update)
 
   fun lazySet(a: Boolean) = atomicRef.lazySet(a)
-
-  override fun toString(): String = atomicRef.toString()
```
JorgeCastilloPrz: Same question here

modules/core/arrow-core/src/main/kotlin/arrow/core/internal/atomic.kt
```diff
@@ -57,11 +53,21 @@ class AtomicIntW(a: Int) {
...
+
+  fun decrementAndGet() = atomicRef.decrementAndGet()
+
   fun updateAndGet(function: (Int) -> Int) = atomicRef.updateAndGet(function)
 
   fun compareAndSet(expect: Int, update: Int) = atomicRef.compareAndSet(expect, update)
 
   fun lazySet(a: Int) = atomicRef.lazySet(a)
-
-  override fun toString(): String = atomicRef.toString()
```
JorgeCastilloPrz: And here

modules/core/arrow-core/src/main/kotlin/arrow/core/internal/atomic.kt
aballano: It seems that they may been doing something behind the scenes, because I checked the source code and `toString` should delegate on the actual value, but the stack trace says differently:

> java.lang.NoSuchMethodError: No virtual method toString(Ljava/lang/Object;)Ljava/lang/String; in class Ljava/util/concurrent/atomic/AtomicIntegerFieldUpdater; or its super classes (declaration of 'java.util.concurrent.atomic.AtomicIntegerFieldUpdater' appears in /system/framework/core-libart.jar)
    

Using `value.toString` seems to work 👍

Issue #1742, opened 2019-10-25
Move arrow-meta-prototype directory to a new repository


rachelcarmena:
Hi **Λrrow Meta contributors**,

`modules/meta/arrow-meta-prototype/` directory is moving to a new repository temporarily. 

Please, follow these guidelines in case of having a WIP development related to that directory.

## Working in a forked repository

0. Create a new fork from [arrow-meta](https://github.com/arrow-kt/arrow-meta)
1. Reduce your local workspace to `modules/meta/arrow-meta-prototype/` directory:
```
git filter-branch --subdirectory-filter modules/meta/arrow-meta-prototype/ -- --all
```
2. Change the origin repository:
```
git config remote.origin.url git@github.com:<your-username>/arrow-meta.git
```
3. Push the branch to the repository to prepare a pull request:
```
git push origin <your-branch>
```

## Working in **arrow** repository

1. Reduce your local workspace to `modules/meta/arrow-meta-prototype/` directory:
```
git filter-branch --subdirectory-filter modules/meta/arrow-meta-prototype/ -- --all
```
2. Change the origin repository:
```
git config remote.origin.url git@github.com:arrow-kt/arrow-meta.git
```
3. Push the branch to the repository to prepare a pull request:
```
git push origin <your-branch>
```

Please, let us know if we can help you.

rachelcarmena:
* Remove arrow-meta-prototype directory in **arrow** repository: #1745 :heavy_check_mark: 
* Move issues about arrow-meta to the new repository :heavy_check_mark:

PR #1760, opened 2019-10-30
arrow-annotations: migrate content from Arrow Meta

rachelcarmena:
Another step for #1742.


rachelcarmena:
> I'll let raul review this one

Thanks @pakoito, I talked to @raulraja about it. 
By the way, I made an automatic task to get the content of this pull request:

```
git diff HEAD rr-meta-prototype-integration -- ./modules/meta/arrow-annotations | git apply
```

I love the power of `git apply` after `git diff` :sweat_smile:

modules/meta/arrow-annotations/src/main/java/arrow/extension.kt
```diff
@@ -4,3 +4,9 @@ package arrow
 @Target(AnnotationTarget.CLASS)
 @MustBeDocumented
 annotation class extension
+
+val given: Nothing
+  get() = TODO("Should have been replaced by Arrow Meta Compiler Plugin provided by [apply plugin: 'arrow.meta.plugin.gradle']")
```
raulraja: Now that we know how a user should include the plugin we should replace this TODO message for a better one that tells them in which place they need to add the plugin

modules/meta/arrow-annotations/src/main/java/arrow/extension.kt
```diff
@@ -4,3 +4,9 @@ package arrow
...
+
+fun <A> given(): A =
+  TODO("Should have been replaced by Arrow Meta Compiler Plugin provided by [apply plugin: 'arrow.meta.plugin.gradle']")
```
raulraja: Same as above

modules/meta/arrow-annotations/src/main/java/arrow/extension.kt
raulraja: Does not need to happen in this PR but this is what the user will see if the plugin is not activated and they attempt to run some code that depends on `given`

raulraja:
@rachelcarmena we are missing the changes for `@extension` and that is most likely what Amanda is facing and using that over a `fun`

PR #1763, opened 2019-10-30
Removed unused Readme files

maruiz81:


pakoito:
Unused here means we now point the inline docs directly to these pages? Some are from the most visited, so we have to make sure to get this one right hahaha

calvellido:
Yes, those entries have been replaced by the ones generated through `dokka`, these are the ones where the menu entries point now:

https://next.arrow-kt.io/docs/apidocs/arrow-core-data/arrow.core/-option/

Eventually, we can think of redirecting the previous entries (https://arrow-kt.io/docs/arrow/core/option/) to this new URLs. This can be done by specifying the appropriate Front Matter metadata in the markdown file generated by `dokka`, e.g.:

```
---
layout: docs
title: Option
permalink: /docs/apidocs/arrow-core-data/arrow.core/-option/
redirect_from:
  - /docs/datatypes/option/
  - /docs/arrow/core/option/
video: q6HpChSq-xc
---
```

https://github.com/jekyll/jekyll-redirect-from#usage

i-walker:
@maruiz81 can we add @calvellido's suggestion in this PR?

pakoito:
We should setup the redirect in this diff, otherwise the page will disappear from the convenient menu to search or somewhere here: 
Screen Shot 2019-10-31 at 17 22 41

maruiz81:
Ok, let me add the @calvellido suggestion.

calvellido:
So, it seems like currently `dokka` doesn't support Front Matter properties, and any of those being included won't be treated as such. The only Front Matter that it gets injected into the generated file is the `title` one:

https://github.com/Kotlin/dokka/blob/de2f32d91fb6f564826ddd7644940452356e2080/core/src/main/kotlin/Formats/JekyllFormatService.kt#L22

Unless we use a custom `dokka`, or propose some changes adding a new feature to `dokka`, right now it won't be possible to redirect previous entries this way.

The only feasible option for now would be to use the `redirect_to` Front Matter option, and that would involve maintaining previous files (removing its content and keeping only redirect info).

maruiz81:
> So, it seems like currently `dokka` doesn't support Front Matter properties, and any of those being included won't be treated as such. The only Front Matter that it gets injected into the generated file is the `title` one:
> 
> https://github.com/Kotlin/dokka/blob/de2f32d91fb6f564826ddd7644940452356e2080/core/src/main/kotlin/Formats/JekyllFormatService.kt#L22
> 
> Unless we use a custom `dokka`, or propose some changes adding a new feature to `dokka`, right now it won't be possible to redirect previous entries this way.
> 
> The only feasible option for now would be to use the `redirect_to` Front Matter option, and that would involve maintaining previous files (removing its content and keeping only redirect info).
> 
> In case you see fit, I've already submitted the changes to your branch @maruiz81.

Ok @calvellido, it fixes the problem for the time being. Thanks!

pakoito:
You can safely ignore the bild failure, it was a flaky test.

PR #1767, opened 2019-11-05
Change generic parameter ordering in mtl classes

1Jajen1:
> Ctr-R many many times...

Right now it is impossible to define abstractions over mtl types that abstract both the monad and the return type leaving only additional type information. For example chaning the base monad (which can be quite useful). Example from: http://hackage.haskell.org/package/mmorph-1.1.3/docs/Control-Monad-Morph.html
```kotlin
interface MFunctor<T> {
  fun <M, N, A> Kind<Kind<T, M>, A>hoist(f: FunctionK<M, N>): Kind<Kind<T, N>, A>
}
```
It is simply impossible to define this for something like `EitherT` or `WriterT` atm because you cannot write `interface EitherTMFunctor<E> : MFunctor<Kind<ForEitherT, E>>`.

So what I did was basically to change the order to have the extra generics that are not the monad or the return type infront so that it is now possible to write the above instance.


modules/mtl/arrow-mtl/src/main/kotlin/arrow/mtl/extensions/kleisli.kt
```diff
@@ -1,173 +1,174 @@
...
-  override fun <A, B> KleisliOf<F, D, A>.map(f: (A) -> B): Kleisli<F, D, B> =
+  override fun <A, B> KleisliOf<D, F, A>.map(f: (A) -> B): Kleisli<D, F, B> =
     fix().map(FF(), f)
 }
 
+/* TODO change conested to represent λ[α => Kind[α, F, C]] instead of λ[α => Kind[F, α, C]]
 @extension
-interface KleisliContravariant<F, D> : Contravariant<Conested<Kind<ForKleisli, F>, D>> {
-  override fun <A, B> Kind<Conested<Kind<ForKleisli, F>, D>, A>.contramap(f: (B) -> A): Kind<Conested<Kind<ForKleisli, F>, D>, B> =
+interface KleisliContravariant<D, F> : Contravariant<Conested<Kind<ForKleisli, D>, F>> {
```
1Jajen1: I am not too familiar with conested. Can someone who is take a quick look on what needs to be done to make this work again? Or is it fine just leaving this?

1Jajen1:
Oh and I also forgot to check docs for this. Will do so later today or tomorrow.

1Jajen1:
Docs are now converted to this pattern as well. So this should be done.

pakoito:
This one. Poor name: https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Cocomposed.kt

1Jajen1:
That feels wrong. Like a hack for something that should not be a problem at all. I don't think I even understand how to use it :/ 
Can you actually implement the interface MFunctor that I listed above with Conested?

1Jajen1:
Thing is `MFunctor` does not simply skip a generic parameter. It omits it completely but in the case of `EitherT` it still needs the left part of the generic parameters to be defined. This is actually how mtl/transformers in haskell define the generic parameter order. I don't know if that is also the reason tho

pakoito:
If you could fix the conflix I promise I'll give it a look.

1Jajen1:
Will do in a sec.
Btw I just found this: https://github.com/scalaz/scalaz/issues/1305

1Jajen1:
@pakoito I don't think `Conested` (in it's current form) can do whats needed. Partial type application of a type that is not the first type. You may be able to cheese that in by partially applying a type to Conested (that requires a few changes to it, if its even possible) but that is really not a great way. I think it is as the issue for scalaz states: The haskell ordering seems to be the more common usage.

pakoito:
I couldn't find the time to do it. I'll remove the blocker, see if @raulraja can review it maybe

pakoito:
There's no remove back to neutral _sigh_

pakoito:
second attempt

pakoito:
that worked

1Jajen1:
I'll close this and do this one by one in another pr (or multiple) in the future

PR #1768, opened 2019-11-05
Add replicate to applicative

1Jajen1:
Just a generally very useful method on applicatives

### DIFF
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Applicative.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Applicative.kt
@@ -22,4 +22,9 @@ interface Applicative<F> : Apply<F> {
   fun <A> Kind<F, A>.replicate(n: Int): Kind<F, List<A>> =
     if (n <= 0) just(emptyList())
     else map(this, replicate(n - 1)) { (a, xs) -> listOf(a) + xs }
+
+  fun <A> Kind<F, A>.replicate(n: Int, MA: Monoid<A>): Kind<F, A> = MA.run {
+    if (n <= 0) just(empty())
+    else map(this@replicate, replicate(n, MA)) { (a, xs) -> a + xs }
+  }
 }
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Applicative.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Applicative.kt
@@ -18,9 +18,8 @@ interface Applicative<F> : Apply<F> {
 
   override fun <A, B> Kind<F, A>.map(f: (A) -> B): Kind<F, B> =
     ap(just(f))
-  
+
   fun <A> Kind<F, A>.replicate(n: Int): Kind<F, List<A>> =
     if (n <= 0) just(emptyList())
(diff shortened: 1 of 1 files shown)


modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Applicative.kt
```diff
@@ -18,4 +18,9 @@ interface Applicative<F> : Apply<F> {
 
   override fun <A, B> Kind<F, A>.map(f: (A) -> B): Kind<F, B> =
     ap(just(f))
+  
+  fun <A> Kind<F, A>.replicate(n: Int): Kind<F, List<A>> =
```
pakoito: Could it be done for `Monoid`.

1Jajen1: You mean to combine to any monoid? Sure, do you have any use case in mind? I'd keep a list version as default tho

1Jajen1: I added an overload that works on any monoid, but kept the list one as I think it is the most useful one. I would have defined it in terms of the more general monoid one, but the List monoid is not available in that module

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Applicative.kt
```diff
@@ -18,4 +18,13 @@ interface Applicative<F> : Apply<F> {
...
   override fun <A, B> Kind<F, A>.map(f: (A) -> B): Kind<F, B> =
     ap(just(f))
+
+  fun <A> Kind<F, A>.replicate(n: Int): Kind<F, List<A>> =
+    if (n <= 0) just(emptyList())
+    else map(this, replicate(n - 1)) { (a, xs) -> listOf(a) + xs }
+
+  fun <A> Kind<F, A>.replicate(n: Int, MA: Monoid<A>): Kind<F, A> = MA.run {
+    if (n <= 0) just(empty())
+    else map(this@replicate, replicate(n - 1, MA)) { (a, xs) -> a + xs }
```
pakoito: Small nit, use `MA.run` just locally in the block inside map. Use `MA.just`. This is because the Android compiler has problem with nested `this` scope, so we play it safe.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Applicative.kt
pakoito: Right, there's a circular dependency problem hahah

Thanks for the change!!

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Applicative.kt
1Jajen1: Done

pakoito: Merge at will! You have rights, IIRC :D

1Jajen1: Will do ^^ Thanks

PR #1769, opened 2019-11-05
Add alternative instances to datatypes

1Jajen1:
Most of these were missing alternative instances, and those that had one via MonadCombine had either no tests or no explicit alternative instance.

I also made combineK default to orElse in alternatives since that is exactly the monoidal behaviour it is supposed to represent.

Btw is there any reason for Alternative laws to be that strict? Looks like a mix of MonadPlus and Alternative laws to me.

### DIFF
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Alternative.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Alternative.kt
@@ -46,4 +46,6 @@ interface Alternative<F> : Applicative<F>, MonoidK<F> {
    * @returns a combination of both computations.
    */
   fun <A> Kind<F, A>.orElse(b: Kind<F, A>): Kind<F, A>
+
+  override fun <A> Kind<F, A>.combineK(y: Kind<F, A>): Kind<F, A> = orElse(y)
 }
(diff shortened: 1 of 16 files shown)


modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Alternative.kt
```diff
@@ -46,4 +46,6 @@ interface Alternative<F> : Applicative<F>, MonoidK<F> {
    * @returns a combination of both computations.
    */
   fun <A> Kind<F, A>.orElse(b: Kind<F, A>): Kind<F, A>
+
+  override fun <A> Kind<F, A>.combineK(y: Kind<F, A>): Kind<F, A> = orElse(y)
```
pakoito: wat


that simple?

1Jajen1: Well yes. orElse and empty form a monoid, that is the very reason it is a monoidk in the first place. Alternative laws are usually just testing if orElse is associative and if empty is the identity. We have a lot more laws on it's interactions with applicatives, but in essence alternative is just a monoid with orElse and empty (over kinds that's why it's MonoidK).

Issue #1777, opened 2019-11-07
Extfun to go from Either to IO


pakoito:
In IO.kt or the io instances file: https://github.com/arrow-kt/arrow/blob/master/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/io.kt

```
fun <E, A> Either<E, A>.toIO(f: (E) -> Throwable): IO<A> = 
  fold({ IO.raiseError(f(it)) }, IO::just)

fun <A> Either<Throwable, A>.toIO(): IO<A> = 
  toIO(::identity)
```

This function helps people with mixed stacks of `suspend () -> Either` and `IO`

abendt:
i'd like to try that. will create a PR

pakoito:
Done and completed! Thank you @abendt !

abendt:
that was quick :-) happy to help

PR #1778, opened 2019-11-07
#1777: add extension function that converts an Either to an IO

abendt:
this PR adds two extension functions to convert an Either to an IO

https://github.com/arrow-kt/arrow/issues/1777


pakoito:
All great, with tests and all. Very thankful that you took this, I bet a lot of people are going to love it <3

PR #1780, opened 2019-11-09
Fix the return type of flatMap

tatsuyafujisaki:
Fix the return type of flatMap


nomisRev:
Thanks for the fix @tatsuyafujisaki! 👍

PR #1781, opened 2019-11-09
Implement Endo Monoid

kutyel:
Since *promises are made to be kept*, here's my PR!

This idea came from a conversation on [Twitter](https://twitter.com/raulraja/status/1176185281470500865) between @raulraja and me 😄 

I think it would also be valuable for Arrow users to have some common predefined monoids, like `All`, `Any` (for booleans) or `Sum` and `Product` (for numbers). You know, like in [Haskell](https://hackage.haskell.org/package/base-4.12.0.0/docs/src/Data.Semigroup.Internal.html#Endo)! 😉 

Also, kudos and thank you very much to @i-walker for helping me so much with the contribution, you rock man! 🎸


kutyel:
I thought you called dibs on `Any` @i-walker 😂 In any case, I'd also like to know @pakoito's opinion to see whether this is useful for the Arrow users or not 😉

modules/core/arrow-core-data/src/main/kotlin/arrow/core/Endo.kt
```diff
@@ -0,0 +1,8 @@
+package arrow.core
+
+/**
+ * The monoid of endomorphisms under composition.
+ */
+data class Endo<A>(val f: (A) -> A) {
```
raulraja: Unless there is an instance of Eq for this it would be more efficient represented as just a class not a data class

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/EndoMonoid.kt
```diff
@@ -0,0 +1,13 @@
...
+
+import arrow.core.Endo
+import arrow.core.compose
+import arrow.core.identity
+import arrow.extension
+import arrow.typeclasses.Monoid
+
+@extension
+interface EndoMonoid<A> : Monoid<Endo<A>> {
+  override fun empty(): Endo<A> = Endo(::identity)
```
raulraja: We should be able to serve a singleton instance of this value to avoid continuous allocations

modules/core/arrow-test/src/main/kotlin/arrow/test/generators/Generators.kt
```diff
@@ -107,6 +108,7 @@ fun Gen.Companion.intPredicate(): Gen<(Int) -> Boolean> =
     )
   }
 
+fun <A> Gen.Companion.endo(gen: Gen<A>): Gen<Endo<A>> = gen.map { a: A -> Endo<A> { a } }
```
raulraja: Kotlin test has an arrow module. We should add all these gen instances there if they make sense

modules/core/arrow-core-data/src/main/kotlin/arrow/core/Endo.kt
raulraja: Can we provide some useful use cases that de demonstrates the use of this new data type ?

sraaphorst:
Then we can compose endos as they should be able to? (Sorry, this is not canonical code... I was just playing around with R3 projections.)

```
typealias R3 = Triple<Double, Double, Double>

val vector: Triple<Double, Double, Double> = Triple(1.0, 2.0, 3.0)

data class Endo<A>(val f: (A) -> A) {
    fun apply(a: A): A = f(a)
    fun compose(other: Endo<A>): Endo<A> = Endo{apply(other.apply(it))}
}

val projection_x: Endo<R3> = Endo {R3(0.0, it.second, it.third)}
val projection_y: Endo<R3> = Endo {R3(it.first, 0.0, it.third)}
val projection_z: Endo<R3> = Endo {R3(it.first, it.second, 0.0)}
val projection_xy = projection_x.compose(projection_y)
val projection_xz = projection_x.compose(projection_z)
val projection_yz = projection_y.compose(projection_z)
```

sraaphorst:
(I'm still a total Kotlin noob if it wasn't obvious, but I'm loving Kotlin and I'm loving this project. We use Scala at work predominantly and I can say that I've learned a lot more about FP from Arrow than I have from Cats / scalaz. I'd love to see a book like Scala with Cats for Arrow. Things just seem much more clear in Kotlin.)

pakoito:
>Shouldn't an `Endo` be a subclass of `Monoid`?

The beauty of typeclasses is that it doesn't rely on inheritance to implement functionality. It is defined outside of the class instead.

sraaphorst:
Ahhh... thanks, @pakoito. I still have a lot to learn 😸

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/EndoMonoid.kt
sraaphorst: Oh, I see here now how it works. Thanks to everyone for your patience with me.

modules/core/arrow-core-data/src/main/kotlin/arrow/core/Endo.kt
kutyel: Maybe the easiest use case is function composition, although there is even a [package](http://hackage.haskell.org/package/endo-0.3.0.1) to showcase some useful examples, maybe we can port some of the as docs for Arrow! 😉

pakoito:
PUSH AND DELETTTT

Issue #1776, opened 2019-11-07
Random errors with multi-thread tests


rachelcarmena:
```
arrow.fx.SingleKTests > Multi-thread Singles finish correctly FAILED
    java.lang.AssertionError: Subscriber still running! (latch = 1, values = 0, errors = 0, completions = 0, timeout!, disposed!)
        at io.reactivex.observers.BaseTestConsumer.fail(BaseTestConsumer.java:189)
        at io.reactivex.observers.BaseTestConsumer.assertTerminated(BaseTestConsumer.java:665)
        at arrow.fx.SingleKTests$2.invokeSuspend(SingleKTests.kt:76)
        at arrow.fx.SingleKTests$2.invoke(SingleKTests.kt)
```

Thanks to @ffgiraldez for the info :+1:

PR #1782, opened 2019-11-09
RxJava ConcurrentSyntax & Dispatchers

nomisRev:
I was investigating #1776  but could not reproduce.

This PR enables `ConcurrentSyntax` syntax for `fx` instead of `AsyncSyntax`.
To do so it adds `Dispatchers` instances for all RxJava types and exposes a `(Scheduler) -> CoroutineContext` function.

### DIFF
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/generators/Generators.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/generators/Generators.kt
@@ -145,3 +145,6 @@ fun Gen.Companion.char(): Gen<Char> =
   Gen.from(('A'..'Z') + ('a'..'z') + ('0'..'9') + "!@#$%%^&*()_-~`,<.?/:;}{][±§".toList())
 
 fun <A> Gen.Companion.genSetK(genA: Gen<A>): Gen<SetK<A>> = Gen.set(genA).map { it.k() }
+
+fun Gen.Companion.unit(): Gen<Unit> =
+  create { Unit }
--- a/modules/fx/arrow-fx-rx2/src/test/kotlin/arrow/fx/SingleKTests.kt
+++ b/modules/fx/arrow-fx-rx2/src/test/kotlin/arrow/fx/SingleKTests.kt
@@ -1,6 +1,5 @@
 package arrow.fx
 
-import arrow.core.Either
 import arrow.fx.rx2.ForSingleK
 import arrow.fx.rx2.SingleK
 import arrow.fx.rx2.SingleKOf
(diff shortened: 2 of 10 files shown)


modules/fx/arrow-fx-rx2/src/test/kotlin/arrow/fx/SingleKTests.kt
```diff
@@ -66,50 +69,61 @@ class SingleKTests : RxJavaSpec() {
...
     "Multi-thread Singles finish correctly" {
-      val value: Single<Long> = SingleK.fx {
-        val a = Single.timer(2, TimeUnit.SECONDS).k().bind()
-        a
-      }.value()
-
-      val test: TestObserver<Long> = value.test()
-      test.awaitDone(5, TimeUnit.SECONDS)
-      test.assertTerminated().assertComplete().assertNoErrors().assertValue(0)
+      forFew(10, Gen.choose(10L, 50)) { delay ->
```
nomisRev: Could not reproduce even when increased to 1000 tries.

modules/fx/arrow-fx-rx2/src/test/kotlin/arrow/fx/SingleKTests.kt
```diff
@@ -66,50 +69,61 @@ class SingleKTests : RxJavaSpec() {
...
+        SingleK.fx {
+          val a = Single.timer(delay, TimeUnit.MILLISECONDS).k().bind()
```
aballano: Subtle but very nice change, I tested locally and got around 16% lower compile time (from ~4:20 to ~3:40) 🎉

modules/fx/arrow-fx-rx2/src/test/kotlin/arrow/fx/SingleKTests.kt
aballano: I think we could close #1776 and reopen if we see this happening again after your changes 👍

nomisRev:
Closes #1776

modules/core/arrow-test/src/main/kotlin/arrow/test/generators/Generators.kt
```diff
@@ -145,3 +145,6 @@ fun Gen.Companion.char(): Gen<Char> =
   Gen.from(('A'..'Z') + ('a'..'z') + ('0'..'9') + "!@#$%%^&*()_-~`,<.?/:;}{][±§".toList())
 
 fun <A> Gen.Companion.genSetK(genA: Gen<A>): Gen<SetK<A>> = Gen.set(genA).map { it.k() }
+
+fun Gen.Companion.unit(): Gen<Unit> =
```
raulraja: We should consider porting these to kotlintest arrrow module

PR #1784, opened 2019-11-10
Add unsafeCancellableRun

### LINKED ISSUE #1735: Add runAsyncCancellable to UnsafeRun
https://github.com/arrow-kt/arrow/blob/master/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/UnsafeRun.kt

Right now it's not possible to do

```
unsafe {
  val cancel = runAsyncCancellable {
  }
}
```

even if 

`io.unsafeRunAsyncCancellable`

is possible

pakoito:
Closes #1735


pakoito:
>Should we also provide instances for Single & Completable?

We need to implement the logic for `OnCancel`, so I've only added the non-cancellation ones.

PR #1787, opened 2019-11-11
Add io pool to dispatchers

nomisRev:


modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IODispatchers.kt
```diff
@@ -13,6 +15,14 @@ import kotlin.coroutines.CoroutineContext
 object IODispatchers {
   // FIXME use ForkJoinPool.commonPool() in Java 8
   val CommonPool: CoroutineContext = ForkJoinPool().asCoroutineContext()
+
+  private val ioCtr = AtomicInteger(0)
```
aballano: I would use `AtomicIntW` if available here just in case we decide to use other functions later on

PR #1791, opened 2019-11-12
Move `Const.kt` to `core` package (#1789)

### LINKED ISSUE #1789: Move Const back with the other datatypes
Const was historically part of the typeclasses module due to some circular dependencies. Now that we've merged data and typeclasses together in core, it can be moved back with the others :D

https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Const.kt

move to

 modules/core/arrow-core-data/src/main/kotlin/arrow/core/

and make sure that all the references to it in the docs reflect the change

rcd27:
Closes #1789


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/TraverseLaws.kt
```diff
@@ -1,29 +1,28 @@
...
 import arrow.core.extensions.const.applicative.applicative
 import arrow.core.extensions.id.applicative.applicative
 import arrow.core.extensions.id.comonad.extract
+import arrow.core.extensions.monoid
 import arrow.core.fix
+import arrow.core.toT
 import arrow.core.value
+import arrow.mtl.typeclasses.ComposedApplicative
+import arrow.mtl.typeclasses.nest
+import arrow.mtl.typeclasses.unnest
```
pakoito: Sorted imports <3

PR #1797, opened 2019-11-14
Arrow-Core language edits

jkmcclellan:
Edits on all the available arrow-core docs.


arrow-site/doc/docs/arrow-core/datatypes/either/README.md
```diff
@@ -34,9 +34,9 @@ val magic = throwsSomeStuff.andThen(throwsOtherThings).andThen(moreThrowing)
 magic
 ```
 
-Assume we happily throw exceptions in our code. Looking at the types of the above functions, any of them could throw any number of exceptions -- we do not know. When we compose, exceptions from any of the constituent
+Assume we happily throw exceptions in our code. Looking at the types of functions above, any could throw a number of exceptions -- we do not know. When we compose, exceptions from any of the constituent
 functions can be thrown. Moreover, they may throw the same kind of exception
-(e.g. `IllegalArgumentException`) and thus it gets tricky tracking exactly where an exception came from.
+(e.g., `IllegalArgumentException`) and, thus, it gets tricky tracking exactly where an exception came from.
```
sraaphorst: This is awesome stuff... I was noticing the lack of commas in the docs while browsing them, so I'm glad someone took the time to rectify this. I'd question the grammar here, though. I think it should read:

Moreover, they may throw the same kind of exception (e.g., `IllegalArgumentException`), and thus, it gets tricky tracking exactly where an exception came from.

arrow-site/doc/docs/arrow-core/datatypes/either/README.md
```diff
@@ -141,12 +141,12 @@ magic("1")
 magic("Not a number")
 ```
 
-In the following exercise we pattern-match on every case the `Either` returned by `magic` can be in.
-Note the `when` clause in the `Left` - the compiler will complain if we leave that out because it knows that
+In the following exercise, we pattern-match on every case in which the `Either` returned by `magic` can be.
+Note the `when` clause in the `Left` - the compiler will complain if we leave that out because it knows that,
```
sraaphorst: The dash here should be a colon.

arrow-site/doc/docs/arrow-core/datatypes/either/README.md
```diff
@@ -187,9 +187,9 @@ fun magic(s: String): Either<Error, String> =
 ```
 
 For our little module, we enumerate any and all errors that can occur. Then, instead of using
-exception classes as error values, we use one of the enumerated cases. Now when we pattern match,
-we are able to comphrensively handle failure without resulting to an `else` branch; moreover
-since Error is sealed, no outside code can add additional subtypes which we might fail to handle.
+exception classes as error values, we use one of the enumerated cases. Now, when we pattern match,
+we are able to comphrensively handle failure without resulting in an `else` branch; moreover,
```
sraaphorst: Comprehensively is misspelled.

arrow-site/doc/docs/arrow-core/datatypes/either/README.md
```diff
@@ -204,7 +204,7 @@ when(x) {
 
 ## Syntax
 
-Either can also map over the `left` value with `mapLeft` which is similar to map but applies on left instances.
+Either can also map over the `left` value with `mapLeft`, which is similar to map, but applies on left instances.
```
sraaphorst: It should be:

the `Left` value

and:

to `Left` instances

arrow-site/doc/docs/arrow-core/datatypes/either/README.md
```diff
@@ -269,10 +269,10 @@ y.fold({ 1 }, { it + 3 })
 ```
 
 The `getOrHandle()` operation allows the transformation of an `Either.Left` value to a `Either.Right` using
-the value of `Left`. This can be useful when a mapping to a single result type is required like `fold()` but without
+the value of `Left`. This can be useful when mapping to a single result type is required like `fold()`, but without
```
sraaphorst: like with `fold()`

(Feel free to ignore any of these: I'm just making suggestions.)

arrow-site/doc/docs/arrow-core/datatypes/eithert/README.md
```diff
@@ -189,7 +189,7 @@ fun getCountryCode(personId: Int): ObservableK<Either<BizError, String>> =
   }
 ```
 
-While we've got the logic working now, we're in a situation where we're forced to deal with the `Left cases`. We also have a ton of boilerplate type conversion with `fold`. The type conversion is necessary because in a monad comprehension you can only use a type of Monad. If we start with `ObservableK`, we have to stay in it’s monadic context by lifting anything we compute sequentially to a `ObservableK` whether or not it's async.
+While we've got the logic working now, we're in a situation where we're forced to deal with the `Left cases`. We also have a ton of boilerplate type conversion with `fold`. The type conversion is necessary because, in a monad comprehension, you can only use a type of Monad. If we start with `ObservableK`, we have to stay in its monadic context by lifting anything we compute sequentially to a `ObservableK`, whether or not it's async.
```
sraaphorst: `Left` cases (not `Left cases`)

arrow-site/doc/docs/arrow-core/datatypes/eithert/README.md
```diff
@@ -240,11 +240,11 @@ fun getCountryCode(personId: Int): ObservableK<Either<BizError, String>> =
   }.value().fix()
 ```
 
-Here we no longer have to deal with the `Left` cases, and the binding to the values on the left side are already the underlying values we want to focus on instead of the potential biz error values. We have automatically `flatMapped` through the `ObservableK` and `Either` in a single expression reducing the boilerplate and encoding the effects concerns in the type signatures.
+Here we no longer have to deal with the `Left` cases, and the binding to the values on the left side are already the underlying values we want to focus on instead of the potential biz error values. We have automatically `flatMapped` through the `ObservableK` and `Either` in a single expression, reducing the boilerplate and encoding the effects concerns in the type signatures.
```
sraaphorst: What is a biz?

arrow-site/doc/docs/arrow-core/datatypes/intro/README.md
```diff
@@ -10,35 +10,35 @@ permalink: /docs/datatypes/intro/
 beginner
 
 A datatype is a an abstraction that encapsulates one reusable coding pattern.
-These solutions have a canonical implementation that is generalised for all possible uses.
+These solutions have a canonical implementation that is generalized for all possible uses.
```
sraaphorst: Generalised / generalized is British vs. US. I don't suppose we have a standard when it comes to documentation. (As a Canadian, my English would be all over both.)

arrow-site/doc/docs/arrow-core/datatypes/intro/README.md
```diff
@@ -10,35 +10,35 @@ permalink: /docs/datatypes/intro/
...
 `Option<A>` is a datatype that represents absence.
 It has one generic parameter `A`, representing the type of the values that `Option` may contain.
 `Option` can be specialized for any type `A` because this type does not affect its behavior.
-`Option` behaves the same for `Int`, `String` or `DomainUserClass`.
-To indicate that `Option` is a [type constructor]({{ '/docs/patterns/glossary/' | relative_url }}) for all values of `A` it implements `OptionOf<A>`, which is a typealias of `Kind<ForOption, A>`.
+`Option` behaves the same for `Int`, `String`, or `DomainUserClass`.
+To indicate that `Option` is a [type constructor]({{ '/docs/patterns/glossary/' | relative_url }}) for all values of `A`, it implements `OptionOf<A>`, which is a typealias of `Kind<ForOption, A>`.
 
-The implementation of `Option<A>` is a sealed class with two subtypes: an object `None` and a data class `Some<A>`.
+The implementation of `Option<A>` is a sealed class with two subtypes: An object `None` and a data class `Some<A>`.
```
sraaphorst: The original "an" was correct: you shouldn't capitalize after a colon.

arrow-site/doc/docs/arrow-core/datatypes/intro/README.md
```diff
@@ -124,17 +124,17 @@ TODO
...
-- [`Sum`]({{ '/docs/arrow/data/moore/' | relative_url }}) - a datatype which holds two comonads and a flag for indicating which one is active. Both sides evolve at the same time.
+- [`Sum`]({{ '/docs/arrow/data/moore/' | relative_url }}) - a datatype that holds two comonads and a flag for indicating which one is active. Both sides evolve at the same time.
 
-- [`Day`]({{ '/docs/arrow/data/day/' | relative_url }}) - a datatype which holds two comonads which evolve independently.
+- [`Day`]({{ '/docs/arrow/data/day/' | relative_url }}) - a datatype that holds two comonads which evolve independently.
 
 #### Effects
 
-All effects are different implementations of the same abstraction: lazy execution of code that can move to other threads and cause exceptions.
+All effects are different implementations of the same abstraction: Lazy execution of code that can move to other threads and cause exceptions.
```
sraaphorst: Again, it should be lazy and not Lazy.

arrow-site/doc/docs/arrow-core/datatypes/option/README.md
```diff
@@ -15,10 +15,10 @@ beginner
 
 [Перевод на русский](/docs/arrow/core/option/ru)
 
-If you have worked with Java at all in the past, it is very likely that you have come across a `NullPointerException` at some time (other languages will throw similarly named errors in such a case). Usually this happens because some method returns `null` when you were not expecting it and thus not dealing with that possibility in your client code. A value of `null` is often abused to represent an absent optional value.
-Kotlin tries to solve the problem by getting rid of `null` values altogether and providing its own special syntax [Null-safety machinery based on `?`](https://kotlinlang.org/docs/reference/null-safety.html).
+If you have worked with Java at all in the past, it is very likely that you have come across a `NullPointerException` at some time (other languages will throw similarly named errors in such a case). Usually this happens because some method returns `null` when you weren't expecting it and, thus, isn't dealing with that possibility in your client code. A value of `null` is often abused to represent an absent optional value.
```
sraaphorst: when you weren't expecting it, and thus, ...

arrow-site/doc/docs/arrow-core/datatypes/option/README.md
```diff
@@ -15,10 +15,10 @@ beginner
...
+Kotlin tries to solve the problem by getting rid of `null` values altogether, and providing its own special syntax [Null-safety machinery based on `?`](https://kotlinlang.org/docs/reference/null-safety.html).
```
sraaphorst: Either "and provides" or just "providing"

arrow-site/doc/docs/arrow-core/datatypes/optiont/README.md
```diff
@@ -12,17 +12,17 @@ video: EWfxL9yBUJo
 {:.advanced}
 advanced
 
-`OptionT` also known as the `Option` monad transformer allows to compute inside the context when `Option` is nested in a different monad.
+`OptionT`, also known as the `Option` monad transformer, allows computation inside the context when `Option` is nested in a different monad.
 
 One issue we face with monads is that they don't compose. This can cause your code to get really hairy when trying to combine structures like `ObservableK` and `Option`. But there's a simple solution, and we're going to explain how you can use Monad Transformers to alleviate this problem.
```
sraaphorst: We shouldn't start a sentence with "But."

Instead we should consider using a comma, and then but.

Also, why is Monad Transformers capitalized?

arrow-site/doc/docs/arrow-core/datatypes/optiont/README.md
```diff
@@ -44,7 +44,7 @@ fun getCountryCode(maybePerson : Option<Person>): Option<String> =
   }
 ```
 
-Nested flatMap calls flatten the `Option` but the resulting function starts looking like a pyramid and can easily lead to callback hell.
+Nested flatMap calls flatten the `Option`, but the resulting function starts looking like a pyramid and can easily lead to callback hell.
```
sraaphorst: This looks copy-pasted from `Either`. Someone should probably work on this sometime to avoid the repetition.

arrow-site/doc/docs/arrow-core/datatypes/optiont/README.md
```diff
@@ -163,7 +163,7 @@ fun getCountryCode(personId: Int): ObservableK<Option<String>> =
       }       
 ```
 
-While we've got the logic working now, we're in a situation where we're forced to deal with the `None cases`. We also have a ton of boilerplate type conversion with `fold`. The type conversion is necessary because in a monad comprehension you can only use a type of Monad. If we start with `ObservableK`, we have to stay in it’s monadic context by lifting anything we compute sequentially to a `ObservableK` whether or not it's async.
+While we've got the logic working now, we're in a situation where we're forced to deal with the `None cases`. We also have a ton of boilerplate type conversion with `fold`. The type conversion is necessary because, in a monad comprehension, you can only use a type of Monad. If we start with `ObservableK`, we have to stay in it’s monadic context by lifting anything we compute sequentially to a `ObservableK`, whether or not it's async.
```
sraaphorst: `None` cases

arrow-site/doc/docs/arrow-core/datatypes/statet/README.md
```diff
@@ -12,17 +12,17 @@ video: 1MJBXKaq0Hc
...
+For our purposes here, we're going to transform a monad that serves as a container that represents branching as an an error (left) or state (right) where computation can be performed. Both `State<S, A>` and `Either<L, R>` would be examples of datatypes that provide instances for the `Monad` typeclasses.
 
-Because [monads don't compose](http://tonymorris.github.io/blog/posts/monads-do-not-compose/), we may end up with nested structures such as `Either<Error, State<Either<Error, State<S, Unit>>, Unit>>` when using `Either` and `State` together. Using Monad Transformers can help us to reduce this boilerplate.
+Because [monads don't compose](http://tonymorris.github.io/blog/posts/monads-do-not-compose/), we may end up with nested structures such as `Either<Error, State<Either<Error, State<S, Unit>>, Unit>>` when using `Either` and `State` together. Using Monad Transformers can help us reduce this boilerplate.
```
sraaphorst: Again, why is Monad Transformers capitalized?

arrow-site/doc/docs/arrow-core/datatypes/statet/README.md
```diff
@@ -211,10 +211,10 @@ fun main() {
 }
 ```
 
-As is immediately clear, this code while properly modelling the errors, has become more complex but our signature now represents a simple `Stack` as a `List` with an error domain.
+As is immediately clear, this code, while properly modelling the errors, has become more complex. But our signature now represents a simple `Stack` as a `List` with an error domain.
```
sraaphorst: Again, a sentence shouldn't start with a conjunction (but). It should be "complex, but"

arrow-site/doc/docs/arrow-core/datatypes/state/README.md
```diff
@@ -129,7 +129,7 @@ fun main() {
 }
 ```
 
-As you can see, since we cannot modify the immutable Stack, we need to create a new instance every time we push or pop values from it. For that same reason we have to return the newly created Stack with every operation.
+As you can see, since we cannot modify the immutable Stack, we need to create a new instance every time we push or pop values from it. For that same reason, we have to return the newly created Stack with every operation.
```
sraaphorst: Should it be `Stack` instead of Stack?

arrow-site/doc/docs/arrow-core/datatypes/statet/README.md
```diff
@@ -315,7 +315,7 @@ fun main() {
 }
 ```
 
-While our code looks very similar to what we had before there are some key advantages. State management is now contained within `State` and we are dealing only with 1 monad instead of 2 nested monads so we can use monad bindings!
+While our code looks very similar to what we had before, there are some key advantages. State management is now contained within `State`, and we are dealing with only 1 monad instead of 2 nested monads, so we can use monad bindings!
```
sraaphorst: Numbers less than 10 should be written out: "one monad instead of two nested monads"

arrow-site/doc/docs/arrow-core/datatypes/sum/README.md
```diff
@@ -10,11 +10,11 @@ redirect_from:
 
 > The sum of two Comonads is also a Comonad.
 
-A Sum is a data structure which holds two [`Comonads`]({{ '/docs/arrow/typeclasses/comonad' | relative_url }}) and a flag for selecting which one is "active".
+A Sum is a data structure that holds two [`Comonads`]({{ '/docs/arrow/typeclasses/comonad' | relative_url }}) and a flag for selecting which one is "active".
```
sraaphorst: A `Sum`

arrow-site/doc/docs/arrow-core/datatypes/sum/README.md
```diff
@@ -10,11 +10,11 @@ redirect_from:
...
 
-A common pattern used when building user interfaces is showing two tabs and allow the user to change between them. We can achieve this behavior with Sum.
+A common pattern used when building user interfaces is showing two tabs and allowing the user to change between them. We can achieve this behavior with Sum.
```
sraaphorst: `Sum`

arrow-site/doc/docs/arrow-core/datatypes/sum/README.md
```diff
@@ -10,11 +10,11 @@ redirect_from:
...
 
-For creating a Sum we need two comonadic datatypes which `extract` result type is the same. We will also need an initial side, by default it is Left.
+For creating a Sum, we need two comonadic datatypes in which the `extract` result type is the same. We will also need an initial side—by default it is Left.
```
sraaphorst: `Sum`

arrow-site/doc/docs/arrow-core/datatypes/sum/README.md
```diff
@@ -33,7 +33,7 @@ We can change the active side with `changeSide`.
 sum.changeSide(Sum.Side.Right).extract(Store.comonad(), Store.comonad())
 ```
 
-As Sum is also a `Comonad` we have access to `coflatmap` and `map` for changing the `extract` result.
+As Sum is also a `Comonad`, we have access to `coflatmap` and `map` for changing the `extract` result.
```
sraaphorst: `Sum`

Okay. I'm going to stop here, but this is a fantastic improvement in the documentation. Thanks so much for this hard work. Instead of making suggestions here, I'll file a PR at a later time.

jkmcclellan: You bet! And thanks for looking over these. You caught some great things I missed.

jkmcclellan:
I've made it through all the available FX and Optics docs too.

nomisRev:
Thank you @jkmcclellan 👏 👏 👏 🎉 🎉

calvellido:
The changes in here were include in #1865

PR #1800, opened 2019-11-15
implement Semialign typeclass

abendt:
implementation of Semialign typeclass with instances for some arrow types


modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
```diff
@@ -279,3 +285,25 @@ interface ListKAlternative : Alternative<ForListK>, ListKApplicative {
...
+    fun <T> maybeGet(list: List<T>, idx: Int) =
+      if (idx >= list.size) Option.empty() else list[idx].some()
+
+    val maxSize = max(l.size, r.size)
+
+    return (0..maxSize).map {
+      maybeGet(l, it) toT maybeGet(r, it)
+    }.filterMap {
+      Ior.fromOptions(it.a, it.b)
+    }.k()
```
1Jajen1: I don't dislike this (especially the use of fromOptions is nice^^), but is there any reason to not write it similar to the haskell one?
So that would be:
```kotlin
fun align(ls: List<Int>, rs: List<Int>): List<Ior<Int, Int>> = when {
    ls.isEmpty() -> rs.map { it.rightIor() }
    rs.isEmpty() -> ls.map { it.leftIor() }
    else -> listOf(Ior.Both(ls.first(), rs.first())) + align(ls.tail(), rs.tail())
}
```
If there is any reason for it then leave it, up to you.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemialignLaws.kt
```diff
@@ -0,0 +1,170 @@
...
+  this.a.a toT (this.a.b toT this.b)
+
+fun <A, B, C> Tuple2<A, Tuple2<B, C>>.unassoc(): Tuple2<Tuple2<A, B>, C> =
+  (this.a toT this.b.a) toT this.b.b
+
+fun <A, T> Const<Const<A, T>, T>.assoc(): Const<A, T> =
+  this.value()
+
+fun <A, T> Const<A, T>.unassoc(): Const<Const<A, T>, A> =
+  Const.just(this)
```
1Jajen1: Are those used anywhere? Maybe we can move them to some utility file or something.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemialignLaws.kt
```diff
@@ -0,0 +1,170 @@
...
+
+    // toListOf (folded . here) (align x y)
+    val middle: List<A> = toList(align(a, b).map { it.justLeft() }).flattenOption()
+
+    // mapMaybe justHere (toList (align x y))
+    val right: List<A> = toList(align(a, b)).filterMap { it.justLeft() }
+
+    left == right && left == middle
+  }
+}
```
1Jajen1: Please move this to a seperate IorLaws.laws function that takes a foldable because not everyone should need to supply it. Just an overload will do

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemialignLaws.kt
```diff
@@ -0,0 +1,170 @@
...
+
+      left.equalUnderTheLaw(right, EQ)
+    }
+
+  fun <F, A> Semialign<F>.semialignAlignedness(
+    G: Gen<Kind<F, A>>,
+    FOLD: Foldable<F>
+  ) = forAll(G, G) { a: Kind<F, A>, b: Kind<F, A> ->
+
+    fun <E> toList(es: Kind<F, E>): List<E> = FOLD.run {
```
1Jajen1: There is (or should be) foldmap which takes a monoid: `toList` = `foldMap(ListK.monoid())` or similar

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemialignLaws.kt
```diff
@@ -0,0 +1,170 @@
...
+      es.foldLeft(emptyList()) { acc, e ->
+        acc + e
```
1Jajen1: I always forget that kotlin plus can work on different types ... so confusing coming from haskell :)

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemialignLaws.kt
1Jajen1: Not sure about order tho

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemialignLaws.kt
```diff
@@ -0,0 +1,170 @@
...
+      }
+    }
+
+    val left: List<A> = toList(a)
+
+    // toListOf (folded . here) (align x y)
+    val middle: List<A> = toList(align(a, b).map { it.justLeft() }).flattenOption()
```
1Jajen1: `flattenOption` is just `filterMap(::identity)` right? So just change `map` to `filterMap` and drop it would work the same I think

1Jajen1: Oh wait that requires `FunctorFilter`, so this is perfectly fine

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
```diff
@@ -279,3 +291,22 @@ interface ListKAlternative : Alternative<ForListK>, ListKApplicative {
...
+    // from arrow-syntax
+    fun <T> List<T>.tail(): List<T> = this.drop(1)
+
+    fun <X, Y> alignRec(ls: List<X>, rs: List<Y>): List<Ior<X, Y>> = when {
+      ls.isEmpty() -> rs.map { it.rightIor() }
+      rs.isEmpty() -> ls.map { it.leftIor() }
+      else -> listOf(Ior.Both(ls.first(), rs.first())).listPlus(alignRec(ls.tail(), rs.tail()))
+    }
+
+    return alignRec(left.fix(), right.fix()).k()
```
1Jajen1: My initial comment was a bit early, wanted that in a review because I often change my thoughts :D

Either way there is no need for functions in here, although I am surprised tail is not available here. Check again after moving the tests.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemialignLaws.kt
```diff
@@ -0,0 +1,170 @@
...
+    Law("Semialign Laws: associativity") { SA.semialignAssociativity(gen, buildEq(iorEq2)) },
+    Law("Semialign Laws: with") { SA.semialignWith(gen, buildEq(String.eq())) },
+    Law("Semialign Laws: functoriality") { SA.semialignFunctoriality(gen, buildEq(Ior.eq(String.eq(), String.eq()))) },
+    Law("Semialign Laws: alignedness") { SA.semialignAlignedness(gen, FOLD) }
+  )
+
+  // Laws ported from https://hackage.haskell.org/package/semialign-1.1/docs/Data-Semialign.html
+
+  fun <F> Semialign<F>.semialignIdempotency() {
+    TODO()
```
1Jajen1: Let me think about this one for a bit. Not sure yet either.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemialignLaws.kt
```diff
@@ -0,0 +1,170 @@
...
+
+object SemialignLaws {
+
+  val iorEq1 = Ior.eq(Int.eq(), Int.eq())
+  val iorEq2 = Ior.eq(Int.eq(), iorEq1)
+
+  fun <F> laws(
+    SA: Semialign<F>,
+    gen: Gen<Kind<F, Int>>,
+    buildEq: (Eq<*>) -> Eq<Kind<F, *>>,
```
1Jajen1: Is there a reason not to just force `Eq<Kind<F, Int>>`?

modules/core/arrow-core/src/test/kotlin/arrow/core/extensions/ListKExtensionTest.kt
```diff
@@ -0,0 +1,33 @@
...
+class ListKExtensionTest : UnitSpec() {
+  init {
+    testLaws(SemialignLaws.laws(ListK.semialign(),
+      Gen.listK(Gen.int()) as Gen<Kind<ForListK, Int>>,
+      { ListK.eq(it) as Eq<Kind<ForListK, *>> },
+      ListK.foldable()
+    ))
+
+    "lists with different lengths are padded" {
+      ListK.semialign().align(
```
1Jajen1: Please use property based testing instead:
```
// pseudo code like stuff
forAll (genList, genList) { l1, l2 ->
 align(l1, l2).size == max(l1.size, l2.size)
}
```
And similar `align(l1, l2).take(max(l1.size, l2.size)).all { isBoth }` and similarly check the rest for left or right

modules/core/arrow-core/src/test/kotlin/arrow/core/extensions/SequencekExtensionTest.kt
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    "align infinite sequences" {
+      val seq1 = sequence {
+        while (true) {
+          yield("A")
+        }
+      }.k()
+
+      val seq2 = sequence {
```
1Jajen1: `generateSequence(0) { it + 1 }` is a nice helper for infinite sequences

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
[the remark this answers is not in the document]
1Jajen1: So github hides this from me when I refresh changes and the files changed, but then puts it here when I submit the review? Weird, ignore this

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
1Jajen1: > Check again after moving the tests.
Ignore that part, I was somewhere else with my thoughts :D
Just use drop(1) instead, its just twice anyway, we have tail as an extension function somewhere tho

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/nonemptylist.kt
```diff
@@ -183,3 +190,28 @@ fun <F, A> Reducible<F>.toNonEmptyList(fa: Kind<F, A>): NonEmptyList<A> =
...
+    val maxSize = max(l.size, r.size)
+
+    val list = (0..maxSize).map {
+      maybeGet(l, it) toT maybeGet(r, it)
+    }.filterMap {
+      Ior.fromOptions(it.a, it.b)
+    }
+
+    return NonEmptyList.fromListUnsafe(list)
+  }
```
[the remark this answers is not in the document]
1Jajen1: If you want to do this safe then just use Nel(Ior.Both(l.head, r.head), align(l.tail, r.tail)) or similar

1Jajen1:
Oh also it'd be nice to have api docs in the typeclass. Just a small explanation and example would be fine. Same for the methods inside.
Oh and add a note to it that either one of the methods needs to be overriden to make it work.

1Jajen1:
An example of api docs is here: https://github.com/arrow-kt/arrow/blob/62f692397ef44e6acde12af817d94a19d2ce9621/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Hash.kt Those will be used to generate docs that look like this: https://arrow-kt.io/docs/arrow/typeclasses/hash/#hash

1Jajen1:
Idempotency law is a mind bender... `join` is from monad (don't think we have it in arrow, but it's just `flatMap(::identity)`. And the monad used is the function monad: `join These` is of type `a -> These a a` so the right hand side is of type `f (These a a)`. And the left hand side is `join align` which is of type `f a -> f (These a a)` and then we apply the `f a` and get `f (These a a)` again.
The function data type in haskell that is relevant here is `Function1` and I admit this is really confusing, until you think about currying the right way: These is a constructor from `a -> b -> These a b` or better put `a -> (b -> These a b)` and `join` on that "flattens" the functions (fixing them to a or b in both cases, haskells type inference is amazing) to `b -> These b b`. The internals for Function1 in arrow are here https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-core-data/src/main/kotlin/arrow/core/Function1.kt
The types in kotlin terms are (Translated to Function1): `Ior.Both` = `Function1<A, Function1<A, Ior<A, A>>>`, with that I think it's trivial to see how `join` flattens to `Function1<A, Ior<A, A>>`. 
And with `join align` = `Function1<Kind<F, A>, Function1<Kind<F, A>, Kind<F, Ior<A, A>>>` which is similar

1Jajen1:
Btw ignore the complicated idempotency law definition: Just define it as: `align(xs, xs) == xs.map { Both(it, it) }` which is the same thing and the only reason haskell made it so complicated is to get it nice and point free...

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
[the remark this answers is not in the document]
abendt: i'm also confused by the GitHub Ui. it did not seem possible to reply to your comments individually until now. so i replied in an extra comment ...

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/mapk.kt
```diff
@@ -127,3 +133,19 @@ interface MapKHash<K, A> : Hash<MapK<K, A>>, MapKEq<K, A> {
...
+    right: Kind<MapKPartialOf<K>, B>
+  ): Kind<MapKPartialOf<K>, Ior<A, B>> {
+    val l = left.fix()
+    val r = right.fix()
+    val keys = l.keys + r.keys
+
+    return keys.map { key ->
+      Ior.fromOptions(l[key].toOption(), r[key].toOption()).map { key toT it }
+    }.flattenOption().toMap().k()
+  }
```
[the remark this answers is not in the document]
abendt: i did not see anything here: https://hackage.haskell.org/package/semialign-1.1/docs/Data-Semialign.html

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
abendt: ok, using drop now directly. the extensions function is in the module arrow-syntax thats not in the dependencies

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemialignLaws.kt
abendt: my idea here was to let SemialignLaws create the Eq and have one generic function to allow tests to put the Eq in a specific context (e.g. ListK.eq). the function is used with different arguments types (see laws construction in SemialignLaws) so the type Eq<Kind<F, Int>> wouldn't fit in all cases. i also think i should enhance this, not yet sure how.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemialignLaws.kt
abendt: assoc/unassoc and justLeft functions in SemialignLaws. yes, it was my thought too to move them somewhere else. parts of them are not used (also did it as an exercise). would it be ok to put them into the respective extensions files?

modules/core/arrow-core/src/test/kotlin/arrow/core/extensions/ListKExtensionTest.kt
abendt: i changed the custom tests to use property based testing

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemialignLaws.kt
1Jajen1: Yeah right, this is actually needed and it's a decent solution. Something better would be lifted equality like eq1 from haskell but that's a different story altogether

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/mapk.kt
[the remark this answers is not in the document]
1Jajen1: Hmm then let's just ignore this, if it's lawful it likely won't matter

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemialignLaws.kt
1Jajen1: I don't think the either assoc/unassoc has any use in current tests, but the tuple one could definitely be used to implement the bijection method for any functor: `map(assoc)` should be equivalent. It is used a few times in some laws for collections (and probably others) and is basically duplicate code. It has to be moved to a utility file tho. If you can come up with a use case for use outside of tests we can also move this to the datatypes themselves.

modules/core/arrow-core/src/test/kotlin/arrow/core/extensions/ListKExtensionTest.kt
1Jajen1: You can also test that everything from `min(s1, s2)` is either left or right in a similar manner. Maps and sequences should also have tests like that. Once that is done the tests are good to go i think :+1:

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemialignLaws.kt
abendt: one of the assoc functions was used in the tests. i implemented the others as an exercise too without using them anywhere.

i removed the unused assoc/unassoc functions for now and changed the visibilty of the remaining two functions to private as they are only used from within the Laws file. is that acceptable for you?

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemialignLaws.kt
abendt: i have changed it

modules/core/arrow-core/src/test/kotlin/arrow/core/extensions/ListKExtensionTest.kt
abendt: i have added more tests for ListK, MapK and SequenceK that should cover these cases now

abendt:
@1Jajen1 for some reason i cannot reply to some of your comments directly.
i added some javadoc to the Semialign interface. thanks also for the Idempotency explanation. Actually i do not understand it yet :-) but i use your simplified version now.

1Jajen1:
> @1Jajen1 for some reason i cannot reply to some of your comments directly.
i added some javadoc to the Semialign interface. thanks also for the Idempotency explanation. Actually i do not understand it yet :-) but i use your simplified version now.

The idempotency law was interesting to break down, the function monad is just confusing when used in haskell because partial application there is not explicit...
It looks good, three small things left:
- drop the inner function in the list and nel instances, I don't think the compiler optimizes them away (btw such a definition is more performant in haskell (without -02) because ghc will better optimize it, but that is sadly not the case in kotlin)
- I think you misunderstood what I meant for the other test cases: Right now we are checking both part twice, but we can actually cover all cases (for lists by dropping min(s1, s2) elements and checking if they are s1 > s2 left or right
- The docs could use an example for align and/or alignWith (best would be a runnable one, check my previous comment for a reference to existing docs that do that). Also the header states "typeclass for zipping and aligning", but isn't semialign only aligning?

I think that's all and after that this is good to go. Thank you for picking this up :+1:

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
```diff
@@ -279,3 +291,17 @@ interface ListKAlternative : Alternative<ForListK>, ListKApplicative {
...
+    left: Kind<ForListK, A>,
+    right: Kind<ForListK, B>
+  ): Kind<ForListK, Ior<A, B>> = alignRec(left.fix(), right.fix()).k()
+
+  private fun <X, Y> alignRec(ls: List<X>, rs: List<Y>): List<Ior<X, Y>> = when {
+    ls.isEmpty() -> rs.map { it.rightIor() }
+    rs.isEmpty() -> ls.map { it.leftIor() }
+    else -> listOf(Ior.Both(ls.first(), rs.first())).listPlus(alignRec(ls.drop(1), rs.drop(1)))
+  }
+}
```
1Jajen1: Why not:
```kotlin
fun align() = when {
    ls.isEmpty() -> rs.map { it.rightIor() }
    rs.isEmpty() -> ls.map { it.leftIor() }
    else -> listOf(Ior.Both(ls.first(), rs.first())).listPlus(alignRec(ls.drop(1), rs.drop(1)))
  }
```
directly? :see_no_evil: Sorry if my comments have been a bit misleading on that

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Semialign.kt
```diff
@@ -0,0 +1,54 @@
...
+   *   println(result)
+   * }
+   * ```
+   */
+  fun <A, B> align(left: Kind<F, A>, right: Kind<F, B>): Kind<F, Ior<A, B>> = alignWith(::identity, left, right)
+
+  /**
+   * Combines two structures by taking the union of their shapes and combining the elements with the given function.
+   *
+   *    *  * {: data-executable='true'}
```
1Jajen1: Not sure how this would render, but can you remove the leading two "*"? Also in the above one there is one extra. I think the line gets removed anyway but this is kind of hard to test beforehand

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
abendt: i tried that but could not make it work initially. issue is that you have fix() ls and rs first, also listPlus is more complicated as i then try to add Kind<ForListK> to List. i will look into it again and see if i can find a way.

1Jajen1:
Btw you can test ank by running the dokka task (`gradle dokka` I think) and then the ank task (`gradle runAnk` afaik) takes a while, but is the only way to check if ank goes right.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Semialign.kt
abendt: yes, thanks that was a copy paste error

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
1Jajen1: ah yeah that's always fun. makes it a bit uglier but you can do it either by calling fix every time or stick a `(ls.fix() toT rs.fix()).let { (ls, rs) -> }` to get it done once before. Does not matter to me and if you think the private function is fine then just keep it. Up to you :)

abendt:
don't understand your point regarding the testcases.

i understood i can change the last ListK test to be more specific:

```
 forAll(Gen.listK(Gen.bool()), Gen.listK(Gen.bool())) { a, b ->
        ListK.semialign().run {
          align(a, b).fix().drop(min(a.size, b.size)).all {
            if (a.size < b.size) {
              it.isRight
            } else {
              it.isLeft
            }
          }
        }
      }
```

however you seem to imply that i can check all cases with just one testcase? could you elaborate here?

1Jajen1:
> don't understand your point regarding the testcases.
> 
> i understood i can change the last ListK test to be more specific:
> 
> ```
>  forAll(Gen.listK(Gen.bool()), Gen.listK(Gen.bool())) { a, b ->
>         ListK.semialign().run {
>           align(a, b).fix().drop(min(a.size, b.size)).all {
>             if (a.size < b.size) {
>               it.isRight
>             } else {
>               it.isLeft
>             }
>           }
>         }
>       }
> ```
> 
> however you seem to imply that i can check all cases with just one testcase? could you elaborate here?

No that is actually exactly what I mean :pray: Currently the tests check length and if the first overlap is Both and then if the non-overlap is not Both, but checking for right or left seems better to me as the whole thing now covers all cases :)

1Jajen1:
This looks good to me, if you want to change more stuff let me know, otherwise we'll merge it. @pakoito Want to take a quick look over if this is good to go?

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemialignLaws.kt
abendt: i looked Eq1 up. As it does not seem to exist in arrow i tried to implement it for Option.
it seems straight forward to do and indeed would enhance the mentioned code as the tests then just would pass in an instance of Eq1.

```
MyLaws.laws(Option.semialign(),
        Gen.option(Gen.int()) as Gen<Kind<ForOption, Int>>,
        Option.eq1() 
      )
```

is there a particular reason why Eq1 does not exist in arrow? it looks quite useful

1Jajen1: It is quite useful :) Same with the other lifted typeclass versions (Show1, Ord1,...)
It's actual uses are quite rare tho, the ones in the tests here are perfect, but the usual case for types is that the inner type is abstract and the outer is concrete, whereas lifted instances like Eq1 come into use when the outer type is abstract and the inner is concrete.
Can't hurt to have it tho :)

abendt:
i merged in the EqK related changes and changed Semialign Laws to use EqK now

1Jajen1:
Looks good to me :+1:

1Jajen1:
Thanks for doing this! I can't wait to replace the custom code for zipping I have all over the place :D

PR #1801, opened 2019-11-15
implement SetK functor instance

abendt:
i noticed that there was no SetK functor. this PR add an implementation


1Jajen1:
Is it lawful? Especially with the second functor law?

I read somewhere that it does not hold it. Will look up the article in a second.

1Jajen1:
Ah here: https://blog.ploeh.dk/2018/12/03/set-is-not-a-functor/ Describes it quite well, just mentally replace select with map

abendt:
thanks for explanation!

PR #1804, opened 2019-11-18
add Eq1 with instances for some arrow types

abendt:
while implementing the Semialign instance tests (https://github.com/arrow-kt/arrow/pull/1800)
i had the need to lift an Eq into contexts in a generic way. my current solution works, however it uses star projection and requires some casts.
@1Jajen1 mentioned that this could be enhanced using lifted equality Eq1 (https://github.com/arrow-kt/arrow/pull/1800#discussion_r347049233) so i went ahead and implemented Eq1 with instances for some types.


pakoito:
I believe Eq1 can be rewritten to feel more idiomatic to Kotlin and with Arrow nomenclature, such as

```
interface EqK<F> {
  fun <A> Kind<F, A>.eqK(other: Kind<F, A>, EQ: Eq<A>)
}
```

Skip `lift` for now, as there are no currying or partial application by default, we tend to apply the values ourselves in a new lambda.

abendt:
ok,  i changed it accordingly

pakoito:
Brilliant, thanks! Next thing: instead of writing a test on each instance, see if you can come up with EqKLaws so we can generalise for all instances :D You can use EqLaws as a baseline, for example!

abendt:
thanks @pakoito. makes sense regarding the generalized test. I have created an EqKLaws based on EqLaws with an additional test to ensure that EqK == Eq respective container specific cases  (e.g. Some.empty(), emptyList() ...) . will look into the doc later

abendt:
i have added some documentation now

PR #1806, opened 2019-11-18
add a SortedMapK Hash instance

abendt:
building on https://github.com/arrow-kt/arrow/pull/1805 this adds a SortedMapK Hash instance

### DIFF
--- a/modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
+++ b/modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
@@ -3,13 +3,16 @@ package arrow.core
 import arrow.Kind2
 import arrow.core.extensions.eq
 import arrow.core.extensions.functor
+import arrow.core.extensions.hash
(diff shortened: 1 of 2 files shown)


modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
```diff
@@ -24,9 +27,9 @@ class SortedMapKTest : UnitSpec() {
   }
 
   init {
-
     testLaws(
       EqLaws.laws(SortedMapK.eq(String.eq(), Int.eq())) { sortedMapOf("$it" to it).k() },
+      HashLaws.laws(SortedMapK.hash(String.hash(), Int.hash()), EQ) { sortedMapOf("key" to it).k() },
```
pakoito: HashLaws already include EqLaws, so you can skip the line above :D https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/HashLaws.kt#L11

PR #1809, opened 2019-11-20
Align with instances

abendt:
typeclass Align with some instances. including docs and tests


modules/docs/arrow-docs/docs/docs/arrow/typeclasses/align/README.md
```diff
@@ -0,0 +1,60 @@
...
+}
+```
+
+### Laws
+
+Arrow provides [`AlignLaws`][functor_laws_source]{:target="_blank"} in the form of test cases for internal verification of lawful instances and third party apps creating their own Align instances.
+
+#### Creating your own `Align` instances
+
+Arrow already provides Align instances for common datatypes (e.g. Option, ListK, SetK). See their implementations
```
1Jajen1: SetK does not have an instance :see_no_evil:

abendt: right, thank you. i remember. Set is not a Functor :D

PR #1810, opened 2019-11-21
Sortedmapk semialign

abendt:
add SortedMapK Semialign instance

### DIFF
--- a/modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
+++ b/modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
@@ -2,6 +2,7 @@ package arrow.core
 
 import arrow.Kind
 import arrow.Kind2
+import arrow.core.extensions.align
 import arrow.core.extensions.eq
 import arrow.core.extensions.foldable
 import arrow.core.extensions.functor
@@ -14,9 +15,9 @@ import arrow.core.extensions.sortedmapk.hash.hash
 import arrow.core.extensions.traverse
 import arrow.test.UnitSpec
 import arrow.test.generators.sortedMapK
+import arrow.test.laws.AlignLaws
 import arrow.test.laws.HashLaws
 import arrow.test.laws.MonoidLaws
-import arrow.test.laws.SemialignLaws
 import arrow.test.laws.ShowLaws
 import arrow.test.laws.TraverseLaws
 import arrow.typeclasses.Eq
@@ -46,7 +47,7 @@ class SortedMapKTest : UnitSpec() {
         SortedMapK.functor<String>(),
         { a: Int -> sortedMapOf("key" to a).k() },
         EQ),
-      SemialignLaws.laws(SortedMapK.semialign<String>(),
+      AlignLaws.laws(SortedMapK.align<String>(),
         Gen.sortedMapK(Gen.string(), Gen.int()) as Gen<Kind<SortedMapKPartialOf<String>, Int>>,
         EQK,
         SortedMapK.foldable<String>()
--- a/modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
+++ b/modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
@@ -46,7 +46,7 @@ class SortedMapKTest : UnitSpec() {
         SortedMapK.functor<String>(),
         { a: Int -> sortedMapOf("key" to a).k() },
         EQ),
-      SemialignLaws.foldablelaws(SortedMapK.semialign<String>(),
+      SemialignLaws.laws(SortedMapK.semialign<String>(),
         Gen.sortedMapK(Gen.string(), Gen.int()) as Gen<Kind<SortedMapKPartialOf<String>, Int>>,
         EQK,
         SortedMapK.foldable<String>()
--- a/modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
+++ b/modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
@@ -22,6 +22,7 @@ import arrow.test.laws.TraverseLaws
 import arrow.typeclasses.Eq
 import arrow.typeclasses.EqK
 import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
 
 class SortedMapKTest : UnitSpec() {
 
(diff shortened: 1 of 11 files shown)


modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
```diff
@@ -32,6 +45,54 @@ class SortedMapKTest : UnitSpec() {
         SortedMapK.traverse<String>(),
         SortedMapK.functor<String>(),
         { a: Int -> sortedMapOf("key" to a).k() },
-        EQ))
+        EQ),
+      SemialignLaws.foldablelaws(SortedMapK.semialign<String>(),
+        Gen.sortedMapK(Gen.string(), Gen.int()) as Gen<Kind<SortedMapKPartialOf<String>, Int>>,
```
nomisRev: This can be avoided if we redefine the generator.

`fun <K : Comparable<K>, V> Gen.Companion.sortedMapK(genK: Gen<K>, genV: Gen<V>): Gen<SortedMapKOf<K, V>> =`

modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
```diff
@@ -32,6 +45,54 @@ class SortedMapKTest : UnitSpec() {
         SortedMapK.traverse<String>(),
         SortedMapK.functor<String>(),
         { a: Int -> sortedMapOf("key" to a).k() },
-        EQ))
+        EQ),
+      SemialignLaws.foldablelaws(SortedMapK.semialign<String>(),
```
nomisRev: Why `foldablelaws` instead of `laws`?

nomisRev:
Looking great! 👏 👏

modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
```diff
@@ -23,6 +31,11 @@ class SortedMapKTest : UnitSpec() {
       fix()["key"] == b.fix()["key"]
   }
 
+  val EQK = object : EqK<SortedMapKPartialOf<String>> {
+    override fun <A> Kind<SortedMapKPartialOf<String>, A>.eqK(other: Kind<SortedMapKPartialOf<String>, A>, EQ: Eq<A>): Boolean =
+      SortedMapK.eq(String.eq(), EQ).run { this@eqK.fix().eqv(other.fix()) }
+  }
+
```
1Jajen1: Can this be added as an instance for SortedMapK directly?

modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
1Jajen1: Semialign has additional laws for functors that are also foldable. For reference: https://hackage.haskell.org/package/semialign-1.1/docs/Data-Semialign.html

1Jajen1: But seeing this I'd prefer an overload instead of a seperate name o.O

nomisRev: > Semialign has additional laws for functors that are also foldable

Got it 👍 It can be renamed to `laws` as well since it has a different signature.

modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
nomisRev: It can and we should :)

nomisRev: Although the `Eq` instance for `SortedMapK` should be parameterized for both the keys and values.

modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
abendt: tried that out.  the types do not match anymore in the  MonoidLaws.laws() function call a few lines above.

modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
abendt: renamed it

modules/core/arrow-core-data/src/test/kotlin/arrow/core/SortedMapKTest.kt
abendt: i have added an Align instance

abendt: > Although the Eq instance for SortedMapK should be parameterized for both the keys and values.

@nomisRev i do not understand what you exactly mean by this? could you explain a bit more please?

nomisRev: We can add an instance where `K` remains generic.

```
interface SortedMapK<K> : EqK<SortedMapKPartialOf<K>> {
   fun EQK(): Eq<K>
   
    override fun <A> Kind<SortedMapKPartialOf<String>, A>.eqK(other: Kind<SortedMapKPartialOf<String>, A>, EQ: Eq<A>): Boolean =
      SortedMapK.eq(EQK(), EQ).run { this@eqK.fix().eqv(other.fix()) }
}
```

abendt: ok now i've got it. didn't get your comment was related to

> Can this be added as an instance for SortedMapK directly?

thanks!

abendt: done

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/sortedmap.kt
```diff
@@ -111,3 +118,36 @@ interface SortedMapKHash<K : Comparable<K>, A> : Hash<SortedMapK<K, A>>, SortedM
...
+  object : SortedMapKSemialign<K> {}
+
+interface SortedMapKAlign<K : Comparable<K>> : Align<SortedMapKPartialOf<K>>, SortedMapKSemialign<K> {
+  override fun <A> empty(): Kind<SortedMapKPartialOf<K>, A> = emptyMap<K, A>().toSortedMap().k()
+}
+
+fun <K : Comparable<K>> SortedMapK.Companion.align(): SortedMapKAlign<K> =
+  object : SortedMapKAlign<K> {}
+
+@extension
```
1Jajen1: Wait, does this work? If so, what is the reason all the others are not using `@extension`? Can you test if this generates the right code and then maybe change the other instances as well?

abendt: adding `@extension` to SortedMapKSemialign leads to

```
internal val semialign_singleton: SortedMapKSemialign<Any?> = object :
    SortedMapKSemialign<java.lang.Comparable> {}

Type argument is not within its bounds: should be subtype of 'Comparable<Any?>'
One type argument expected for interface Comparable<T : Any!>
```

similiar effect with the Align instance

1Jajen1: Ah ok, hopefully those things will be gone with arrow-meta in the future...

1Jajen1:
Guess it's time to merge :) Thanks :clap:

Issue #1749, opened 2019-10-26
Traverse is fundamentally broken for infinite data structures


1Jajen1:
Currently code like this will loop forever:
```kotlin
fun main() {
    val xs = generateSequence(0) { println("..."); it + 1 }
    val res = xs.k().traverse(Option.applicative()) {
        if (it < 5) it.some() else None
    }.fix()
    println(res)
}
```
This is because the `Applicative` typeclass takes a strict argument for `ap` and it's derivatives. That is fundamentally the root problem...

A lazy version like this however completes just fine.
```kotlin
fun <A, B>Option<A>.lazyAp(x: Eval<Option<(A) -> B>>): Option<B> = flatMap { a ->
    x.value().map { f -> f(a) }
}

fun main() {
    val xs = generateSequence(0) { println("..."); it + 1 }
    val f = { i: Int -> if (i < 5) i.some() else None }
    // val res = xs.k().traverse(Option.applicative(), f).fix()
    
    xs.foldRight(Eval.now(Option.just(emptyList<Int>()))) { v, acc ->
        Eval.later {
            f(v).lazyAp(
                acc.map { optL -> optL.map { l -> { i: Int -> listOf(i) + l } } }
            )
        }
    }.value()
}
```

`traverse` should definitly be implemented similar to that! This also changes the order to left-to-right and basically allows dealing with any data-structure, finite or not.

So the change that I'd make is: Add `lazyAp` to `Applicative` (or change normal ap, although I'd be fine keeping it) and implement `traverse` for every (collection) instance using this method.
This also gives some insight on how to fix #1734 . This method can also be used in `Foldable` to get a lazy variant of `get` which works on infinite structures.

Issue #1734, opened 2019-10-23
Short circuit foldable operations


1Jajen1:
Currently a call to `firstOption` for example forces the entire fold to go through, which is especially bad if it is infinite (this has bitten me more than once with sequences...). This is unnecessary and should be easy to fix by short circuiting in right folds.

Any opinions?

Edit: The only method that does not use a right-fold (and the one firstOption is implemented with) is `get`. Changing that to a right fold will do the trick

pakoito:
Sounds like a job for `tailrecM` and `traverse` :D

1Jajen1:
Hmm, it does need something other than `foldRight` because thats not gonna work for infinite sequences. `get` needs a left fold implemented using eval (similar to the right fold). Does that exist somewhere already?

pakoito:
so what you want is a `traverse` that starts from the right? hmmmm

1Jajen1:
if we were in `Traversable` then yes, traverse with Option would work just fine, but in foldable not so much. (Also depends on Traversable's implementation for possibly infinite structures)

`get` just needs to return as soon as it hits its index and not continue folding the rest of the structure, which is what it is doing right now. 

Also I just noticed, all first/firstOrNull methods from the stdlib (for sequences) return the first matching element, whereas firstOption with a predicate returns only the first element if it matches. Kinda weird as well.

1Jajen1:
So the only two implementations I can think of to accomplish this, within Foldable, are:
- add a foldLeft variant with Eval
- throw an exception with the result inside foldLeft -> catch it and return it

The first one is definitly better, but forces all foldables to implement yet another method (but maybe we can implement the current foldLeft with the new one to make this less painful. This would also be great for a right fold without eval)

pakoito:
I'd go back to scala cats/scalaz and try to find if there's an equivalent. It rings a bell as a different typeclass tbh. Hoogle the signature maybe?

1Jajen1:
I think it's a problem with missing lazyness, haskell cannot have that issue. Not really familiar with cats/scalaz, but I'll try

1Jajen1:
Current `get` is based on cats, scalaz is similar. Both are not lazy folds, which makes it unusable on infinite sequences sadly. Another, unsatisfying, solution would be to just specialize `get` for `Sequence`. But then again that sucks, because if `get` would be lazy all `Foldables`, finite or not, would benefit from it

1Jajen1:
Closed via #1816

PR #1816, opened 2019-11-25
Override some typeclass methods for better laziness

### LINKED ISSUE #1815: StackOverflow in SequenceK.filter
Using Arrow’s `SequenceK` and the `filter` extension function seems to filter recursively, which causes a StackOverflow when too many consecutive elements don’t satisfy the predicate.
This does not occur with standard Kotlin sequences.

Example:
```kotlin
import arrow.core.extensions.sequencek.functorFilter.filter
import arrow.core.k

fun main() {
  (1..10000).asSequence().k()
      .filter { it > 5000 }
      .first()
}
```
will result in:
```
Exception in thread "main" java.lang.StackOverflowError
	at arrow.core.SequenceK.iterator(SequenceK.kt)
	at arrow.core.SequenceK.iterator(SequenceK.kt)
	at arrow.core.SequenceK.iterator(SequenceK.kt)
	at arrow.core.SequenceK.iterator(SequenceK.kt)
	at arrow.core.SequenceK.iterator(SequenceK.kt)
	at arrow.core.SequenceK.iterator(SequenceK.kt)
	at arrow.core.SequenceK.iterator(SequenceK.kt)
	at arrow.core.SequenceK.iterator(SequenceK.kt)
        ...
```

Using Kotlin 1.3.60, Arrow 10.3, and openJDK 13.

1Jajen1:
Fixes #1815 and #1734 . It does not fix traverse being broken #1749 because that requires a lot more than just overrides in sequence...

### DIFF
--- a/modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/sequenceK.kt
+++ b/modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/sequenceK.kt
@@ -5,17 +5,24 @@ import arrow.core.Either
 import arrow.core.Eval
 import arrow.core.ForSequenceK
 import arrow.core.Ior
+import arrow.core.None
 import arrow.core.Option
 import arrow.core.SequenceK
 import arrow.core.SequenceKOf
 import arrow.core.Tuple2
+import arrow.core.extensions.eval.applicative.applicative
+import arrow.core.extensions.sequence.foldable.firstOption
+import arrow.core.extensions.sequence.foldable.foldLeft
+import arrow.core.extensions.sequence.foldable.foldRight
 import arrow.core.extensions.sequence.foldable.isEmpty
 import arrow.core.extensions.sequence.monadFilter.filterMap
+import arrow.core.extensions.sequencek.foldable.firstOption
 import arrow.core.extensions.sequencek.monad.map
 import arrow.core.extensions.sequencek.monad.monad
 import arrow.core.fix
 import arrow.core.k
 import arrow.core.some
+import arrow.core.toOption
 import arrow.core.toT
 import arrow.extension
 import arrow.typeclasses.Align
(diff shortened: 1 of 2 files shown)


1Jajen1:
I think alternative has some methods as well, but thats something for tomorrow :sleeping:

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/sequenceK.kt
```diff
@@ -71,10 +76,15 @@ interface SequenceKEq<A> : Eq<SequenceK<A>> {
...
-    }
+  /**
+   * This only evaluates up to the first element that differs or to the first element at the index where the other
+   *  sequence is empty
+   */
+  override fun SequenceK<A>.eqv(b: SequenceK<A>): Boolean = object: SequenceKSemialign {}.run {
+    alignWith(this@eqv, b) { ior ->
+      ior.fold({ false }, { false }, { l, r -> EQ().run { l.eqv(r) } })
+    }.firstOption { it.not() }.isEmpty()
+  }
```
1Jajen1: This is kind of concerning, not the old implementation, but the fact that this never failed a test. (Old one should have failed for `[0, 1] != [0]` for example. Will check that later as well.

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/sequenceK.kt
```diff
@@ -71,10 +76,15 @@ interface SequenceKEq<A> : Eq<SequenceK<A>> {
 
   fun EQ(): Eq<A>
 
-  override fun SequenceK<A>.eqv(b: SequenceK<A>): Boolean =
-    zip(b) { aa, bb -> EQ().run { aa.eqv(bb) } }.fold(true) { acc, bool ->
-      acc && bool
-    }
```
1Jajen1: This is kind of concerning, this should have failed tests, for different sized lists. Will check this later on...

1Jajen1: Welp our tests are flawed... Constructor functions like `cf: (Int) -> Kind<F, Int>` are heavily misused, similar to applicative instances. This needs to change because currently the guarantees that our tests provide are very very weak. I'll open a ticket for discussion sometime later today...

1Jajen1:
This should be ready now. Sequences are now as lazy as possible (except for traverse). Not great, but I don't think there is a good solution other than this.

Issue #1747, opened 2019-10-25
Safe zipping using Ior


1Jajen1:
I read a bit about, and used, [these](http://hackage.haskell.org/package/these) in a small project lately and it's a nice improvement over the usual zipping operators.

In arrow `Ior` is isomorphic to `These` from haskell and as far as I can tell also implements the typeclasses the same way.

Given that `Ior` already exists the missing parts are:
- ~~[Align](https://hackage.haskell.org/package/semialign-1.1/docs/Data-Align.html) typeclass and instances~~ Done in #1809 (Thanks @abendt )
- ~~[Semialign](https://hackage.haskell.org/package/semialign-1.1/docs/Data-Semialign.html) typeclass and instances~~ Done in #1800 (Thanks @abendt )
- [Unalign](https://hackage.haskell.org/package/semialign-1.1/docs/Data-Align.html) typeclass and instances
- [Zip/Repeat/Unzip](https://hackage.haskell.org/package/semialign-1.1/docs/Data-Zip.html) typeclasses and instances
- [Crosswalk/Bicrosswalk](https://hackage.haskell.org/package/semialign-1.1/docs/Data-Crosswalk.html) typeclasses and instaces 

I'd be fine with skipping Crosswalk, but the others would be quite useful.

abendt:
@1Jajen1 i also started to look at the Crosswalk implementation and have difficulties to understand the laws definition: `crosswalk (const nil) = const nil` (http://hackage.haskell.org/package/semialign-1.1/docs/Data-Crosswalk.html#v:crosswalk)

i interpret the types like this

left side:
`const nil : a -> f b`
`crosswalk (const nil) : t a -> f (t b)`


right side:
`const nil : a -> fb`



as these types do not seem to match i am most likely misunderstanding somthing here. can you help me understand this law and how the kotlin version should look?

```
forAll(G) { a: Kind<T, A> ->
 val constNil: (A) -> Kind<F, B> = { _: A -> ALIGN.empty<B>() }
    val ls: (Kind<T, A>) -> Kind<F, Kind<T, B>> = { ta: Kind<T, A> -> crosswalk(ALIGN, constNil, ta) }
    val rs: (Kind<T, A>) -> Kind<F, Kind<T, B>> = constNil // wrong type here!

    ls(a).equalUnderTheLaw(rs(a), EQ)
}
```

1Jajen1:
Your types are correct, it becomes a bit easier with explicit foralls:
[...]

PR #1820, opened 2019-11-27
this adds Crosswalk/Bicrosswalk typeclasses with some instances

abendt:
this PR adds the Crosswalk and the Bicrosswalk typeclass together with some instances for Arrow types (#1747)

### DIFF
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Bicrosswalk.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Bicrosswalk.kt
@@ -18,15 +18,15 @@ interface Bicrosswalk<T> : Bifunctor<T>, Bifoldable<T> {
    *
    * Either.bicrosswalk().run {
    *   val either = Either.Right("arrow")
-   *   bicrosswalk(ListK.align(), {ListK.just("fa($it)")}, {ListK.just("fb($it)")}, either)
+   *   bicrosswalk(ListK.align(), either, {ListK.just("fa($it)")}) {ListK.just("fb($it)")}
    * }
    * ```
    */
   fun <F, A, B, C, D> bicrosswalk(
     ALIGN: Align<F>,
+    tab: Kind2<T, A, B>,
     fa: (A) -> Kind<F, C>,
-    fb: (B) -> Kind<F, D>,
-    tab: Kind2<T, A, B>
+    fb: (B) -> Kind<F, D>
   ): Kind<F, Kind2<T, C, D>> =
     bisequenceL(ALIGN, tab.bimap(fa, fb))
 
@@ -41,7 +41,7 @@ interface Bicrosswalk<T> : Bifunctor<T>, Bifoldable<T> {
    *
    * Either.bicrosswalk().run {
    *   val either: Either<ListK<Int>, ListK<String>> = Either.Right(listOf("hello", "arrow").k())
-   *   bisequencek(ListK.align(), either)
+   *   bisequenceL(ListK.align(), either)
    * }
    * ```
    */
@@ -49,5 +49,5 @@ interface Bicrosswalk<T> : Bifunctor<T>, Bifoldable<T> {
     ALIGN: Align<F>,
     tab: Kind2<T, Kind<F, A>, Kind<F, B>>
   ): Kind<F, Kind2<T, A, B>> =
-    bicrosswalk(ALIGN, ::identity, ::identity, tab)
+    bicrosswalk(ALIGN, tab, ::identity, ::identity)
 }
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Bicrosswalk.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Bicrosswalk.kt
@@ -5,6 +5,23 @@ import arrow.Kind2
 import arrow.core.identity
 
 interface Bicrosswalk<T> : Bifunctor<T>, Bifoldable<T> {
+
+  /**
+   * {: data-executable='true'}
+   *
+   * ```kotlin:ank
+   *
+   * import arrow.core.extensions.*
+   * import arrow.core.extensions.bicrosswalk
+   * import arrow.core.extensions.listk.align.align
+   * import arrow.core.*
+   *
+   * Either.bicrosswalk().run {
+   *   val either = Either.Right("arrow")
+   *   bicrosswalk(ListK.align(), {ListK.just("fa($it)")}, {ListK.just("fb($it)")}, either)
+   * }
+   * ```
+   */
   fun <F, A, B, C, D> bicrosswalk(
     ALIGN: Align<F>,
     fa: (A) -> Kind<F, C>,
@@ -13,6 +30,21 @@ interface Bicrosswalk<T> : Bifunctor<T>, Bifoldable<T> {
   ): Kind<F, Kind2<T, C, D>> =
     bisequenceL(ALIGN, tab.bimap(fa, fb))
 
+  /**
+   * {: data-executable='true'}
+   *
+   * ```kotlin:ank
+   * import arrow.core.extensions.*
+   * import arrow.core.extensions.bicrosswalk
+   * import arrow.core.extensions.listk.align.align
+   * import arrow.core.*
+   *
+   * Either.bicrosswalk().run {
+   *   val either: Either<ListK<Int>, ListK<String>> = Either.Right(listOf("hello", "arrow").k())
+   *   bisequencek(ListK.align(), either)
+   * }
+   * ```
+   */
   fun <F, A, B> bisequenceL(
     ALIGN: Align<F>,
     tab: Kind2<T, Kind<F, A>, Kind<F, B>>
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Bicrosswalk.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Bicrosswalk.kt
@@ -12,7 +12,7 @@ interface Bicrosswalk<T> : Bifunctor<T>, Bifoldable<T> {
    * ```kotlin:ank
    *
    * import arrow.core.extensions.*
-   * arrow.core.extensions.either.bicrosswalk.bicrosswalk
+   * import arrow.core.extensions.either.bicrosswalk.bicrosswalk
    * import arrow.core.extensions.listk.align.align
    * import arrow.core.*
    *
@@ -35,7 +35,7 @@ interface Bicrosswalk<T> : Bifunctor<T>, Bifoldable<T> {
    *
    * ```kotlin:ank
    * import arrow.core.extensions.*
-   * arrow.core.extensions.either.bicrosswalk.bicrosswalk
+   * import arrow.core.extensions.either.bicrosswalk.bicrosswalk
    * import arrow.core.extensions.listk.align.align
    * import arrow.core.*
    *
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Bicrosswalk.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Bicrosswalk.kt
@@ -12,7 +12,7 @@ interface Bicrosswalk<T> : Bifunctor<T>, Bifoldable<T> {
    * ```kotlin:ank
    *
    * import arrow.core.extensions.*
-   * import arrow.core.extensions.bicrosswalk
+   * arrow.core.extensions.either.bicrosswalk.bicrosswalk
    * import arrow.core.extensions.listk.align.align
    * import arrow.core.*
    *
@@ -35,7 +35,7 @@ interface Bicrosswalk<T> : Bifunctor<T>, Bifoldable<T> {
    *
    * ```kotlin:ank
    * import arrow.core.extensions.*
-   * import arrow.core.extensions.bicrosswalk
+   * arrow.core.extensions.either.bicrosswalk.bicrosswalk
    * import arrow.core.extensions.listk.align.align
    * import arrow.core.*
    *
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Bicrosswalk.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Bicrosswalk.kt
@@ -1,0 +1,21 @@
+package arrow.typeclasses
+
+import arrow.Kind
+import arrow.Kind2
+import arrow.core.identity
+
+interface Bicrosswalk<T> : Bifunctor<T>, Bifoldable<T> {
+  fun <F, A, B, C, D> bicrosswalk(
+    ALIGN: Align<F>,
+    fa: (A) -> Kind<F, C>,
+    fb: (B) -> Kind<F, D>,
+    tab: Kind2<T, A, B>
+  ): Kind<F, Kind2<T, C, D>> =
+    bisequenceL(ALIGN, tab.bimap(fa, fb))
+
+  fun <F, A, B> bisequenceL(
+    ALIGN: Align<F>,
+    tab: Kind2<T, Kind<F, A>, Kind<F, B>>
+  ): Kind<F, Kind2<T, A, B>> =
+    bicrosswalk(ALIGN, ::identity, ::identity, tab)
+}
--- a/modules/core/arrow-core-data/src/test/kotlin/arrow/core/IorTest.kt
+++ b/modules/core/arrow-core-data/src/test/kotlin/arrow/core/IorTest.kt
@@ -1,6 +1,7 @@
 package arrow.core
 
 import arrow.Kind
+import arrow.Kind2
 import arrow.core.Ior.Right
 import arrow.core.extensions.eq
 import arrow.core.extensions.hash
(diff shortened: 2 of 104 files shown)


modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Bicrosswalk.kt
```diff
@@ -0,0 +1,53 @@
+package arrow.typeclasses
+
+import arrow.Kind
+import arrow.Kind2
+import arrow.core.identity
+
+interface Bicrosswalk<T> : Bifunctor<T>, Bifoldable<T> {
```
1Jajen1: "Unfortunately it is too perfect an abstraction to be useful. - elliott" I love this quote :D
https://teh.id.au/posts/2017/03/29/these-align-crosswalk/

I wonder when the first time comes around where I actually have a legitimate use of this typeclass :)
Either way it's not wrong having it anyway.

modules/core/arrow-core-data/src/test/kotlin/arrow/core/IorTest.kt
```diff
@@ -140,3 +149,16 @@ class IorTest : UnitSpec() {
...
+    override fun constants(): Iterable<Kind<ForIor, A>> =
+      gen.constants().map {
+        Ior.Both(it, it) as Kind<ForIor, A>
+      }
+
+    override fun random(): Sequence<Kind<ForIor, A>> =
+      gen.random().map {
+        Ior.Both(it, it) as Kind<ForIor, A>
+      }
+  }
```
1Jajen1: This is not a good generator. A good generator should cover all cases of `Ior` and thus require two generators as input. But tbh wait with this untill the zip pr is merged and just use `LiftGen`

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/either.kt
```diff
@@ -226,3 +233,34 @@ interface EitherHash<L, R> : Hash<Either<L, R>>, EitherEq<L, R> {
...
+        fa(e.a).map { it.left() }
+      }
+      is Either.Right -> ALIGN.run {
+        fb(e.b).map { it.right() }
+      }
+    }
+}
+
+fun Either.Companion.bicrosswalk() = object : EitherBicrosswalk {
+}
```
1Jajen1: Does `@extension` not work here? :/

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/ior.kt
```diff
@@ -169,3 +176,46 @@ interface IorHash<L, R> : Hash<Ior<L, R>>, IorEq<L, R> {
...
+      is Ior.Right -> ALIGN.run {
+        fb(e.value).map { it.rightIor() }
+      }
+      is Ior.Both -> ALIGN.run {
+        align(fa(e.leftValue), fb(e.rightValue))
+      }
+    }
+}
+
+fun Ior.Companion.bicrosswalk() = object : IorBicrosswalk {}
```
1Jajen1: `@extension`? Ignore this if it does not work, but if it does it's usually much better as it generates methods as well.

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
```diff
@@ -339,3 +340,33 @@ interface ListKUnalign : Unalign<ForListK>, ListKSemialign {
...
+        ALIGN.run {
+          alignWith(ls, rs) { ior: Ior<B, Kind<ForListK, B>> ->
+            when (ior) {
+              is Ior.Left -> ListK.just(ior.value)
+              is Ior.Right -> ior.value
+              is Ior.Both -> ListK.just(ior.leftValue) + ior.rightValue.fix()
+            }
+          }
+        }
+      }
```
1Jajen1: This looks wrong. Not in the sense that the implementation might be wrong, but I feel like there has to be a better/cleaner way to write this. I'll take a look at this later with a bit more time :)

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/sequenceK.kt
```diff
@@ -332,3 +335,35 @@ interface SequenceKUnalign : Unalign<ForSequenceK>, SequenceKSemialign {
...
+                is Ior.Right -> ior.value
+                is Ior.Both -> (SequenceK.just(ior.leftValue) + ior.rightValue.fix().sequence).k()
+              }
+          }
+        }
+      } else {
+        ALIGN.run { empty<B>().map { SequenceK.empty<B>() } }
+      }
+
+    return crosswalkSequence(a.fix().sequence.iterator())
```
1Jajen1: Similar to the list one, there must be a nicer way to write this :/

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BicrosswalkLaws.kt
```diff
@@ -0,0 +1,76 @@
...
+    EQ: Eq<Kind2<T, String, String>>
+  ): List<Law> {
+
+    val funGen = object : Gen<(Int) -> Kind<ForListK, String>> {
+      override fun constants(): Iterable<(Int) -> ListK<String>> = listOf(
+        { _: Int -> ListK.empty<String>() }, { _: Int -> ListK.just("value") }
+      )
+
+      override fun random(): Sequence<(Int) -> ListK<String>> = generateSequence({ int: Int -> List(abs(int % 1000)) { "$it" }.k() }) { it }
+    }
```
1Jajen1: I think there are primitive function generators in our tests already :) Although those are literally worthless imo, as the author of a property testing library I might be biased tho :)

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/bicrosswalk/README.md
```diff
@@ -0,0 +1,76 @@
+---
+layout: docs
+title: Bicrosswalk
+permalink: /docs/arrow/typeclasses/bicrosswalk/
+redirect_from:
+  - /docs/typeclasses/bicrosswalk/
+---
+
+## Align
```
1Jajen1:
```suggestion
## Bicrosswalk
```

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/crosswalk/README.md
```diff
@@ -0,0 +1,80 @@
+---
+layout: docs
+title: Crosswalk
+permalink: /docs/arrow/typeclasses/crosswalk/
+redirect_from:
+  - /docs/typeclasses/crosswalk/
+---
+
+## Align
```
1Jajen1:
```suggestion
## Crosswalk
```

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/crosswalk/README.md
```diff
@@ -0,0 +1,80 @@
...
+### Data types
+
+```kotlin:ank:replace
+import arrow.reflect.*
+import arrow.typeclasses.Crosswalk
+
+TypeClass(Crosswalk::class).dtMarkdownList()
+```
+
+ank_macro_hierarchy(arrow.typeclasses.Align)
```
1Jajen1:
```suggestion
ank_macro_hierarchy(arrow.typeclasses.Crosswalk)
```

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Bicrosswalk.kt
abendt: yes, i also came across that article :-D

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/either.kt
abendt: my impression was that the generated code did not compile for Kind2 types. i think i tried to use `@extension` first i am not sure if in this particluar instance. will recheck later ...

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
abendt: looking forward for it!

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BicrosswalkLaws.kt
abendt: what lib are your the author of?

1Jajen1: https://github.com/1Jajen1/propCheck
It's undergoing quite the rewrite tho. (the pr for 1.0)The current one is good, but the rewrite aims to make it easier to use along with giving it more power. This is arguably better at property based testing than kotlin test if you only consider property testing. I plan on having the rewrite done by 2020 with better docs etc. But that may be delayed depending on work and/university :/

modules/core/arrow-core-data/src/test/kotlin/arrow/core/IorTest.kt
abendt: true, i went ahead and changed it. the code later can be moved to GenK

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/either.kt
abendt: you were right it worked. think the problem was not with Kind2 but when one type had a bound (`K: Comparable<K>` with SortedMapK). i have changed it

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/ior.kt
abendt: same

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
abendt: i would be really curious here to hear what you are looking for in a good FP implementation.as, coming from an OO background this would be really interesting for me to hear. so it would be great if you could share a bit of your thought process regarding the current implementation and the enhanced one.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BicrosswalkLaws.kt
abendt: looks interesting. i don't have much experience yet with property based testing. 
do you plan to introduce propCheck to the arrow codebase?

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
1Jajen1: Main motivation is usually just to avoid _magic_ as much as possible and work on pure values.
That usually just involves avoiding language operators like `if`, loops, `return`, `throw` as much as possible and stick to `fold` functions and pattern matching over `when` whenever possible. `if` can have good uses, but the other constructs should be avoided at all cost. There is nothing inherently wrong from a functional setting with this code. It just feels overly verbose and is build from lots of small operations that feel like they should have some more general primitve that makes this code more concise.

1Jajen1: That is for low level code, higher level design of functional programs is a bit more and also very much opinion based. Usually sticking to pure functions, functions over data and no-magic is a good route forward.

1Jajen1: I kind of wanna write more about design in functional programming, but so much of what I want to do is only possible in haskell :/ I usually only use kotlin when I am bound to the jvm by previous work or by libraries that I have to use...

1Jajen1: > I kind of wanna write more about design in functional programming, but so much of what I want to do is only possible in haskell :/ I usually only use kotlin when I am bound to the jvm by previous work or by libraries that I have to use...

Which is most of the time at work for me, so I still use it a lot :)

abendt: thanks for the insight! i think i will learn a bit more about haskell :-)

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BicrosswalkLaws.kt
1Jajen1: I'd love to, but don't have the time for it atm. Raul and I were talking about this already half a year ago, but I had an annoying hospital stay + time to recover and catch up with work and university made that impossible.

Maybe another attempt next year when the rewrite is done, but that also depends on what the other maintainers think.

abendt: sorry to hear that about your hospital stay. i hope you are fully recovered! i will definitely check out propCheck.

1Jajen1: Yeah all good now :)
Property based testing is **amazing** if you can get it right, some really really helpful resources: https://williamyaoh.com/posts/2019-11-24-design-and-testing-articles.html
Everything related to property based testing on the testing part is great, especially the ones by Oskar Wickström

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
```diff
@@ -358,3 +360,32 @@ interface ListKUnzip : Unzip<ForListK>, ListKZip {
...
+
+        ALIGN.run {
+          alignWith(ls, rs) { ior ->
+            ior.fold(
+              { ListK.just(it) },
+              ::identity,
+              { l, r -> ListK.just(l) + r.fix() })
+          }
+        }
+      }
```
1Jajen1: This is a fold so it should be represented by something like this: `list.foldLeft(empty<ListK<B>>()) { xs, x -> alignWith(fa(x), xs) { ... } }`

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/sequenceK.kt
```diff
@@ -372,3 +376,33 @@ interface SequenceKUnzip : Unzip<ForSequenceK>, SequenceKZip {
...
+              ::identity,
+              { l, r -> (SequenceK.just(l) + r.fix().sequence).k() })
+          }
+        }
+      } else {
+        ALIGN.run { empty<B>().map { SequenceK.empty<B>() } }
+      }
+
+    return crosswalkSequence(a.fix().sequence.iterator())
+  }
```
1Jajen1: Same as with the list. This cannot be fully lazy anyway so it might as well use a fold.

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/bicrosswalk/README.md
```diff
@@ -0,0 +1,76 @@
...
+Arrow provides [`BicrosswalkLaws`][functor_laws_source]{:target="_blank"} in the form of test cases for internal verification of lawful instances and third party apps creating their own Bicrosswalk instances.
+
+#### Creating your own `Bicrosswalk` instances
+
+Arrow already provides Bicrosswalk instances for common datatypes (e.g. Either, Ior). See their implementations
+and accompanying testcases for reference.
+
+See [Deriving and creating custom typeclass]({{ '/docs/patterns/glossary' | relative_url }})
+
+Additionally all instances of [`Bicrosswalk`]({{ '/docs/arrow/typeclasses/bicrosswalk' | relative_url }}) implement the `Bifucntor` and `Bifoldable` typeclass directly.
```
1Jajen1:
```suggestion
Additionally all instances of [`Bicrosswalk`]({{ '/docs/arrow/typeclasses/bicrosswalk' | relative_url }}) implement the `Bifunctor` and `Bifoldable` typeclass directly.
```

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/bicrosswalk/README.md
```diff
@@ -0,0 +1,76 @@
...
+layout: docs
+title: Bicrosswalk
+permalink: /docs/arrow/typeclasses/bicrosswalk/
+redirect_from:
+  - /docs/typeclasses/bicrosswalk/
+---
+
+## Bicrosswalk
+
+{:.beginner}
```
1Jajen1: This is very optimistic ;) Probably bump this up a bit

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/crosswalk/README.md
```diff
@@ -0,0 +1,80 @@
...
+layout: docs
+title: Crosswalk
+permalink: /docs/arrow/typeclasses/crosswalk/
+redirect_from:
+  - /docs/typeclasses/crosswalk/
+---
+
+## Crosswalk
+
+{:.beginner}
```
1Jajen1: Same as for Bicrosswalk, this is not an easy abstraction by any means

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
abendt: ok, changed it

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/sequenceK.kt
abendt: ok, it uses the list implementation now

1Jajen1:
Ok I think this is also good to go. This also finishes #1747 :tada: :tada: :tada: 
Thanks so much for doing this, the align/zip hierarchy is so useful to have! (Not so sure about crosswalk, but I will definitly write a small post about it when I find a good use :joy: and it's good to have for completeness)

abendt:
@1Jajen1 happy to help and i learned some new stuff. thanks a lot for your support! looking forward to the post about crosswalk :-D

PR #1831, opened 2019-12-05
Specialized aligns

abendt:
this PR adds some specialized align functions that build on top of Semialign (#1747)

### DIFF
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/core/ListK.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/core/ListK.kt
@@ -173,6 +173,65 @@ data class ListK<out A>(private val list: List<A>) : ListKOf<A>, List<A> by list
 
   override fun hashCode(): Int = list.hashCode()
 
+  /**
+   * Align two Lists as in zip, but filling in blanks with None.
+   */
+  fun <B> padZip(
+    other: ListK<B>
+  ): ListK<Tuple2<Option<A>, Option<B>>> =
+    alignWith(this, other) { ior ->
+      ior.bimap({ Option.just(it) }, { Option.just(it) })
+        .fold(
+          { it toT Option.empty<B>() },
+          { Option.empty<A>() toT it },
+          { a, b -> a toT b })
+    }
+
+  /**
+   * Align two Lists as in zipWith, but filling in blanks with None.
+   */
+  fun <B, C> padZipWith(
+    other: ListK<B>,
+    fa: (Option<A>, Option<B>) -> C
+  ): ListK<C> =
+    padZip(other).map { fa(it.a, it.b) }
+
+  /**
+   * Left-padded zipWith.
+   */
+  fun <B, C> lpadZipWith(
+    other: ListK<B>,
+    fab: (Option<A>, B) -> C
+  ): ListK<C> =
+      padZipWith(other) { a, b -> b.map { fab(a, it) } }.filterMap(::identity)
+
+  /**
+   * Left-padded zip.
+   */
+  fun <B> lpadZip(
+    other: ListK<B>
+  ): ListK<Tuple2<Option<A>, B>> =
+    this.lpadZipWith(other) { a, b -> a toT b }
+
+  /**
+   * Right-padded zipWith.
+   */
+  fun <B, C> rpadZipWith(
+    other: ListK<B>,
+    fa: (A, Option<B>) -> C
+  ): ListK<C> =
+    other.lpadZipWith(this) { a, b -> fa(b, a) }
+
+  /**
+   * Right-padded zip.
+   */
+  fun <B> rpadZip(
+    other: ListK<B>
+  ): ListK<Tuple2<A, Option<B>>> =
+    this.rpadZipWith(other) { a, b ->
+      a toT b
+    }
+
   companion object {
 
     fun <A> just(a: A): ListK<A> = listOf(a).k()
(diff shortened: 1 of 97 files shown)


modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Semialign.kt
```diff
@@ -55,4 +58,40 @@ interface Semialign<F> : Functor<F> {
...
+   * Align two structures as in zip, but filling in blanks with None.
+   */
+  fun <A, B> Kind<F, A>.padZip(
+    other: Kind<F, B>
+  ): Kind<F, Tuple2<Option<A>, Option<B>>> =
+    alignWith(this, other) { ior ->
+      when (val e = ior.bimap({ Option.just(it) }, { Option.just(it) })) {
+        is Ior.Right -> Option.empty<A>() toT e.value
+        is Ior.Left -> e.value toT Option.empty()
+        is Ior.Both -> e.leftValue toT e.rightValue
```
1Jajen1: This would be cleaner with `Ior.fold`.

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
```diff
@@ -321,6 +323,44 @@ interface ListKSemialign : Semialign<ForListK>, ListKFunctor {
   }
 }
 
+/**
+ * Left-padded zipWith.
+ */
+fun <A, B, C> ListK<A>.lpadZipWith(
```
1Jajen1: I understand why this has to be in this module, I am just not sure if this is the right file for this. I am also not aware of any other methods like this on other datatypes. @pakoito Is this in the right place?

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/sequenceK.kt
```diff
@@ -317,6 +318,11 @@ interface SequenceKSemialign : Semialign<ForSequenceK>, SequenceKFunctor {
     }.k()
 }
 
+fun <A, B, C> SequenceK<A>.alignSequenceWith(other: SequenceK<B>, fab: (Ior<A, B>) -> C): SequenceK<C> =
+  SequenceK.semialign().run {
+    alignWith(this@alignSequenceWith, other, fab)
+  }.fix()
+
```
1Jajen1: What is this needed for? Is this just an specialisation for the semialign alignWith?

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/listk.kt
pakoito: I'd put it in the original ListK file tbh, unless there's an import issue

1Jajen1: Well yes, it uses methods from the semialign instance. I think the current workaround is to specialise all methods needed to list and implement the instance using those. This means moving the implementation of all of those methods to `ListK`. So `padZipWith`, `padZip` and `alignWith` and while at it probably also `align` itself.

abendt: ok. i have moved the functions + align to ListK

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/sequenceK.kt
abendt: i have removed it

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Semialign.kt
1Jajen1: I think this is still open and after that this pr is good to merge

abendt: true, i missed that ...

modules/core/arrow-core-data/src/main/kotlin/arrow/core/ListK.kt
```diff
@@ -173,6 +173,65 @@ data class ListK<out A>(private val list: List<A>) : ListKOf<A>, List<A> by list
...
+  fun <B> padZip(
+    other: ListK<B>
+  ): ListK<Tuple2<Option<A>, Option<B>>> =
+    alignWith(this, other) { ior ->
+      ior.bimap({ Option.just(it) }, { Option.just(it) })
+        .fold(
+          { it toT Option.empty<B>() },
+          { Option.empty<A>() toT it },
+          { a, b -> a toT b })
+    }
```
1Jajen1: This is not quite what I meant.
```kotlin
ior.fold({ it.some() toT None }, { None toT it.some() }, { l, r -> l.some() toT r.some() })
```
Does the same thing and skips the call to bimap. For a small function like this I think this is better, especially if the fold is needed anyway.

abendt: true, this is more concise

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Semialign.kt
```diff
@@ -55,4 +58,40 @@ interface Semialign<F> : Functor<F> {
...
+  fun <A, B> Kind<F, A>.padZip(
+    other: Kind<F, B>
+  ): Kind<F, Tuple2<Option<A>, Option<B>>> =
+    alignWith(this, other) { ior ->
+      ior.bimap({ Option.just(it) }, { Option.just(it) }).fold(
+        { it toT Option.empty<B>() },
+        { Option.empty<A>() toT it },
+        ::Tuple2
+      )
+    }
```
1Jajen1: I forgot to mark this one, let's change this as well for parity and then merge this (this duplication will all be gone when we have arrow-meta so it doesn't matter atm)

PR #1832, opened 2019-12-08
Adds Sliding, Dropping and Unbounded strategies to arrow.fx.Queue

scott-christopher:
This change adds Sliding, Dropping and Unbounded strategies to the `arrow.fx.Queue` implementation.

**Sliding** - *acts as a sliding window when at capacity, dropping oldest queue values to make room for newly offered values*

**Dropping** - *simply drops offered values if the queue is already at capacity*

**Unbounded** - *has no notion of capacity and will happily exhaust all memory of the runtime if given the opportunity*

Please be mindful that I am somewhat new to both Arrow and Kotlin, so I would welcome any feedback, especially regarding conventions :)

### DIFF
--- a/modules/docs/arrow-docs/docs/docs/arrow-fx/queue/README.md
+++ b/modules/docs/arrow-docs/docs/docs/arrow-fx/queue/README.md
@@ -1,0 +1,213 @@
+---
+layout: docs
+title: Queue
+permalink: /docs/effects/queue/
+---
+
+## Queue
+
+{:.intermediate}
+intermediate
+
+A `Queue` is a lightweight, asynchronous first-in-first-out queue for holding arbitrary values within a `Concurrent`
+context. There are two primary means of working with an instance of a `Queue`: placing items via `offer` and removing
+items via `take`.
+
+```kotlin:ank:playground
+import arrow.fx.*
+import arrow.fx.extensions.fx
+import arrow.fx.extensions.io.concurrent.concurrent
+
+fun main(args: Array<String>) {
+  val result =
+  //sampleStart
+    IO.fx {
+      val q = !Queue.bounded<ForIO, Int>(10, IO.concurrent())
+      !q.offer(42)
+      !q.take()
+    }
+  //sampleEnd
+  println(result.unsafeRunSync())
+}
+```
+
+Attempting to take a value from an empty `Queue` will cause the calling fiber to be suspended and will resume upon a new
+value being placed on the `Queue`.
+
+```kotlin:ank:playground
+import arrow.fx.*
+import arrow.fx.extensions.fx
+import arrow.fx.extensions.io.concurrent.concurrent
+
+fun main(args: Array<String>) {
+  val result =
+  //sampleStart
+    IO.fx {
+      val q = !Queue.bounded<ForIO, Int>(10, IO.concurrent())
+      val waiting = !q.take().fork()
+      !q.offer(42)
+      !waiting.join()
+    }
+  //sampleEnd
+  println(result.unsafeRunSync())
+}
+```
+
+### Construction and capacity strategies
+
+When constructing a new `Queue` there are four different options to choose from which behave differently with respect to
+overflowing the configured capacity of the `Queue`:
+
+ * **Bounded**: Offering to a `bounded` queue at capacity will cause the fiber making the call to be suspended until the
+ queue has space to receive the offer value.
+
+```kotlin:ank:playground
+import arrow.fx.*
+import arrow.fx.extensions.fx
+import arrow.fx.extensions.io.concurrent.concurrent
+
+fun main(args: Array<String>) {
+  val result =
+  //sampleStart
+    IO.fx {
+      val capacity = 2
+      val q = !Queue.bounded<ForIO, Int>(capacity, IO.concurrent())
+      !q.offer(42)
+      !q.offer(43)
+      !q.offer(44).fork() // <-- This `offer` exceeds the capacity and will be suspended
+      val fortyTwo   = !q.take()
+      val fortyThree = !q.take()
+      val fortyFour  = !q.take()
+      listOf(fortyTwo, fortyThree, fortyFour)
+    }
+  //sampleEnd
+  println(result.unsafeRunSync())
+}
+```
+
+ * **Dropping**: Offering to a `dropping` queue at capacity will cause the offered value to be discarded.
+
+```kotlin:ank:playground
+import arrow.fx.*
+import arrow.fx.extensions.fx
+import arrow.fx.extensions.io.concurrent.concurrent
+
+fun main(args: Array<String>) {
+  val result =
+  //sampleStart
+    IO.fx {
+      val capacity = 2
+      val q = !Queue.dropping<ForIO, Int>(capacity, IO.concurrent())
+      !q.offer(42)
+      !q.offer(43)
+      !q.offer(44) // <-- This `offer` exceeds the capacity and will be dropped immediately
+      val fortyTwo   = !q.take()
+      val fortyThree = !q.take()
+      !q.offer(45)
+      val fortyFive  = !q.take()
+      listOf(fortyTwo, fortyThree, fortyFive)
+    }
+  //sampleEnd
+  println(result.unsafeRunSync())
+}
+```
+
+ * **Sliding**: Offering to a `sliding` queue at capacity will cause the oldest value at the front of the queue to be
+discarded, making room for the offered value. n.b. A `sliding` queue must have a capacity of at least 1.
+
+```kotlin:ank:playground
+import arrow.fx.*
+import arrow.fx.extensions.fx
+import arrow.fx.extensions.io.concurrent.concurrent
+
+fun main(args: Array<String>) {
+  val result =
+  //sampleStart
+    IO.fx {
+      val capacity = 2
+      val q = !Queue.sliding<ForIO, Int>(capacity, IO.concurrent())
+      !q.offer(42)
+      !q.offer(43)
+      !q.offer(44) // <-- This `offer` exceeds the capacity, causing the oldest value to be removed
+      val fortyThree = !q.take()
+      val fortyFour  = !q.take()
+      !q.offer(45)
+      val fortyFive  = !q.take()
+      listOf(fortyThree, fortyFour, fortyFive)
+    }
+  //sampleEnd
+  println(result.unsafeRunSync())
+}
+```
+
+ * **Unbounded**: An `unbounded` queue has no notion of capacity and is bound only by exhausting the memory limits of
+the runtime.
+
+```kotlin:ank:playground
+import arrow.fx.*
+import arrow.fx.extensions.fx
+import arrow.fx.extensions.io.concurrent.concurrent
+
+fun main(args: Array<String>) {
+  val result =
+  //sampleStart
+    IO.fx {
+      val q = !Queue.unbounded<ForIO, Int>(IO.concurrent())
+      !q.offer(42)
+      // ...
+      !q.offer(42000000)
+      !q.take()
+    }
+  //sampleEnd
+  println(result.unsafeRunSync())
+}
+```
+
+### Shutting down
+
+A `Queue` also has the ability to be `shutdown`, interrupting any future calls to `take` or `offer` with a
+`QueueShutdown` error and cancelling any suspended fibers waiting to `take` or `offer`.
+
+```kotlin:ank:playground
+import arrow.fx.*
+import arrow.fx.extensions.fx
+import arrow.fx.extensions.io.concurrent.concurrent
+
+fun main(args: Array<String>) {
+  val result =
+  //sampleStart
+    IO.fx {
+      val q = !Queue.bounded<ForIO, Int>(10, IO.concurrent())
+      val t = !q.take().fork()
+      !q.shutdown()
+      !t.join() // Attempting to `join` after `shutdown` results in a `QueueShutdown` error
+    }
+  //sampleEnd
+  println(result.attempt().unsafeRunSync())
+}
+```
+
+Consumers of the `Queue` can also track the event of a shutdown by calling `awaitShutdown` to receive a suspended
+`Concurrent<F>` that will resume once the `Queue` has been shutdown.
+
+```kotlin:ank:playground
+import arrow.fx.*
+import arrow.fx.extensions.fx
+import arrow.fx.extensions.io.concurrent.concurrent
+
+fun main(args: Array<String>) {
+  val result =
+  //sampleStart
+    IO.fx {
+      val q = !Queue.bounded<ForIO, Int>(10, IO.concurrent())
+      val onShutdown = !q.awaitShutdown().fork()
+      !q.offer(42)
+      val fortyTwo = !q.take()
+      !q.shutdown()
+      !onShutdown.join()
+      fortyTwo
+    }
+  //sampleEnd
+  println(result.unsafeRunSync())
+}
+```
--- a/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
+++ b/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
@@ -13,10 +13,20 @@ import arrow.typeclasses.ApplicativeError
  * Lightweight, asynchronous queue for values of A in a Concurrent context F
  * A Queue can be implemented using 4 different strategies
  *
- * Bounded: Offering to a queue at capacity will cause the fiber making the call
- * to be suspended until the queue has space to receive the offer value
+ * Bounded: Offering to a bounded queue at capacity will cause the fiber making
+ * the call to be suspended until the queue has space to receive the offer value
  *
- * ported from Scala ZIO Queue implementation
+ * Dropping: Offering to a dropping queue at capacity will cause the offered
+ * value to be discarded
+ *
+ * Sliding: Offering to a sliding queue at capacity will cause the value at the
+ * front of the queue to be discarded to make room for the offered value
+ *
+ * Unbounded: An unbounded queue has no notion of capacity and is bound only by
+ * exhausting the memory limits of the runtime
+ *
+ * ported from [Scala ZIO Queue](https://zio.dev/docs/datatypes/datatypes_queue)
+ * implementation
  */
 class Queue<F, A> private constructor(private val strategy: SurplusStrategy<F, A>, private val ref: Ref<F, State<F, A>>, private val CF: Concurrent<F>) :
   Concurrent<F> by CF {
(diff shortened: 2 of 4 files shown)


modules/fx/arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -198,6 +208,61 @@ class Queue<F, A> private constructor(private val capacity: Int, private val ref
...
+
+    internal data class Shutdown<F>(val AE: ApplicativeError<F, Throwable>) : State<F, Nothing>() {
+      override fun size(): Kind<F, Int> = AE.raiseError(QueueShutdown)
+    }
+  }
+
+  private sealed class SurplusStrategy<F, A> {
+    abstract fun handleSurplus(p: Promise<F, Unit>, surplus: State.Surplus<F, A>, a: A): Tuple2<Kind<F, Unit>, State<F, A>>
+
+    internal data class Blocking<F, A>(val capacity: Int, val AP: Applicative<F>) : SurplusStrategy<F, A>() {
```
nomisRev: All these `internal` modifiers are unnecessary since the sealed class is `private` all its sub types also are also `private`.

```suggestion
    data class Blocking<F, A>(val capacity: Int, val AP: Applicative<F>) : SurplusStrategy<F, A>() {
```

nomisRev: We should probably name this the same as the constructor like the others.

```suggestion
    internal data class Bounded<F, A>(val capacity: Int, val AP: Applicative<F>) : SurplusStrategy<F, A>() {
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -198,6 +208,61 @@ class Queue<F, A> private constructor(private val capacity: Int, private val ref
         ifShutdown = ::identity
       )
     }
+
+  private sealed class State<F, out A> {
+    abstract fun size(): Kind<F, Int>
+
+    internal data class Deficit<F, A>(val takers: IQueue<Promise<F, A>>, val AP: Applicative<F>, val shutdownHook: Kind<F, Unit>) : State<F, A>() {
```
nomisRev: Same here. I missed this last time.
```suggestion
    data class Deficit<F, A>(val takers: IQueue<Promise<F, A>>, val AP: Applicative<F>, val shutdownHook: Kind<F, Unit>) : State<F, A>() {
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -198,6 +208,61 @@ class Queue<F, A> private constructor(private val capacity: Int, private val ref
...
+      override fun size(): Kind<F, Int> = AP.just(-takers.length())
+    }
+
+    internal data class Surplus<F, A>(val queue: IQueue<A>, val putters: IQueue<Tuple2<A, Promise<F, Unit>>>, val AP: Applicative<F>, val shutdownHook: Kind<F, Unit>) : State<F, A>() {
```
nomisRev:
```suggestion
    data class Surplus<F, A>(val queue: IQueue<A>, val putters: IQueue<Tuple2<A, Promise<F, Unit>>>, val AP: Applicative<F>, val shutdownHook: Kind<F, Unit>) : State<F, A>() {
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -198,6 +208,61 @@ class Queue<F, A> private constructor(private val capacity: Int, private val ref
...
+      override fun size(): Kind<F, Int> = AP.just(queue.length() + putters.length())
+    }
+
+    internal data class Shutdown<F>(val AE: ApplicativeError<F, Throwable>) : State<F, Nothing>() {
```
nomisRev:
```suggestion
    data class Shutdown<F>(val AE: ApplicativeError<F, Throwable>) : State<F, Nothing>() {
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -198,6 +208,61 @@ class Queue<F, A> private constructor(private val capacity: Int, private val ref
...
+      override fun handleSurplus(p: Promise<F, Unit>, surplus: State.Surplus<F, A>, a: A): Tuple2<Kind<F, Unit>, State<F, A>> =
+        surplus.run {
+          if (queue.length() < capacity && putters.isEmpty())
+            p.complete(Unit) toT copy(queue = queue.enqueue(a))
+          else
+            AP.unit() toT copy(putters = putters.enqueue(a toT p))
+        }
+    }
+
+    internal data class Sliding<F, A>(val capacity: Int, val AP: Applicative<F>) : SurplusStrategy<F, A>() {
```
nomisRev:
```suggestion
    data class Sliding<F, A>(val capacity: Int, val AP: Applicative<F>) : SurplusStrategy<F, A>() {
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -198,6 +208,61 @@ class Queue<F, A> private constructor(private val capacity: Int, private val ref
...
+      override fun handleSurplus(p: Promise<F, Unit>, surplus: State.Surplus<F, A>, a: A): Tuple2<Kind<F, Unit>, State<F, A>> =
+        surplus.run {
+          val nextQueue =
+            if (queue.length() < capacity) queue.enqueue(a)
+            else queue.dequeue().b.enqueue(a)
+          p.complete(Unit) toT copy(queue = nextQueue)
+        }
+    }
+
+    internal data class Dropping<F, A>(val capacity: Int, val AP: Applicative<F>) : SurplusStrategy<F, A>() {
```
nomisRev:
```suggestion
    data class Dropping<F, A>(val capacity: Int, val AP: Applicative<F>) : SurplusStrategy<F, A>() {
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -198,6 +208,61 @@ class Queue<F, A> private constructor(private val capacity: Int, private val ref
...
+      override fun handleSurplus(p: Promise<F, Unit>, surplus: State.Surplus<F, A>, a: A): Tuple2<Kind<F, Unit>, State<F, A>> =
+        surplus.run {
+          val nextQueue = if (queue.length() < capacity) queue.enqueue(a) else queue
+          p.complete(Unit) toT copy(queue = nextQueue)
+        }
+    }
+
+    internal data class Unbounded<F, A>(val AP: Applicative<F>) : SurplusStrategy<F, A>() {
```
nomisRev:
```suggestion
    data class Unbounded<F, A>(val AP: Applicative<F>) : SurplusStrategy<F, A>() {
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -165,7 +151,31 @@ class Queue<F, A> private constructor(private val capacity: Int, private val ref
      */
     fun <F, A> bounded(capacity: Int, CF: Concurrent<F>): Kind<F, Queue<F, A>> = CF.run {
       Ref<State<F, A>>(State.Surplus(IQueue.empty(), IQueue.empty(), this, unit())).map {
-        Queue(capacity, it, this)
+        Queue(SurplusStrategy.Blocking(capacity, this), it, this)
```
nomisRev:
```suggestion
        Queue(SurplusStrategy.Bounded(capacity, this), it, this)
```

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
```diff
@@ -214,7 +163,149 @@ class QueueTest : UnitSpec() {
       }
     }
 
-    tests("BoundedQueue", queue =
+    fun boundedStrategyTests(
+      ctx: CoroutineContext = Dispatchers.Default,
+      queue: (Int) -> IO<Queue<ForIO, Int>>
+    ) {
```
nomisRev: You can call `allStrategyTests` within `boundedStrategyTests` and that will ensure that those test run for `boundedStrategyTests` to pass. That also reduces calls sites.

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
```diff
@@ -24,20 +27,20 @@ class QueueTest : UnitSpec() {
 
   init {
 
-    fun tests(
+    fun allStrategyTests(
       label: String,
       ctx: CoroutineContext = Dispatchers.Default,
```
nomisRev: We should probably use `IO.dispatchers().default()` instead here.

scott-christopher:
Apologies for the radio silence on this, I've been slammed with work for the past couple of weeks. I've got the requested changes ready along with some docs, however I've encountered some issues along the way.

1. I figured that it still makes sense to allow a dropping strategy with 0 capacity, effectively dropping all messages unless a taker was waiting. Unfortunately this introduces a race condition in one of the tests that I'm not currently aware how to resolve. The test has been disabled to prevent it locking up the CI builds until solved. Introducing a sleep or throwing a number of `offer` calls after the `take` prevents it from hanging so I suspect that the offer is executing before the take, leaving it hanging waiting for a second offer that will never arrive. Is there a way to guarantee that the `taker` fiber above has suspended before attempting to `offer`?

2. I was having issues with `runAnk` which would hang when trying to evaluate any of the `offer` or `take` calls in the docs running as `kotlin:ank` snippets. I couldn't determine why, so I have wrapped them in a `main` function following the patterns of the other `kotlin:ank:playground` snippets. Not sure which approach is preferable here, but happy to take advice. Looks like some changes have since landed in master too moving some of the docs locations around, so please confirm that I've created the docs in the right spot.

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
```diff
@@ -214,7 +163,166 @@ class QueueTest : UnitSpec() {
...
+
+    fun droppingStrategyTests(
+      ctx: CoroutineContext = IO.dispatchers().default(),
+      queue: (Int) -> IO<Queue<ForIO, Int>>
+    ) {
+      val label = "DroppingQueue"
+
+      allStrategyTests(label, ctx, queue)
+
+      "$label - offering to a zero capacity queue with a pending taker".config(enabled = false) {
```
scott-christopher: This is the test that was causing problems for me.

nomisRev:
Hey @scott-christopher,

Thanks for the update and thanks for the effort!

No worries, if you allow PR access to the maintainers I'd be happy to look into the tests and Ank.

scott-christopher:
Hey @nomisRev, I believe it should already be enabled.

scott-christopher:
Some of the tests are failing due to `NonEmptyList.traverse` not currently maintaining order of effects, though I have submitted a change to resolve that here: #1879

modules/docs/arrow-docs/docs/docs/arrow-fx/queue/README.md
```diff
@@ -0,0 +1,213 @@
...
+      val q = !Queue.bounded<ForIO, Int>(10, IO.concurrent()).fix()
+      !q.offer(42)
+      !q.take()
+    }
+  //sampleEnd
+  println(result.unsafeRunSync())
+}
+```
+
+Attempting to take a value from an empty `Queue` will cause the calling fibre to be suspended and will resume upon a new
```
nomisRev:
```suggestion
Attempting to take a value from an empty `Queue` will cause the calling fiber to be suspended and will resume upon a new
```

nomisRev:
Hey @scott-christopher,

I fixed the test but couldn't push to your branch.
I've done it successfully in the past when following [this](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork).

Anyway, this is the updated test. I'd be happy to try again if you enabled the above, in case you're busy.

```kotlin
"$label - offering to a zero capacity queue with a pending taker" {
  forAll(Gen.int()) { x ->
    IO.fx {
      val q = !queue(0)
      val taker = !q.take().fork(ctx)
      // Sleep 10 milliseconds, this forces 2 shifts
      // which guarantee `take` to schedule before `offer`.
      !sleep(10.milliseconds)
      !q.offer(x)
      !taker.join()
    }.unsafeRunSync() == x
  }
}
```

The issue as you already mentioned in the comment, we can't synchronize `take` and `offer` since we don't know when `take` is scheduled. Afaik, this is the best we can do. `sleep` requires 2 shifts internally, one to the `ScheduledExecutor` and one back to the default context, this ensures that `take` will gets scheduled before `offer` triggers.

scott-christopher:
Hey @nomisRev, I've pushed up those changes to add the `sleep` to the test (I suspect you may not have been able to push because I had amended the commit for the last batch of changes).

I spent a little while looking at a way to add a `Queue::takeAsync` or similar that could either take a callback or return a promise instead of suspending the fiber that `take` currently does, though I'm not sure how best to go about forking from within the library code to ensure the promise is cleaned up from the `takers` after it fires. An approach like `takeAsync` could allow the state to be updated in the `Ref` prior to the call to `offer` being evaluated, which I believe would offer an alternative to the `sleep` in the tests if desired. In any case, the tests are now all green again :)

scott-christopher:
I have pushed up another change that switches the `sleep` in that particular test with the following to ensure forked call to `take` has occurred before we `offer`.

```kotlin
q.size()
  .repeat<ForIO, Int, Int>(IO.concurrent(), Schedule.doUntil(IO.monad()) { it == -1 })
```

modules/docs/arrow-docs/docs/docs/arrow-fx/queue/README.md
```diff
@@ -0,0 +1,213 @@
...
+```kotlin:ank:playground
+import arrow.fx.*
+import arrow.fx.extensions.fx
+import arrow.fx.extensions.io.concurrent.concurrent
+
+fun main(args: Array<String>) {
+  val result =
+  //sampleStart
+    IO.fx {
+      val q = !Queue.bounded<ForIO, Int>(10, IO.concurrent()).fix()
```
nomisRev: Shouldn't be necessary to fix here since `!` is defined for `Kind<ForIO, A>`.

```suggestion
      val q = !Queue.bounded<ForIO, Int>(10, IO.concurrent())
```

modules/docs/arrow-docs/docs/docs/arrow-fx/queue/README.md
```diff
@@ -0,0 +1,213 @@
...
```
nomisRev:
```suggestion
      val q = !Queue.bounded<ForIO, Int>(10, IO.concurrent())
```

modules/docs/arrow-docs/docs/docs/arrow-fx/queue/README.md
```diff
@@ -0,0 +1,213 @@
...
+import arrow.fx.*
+import arrow.fx.extensions.fx
+import arrow.fx.extensions.io.concurrent.concurrent
+
+fun main(args: Array<String>) {
+  val result =
+  //sampleStart
+    IO.fx {
+      val capacity = 2
+      val q = !Queue.bounded<ForIO, Int>(capacity, IO.concurrent()).fix()
```
nomisRev:
```suggestion
      val q = !Queue.bounded<ForIO, Int>(capacity, IO.concurrent())
```

modules/docs/arrow-docs/docs/docs/arrow-fx/queue/README.md
```diff
@@ -0,0 +1,213 @@
...
+import arrow.fx.*
+import arrow.fx.extensions.fx
+import arrow.fx.extensions.io.concurrent.concurrent
+
+fun main(args: Array<String>) {
+  val result =
+  //sampleStart
+    IO.fx {
+      val capacity = 2
+      val q = !Queue.dropping<ForIO, Int>(capacity, IO.concurrent()).fix()
```
nomisRev:
```suggestion
      val q = !Queue.dropping<ForIO, Int>(capacity, IO.concurrent())
```

modules/docs/arrow-docs/docs/docs/arrow-fx/queue/README.md
```diff
@@ -0,0 +1,213 @@
...
+import arrow.fx.*
+import arrow.fx.extensions.fx
+import arrow.fx.extensions.io.concurrent.concurrent
+
+fun main(args: Array<String>) {
+  val result =
+  //sampleStart
+    IO.fx {
+      val capacity = 2
+      val q = !Queue.sliding<ForIO, Int>(capacity, IO.concurrent()).fix()
```
nomisRev:
```suggestion
      val q = !Queue.sliding<ForIO, Int>(capacity, IO.concurrent())
```

modules/docs/arrow-docs/docs/docs/arrow-fx/queue/README.md
```diff
@@ -0,0 +1,213 @@
...
+```kotlin:ank:playground
+import arrow.fx.*
+import arrow.fx.extensions.fx
+import arrow.fx.extensions.io.concurrent.concurrent
+
+fun main(args: Array<String>) {
+  val result =
+  //sampleStart
+    IO.fx {
+      val q = !Queue.unbounded<ForIO, Int>(IO.concurrent()).fix()
```
nomisRev:
```suggestion
      val q = !Queue.unbounded<ForIO, Int>(IO.concurrent())
```

modules/docs/arrow-docs/docs/docs/arrow-fx/queue/README.md
```diff
@@ -0,0 +1,213 @@
...
+```kotlin:ank:playground
+import arrow.fx.*
+import arrow.fx.extensions.fx
+import arrow.fx.extensions.io.concurrent.concurrent
+
+fun main(args: Array<String>) {
+  val result =
+  //sampleStart
+    IO.fx {
+      val q = !Queue.bounded<ForIO, Int>(10, IO.concurrent()).fix()
```
nomisRev:
```suggestion
      val q = !Queue.bounded<ForIO, Int>(10, IO.concurrent())
```

modules/docs/arrow-docs/docs/docs/arrow-fx/queue/README.md
```diff
@@ -0,0 +1,213 @@
...
```
nomisRev:
```suggestion
      val q = !Queue.bounded<ForIO, Int>(10, IO.concurrent())
```

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
```diff
@@ -214,7 +164,163 @@ class QueueTest : UnitSpec() {
...
+      allStrategyTests(label, ctx, queue)
+
+      "$label - offering to a zero capacity queue with a pending taker" {
+        forAll(Gen.int()) { x ->
+          IO.fx {
+            val q = !queue(0)
+            val taker = !q.take().fork(ctx)
+            // Wait for the forked `take` to complete by checking the queue `size`,
+            // otherwise the test will suspend indefinitely if `take` occurs after `offer`.
+            !q.size().repeat<ForIO, Int, Int>(IO.concurrent(), Schedule.doUntil(IO.monad()) { it == -1 })
```
nomisRev: 🥇

scott-christopher:
Alright, we should be good now. Thanks @nomisRev.

nomisRev:
Thanks for the awesome contribution @scott-christopher! 🎉

PR #1834, opened 2019-12-08
Deprecate KindConnection from public API

nomisRev:
This PR deprecates `KindConnection` from the public API so it can be completely removed in the next major release.

Exposing manual cancellation mechanism likes this has proofed to be error prone.


modules/fx/arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxK.kt
```diff
@@ -243,6 +262,66 @@ data class FluxK<out A>(val flux: Flux<out A>) : FluxKOf<A> {
...
+        }
+
+        val fa2 = try {
+          fa(cb)
+        } catch (t: Throwable) {
+          cb(Left(t.nonFatalOrThrow()))
+          just(just(Unit))
+        }
+
+        val cancelOrToken = AtomicRefW<Either<Unit, CancelToken<ForFluxK>>?>(null)
```
pakoito: jeeeeeesus christ

PR #1835, opened 2019-12-11
Fix Kleisli fx

nomisRev:
Incorrect receiver


raulraja:
Thanks, and sorry, my bad 🙏

Issue #1819, opened 2019-11-26
Stop using applicative/constructor functions to test typeclasses


1Jajen1:
Currently we have tests like this:
`testLaws(cf: (Int) -> Kind<F, Int>)` or similar `testLaws(AP: Applicative<F>)` for our tests. This is awful for several reasons:
`cf` has no source of randomness, hence the `Kind<F, Int>` will only ever differ in the content `Int`. That might not sound too bad for basic things like `Id`, but easily fails with something like `Option`, where we can only ever get `Some`. This is just bad and caused at least one instance where a test should have failed but did not: https://github.com/arrow-kt/arrow/pull/1816#discussion_r350476512
The `AP` approach has exactly the same problem.

A solution would be to instead take either `cf: (Int) -> Gen<Kind<F, Int>>` or `cf: (Gen<A>) -> Gen<Kind<F, A>>` as arguments. This can also be written as a typeclass like this pr did: https://github.com/abendt/arrow/blob/zip/modules/core/arrow-test/src/main/kotlin/arrow/test/generators/LiftGen.kt
(I'd change a few small things, but that is basically what I think is the best solution to the problem)

This should be rather easy to change in almost all places and will significantly improve the coverage of our tests.

pakoito:
Thing is, `None` is tested through `raiseError` for typeclasses that support it. If we'd like to use `ApplicativeError` for anything above it on the typeclass hierarchy I'm not opposed.

nomisRev:
> cf has no source of randomness

Thank you @1Jajen1! This has already caused more false positives in the passed.
I'm sure there is more low hanging fruit in the tests. They definitely deserve more love for all the work they're doing.

1Jajen1:
> Thing is, `None` is tested through `raiseError` for typeclasses that support it. If we'd like to use `ApplicativeError` for anything above it on the typeclass hierarchy I'm not opposed.

This can work for `Option`, `Either` and anything else that is a sum type of 2 with one error and one success case. But it would be wrong to do so in a generic setting when you want to test all cases. 

Btw a quick overview on just how bad this is: https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-core-data/src/test/kotlin/arrow/core/OptionTest.kt
The only laws that actually test with `None` are `Monoid` and the recently added `Align`, `EqK` and `Unalign`. There are 7 more groups of laws that literally never see the `None` case.
[...]

PR #1838, opened 2019-12-13
refactor EqLaws, HashLaws, OrderLaws and ShowLaws to use Gen for testdata

abendt:
refactor EqLaws, HashLaws, OrderLaws and ShowLaws to use Gen for testdata creation (#1819)

### DIFF
--- a/modules/core/arrow-core-data/src/test/kotlin/arrow/core/ConstTest.kt
+++ b/modules/core/arrow-core-data/src/test/kotlin/arrow/core/ConstTest.kt
@@ -11,6 +11,7 @@ import arrow.test.laws.EqLaws
 import arrow.test.laws.ShowLaws
 import arrow.test.laws.TraverseFilterLaws
 import arrow.typeclasses.Eq
+import io.kotlintest.properties.Gen
 
 class ConstTest : UnitSpec() {
   init {
@@ -18,9 +19,14 @@ class ConstTest : UnitSpec() {
       testLaws(
         TraverseFilterLaws.laws(Const.traverseFilter(), Const.applicative(this), { Const(it) }, Eq.any()),
         ApplicativeLaws.laws(Const.applicative(this), Eq.any()),
-        EqLaws.laws(Const.eq<Int, Int>(Eq.any())) { Const(it) },
-        ShowLaws.laws(Const.show(), Const.eq<Int, Int>(Eq.any())) { Const(it) }
+        EqLaws.laws(Const.eq<Int, Int>(Eq.any()), Gen.const(Gen.int())),
+        ShowLaws.laws(Const.show(), Const.eq<Int, Int>(Eq.any()), Gen.const(Gen.int()))
       )
     }
   }
 }
+
+private fun <A> Gen.Companion.const(gen: Gen<A>): Gen<Const<A, A>> =
+  gen.map {
+    Const<A, A>(it)
+  }
--- a/modules/core/arrow-core-data/src/test/kotlin/arrow/core/ConstTest.kt
+++ b/modules/core/arrow-core-data/src/test/kotlin/arrow/core/ConstTest.kt
@@ -6,6 +6,7 @@ import arrow.core.extensions.const.show.show
 import arrow.core.extensions.const.traverseFilter.traverseFilter
 import arrow.core.extensions.monoid
 import arrow.test.UnitSpec
+import arrow.test.generators.genConst
 import arrow.test.laws.ApplicativeLaws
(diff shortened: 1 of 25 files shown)


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/HashLaws.kt
```diff
@@ -7,23 +7,23 @@ import io.kotlintest.properties.forAll
...
-    forAll(Gen.int()) { f ->
-      val a = cf(f)
-      val b = cf(f)
-      EQ.run { a.eqv(b) } && HF.run { a.hash() == b.hash() }
+  private fun <F> equalHash(HF: Hash<F>, EQ: Eq<F>, G: Gen<F>) {
+    forAll(G, G) { a, b ->
+      if (EQ.run { a.eqv(b) })
+        HF.run { a.hash() == b.hash() }
+      else
+        true
```
1Jajen1: I love this test. The reason it takes an additional `Eq` is to still work correct on hash collisions, but nothing stops one from actually just passing the hash instance twice :D

modules/free/arrow-free-data/src/test/kotlin/arrow/free/FreeApplicativeTest.kt
```diff
@@ -70,3 +71,8 @@ class FreeApplicativeTest : UnitSpec() {
     }
   }
 }
+
+private fun Gen.Companion.opsAp(gen: Gen<Int>) =
+  gen.map {
+    OpsAp.value(it)
+  }
```
1Jajen1: I'd love to make this one a recursive generator that can generate the whole opsAst, but kotlintest is no good at recursive gens, so I guess this is fine...

modules/core/arrow-core-data/src/test/kotlin/arrow/core/ConstTest.kt
```diff
@@ -11,16 +11,22 @@ import arrow.test.laws.EqLaws
...
+        ShowLaws.laws(Const.show(), Const.eq<Int, Int>(Eq.any()), Gen.const(Gen.int()))
       )
     }
   }
 }
+
+private fun <A> Gen.Companion.const(gen: Gen<A>): Gen<Const<A, A>> =
+  gen.map {
+    Const<A, A>(it)
+  }
```
1Jajen1: I think this would fit better if it was parameterized to `A` and `B` since they don't have to be the same

1Jajen1: Also let's move this and the other common generators to the generators file

abendt: ok

which other generators do you mean?  in the other cases it does not seem to make sense to move it

modules/free/arrow-free-data/src/test/kotlin/arrow/free/FreeApplicativeTest.kt
abendt: would something like this be better?

```

fun Gen.Companion.opsAp() =
  Gen.oneOf(valueGen, addGen, subtractGen)

val valueGen = Gen.bind(Gen.int()) {
  OpsAp.value(it)
}

val addGen = Gen.bind(Gen.int(), Gen.int()) { a, b ->
  OpsAp.add(a, b)
}

val subtractGen = Gen.bind(Gen.int(), Gen.int()) { a, b ->
  OpsAp.subtract(a, b)
}
```

1Jajen1: Oh right, I somehow thought ops was a recursive type where `class Add(val a: Ops, val b: Ops)` or similar. Yes, that is much better

modules/core/arrow-core-data/src/test/kotlin/arrow/core/ConstTest.kt
1Jajen1: Wrote this thinking there was another one, but you are right, that is the only one that makes sense :+1:

modules/free/arrow-free-data/src/test/kotlin/arrow/free/FreeApplicativeTest.kt
abendt: ok, i have changed it

1Jajen1:
I think this is good to merge. Thank you :clap: :clap:

PR #1840, opened 2019-12-15
Update Gradle to latest version (6.0.1)

LordRaydenMK:
My first failed attempt to do this was in October.

Then it was enabled by merging this: https://github.com/arrow-kt/arrow/pull/1687

Weeks later I tried again but `animalsniffer` was failing. Arrow wasn't using the latest version so I updated. That broke jmh so I updated that.

Then the `arrow-banchmarks-fx` project failed, it didn't recognize scala classes. So I updated to Scala 2.13. Which didn't fix the problem.

Then I realized the problem was using `implementation` and the `arrow-banchmarks-fx` could not see the scala classes. Using `api` for those dependencies makes sure the benchmarks project can access them. (the alternative would be do declare them explicitly in `arrow-banchmarks-fx` using `implementation`).

I could revert the Scala/Cats/ZIO update (they are not related here) but why not keep them. :smile:


rachelcarmena:
> LGTM specially since all tests passed, @arrow-kt/maintainers does someone know if we would, otherwise should, re-run the benchmarks for possible changes between lib versions?

@aballano , I think they are included in the `Build` check because it runs these steps:

```
./gradlew clean build :arrow-benchmarks-fx:jmhClasses
./gradlew check
```

pakoito:
I hope build speed goes up too

aballano:
The tests passed 3 times in a row already so I'll merge now and keep an eye 👀

PR #1843, opened 2019-12-16
Adding *> and <* operations to Apply

### LINKED ISSUE #1839: ["Request"] Add applicative *> and <* to arrow
What would you like to see?

A version of `*>` and `<*` in the `Apply` typeclass. This has been asked more than once in [slack](https://kotlinlang.slack.com/archives/C5UPMM0A0/p1576411952063600) now, so I think it'd be a good idea to add it.

It can be implemented with default implementations and should all be done inside `Apply`. 

We do need to think about good names because we cannot use operators, apart from that the implementation should be straightforward:  `map(left, right) { (l, _) -> l }` and `map(left, right) { (_, r) -> r }`

We also have a similar operator in `Monad` which is called `followedBy` (which is a good name for `*>`) so that should be replaced, but the implementation in `Monad` using `flatMap { it }` could be better than the default for `*>` so it might make sense to explicitly override it in `Monad`

Cotel:
Close #1839

I've used `apIgnore` as a placeholder name for `<*`. Some alternatives I've thought are **mapReplace** (although this one fits *> better) or **alsoDo**. But we can discuss names here. I would also like to place better examples in the docs for the usage of these operators 🤔.

Looking for documentation in Haskell for these functions I saw they are implemented using `$>` and `<$` from Functor and I think we are lacking those operators too. Maybe this is a good time for adding them 😄.


modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Apply.kt
```diff
@@ -315,4 +315,28 @@ interface Apply<F> : Functor<F> {
...
+  /**
+   * Given two actions, it performs them sequentially.
+   * Ignores the result of the first action.
+   *
+   * This is equivalent to *> in Haskell.
+   */
+  fun <A, B> Kind<F, A>.followedBy(fb: Kind<F, B>): Kind<F, B> =
+    map(this, fb) { (_, right) -> right }
+
+  fun <A, B> Kind<F, A>.followedByEval(fb: Eval<Kind<F, B>>): Kind<F, B> =
```
1Jajen1: There is sadly no point to having this on `Apply` without having a lazy `ap` method. Maybe move this back to `Monad` where it actually made sense or remove it :/

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Apply.kt
```diff
@@ -315,4 +315,28 @@ interface Apply<F> : Functor<F> {
...
+  /**
+   * Given two actions, it performs them sequentially.
+   * Discards the result of the second action.
+   *
+   * This is equivalent to <* in Haskell.
+   */
+  fun <A, B> Kind<F, A>.apIgnore(fb: Kind<F, B>): Kind<F, A> =
+    map(this, fb) { (left, _) -> left }
+
+  fun <A, B> Kind<F, A>.apIgnoreEval(fb: Eval<Kind<F, B>>): Kind<F, A> =
```
1Jajen1: Same as above

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Monad.kt
```diff
@@ -50,12 +50,6 @@ interface Monad<F> : Selective<F> {
   fun <A> Kind<F, Kind<F, A>>.flatten(): Kind<F, A> =
     flatMap(::identity)
 
-  fun <A, B> Kind<F, A>.followedBy(fb: Kind<F, B>): Kind<F, B> =
-    flatMap { fb }
-
-  fun <A, B> Kind<F, A>.followedByEval(fb: Eval<Kind<F, B>>): Kind<F, B> =
-    flatMap { fb.value() }
-
```
1Jajen1: Can you overwrite those here? I think a single flatMap is more performant than `Apply.map` which calls a few other methods

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/applicative/README.md
```diff
@@ -98,6 +98,23 @@ Computation happens when `.value()` is invoked.
...
+```
+
+#### Kind<F, A>#apIgnore / apIgnoreEval
+
+This is a reverse for `followedBy`. Sequences actions but discarding the value of the second argument.
+
+```kotlin:ank
+Option.applicative().run { Some(1).apIgnore(Some(2)) } 
+```
+
```
1Jajen1: This should be under `Apply` shouldn't it?

1Jajen1:
Regarding names, alsoDo doesn't seem to bad, kind of inline with also from the stdlib, just not sure about the Do part :D

> Looking for documentation in Haskell for these functions I saw they are implemented using $> and <$ from Functor and I think we are lacking those operators too. Maybe this is a good time for adding them :smile:.

Sure, I won't be against it :) Just more funny names to figure out :P

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Apply.kt
Cotel: I just move it by inertia 😅

modules/docs/arrow-docs/docs/docs/arrow/typeclasses/applicative/README.md
Cotel: In fact, all the documented methods in this file belong to Apply, there is a section explaining what Apply is just below this fragment. I can just copy these two at the Apply docs too 🤔

1Jajen1: Ah ok, then leave it as it is. Thats probably less confusing

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Monad.kt
```diff
@@ -50,9 +50,12 @@ interface Monad<F> : Selective<F> {
   fun <A> Kind<F, Kind<F, A>>.flatten(): Kind<F, A> =
     flatMap(::identity)
 
-  fun <A, B> Kind<F, A>.followedBy(fb: Kind<F, B>): Kind<F, B> =
+  override fun <A, B> Kind<F, A>.followedBy(fb: Kind<F, B>): Kind<F, B> =
     flatMap { fb }
 
+  override fun <A, B> Kind<F, A>.apIgnore(fb: Kind<F, B>): Kind<F, A> =
+    fb.flatMap { this }
+
```
1Jajen1: This is no good! This reverses the order! I think this is also something we already have: `flatTap` does the same right?

Cotel: Not exactly because `flatTap` has a `f: (A) -> Kind<F, B>`. In this case we ignore the value of the monad completely. I made this implementation looking at the overriden version of `ap` for Monad (Line 48)

1Jajen1: Thats true the signature is different. :+1:  

But still this override is no good as it changes the behaviour:
`IO { println(1) }.apIgnore(IO { println(2) })` will print 1 and then 2 with the apply instance, but 2 and then 1 with the monad instance. 

Just implement it as `flatTap { fb }` which should work.

1Jajen1: I guess we should somehow try to enforce the monad-applicative consistency law by adding a test to the MonadLaws that takes both a Monad and an explicit Applicative instance and checks if they do the same thing. Will open a ticket for that, as I find this very important!

1Jajen1: https://github.com/arrow-kt/arrow/issues/1847 Added this as a ticket.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Apply.kt
```diff
@@ -315,4 +315,22 @@ interface Apply<F> : Functor<F> {
...
+  fun <A, B> Kind<F, A>.followedBy(fb: Kind<F, B>): Kind<F, B> =
+    map(this, fb) { (_, right) -> right }
+
+  /**
+   * Given two actions, it performs them sequentially.
+   * Discards the result of the second action.
+   *
+   * This is equivalent to <* in Haskell.
+   */
+  fun <A, B> Kind<F, A>.apIgnore(fb: Kind<F, B>): Kind<F, A> =
```
pakoito: we called this `flatTap`

pakoito: Oh, I saw the other discussion. I see!

Let's come up with a name for it.

1Jajen1: Well `flatTap` "taps" into the value but does not change it after a `flatMap` which is why it's actually a really good name imo.

This function applies an effect, but ignores the result, so I understand where `apIgnore` comes from, it just does not sound good :smile: 
`apEffect` might be something, `withEffect` also comes to mind. `alsoDo` sounds weird because of the "do", but maybe `apAlso` or `alsoAp` might be better.

pakoito: `apTap`

pakoito: I used to call flatTap, effectM. We agreed to use effect for other purposes, sadly.

Cotel: I'll use apTap so we don't stop this PR but following the reasoning above we are not tapping the value in the monad 🤔

1Jajen1:
Please read the comment I added to #1847 . I did not see that you already started this, otherwise I would have added it here :)

Cotel:
I'm not sure if this is what you were thinking. Check it out pls @1Jajen1

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadLaws.kt
```diff
@@ -28,11 +28,14 @@ object MonadLaws {
...
         Law("Monad Laws: stack safe") { M.stackSafety(5000, EQ) },
-        Law("Monad Laws: selectM == select when Selective has a monad instance") { M.selectEQSelectM(EQ) },
         Law("Monad Laws: monad instance should preserve behavior of Functor") { M.preservesFunctor(FF, EQ) },
         Law("Monad Laws: monad instance should preserve behavior of Applicative") { M.preservesApplicative(AP, EQ) }
       )
 
+  fun <F> laws(M: Monad<F>, FF: Functor<F>, AP: Applicative<F>, SL: Selective<F>, EQ: Eq<Kind<F, Int>>): List<Law> =
+    laws(M, FF, AP, EQ) +
+      Law("Monad Laws: monad instance should preserve behavior of Selective") { M.preservesSelective(SL, EQ) }
+
```
1Jajen1: Yes this is exactly what I meant, please also move over the tests for applicative and functor to here and revert the original `laws` method back to its previous state. This way the consistency laws are opt-in. We should test them with all datatypes we have in arrow, but as I said in my comment, there are valid reasons to not follow this sometimes.

Cotel: Perfect 👌, although I'll leave usages of this method in tests as they seem to work.

Cotel: If I move functor and applicative to this method we will only be able to test consistency for datatypes which can provide a selective instance. Maybe we can do another overload? We will then have basic monad laws, basic + functor/applicative and basic + functor/ap + selective 🤔

1Jajen1: Yes we could do another overload, but that will grow in size quite a bit ^-^

Just fill the selective instance in with the monad instance in those cases. The tests, although meaningless, should pass as they are literally the same methods.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadLaws.kt
```diff
@@ -72,9 +81,25 @@ object MonadLaws {
...
+  fun <F> Monad<F>.preservesApplicative(AP: Applicative<F>, EQ: Eq<Kind<F, Int>>): Unit =
+    forAll(Gen.int(), Gen.functionAToB<Int, Int>(Gen.int())) { num, f ->
+      val ff = just(f)
+      AP.run { just(num).ap(ff) }.equalUnderTheLaw(just(num).ap(ff), EQ)
+
+      val fg = just(1)
+      AP.run { just(num).followedBy(fg) }.equalUnderTheLaw(just(num).followedBy(fg), EQ)
+
+      AP.run { just(num).apTap(fg) }.equalUnderTheLaw(just(num).apTap(fg), EQ)
```
1Jajen1: Can you name this "preservesApply" and change the constraint? We can be less strict about this in docs because most datatypes are both and Applicative is just more common, but in the api we should always use the lowest possible constraint. With that I think this is good to go.

Cotel:
I've checked Functor and we already have an equivalent for `*>` named \`as`:

```kotlin
fun <A, B> Kind<F, A>.`as`(b: B): Kind<F, B> = map { b } 
```

We could add another function for `<*` named \`as` as well:

```kotlin
fun <A, B> A.`as`(fb: Kind<F, B>): Kind<F, A> = fb.`as`(this)
```

On another note, I don't like that name because of the \` symbols and because is not clear enough but I can't think a better name for it.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadLaws.kt
1Jajen1: Btw this will result in false-positives until #1844 comes along and changes the way test-data is generated. Why? Because this currently only tests cases equal to `Id(num).applyAp(constFn) == Id(num).monadAp(constFn)`. You can ignore this for now as it's quite the large change to properly do away with that issue, but please change the monad version of ap to `flatMap { a -> ff.map { f -> f(a) } }` instead of the reverse order.

1Jajen1:
Hmm using `as` will be annoying. `mapConst`, `applyValue`, `replaceValue` are some ideas. Not sure I like them, but who knows ^-^

Cotel:
> Hmm using `as` will be annoying. `mapConst`, `applyValue`, `replaceValue` are some ideas. Not sure I like them, but who knows ^-^

My personal vote is for `mapConst` given the default implementation in Haskell is `fmap . const` 👍

1Jajen1:
>> Hmm using as will be annoying. mapConst, applyValue, replaceValue are some ideas. Not sure I like them, but who knows ^-^

> My personal vote is for mapConst given the default implementation in Haskell is fmap . const +1

Right, that's actually a good reason to name it that way. Do you want to add this here or in a new pr? I think this is good to go, so it's up to you ^^

Cotel:
I'll continue here. I'm trying to use the IDE rename but the project is so big it immediately hangs 😅

1Jajen1:
I think this is good to go. I have two more name suggestions for `*>` and `<*` tho: `zipLeft` and `zipRight`. That is what zio called them in `Schedule` and they kind of make sense. I like `followedBy` a lot tho, so not sure if we should change it (at least keep an alias).

raulraja:
`as` is also known as widden, in case it helps

1Jajen1:
Something weird is happening with the ci builds for this pr. I canceled one, retried it and had to cancel it after an hour again. Is this on latest master?

Cotel:
There were some conflicts with masters and it seems to hang with Kleisli concurrent laws. Before those conflicts Kleisli was testing another laws for the monad hierarchy 🤔

modules/mtl/arrow-mtl-data/.kotlintest/spec_failures
```diff
@@ -0,0 +1 @@
+arrow.mtl.EitherTTest
```
1Jajen1: I don't think this should be here.

1Jajen1:
Yes, imo just fix the conflicts and merge it when the build passes.

Cotel:
Should we wait for another approval or merge it already? 🤔

1Jajen1:
> Should we wait for another approval or merge it already? :thinking:

Up to you :)

PR #1845, opened 2019-12-16
Removes level tags

AntonioMateoGomez:
This PR removes the level tags from docs. Also reorganizes the lastest included documentation to its proper sidebar file.


modules/docs/arrow-docs/docs/_data/sidebar-core.yml
```diff
@@ -241,6 +241,18 @@ options:
       - title: Unalign
         url: /docs/arrow/typeclasses/unalign/
 
+      - title: Zip
```
calvellido: Are these changes expected?

calvellido: Ahm, seems like they are :+1:

AntonioMateoGomez: Yes, those docs were the last ones included in Arrow. They must be placed in the sidebar-core section

PR #1848, opened 2019-12-17
Make suspended IO cancelation aware

nomisRev:
This PR fixes the issue were IO loses its connection across `suspended` boundaries.

 - Currently if you transform an `IO` to `suspended` and then wrap it into another `IO` using `effect` or the default constructor it becomes uncancelable.
- This PR makes the CoroutineContext `IOContext` aware, so it can carry its `IOConnection` within the `CoroutineContext` and share it across multiple `IORunLoop` for the same `IO`.

This makes `suspendCoroutine` the equivalent of `RestartCallback` on the language level, where we can move between callbacks and direct values. This allows us to bridge the connection across the `IO` program since the `Continuation` instance is shared across the whole IO program just like `RestartCallback`.

This also allows for working with `IO` in a style more similar to `KotlinX`.


pakoito:
Damn, and merged in no time. Great find :D

PR #1851, opened 2019-12-18
BIO - Part 1

nomisRev:
- [x] Add `E` Param
- [x] Add RaiseLeft case

### DIFF
--- a/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
+++ b/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
@@ -1013,7 +1013,7 @@ sealed class BIO<out E, out A> : IOOf<A> {
     override fun unsafeRunTimedTotal(limit: Duration): Option<A> = unsafeResync(this, limit)
   }
 
-  internal data class Bind<A, E, out B, out E2: E>(val cont: BIO<E, A>, val g: (A) -> BIO<E, B>) : BIO<E2, B>() {
+  internal data class Bind<A, E, out B, out E2 : E>(val cont: BIO<E, A>, val g: (A) -> BIO<E, B>) : BIO<E2, B>() {
     override fun unsafeRunTimedTotal(limit: Duration): Option<B> = throw AssertionError("Unreachable")
   }
 
@@ -1098,7 +1098,6 @@ fun <A> IOOf<A>.handleError(f: (Throwable) -> A): IO<A> =
  */
 fun <A> IOOf<A>.handleErrorWith(f: (Throwable) -> IOOf<A>): IO<A> =
   BIO.Bind(fix(), IOFrame.Companion.ErrorHandler(f))
-
 
 /**
  * Executes the given `finalizer` when the source is finished, either in success or in error, or if canceled, allowing
--- a/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
+++ b/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
@@ -1,9 +1,11 @@
 package arrow.fx
 
+import arrow.Kind
 import arrow.core.Either
 import arrow.core.Either.Left
 import arrow.core.Eval
 import arrow.core.NonFatal
+import arrow.core.None
 import arrow.core.Option
 import arrow.core.Some
 import arrow.core.andThen
@@ -35,22 +37,24 @@ import kotlin.coroutines.resume
 import kotlin.coroutines.resumeWithException
 import kotlin.coroutines.suspendCoroutine
 
-class ForIO private constructor() {
+class ForBIO private constructor() {
   companion object
 }
-typealias IOOf<A> = arrow.Kind<ForIO, A>
+typealias BIOOf<E, A> = arrow.Kind2<ForBIO, E, A>
+typealias BIOPartialOf<E> = Kind<ForBIO, E>
 
-@Suppress("UNCHECKED_CAST", "NOTHING_TO_INLINE")
-inline fun <A> IOOf<A>.fix(): IO<A> =
-  this as IO<A>
+inline fun <E, A> BIOOf<E, A>.fix(): BIO<E, A> =
+  this as BIO<E, A>
+
+typealias IO<A> = BIO<Nothing, A>
+typealias IOOf<A> = BIOOf<Nothing, A>
+typealias ForIO = BIOPartialOf<Nothing>
 
 typealias IOProc<A> = ((Either<Throwable, A>) -> Unit) -> Unit
 typealias IOProcF<A> = ((Either<Throwable, A>) -> Unit) -> IOOf<Unit>
 
-typealias IO<A> = BIO<Nothing, A>
-
 @Suppress("StringLiteralDuplication")
-sealed class BIO<out E, out A> : IOOf<A> {
+sealed class BIO<out E, out A> : BIOOf<E, A> {
 
   companion object : IOParMap2, IOParMap3, IORacePair, IORaceTriple {
 
@@ -121,20 +125,38 @@ sealed class BIO<out E, out A> : IOOf<A> {
     fun <A> just(a: A): IO<A> = Pure(a)
 
     /**
-     * Raise an error in a pure way without actually throwing.
+     * Raise an exception in a pure way without actually throwing.
      *
      * ```kotlin:ank:playground
      * import arrow.fx.IO
      *
      * fun main(args: Array<String>) {
      *   //sampleStart
-     *   val result: IO<Int> = IO.raiseError<Int>(RuntimeException("Boom"))
+     *   val result: IO<Int> = IO.raiseException<Int>(RuntimeException("Boom"))
      *   //sampleEnd
      *   println(result.unsafeRunSync())
      * }
      * ```
      */
-    fun <A> raiseError(e: Throwable): IO<A> = RaiseError(e)
+    fun <A> raiseException(e: Throwable): BIO<Nothing, A> = RaiseException(e)
+
+    /**
+     * Raise an error in a pure way
+     *
+     * ```kotlin:ank:playground
+     * import arrow.fx.IO
+     *
+     * object NetworkError
+     *
+     * fun main(args: Array<String>) {
+     *   //sampleStart
+     *   val result: BIO<NetworkError, Int> = IO.raiseError(NetworkError)
+     *   //sampleEnd
+     *   println(result.unsafeRunSync())
+     * }
+     * ```
+     */
+    fun <E, A> raiseError(e: E): BIO<E, A> = RaiseError(e)
 
     /**
      *  Sleeps for a given [duration] without blocking a thread.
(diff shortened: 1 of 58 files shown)


modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IOParMap2.kt
```diff
@@ -12,7 +12,7 @@ import kotlin.coroutines.CoroutineContext
 /** Mix-in to enable `parMapN` 2-arity on IO's companion directly. */
 interface IOParMap2 {
 
-  fun <A, B, C> parMapN(ctx: CoroutineContext, fa: IOOf<A>, fb: IOOf<B>, f: (A, B) -> C): IO<C> = IO.Async { conn, cb ->
+  fun <A, B, C> parMapN(ctx: CoroutineContext, fa: IOOf<A>, fb: IOOf<B>, f: (A, B) -> C): IO<C> = BIO.Async { conn, cb ->
```
pakoito: This is why I'd rather we kept the name as `IO`. We have the examples and people don't mind, it'll make the migration easier, and we don't have to rename to RIO or something once we add reader.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -47,8 +47,10 @@ inline fun <A> IOOf<A>.fix(): IO<A> =
 typealias IOProc<A> = ((Either<Throwable, A>) -> Unit) -> Unit
 typealias IOProcF<A> = ((Either<Throwable, A>) -> Unit) -> IOOf<Unit>
 
+typealias IO<A> = BIO<Nothing, A>
```
pakoito: `IO<A>` is `Throwable` for compatibility, right?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IOParMap2.kt
nomisRev: Oh actually you're happy you say that! I would also be in favor of keeping `IO` to be honest, I thought we wanted to keep the current API alive as much as possible.

It seems that'll far apart since a lot of functions need to become extensions since Kotlin doesn't have variance for functions.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
nomisRev: Yes.
- `IO<A> <~> suspend () -> A`
-  `BIO<E, A> <~> EitherT<ForIO, E, A> <~> suspend () -> Either<E, A>`

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -1024,22 +1013,22 @@ sealed class IO<out A> : IOOf<A> {
     override fun unsafeRunTimedTotal(limit: Duration): Option<A> = unsafeResync(this, limit)
   }
 
-  internal data class Bind<E, out A>(val cont: IO<E>, val g: (E) -> IO<A>) : IO<A>() {
-    override fun unsafeRunTimedTotal(limit: Duration): Option<A> = throw AssertionError("Unreachable")
+  internal data class Bind<A, E, out B, out E2 : E>(val cont: BIO<E, A>, val g: (A) -> BIO<E, B>) : BIO<E2, B>() {
```
aballano: maybe dumb question, but wouldn't you need to provide a transformation to E2 somehow? 

```suggestion
  internal data class Bind<A, E, out B, out E2 : E>(val cont: BIO<E, A>, val g: (A) -> BIO<E2, B>) : BIO<E2, B>() {
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -1108,4 +1097,18 @@ fun <A> IOOf<A>.handleError(f: (Throwable) -> A): IO<A> =
...
  */
 fun <A> IOOf<A>.handleErrorWith(f: (Throwable) -> IOOf<A>): IO<A> =
-  IO.Bind(fix(), IOFrame.Companion.ErrorHandler(f))
+  BIO.Bind(fix(), IOFrame.Companion.ErrorHandler(f))
+
+/**
+ * Executes the given `finalizer` when the source is finished, either in success or in error, or if canceled, allowing
+ * for differentiating between exit conditions. That's thanks to the [ExitCase] argument of the finalizer.
+ *
+ * As best practice, it's not a good idea to release resources via `guaranteeCase` in polymorphic code.
```
raulraja: It would be good to say here why is not a good idea, as reader unfamiliar with Bracket, they will be left wondering.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
nomisRev: No, you are allowed to return an `BIO` that returns a *more specific* error.

Example where we go from `ErrorDomain` to `Error1`.

```
sealed class ErrorDomain {
  object Error1 : ErrorDomain()
  object Error2 : ErrorDomain()
}
val test: BIO<ErrorDomain, Int> = BIO.just(1)
val other: BIO<ErrorDomain.Error2, String> = BIO.just("other")
val result: BIO<ErrorDomain.Error2, String> = test.flatMap { other }
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
nomisRev: Yes, the documentation keep use a refresher. I'm hoping to go over all snippets in the final PR which where we'll need to tackle all broken docs.. 😭

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IOParMap2.kt
nomisRev: I.e. `flatMap` became an extension function so that now requires an import.
That will be true for all combinators that have E` in the `in`position.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -1108,4 +995,148 @@ fun <A> IOOf<A>.handleError(f: (Throwable) -> A): IO<A> =
...
+ * fun main(args: Array<String>) {
+ *   val result =
+ *   //sampleStart
+ *   IO.just("Hello").flatMap { IO { "$it World" } }
+ *   //sampleEnd
+ *   println(result.unsafeRunSync())
+ * }
+ * ```
+ */
+fun <E, A, B, E2 : E> BIOOf<E, A>.flatMap(f: (A) -> BIOOf<E2, B>): BIO<E2, B> =
```
JorgeCastilloPrz: Where's `IO.flatMap` now?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -1108,4 +995,148 @@ fun <A> IOOf<A>.handleError(f: (Throwable) -> A): IO<A> =
...
+ *
+ * fun main(args: Array<String>) {
+ *   //sampleStart
+ *   val safeComputation = openFile("data.json").bracket({ file: File -> closeFile(file) }, { file -> fileToString(file) })
+ *   //sampleEnd
+ *   println(safeComputation.unsafeRunSync())
+ * }
+ * ```
+ */
+fun <A, B> IOOf<A>.bracket(release: (A) -> IOOf<Unit>, use: (A) -> IOOf<B>): IO<B> =
```
JorgeCastilloPrz: Since both `IO` and `BIO` will implement all these methods, would it make sense to separate both in different files even if `IO` is an alias for `BIO`? We could keep here `IO` alias an all its combinators and move all BIO stuff to a separate file called `BIO.kt` so it's also easier to search. Feels like we'll have a big bunch of code within this file otherwise.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IOParMap2.kt
JorgeCastilloPrz: If possible, I'd also rather keeping the `IO` name for all (IO, BIO and RIO).

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
nomisRev: `typalias IO<A> = BIO<Nothing, A>`, this is the new `flatMap` that works for both.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
nomisRev: Just like `flatMap` these will all be refactored to work for `BIO<E, A>` which means they'll work for both. This is what we'll be splitting over multiple PRs since it's a quite large diff. (~3k loc for refactoring all combinators and adding a couple new ones to work on `E`)

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
JorgeCastilloPrz: Oh, I've seen some methods defined over `BIO` and other's over `IO` so that got me confused for a second.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
JorgeCastilloPrz: Perfect. Much more clear now. It got me wondering why some methods were defined over `IO` and some others over `BIO`.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -862,157 +865,41 @@ sealed class IO<out A> : IOOf<A> {
...
     override fun <B> map(f: (A) -> B): IO<B> = Suspend { Pure(f(a)) }
 
-    // Pure can be replaced by its value
-    override fun <B> flatMap(f: (A) -> IOOf<B>): IO<B> = Suspend { f(a).fix() }
-
     override fun unsafeRunTimedTotal(limit: Duration): Option<A> = Some(a)
   }
 
-  internal data class RaiseError(val exception: Throwable) : IO<Nothing>() {
+  internal data class RaiseException(val exception: Throwable) : IO<Nothing>() {
```
JorgeCastilloPrz: Is this gonna be redefined for BIO instead also?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IOParMap2.kt
raulraja: Agreed on keeping IO as single name. f we do that we need to write a small migration guide to broken apis that would now require E specified

JorgeCastilloPrz: `IO<A>` is aliased to `BIO<Nothing, A>` so they shouldn't need to, right?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
nomisRev: Currently in `0.10.3` there was only 1 faulty scenario for `IO` which was `Throwable`. Now, there is 2. `Throwable` and `E`, I decided to keep `RaiseError` for `E` and to refactor the original `Throwable` to `RaiseException` instead. I felt that better reflected the internals.

An `IO` for `<E, A>` is right-biased towards `A` and it short-circuits errors `E` and exceptions.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IOParMap2.kt
nomisRev: @pakoito after making the change. Not sure if that'll make it easier to migrate.

nomisRev: @JorgeCastilloPrz I think Paco wants `IO<Nothing, A>` and `IO<E, A>`.

PR #1853, opened 2019-12-18
Generates Dokka API docs links

franciscodr:
There is a couple of functions called `tcMarkdownList` and `dtMarkdownList` that, given a `DataType` or a `TypeClass`, generate a table containing all the typleclasses or data types that support them.

Those functions generate wrong links so the Arrow site has a lot of broken links.

This pull request changes those changes to generate the links in the expected format.


franciscodr:
Closing it because it was created against a wrong base branch

PR #1859, opened 2019-12-19
[BIO]: include `E` into IOFrame & IORunLoop

nomisRev:


modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -991,8 +915,59 @@ fun <A> IOOf<Nothing, A>.handleError(f: (Throwable) -> A): IO<Nothing, A> =
  *
  * @see handleErrorWith for a version that can resolve the error using an effect
  */
+fun <E, A, E2 : E> IOOf<E, A>.handleErrorWith(f: (Throwable) -> IOOf<E2, A>, fe: (E) -> IOOf<E2, A>): IO<E2, A> =
```
pakoito: I believe we should split the API here

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORunLoop.kt
```diff
@@ -12,22 +9,35 @@ import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.EmptyCoroutineContext
 import kotlin.coroutines.startCoroutine
 
+sealed class IOResult<out E, out A> {
```
pakoito: This is public API, it should be in IO.kt

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORunLoop.kt
```diff
@@ -12,22 +9,35 @@ import kotlin.coroutines.CoroutineContext
...
+  data class Right<A>(val value: A) : IOResult<Nothing, A>()
```
pakoito: We either use left-right analogy or success-failure-error. Now that Try is out those names are free.

nomisRev: Yes, this should probably be `Success` in analogy to success-failure-error`.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
nomisRev:
```
fun <E, A, E2 : E> IOOf<E, A>.handleErrorWith(f: (Throwable) -> IOOf<E2, A>, fe: (E) -> IOOf<E2, A>): IO<E2, A> =
fun <A> IOOf<Nothing, A>.handleErrorWith(f: (Throwable) -> IOOf<Nothing, A>): IO<Nothing, A> =
fun <E, A> IOOf<E, A>.fallbackTo(f: () -> A): IOOf<E, A> =
fun <E, A, E2> IOOf<E, A>.fallbackWith(fa: IOOf<E2, A>): IO<E2, A> =
```

I am not sure I understand what you mean by split up the API. You mean you want more overload than these? Any suggestions which ones you're missing here? I am just including the most basic stuff here, i.e. `fallbackTo` was an attempt to keep something analog to the old API. We're currently not including a lot of derived operators on concretely on `IO`.

`fallbackWith` is similar to the original `handleErrorWith`, we could introduce a `Either<Throwable, E>` lambda param there which we could later upgrade to a lighter `Union<Throwable, E>`?

pakoito: I mean handleErrorWith, handleExceptionWith, and handleAllWith, or something like that. Installing an error handler while keeping the existing exception handler should be possible, as there's no real way to retrieve it.

pakoito: Same for the other APIs like mapError & mapException

nomisRev: 👍 Yes, that's definitely possible, those would simply be overloads of what we have now.

We can probably make `IOFrame` more complex to be able to distinct between the 2 handlers but I doubt it's worth it. Unless you're going to do *a lot*  computations on `E` with *a lot* exception handlers in between.

nomisRev: @pakoito I'm going to leave this until the end when we have an overview of the final minimal API and we can introduce more combinators. I'm running into some erasures issues as we speak 😅

Maybe it's easier to review the final public API in the end? I'm going to write a migration guide, so I should cover those things there again anyway.

I.e. these get erased to the same thing. We can probably walk around these things using `@JvmName` which is something we haven't done often in the past yet.

```
fun <E, A, E2 : E> IOOf<E, A>.handleErrorWith(fe: (E) -> IOOf<E2, A>): IO<E2, A>
fun <A> IOOf<Nothing, A>.handleErrorWith(f: (Throwable) -> IOOf<Nothing, A>): IO<Nothing, A>
```

pakoito: >Yes, that's definitely possible, those would simply be overloads of what we have now.

they shouldn't be, that's the point. If we have an exception handler installed it should be replaced by a re-raise one...or should it? 🤔

pakoito: Seems wasteful, although now that I wrote it it may work.

nomisRev: >  If we have an exception handler installed it shouldn't be replaced by a re-raise one...or should it?

Not 100% sure I follow what you mean. An `IOFrame` is essentially a `Bind` that *always* executes. *Also* if no error or exception occurred, that's why `Redeem` also exists which in the context of the old `IO` meant `handleErrorWith + flatMap` which can change `A` -> `B` while resolving the error.

In contrast to `handleErrorWith` which applies `(Throwable) -> IO<A>` *OR* `{ IO.Pure(it) }`.

This model is kept the same except for that `IOFrame` now has an additonal `E` handler.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -593,11 +604,11 @@ sealed class IO<out E, out A> : IOOf<E, A> {
    *
    * **BEWARE** this does **not** support cancelation since Kotlin has no cancelation support for `suspend` on the language level.
    */
-  suspend fun suspended(): A = suspendCoroutine { cont ->
+  suspend fun suspended(): Either<E, A> = suspendCoroutine { cont ->
```
JorgeCastilloPrz: This might be non sense but, shouldn't we have a suspended version for the `IO` alias that still returns `A` (or throws)?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -893,45 +827,45 @@ sealed class IO<out E, out A> : IOOf<E, A> {
...
 
-    override fun unsafeRunTimedTotal(limit: Duration): Option<A> = throw AssertionError("Unreachable")
+    override fun unsafeRunTimedTotal(limit: Duration): Option<Nothing> = throw AssertionError("Unreachable")
   }
 
   internal data class ContextSwitch<E, A>(
     val source: IO<E, A>,
     val modify: (IOConnection) -> IOConnection,
-    val restore: ((Any?, Throwable?, IOConnection, IOConnection) -> IOConnection)?
+    val restore: ((a: Any?, e: Any?, t: Throwable?, old: IOConnection, new: IOConnection) -> IOConnection)?
```
JorgeCastilloPrz: thanks 👍

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -850,7 +784,7 @@ sealed class IO<out E, out A> : IOOf<E, A> {
    * @see [unsafeRunAsync] or [unsafeRunAsyncCancellable] that run the value as [Either].
    * @see [runAsync] to run in a referential transparent manner.
    */
-  fun unsafeRunSync(): A =
+  fun unsafeRunSync(): Either<E, A> =
```
JorgeCastilloPrz: What happens with `IO<A>` then?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
JorgeCastilloPrz: > An IOFrame is essentially a Bind that always executes unrelated to an Exception, Error or Success occured. 

Is this another way to put "you can't pass already installed handlers between binds?" I think Paco means what'd happen if you already installed an exception handler and you want to stay with it but still hook a different error (`E`) handler. Probably it's me the one not following tho.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
nomisRev: Well that is a very good question! :D We have a problem with distinct-ing between `IO<Nothing, A>` and `IO<E, A>`since due to erasure they're the same. Which leaves us with the following, same problem with the other runners.

```
fun <E, A> IOOf<E, A>.suspended(): Either<E, A>
fun Either<Nothing, A>.value(): A = fold(::identity, ::identity)
fun <A> IOOf<Nothing, A>.suspendedGet(): A = fix().suspended().value()
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
nomisRev: Same reply as above. We need to think of a strategy for the API to distinct the 2 in an ergonomic way.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
nomisRev: Yes, for a hook that is not applicable nothing changes.

1. First `IO` binds `handleExceptionWith` and applies an identity function `IO.RaiseError(it)` since `handleExceptionWith` doesn't deal with `Error`.
2. Following `IO` binds `handleErrorWith` and applies the supplied function to the `Error`.

It might be possible to optimize and differentiate between `Exception` handlers and `Error` handlers but not sure that's worth it. Currently all we treat `Success` as `identity` for all error handlers as well, which we do to support `Redeem`.

```
IO.raiseError<Int, String>(1)
    .handleExceptionWith { IO.just(it + 1)  }
    .handleErrorWith { IO.just(it + 2) }
    .suspended() // Left(3)
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
JorgeCastilloPrz: So what's on the left side if you just do `suspended()` for an `IO<A>`?

nomisRev: `Nothing` so you can safely extract `A` using `fold(::identity, ::identity)`

JorgeCastilloPrz: I mean, `Nothing` is the type but what's the value of the left side in that case? Mostly thinking on what's gonna happen if they try to read it.

nomisRev: If the type is `Nothing` then there can be no value on the left side. Or in other words `Either<Nothing, A>` is isomorphic to `Id<A>` and `A`.

PR #1868, opened 2019-12-20
Add lazyAp and use it in traverse implementations

### LINKED ISSUE #1749: Traverse is fundamentally broken for infinite data structures

### LINKED ISSUE #1755: Alternative some and many default implementation stack overflows
The current implementation of some and many:
```kotlin
fun <A> Kind<F, A>.some(): Kind<F, SequenceK<A>> = map(this, many()) { (v, vs) -> (sequenceOf(v) + vs).k() }
fun <A> Kind<F, A>.many(): Kind<F, SequenceK<A>> = some().orElse(just(emptySequence<A>().k()))
```
stackoverflows because `map` from applicative is strict in its arguments. Hence this is related to #1749, where the issue is `ap` (which `map` is implemented with) is strict.. This is untested because we have no instances for `Alternative` atm (except composed). `Applicative` is the root cause again and needs a lazy variant of it's methods. Ideas?

1Jajen1:
This will fix #1749 and also fix #1755 .

I added `lazyAp` with a default implementation in terms of `ap`, so only applicative functors that can actually short circuit will get an override for it.

Adding `lazyAp` had a few effects:
- `traverse` now truly short-circuits. It even works on infinite sequences.
- `traverse` with for example `IO` will start "running" code directly instead of folding the entire functor and then running the `IO`.
- `traverse_` has the same benefits
- `some` and `many` from `Alternative` will now work without someone explicitly overriding one of them.

Few things:
- ~~Both `traverse` and `traverse_` are *really* slow. Is this because of `foldRight` and thus by extension `Eval`? Because having, imo, the most useful function that slow is kind of annoying :)~~ This has been attributed to `foldRight` being slow. `traverse` is plenty fast :) 
- ~~`SequenceK.traverse` still has one problem left: calling `.last()` on the resulting sequence will take a long long time and eventually fail with a stackoverflow which happens because of `Sequence.plus` not handling things nicely.~~


modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Apply.kt
```diff
@@ -38,6 +38,12 @@ interface Apply<F> : Functor<F> {
    */
   fun <A, B> Kind<F, A>.ap(ff: Kind<F, (A) -> B>): Kind<F, B>
 
+  /**
+   * Lazy version of ap, useful for datatypes which can short circuit.
+   * This will at some point be removed for a compiler plugin
+   */
+  fun <A, B> Kind<F, A>.lazyAp(ff: () -> Kind<F, (A) -> B>): Kind<F, B> = ap(ff())
```
1Jajen1: The default implementation is just `ap` because most applicatives don't benefit from `lazyAp` only those with a short-circuit pattern do and forcing those to implement a pointless method is more work than having the others explicitly implement it

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Monad.kt
```diff
@@ -99,6 +99,9 @@ interface Monad<F> : Selective<F> {
     flatMap { it.fold({ a -> f.map { ff -> ff(a) } }, { b -> just(b) }) }
 
   override fun <A, B> Kind<F, Either<A, B>>.select(f: Kind<F, (A) -> B>): Kind<F, B> = selectM(f)
+
+  override fun <A, B> Kind<F, A>.lazyAp(ff: () -> Kind<F, (A) -> B>): Kind<F, B> =
```
1Jajen1: Because flatMap is actually lazy for all types that can short circuit it is a really good fit for a default and because of monad-applicative consistency laws this is also always the same behaviour

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/ParApplicative.kt
```diff
@@ -20,6 +20,8 @@ internal fun <F> Concurrent<F>.ParApplicative(ctx: CoroutineContext? = null): Ap
   override fun <A, B> Kind<F, A>.ap(ff: Kind<F, (A) -> B>): Kind<F, B> =
     _ctx.parMapN(ff, this@ap) { f, a -> f(a) }
 
+  override fun <A, B> Kind<F, A>.lazyAp(ff: () -> Kind<F, (A) -> B>): Kind<F, B> = ap(ff())
```
1Jajen1: @nomisRev This is needed to have `parTraverse` pass tests. This implies the old behaviour of `traverse` where the traversable functor gets folded first to build up an `IO` action which then gets executed. Is this a necessity when working with `parTraverse` or can we somehow change this to a step by step process as well? This would mean having a sensible implementation of `lazyAp` which still preserves `parTraverse`'s functionality. (The benefit would be the same as the benefits `traverse` gets from using `lazyAp`)

1Jajen1:
This also changes the order of `traverse` to left->right (as it should be!) for some datatypes.This meant two tests failed because they assumed the other order (Resource, which was an easy fix, and an optics test. The optics test was caused by `ConstApplicative` evaluating `ap` in reverse order)

1Jajen1:
So I figured out why using traverse on sequences was awfully slow:
It internally uses `foldRight` which has it's implementation copied from `ListK`:
```kotlin
fun <B> foldRight(lb: Eval<B>, f: (A, Eval<B>) -> Eval<B>): Eval<B> {
    fun loop(fa_p: ListK<A>): Eval<B> = when {
      fa_p.list.isEmpty() -> lb
      else -> f(fa_p.fix().list.first(), Eval.defer { loop(fa_p.list.drop(1).k()) })
    }
  }
```
This looks perfectly reasonable however when used with sequences (which is almost the same code, the differences do not matter for this) it slowed down drastically after about 1-2k iterations. The reason is the call to `drop`, which for lists creates a new list with one item less. For sequences it creates a `DroppingSequence` which keeps track of how many things do drop. This happens at each level (Dropping sequences flatten out when you call drop on a droppingsequence). What this means is: The call to `first()` has to start from 0 every single time. I changed `foldRight` to directly use the `Iterator` of a sequence to bypass all of this bs.

1Jajen1:
@arrow-kt/maintainers This now solves all problems `traverse` with sequences has:
- It is now fast
- No stackoverflows because accumulation is done as a list
- Works with infinite sequences (only with datatypes that implement `lazyAp`, but that's all relevant ones)

So I think this is now done, please review. Improving the `parApplicative` would be nice, but that can be done later on.

raulraja:
@1Jajen1 thanks so much for adressing and untangling this issue!

PR #1870, opened 2019-12-20
IO: IOConnection & IOBracket

nomisRev:
- Updates IOBracket
- Removes KindConnection and refactors to internal IOConnection
- Refactor ForwardCancelable
- Implement `IO<E, ?>` typeclasses up to Async

Nit: wishing for a nicer `ExitCase2` solution.


modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -862,7 +863,7 @@ sealed class IO<out E, out A> : IOOf<E, A> {
 
     companion object {
       // Internal reusable reference.
-      internal val makeUncancelable: (IOConnection) -> IOConnection = { KindConnection.uncancelable }
```
JorgeCastilloPrz: How does cancellation work for any `F` without `KindConnection`?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/KindConnection.kt
```diff
@@ -1,346 +0,0 @@
-package arrow.fx
```
JorgeCastilloPrz: Feels like you're bringing the old `IOConnection` back and no `KindConnection` anymore. Could you give more details on this?

modules/fx/arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxK.kt
```diff
@@ -207,27 +206,6 @@ data class FluxK<out A>(val flux: Flux<out A>) : FluxKOf<A> {
      * }
      * ```
      */
-    @Deprecated(message =
```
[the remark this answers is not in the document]
nomisRev: They were deprecated in `0.10.3`

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
nomisRev: A while ago we decided not to expose `KindConnection` anymore because it's extremely error prone due to the user being responsible for `push`/`popping` cancelation tasks.

Instead we offer a couple other combinators such as `onCancel` which you can use to register a cancel token, or `cancelable`. This way we can take away that complexity from the user and we can offer a simpler API.

I've been considering merging all those APIs tho in something similar to RxJava/Flow's `create`. We can keep track of how many tasks the  user registers, and cancel them ourselves, or pop them when the effect finishes. We can also offer them a more idiomatic API (I think) compared to `cb: (Either<Throwable, A> -> Unit) ->` and `cb(Right("Hello World!"))`

```
val io = IO.create<DomainError, String> { // Receiver enabled syntax 
      success("Hello World")
      //error(DomainError)
     //exception(RuntimException("Boom"))
    onCancel(IO { println("I get printed if this task gets cancelled") })
    onCancel(IO { println("multiple tasks can be registered") })
}
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/KindConnection.kt
nomisRev: That's correct. `KindConnection` is removed from public API and `IOConnection` becomes internal in `arrow.fx`. See other reply about why `KindConnection` was too error-prone.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
raulraja: Anything that improves in the state of `IO.async { conn, cb ->` is welcome. Many users including myself have ran into the issue where the async constructor blindly does not return because we forget to call `cb`

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IOConnection.kt
```diff
@@ -1,48 +1,98 @@
...
-typealias IOConnection = KindConnection<ForIO>
+enum class OnCancel {
+  ThrowCancellationException, Silent;
+
+  companion object {
+    val CancellationException = ConnectionCancellationException
+  }
+}
+
+object ConnectionCancellationException : JavaCancellationException("User cancellation")
```
pakoito: Could we double check that these can be compared to themselves?

nomisRev: You mean add a test that checks for reference checking?

```
//IOTest.kt

"ConnectionCancellationException can be reference checked" {
  ConnectionCancellationException shouldBe ConnectionCancellationException
}
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/io.kt
```diff
@@ -43,127 +48,156 @@ import arrow.typeclasses.MonadThrow
...
-
-  override fun <A> IOOf<A>.guarantee(finalizer: IOOf<Unit>): IO<Nothing, A> =
-    fix().guarantee(finalizer)
-
-  override fun <A> IOOf<A>.guaranteeCase(finalizer: (ExitCase<Throwable>) -> IOOf<Unit>): IO<Nothing, A> =
-    GuaranteeCase(finalizer)
+interface IOBracket<E> : Bracket<IOPartialOf<E>, Throwable>, IOMonadThrow<E> {
+  override fun <A, B> IOOf<E, A>.bracketCase(release: (A, ExitCase<Throwable>) -> IOOf<E, Unit>, use: (A) -> IOOf<E, B>): IO<E, B> =
+    // Capture `E` into `Either`
+    RedeemWith({ t -> IO.raiseException<Either<E, A>>(t) }, { e -> IO.just(Left(e)) }, { a -> IO.just(Right(a)) })
```
pakoito: Maybe this should be implemented outside the extensions file.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/io.kt
```diff
@@ -43,127 +48,156 @@ import arrow.typeclasses.MonadThrow
...
+      .BracketCase<Either<E, A>, E, Either<E, B>>(release = { a, ex ->
+        when (a) {
+          is Either.Right -> { // Release resource.
+            when (ex) {
+              ExitCase2.Completed -> release(a.b, ExitCase.Completed).fix()
+              ExitCase2.Canceled -> release(a.b, ExitCase.Canceled).fix()
+              is ExitCase2.Exception -> release(a.b, ExitCase.Error(ex.exception)).fix()
+              is ExitCase2.Error -> throw AssertionError("Unreachable") // E is `Nothing`.
```
pakoito: Maybe worth setting the original exception as the origin of this one.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/internal/IOBracket.kt
```diff
@@ -1,23 +1,24 @@
...
+import arrow.fx.typeclasses.ExitCase2
 import kotlinx.atomicfu.atomic
 
 internal object IOBracket {
 
   /**
    * Implementation for `IO.bracketCase`.
    */
-  operator fun <A, B> invoke(acquire: IOOf<A>, release: (A, ExitCase<Throwable>) -> IOOf<Unit>, use: (A) -> IOOf<B>): IO<Nothing, B> =
+  operator fun <E, E2 : E, A, B> invoke(acquire: IOOf<E, A>, release: (A, ExitCase2<E>) -> IOOf<E2, Unit>, use: (A) -> IOOf<E2, B>): IO<E2, B> =
```
pakoito: I'm confuesed here, is it because E is invariant but we need it in a contravariant position?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/Bracket.kt
```diff
@@ -22,6 +22,19 @@ sealed class ExitCase<out E> {
 fun <E> Either<E, *>.toExitCase() =
   fold(::Error) { ExitCase.Completed }
 
+sealed class ExitCase2<out E> {
```
pakoito: Not a fan of versioning in the name, if could be `ExitCases` or something similar.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/io.kt
nomisRev: This is not an exception but an `Error` of type `Nothing`.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/Bracket.kt
nomisRev: Not a fan of the name either 😅

I was hoping to do something with `sealed class ExitCase2<out E, out E2> : ExitCase<E>()` but that's weird since `ExitCase` has `Error` which should be `E2` according to the current `IO` terminology 🤕

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IOConnection.kt
pakoito: Yes, something like that :D `object` have comparison overriden anyway, right? There was some edge case where you could instantiate a second one after a rotation or something IIRC.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/io.kt
pakoito: What's the type in the field in ExitCase2.Error then?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
aballano: > val io = IO.create<DomainError, String> { // Receiver enabled syntax 
>      onCancel(IO { println("I get printed if this task gets cancelled") })
>      onCancel(IO { println("multiple tasks can be registered") })
> 
>       success("Hello World")
>       //error(DomainError)
>      //exception(RuntimException("Boom"))
> }

My 2 cents here regarding receiver syntax: having to guess/learn from the devs perspective which functions to call is a bit cumbersome and something I myself find more attractive from IO or Observable's creation function, where you simply get a function/interface callback parameter to call over.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/io.kt
nomisRev: `Nothing` so can also return `ex.error` which is `Nothing` so no need to throw.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IOConnection.kt
nomisRev: Yes, I think it's possible to run in a weird double instance scenario but thought that was only possible in Android user land. When you have nested objects or inherit from certain lifecycle aware classes? 🤔

They should never be able to break a library object like `ConnectionCancellationException`, right? That'd be very scary if that'd be possible by accident.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
nomisRev: @aballano that's a very valid comment and something I was also weary about. Thanks for bringing it up ;)

So, in this case, it would be better? That'd be more in line with RxJava, and the example above is in line with `Flow`. We can also do a mix and match, so people can choose. The syntax is available implicit but also available concretely on `emitter` or whatever it's named.

```kotlin
IO.create { sink ->
   sink.success("Hello World")
  //sink.error(DomainError)
 //sink.exception(RuntimException("Boom"))
}
```

Issue #1858, opened 2019-12-19
Remove last usages of cf / applicative just in tests


1Jajen1:
As per #1819 and as a follow up to #1844 (which already did a lot of work there :heart: ).

Currently a few laws still work with either `cf: (Int) -> Kind<F, Int>` or `Applicative.just` to generate test data, most notably the entire `Applicative` and upwards typeclasses.

There are two difficulties in doing this:
- It's quite a bit of work, this touches a lot of files
- Some GenK/EqK instances are not trivial (I can try help with those)

Also something else to keep in mind: This will probably fail a few tests, especially after #1843 is merged which brings the monad-x-consistency laws, which is a good thing ;)

1Jajen1:
I think this is mostly (if not entirely) done, so I'll close this now, so we don't have to move it.

PR #1844, opened 2019-12-16
provide testdata with Gen/GenK instead of construction functions

abendt:
regarding #1819 these PR changes some of the test laws (e.g FunctorLaws, FoldableLaws, etc) to use `Gen/GenK` for testdata generation.

In general i changed the actual tests of a Law to use `Gen` for testdata. 
Next i changed the `laws` function to accept `GenK`. Whereever i could not make that work from the test i added another `laws` function that accepts `Gen` to allow more control from the test side.

One example for that is `SetKTest` (MonoidalLaws): using GenK leads to big testdata sets and long run times. 
In other places i just didn't know how to properly use GenK instances there ...

PR #1875, opened 2019-12-21
ComonadLaws uses GenK for testdata generation

abendt:
this PR changes ComonadLaws to use GenK for test data generation (#1858)

i tried to come up with a better GenK for Cofree as suggested here (https://github.com/arrow-kt/arrow/pull/1844/files#r358260447) 

However the tests fail if the Nel has a size > 1. Not sure if this is related to a wrong GenK or is actually a bug uncovered ...


1Jajen1:
Yes, that is another bug ^-^. Number 3!

Please change `coflatMap` and `duplicate` in `Cofree` to this:
```kotlin
  fun <B> coflatMap(f: (Cofree<S, A>) -> B): Cofree<S, B> = Cofree(FS, f(this), tail.map { it.map { it.coflatMap(f) } })

  fun duplicate(): Cofree<S, Cofree<S, A>> = Cofree(FS, this, tail.map { it.map { it.duplicate() } })
```

Then you can remove the filter on the genK instance and run it with any nel again :)

1Jajen1:
Time to merge! Thanks :clap:

PR #1878, opened 2019-12-22
Minor update for Ank

i-walker:
updates ScrpitEngine version to 1.3.61


1Jajen1:
Should the script engine version be on par with the compiler version we are using project wide?
We are currently using "1.3.11" as the kotlin version (at least according to this: https://github.com/arrow-kt/arrow/blob/master/build.gradle#L22)

i-walker:
AFAIK we're using 1.3.50 here https://github.com/arrow-kt/arrow/blob/b79010ccc235b2bfff9523fa6909474eab89648c/build.gradle#L32. But yes, 
I can update both versions if that was the intention @1Jajen1.

1Jajen1:
Right, had the wrong line :)
I think we should have both always on the same version if possible. Don't think its a requirement tho

build.gradle
```diff
@@ -52,7 +51,7 @@ buildscript {
         classpath "com.github.ben-manes:gradle-versions-plugin:$gradleVersionsPluginVersion"
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.4'
-        classpath 'io.arrow-kt:ank-gradle-plugin:0.7.1'
+        classpath "io.arrow-kt:ank-gradle-plugin:$VERSION_NAME"
```
rachelcarmena: Hi @i-walker ! This change makes build-check fail because `0.10.5-SNAPSHOT` version is not found in the repositories.

However, the solution wouldn't be to add OSS repository to find that dependency because that `VERSION_NAME` will cause errors when updating the value of `VERSION_NAME` during release time (the source code cannot depend on an artifact that is deployed from that source code). 

Maybe that's the reason why that version is fixed from `0.7.1` because the Ank Gradle Plugin hasn't evolved since then. Some alternatives:

- Don't apply Arrow Ank Gradle Plugin. Add the `runAnk` task in `arrow-docs/build.gradle`.
- Exclude Arrow Ank Gradle Plugin. Change build and publish checks to control the order.

i-walker:
Closing this, due to ScriptEngine Errors in new versions of `kotlin-script-utils`. Either, the compilation process fails, because of unregistered Extensions prior to execution or unresolved errors.

i-walker:
Thanks @rachelcarmena and @1Jajen1 :)

PR #1879, opened 2019-12-22
Ensure effect order is preserved with Traversable

scott-christopher:
I discovered as part of #1832 that the recent changes to introduce `lazyAp` has resulted in the effects of `NonEmptyList.traverse` to no longer maintain their order, causing some of the `Queue` tests to fail.

My assumption was that it was desirable to maintain the effect order with the order of the elements being traversed, so I have included a new law to `Traversable` to compare effect order using a mutable `Ref` with evaluation order of the resulting values.

Upon introducing the new law, `FluxK`, ObservableK`, `FlowableK` were also failing to preserve the ordering of the effects, so I have included changes for those too. I'm not too familiar with these specific types, so it would be worth double checking that the intended semantics have been maintained.

### DIFF
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/TraverseLaws.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/TraverseLaws.kt
@@ -15,6 +15,9 @@ import arrow.core.extensions.monoid
 import arrow.core.fix
 import arrow.core.toT
 import arrow.core.value
+import arrow.fx.IO
+import arrow.fx.extensions.io.applicative.applicative
+import arrow.fx.extensions.io.concurrent.concurrent
 import arrow.mtl.typeclasses.ComposedApplicative
 import arrow.mtl.typeclasses.nest
 import arrow.mtl.typeclasses.unnest
@@ -64,7 +67,8 @@ object TraverseLaws {
         Law("Traverse Laws: Identity") { TF.identityTraverse(TF, GEN, EQ) },
         Law("Traverse Laws: Sequential composition") { TF.sequentialComposition(GEN, EQ) },
         Law("Traverse Laws: Parallel composition") { TF.parallelComposition(GEN, EQ) },
-        Law("Traverse Laws: FoldMap derived") { TF.foldMapDerived(GEN) }
+        Law("Traverse Laws: FoldMap derived") { TF.foldMapDerived(GEN) },
+        Law("Traverse Laws: Effect order preserved") { TF.effectOrderPreserved(GEN) }
       )
   }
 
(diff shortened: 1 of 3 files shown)


1Jajen1:
Btw what was the encountered order? Because that actually depends on the applicative instance one uses and not on the traverse implementation (at least now it does, before it was kind of weird)

scott-christopher:
It seems like the effect from the head of the non-empty list was occurring after the tail.

e.g.
```kotlin
val l = NonEmptyList(1, listOf(2, 3, 4, 5))
IO.fx {
  val q = !queue(l.size)
  !l.traverse(IO.applicative(), q::offer)
  val nl = !(1..l.size).toList().traverse(IO.applicative()) { q.take() }
  nl.fix()
}.unsafeRunSync() shouldBe l.toList()
```
```
expected: [1, 2, 3, 4, 5] but was: [2, 3, 4, 5, 1]
```

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/TraverseLaws.kt
```diff
@@ -121,4 +125,14 @@ object TraverseLaws {
       val mapped = fa.foldMap(Int.monoid(), f)
       mapped.equalUnderTheLaw(traversed, Eq.any())
     }
+
+  fun <F, A> Traverse<F>.effectOrderPreserved(GEN: Gen<Kind<F, A>>) = IO.concurrent().run {
+    forAll(GEN) { fa: Kind<F, A> ->
+      val foldableOrder = fa.foldLeft(emptyList()) { xs: List<A>, x: A -> xs + x }
```
scott-christopher: By the way, does some form of `Foldable.toList` exist anywhere?

1Jajen1:
> It seems like the effect from the head of the non-empty list was occurring after the tail.

That is a bug in `map2Eval`. https://github.com/arrow-kt/arrow/blob/3051b67bcc4298c5b1553be134bd4222da04f3c6/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Apply.kt#L152-L153
It switches the order when calling `map2`. It's also a pointless function because the lazy argument needs to get evaluated in all cases when forcing the result, which means it's never lazy... Probably should be removed.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/TraverseLaws.kt
1Jajen1: Not that I know of. That'd be an alias for `combineAll(ListK.monoid())` right?

scott-christopher:
> It is actually `product` where this happens

Ah, so it is. Shall I include a fix and test for that in this PR too?

scott-christopher:
This seems to fix the issue:

```diff
fun <A, B> Kind<F, A>.product(fb: Kind<F, B>): Kind<F, Tuple2<A, B>> =
-  fb.ap(this.map { a: A -> { b: B -> Tuple2(a, b) } })
+  ap(fb.map { b: B -> { a: A -> Tuple2(a, b) } })
```

But in doing so has identified another issue downstream in `Validated` that was seemingly depending on that behaviour, resulting in the combined error messages being concatenated in the wrong order. The following is also needed to resolve that:

```diff
fun <E, A, B> ValidatedOf<E, A>.ap(SE: Semigroup<E>, f: Validated<E, (A) -> B>): Validated<E, B> =
  fix().fold(
-   { e -> f.fold({ Invalid(SE.run { it.combine(e) }) }, { Invalid(e) }) },
+   { e -> f.fold({ Invalid(SE.run { e.combine(it) }) }, { Invalid(e) }) },
    { a -> f.fold(::Invalid) { Valid(it(a)) } }
  )
```

Perhaps I should replace the fixes in this current PR with those above?

scott-christopher:
I've replaced the commit with the change discussed above, as the original change was not addressing the underlying issue.

1Jajen1:
The ordering issues in applicative are probably due to most very old stuff from arrow coming from scala cats and there the parameter order on `ap` is reversed. The fix there is fine, there is a lot of rework in the tests going on atm trying to uncover more and more of these inconsistencies.
Anyway, thanks for doing this :clap: :clap:

PR #1889, opened 2019-12-25
Arrow jackson support

alphaho:
Ported NonEmptyListModule and OptionModule from [Jackson Module Scala](https://github.com/FasterXML/jackson-module-scala)

### DIFF
--- a/modules/integrations/arrow-integrations-jackson-module/build.gradle
+++ b/modules/integrations/arrow-integrations-jackson-module/build.gradle
@@ -1,0 +1,14 @@
+apply plugin: 'kotlin'
+
+dependencies {
+    compile project(':arrow-core-data')
+    compile project(':arrow-syntax')
+    compile "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlinVersion"
+    compile "com.fasterxml.jackson.module:jackson-module-kotlin:2.9.7"
+    testRuntime("org.junit.vintage:junit-vintage-engine:$jUnitVintageVersion")
+    testCompile project(':arrow-test')
+    testCompile "io.kotlintest:kotlintest-runner-junit5:$kotlinTestVersion"
+    testCompile "io.kotlintest:kotlintest-assertions-json:$kotlinTestVersion"
+}
+
+apply from: rootProject.file('gradle/gradle-mvn-push.gradle')
--- a/modules/integrations/arrow-integrations-jackson-module/gradle.properties
+++ b/modules/integrations/arrow-integrations-jackson-module/gradle.properties
@@ -1,0 +1,4 @@
+# Maven publishing configuration
+POM_NAME=Arrow-Integrations-Jackson-Module
+POM_ARTIFACT_ID=arrow-integration-jackson-module
+POM_PACKAGING=jar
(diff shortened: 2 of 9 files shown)


alphaho:
That sounds good. 

By the way, I'm not quite sure whether we should provide a Jackson serializer/deserializer for `Either`. As the JSON encoding for `Either` may be different from different people's views.

For example, in the beginning, I think it should be encoded as the underlying values, then I found out the [Jackson Module Scala](https://github.com/FasterXML/jackson-module-scala/blob/master/src/main/scala/com/fasterxml/jackson/module/scala/ser/EitherSerializer.scala) encodes it as a wrapped JSON object with field name as `l` or `r`.

And if we are encoding it as the underlying value, we may not be able to tell whether it's a left or a right value when it's `Either<T, T>` (both left and right are of the same type T)

raulraja:
I would look into how they are encoding the Result data type but I think following what others have done in other langs can help adoption since its already an established pattern and would work out of the box for those cases

alphaho:
aha, that makes sense. I will raise another PR for porting the Either module from Jackson Module Scala later

MarioAriasC:
Hi @alphaho, first of all, thanks for taking care of this, excellent job.

There is a bug on your implementation, probably ported from the scala code, but more likely is on the interaction within your code and the `KotlinModule` from Jackson itself (I know because I was working on a previous implementation for this and I had many troubles).

Your code is failing on `Option<Option<T>>` type if the value is something like `Some(None)`

```kotlin
"serializing Option and then deserialize it should be the same as before the deserialization" {
      assertAll(
        Gen.oneOf(
          Gen.option(Gen.someObject()).map { it to jacksonTypeRef<Option<SomeObject>>() },
          Gen.option(Gen.int()).map { it to jacksonTypeRef<Option<Int>>() },
          Gen.option(Gen.string()).map { it to jacksonTypeRef<Option<String>>() },
          Gen.option(Gen.option(Gen.bool())).map { it to jacksonTypeRef<Option<Option<Boolean>>>() },
          Gen.option(Gen.bool()).map { it to jacksonTypeRef<Option<Boolean>>() }
        )
      ) { (option, typeReference) ->
        val roundTripped = mapper.writeValueAsString(option) pipe { mapper.readValue<Option<*>>(it, typeReference) }

        roundTripped shouldBe option
      }
    }
```

Again, good work and I hope that you can find a solution.

Cheers

alphaho:
Hi @MarioAriasC , thanks for raising this bug.

I haven't thought of this before. It reminds me of an article from[ Roman Elizarov talking about tagged and untagged union](https://medium.com/@elizarov/dealing-with-absence-of-value-307b80534903).

It seems the current encoding to turn `None` into JSON's `null` would undoubtedly lose the information if the source value is `Some(None)`.
Because `T??` is the same as `T?`, but `Option<Option<T>` is different from `Option<T>`. There can't be a one to one matching rule if we are serializing `Option<Option<T>` into JSON. Unless we change the way we serialize the Option value. :-(

However, as I learned from some arrow-meta discussions, it seems the direction Arrow is heading towards is to treat `Option<T>` "the same" as `T?`. So I think maybe it's ok to stick with the current JSON encoding. But feel free to correct me if I'm wrong.

PR #1891, opened 2019-12-26
use GenK for testdata generation

abendt:
use GenK to generate testdata. this is a part of (#1858)

modified laws are:
* BicrosswalkLaws
* BifoldableLaws
* BitraverseLaws

### DIFF
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Eq2K.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Eq2K.kt
@@ -7,8 +7,45 @@ import arrow.Kind2
  */
 interface Eq2K<F> {
 
+  /**
+   * Lifts the equality check provided by given Eq<A> and Eq<B> instances to Eq<Kind2<F, A, B>>
+   *
+   * {: data-executable='true'}
+   *
+   * ```kotlin:ank
+   * import arrow.core.extensions.either.eq2K.eq2K
+   * import arrow.core.extensions.*
+   * import arrow.core.*
+   *
+   * fun main(args: Array<String>) {
+   *    // sampleStart
+   *    val result = Either.eqK().run { Either.right("hello").eqK(Either.right("kotlin"), String.eq(), String.eq()) }
+   *    // sampleEnd
+   *    println(result)
+   * }
+   * ```
+   */
   fun <A, B> Kind2<F, A, B>.eqK(other: Kind2<F, A, B>, EQA: Eq<A>, EQB: Eq<B>): Boolean
 
+  /**
+   * Lifts the equality check provided by given Eq<A> and Eq<B> instances to Eq<Kind2<F, A, B>>
+   *
+   * {: data-executable='true'}
+   *
+   * ```kotlin:ank
+   * import arrow.core.extensions.either.eq2K.eq2K
+   * import arrow.core.extensions.*
+   * import arrow.core.*
+   *
+   * fun main(args: Array<String>) {
+   *    // sampleStart
+   *    val EQ = Either.eqK().liftEq(String.eq(), String.eq())
+   *    val result = EQ.run { Either.right("hello").eqv(Either.right("kotlin")) }
+   *    // sampleEnd
+   *    println(result)
+   * }
+   * ```
+   */
   fun <A, B> liftEq(EQA: Eq<A>, EQB: Eq<B>): Eq<Kind2<F, A, B>> =
     Eq { a, b ->
       this@Eq2K.run { a.eqK(b, EQA, EQB) }
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Eq2K.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Eq2K.kt
@@ -19,7 +19,7 @@ interface Eq2K<F> {
    *
    * fun main(args: Array<String>) {
    *    // sampleStart
-   *    val result = Either.eqK().run { Either.right("hello").eqK(Either.right("kotlin"), String.eq(), String.eq()) }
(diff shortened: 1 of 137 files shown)


abendt:
one question regarding the use of GenK/EqK for Kind2 types. what i like about the tests using unary type constructors the actual tests do not need any knowledge about the actual inner type.
in constrast the tests here always need to know of one of the inner types.
wouldn't it be better to use something like Gen2K/Eq2K here?

1Jajen1:
Good idea :+1:. With arity 2 we should have basically every typeclass covered.

abendt:
> Good idea 👍. With arity 2 we should have basically every typeclass covered.

i have prepared a quick spike so you can see how it would look like: https://github.com/abendt/arrow/tree/gen2k

seems to be a straightforward change

abendt:
i changed the laws to use Gen2K/Eq2K now

abendt:
i have merged in my other PR with CategoryLaws and ProfunctorLaws and changed the laws to use Gen2K/Eq2K too

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Eq2K.kt
```diff
@@ -0,0 +1,53 @@
+package arrow.typeclasses
+
+import arrow.Kind2
+
+/**
+ * The `Eq2K` typeclass abstracts the ability to lift the Eq class to binary type constructors.
+ */
+interface Eq2K<F> {
```
1Jajen1: I think `EqK2` is a better name as I would read it as "EqualityKind2".

modules/core/arrow-core-data/src/test/kotlin/arrow/core/AndThenTest.kt
```diff
@@ -119,9 +122,24 @@ private fun <A> AndThen.Companion.eqK() = object : EqK<AndThenPartialOf<A>> {
     }
 }
 
+private fun AndThen.Companion.eq2K() = object : Eq2K<ForAndThen> {
+  override fun <A, B> Kind2<ForAndThen, A, B>.eqK(other: Kind2<ForAndThen, A, B>, EQA: Eq<A>, EQB: Eq<B>): Boolean =
+    (this.fix() to other.fix()).let { (ls, rs) ->
+      EQB.run {
+        ls(1).eqv(rs(1))
```
1Jajen1: How does this even typecheck? o.O Shouldn't the input be `A` and not `Int`?

modules/docs/arrow-docs/docs/docs/arrow-core/typeclasses/eq2k/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Eq2K
+
+The `Eq2K` typeclass abstracts the ability to lift the Eq class to binary type constructors.
+
+### Main Combinators
+
+#### Kind2<F, A, B>.eqK(other: Kind2<F, A, B>, EQA: Eq<A>, EQB: Eq<B>): Boolean
+
+Compares two instances of a type with binary typeconstructor in the context of `F` using the provided `Eq<A>` and `Eq<B>`. Returns true if they're considered equal.
```
1Jajen1:
```suggestion
Compares two instances of a type `F` with binary a type constructor using the provided `Eq<A>` and `Eq<B>`. Returns true if they're considered equal.
```

modules/docs/arrow-docs/docs/docs/arrow-core/typeclasses/eq2k/README.md
```diff
@@ -0,0 +1,46 @@
...
+// lift String Eq to Either
+Either.eq2K().run {
+    Either.right("hello").eqK(Either.right("arrow"), String.eq(), String.eq())
+}
+```
+
+### Laws
+
+Arrow provides `Eq2KLaws` in the form of test cases for internal verification of lawful instances and third party apps creating their own `Eq2K` instances.
+See the existing EqK instances implementations and accompanying tests for reference.
```
1Jajen1:
```suggestion
See the existing EqK2 instances implementations and accompanying tests for reference.
```
Change this to `Eq2K` if we keep that name

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Eq2K.kt
abendt: yes makes sense. i have changed it

modules/core/arrow-core-data/src/test/kotlin/arrow/core/AndThenTest.kt
abendt: i just adapted the existing code 😁 isn't 'A' just resolved to 'Int' here?

modules/docs/arrow-docs/docs/docs/arrow-core/typeclasses/eqk2/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## EqK2
+
+The `EqK2` typeclass abstracts the ability to lift the Eq class to binary type constructors.
+
+### Main Combinators
+
+#### Kind2<F, A, B>.eqK(other: Kind2<F, A, B>, EQA: Eq<A>, EQB: Eq<B>): Boolean
+
+Compares two instances of a type `F` with binary a type constructor using the provided `Eq<A>` and `Eq<B>`. Returns true if they're considered equal.
```
1Jajen1:
```suggestion
Compares two instances of a type `F` with a binary type constructor using the provided `Eq<A>` and `Eq<B>`. Returns true if they're considered equal.
```
My bad :man_facepalming:

1Jajen1:
```suggestion
Compares two instances of a type `F` with a binary type constructor using the provided `Eq<A>` and `Eq<B>`. Returns true if they're considered equal.
```
My bad :man_facepalming:

PR #1892, opened 2019-12-27
IO Concurrency combinators & Green Build

nomisRev:
This PR rewrite the old `IO<A>` concurrency operators to their new `IO<E, A>` equivalent.
Since this is the last set of combinators to be refactored I'm also fixing the tests in this PR.

 - [x] Rewrite Fork
 - [x] Rewrite ParMapN
 - [x] Rewrite RaceN
 - [x] Fix instances
 - [x] Arrow Fx compiles
 - [x] Arrow Fx Test compiles
 - [x] Arrow Fx Tests pass

That would leave the runners, documentation and finalizing the final API for the subsequent PRs.


modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -764,8 +699,16 @@ sealed class IO<out E, out A> : IOOf<E, A> {
    * @see [unsafeRunAsyncCancellable] to run in a cancellable manner.
```
pakoito: This link may be broken now

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -764,8 +699,16 @@ sealed class IO<out E, out A> : IOOf<E, A> {
...
    * @see [runAsync] to run in a referential transparent manner.
    */
-  fun unsafeRunAsyncCancellable(onCancel: OnCancel = Silent, cb: (IOResult<E, A>) -> Unit): Disposable =
-    runAsyncCancellable(onCancel, cb andThen { unit }).unsafeRunSync().fold(::identity, ::identity)
+  fun unsafeRunAsyncCancellableEither(onCancel: OnCancel = Silent, cb: (IOResult<E, A>) -> Unit): Disposable =
```
pakoito: New name for binary comp purposes?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -779,7 +722,7 @@ sealed class IO<out E, out A> : IOOf<E, A> {
    * @see [unsafeRunAsync] or [unsafeRunAsyncCancellable] that run the value as [Either].
    * @see [runAsync] to run in a referential transparent manner.
    */
-  fun unsafeRunSync(): Either<E, A> =
+  fun unsafeRunSyncEither(): Either<E, A> =
```
pakoito: Same, signature in name?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -1173,3 +1116,67 @@ fun <A, E, B> IOOf<E, A>.bracketCase(release: (A, ExitCase2<E>) -> IOOf<E, Unit>
...
+    // A new IOConnection, because its cancellation is now decoupled from our current one.
+    val conn = IOConnection()
+    IORunLoop.startCancelable(IOForkedStart(this, ctx), conn, promise::complete)
+    cb(IOResult.Success(IOFiber(promise, conn)))
+  }
+
+fun <A> IOOf<Nothing, A>.unsafeRunSync(): A =
+  fix().unsafeRunSyncEither()
+    .fold({
+      TODO()
```
pakoito: TODO

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -1173,3 +1116,67 @@ fun <A, E, B> IOOf<E, A>.bracketCase(release: (A, ExitCase2<E>) -> IOOf<E, Unit>
...
+    }, ::identity)
+
+fun <A> IOOf<Nothing, A>.unsafeRunAsync(f: (Either<Throwable, A>) -> Unit): Unit =
```
pakoito: I seeeeee

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
nomisRev: Needs to be replaced by `throw ArrowException("Impossible case: ...")`

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/EffectsSuspendDSLTests.kt
```diff
@@ -80,7 +81,7 @@ class EffectsSuspendDSLTests : UnitSpec() {
     "raiseError" {
       shouldThrow<TestError> {
         fxTest {
-          IO.fx {
+          IO.fx<Nothing, Int> {
```
pakoito: <Int, Nothing>?

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/IOTest.kt
```diff
@@ -187,56 +187,6 @@ class IOTest : UnitSpec() {
       }
     }
 
-    "should complete when running a pure value with runAsync" {
```
pakoito: These should be kept or updated to explain the behavior IMO

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
pakoito: Ye, good idea to declare these on the IOOf to avoid confusion. The suffix Either doesn't thrill me, still, I don't have a better name for it

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/EffectsSuspendDSLTests.kt
nomisRev: It respects the order of the type.

```
IO.fx<Error, Value> { ... }
Either.fx<Left, Right> { ... }
Option.fx<Value> { ... } <--- Single type arguments can be infered.
```

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/IOTest.kt
nomisRev: 👍 I'll revert these changes and we can keep them until we can fix them

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
nomisRev: Hoping to get docs green in following PR and then we can move on to polishing final API and elaborate in docs.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORaceTriple.kt
```diff
@@ -48,103 +45,137 @@ interface IORaceTriple {
    *
    * @see [arrow.fx.typeclasses.Concurrent.raceN] for a simpler version that cancels losers.
    */
-  fun <A, B, C> raceTriple(ctx: CoroutineContext, ioA: IOOf<A>, ioB: IOOf<B>, ioC: IOOf<C>): IO<Nothing, RaceTriple<ForIO, A, B, C>> =
+  fun <E, A, B, C> raceTriple(ctx: CoroutineContext, ioA: IOOf<E, A>, ioB: IOOf<E, B>, ioC: IOOf<E, C>): IO<E, RaceTriple<IOPartialOf<E>, A, B, C>> =
```
JorgeCastilloPrz: Probably not the time for this, but these race methods seem to have a lot of code replicated for each one of the cancellables started. It'd be a good idea to pull things out and pass the connections as args.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
nomisRev: > The suffix Either doesn't thrill me, still, I don't have a better name for it
So we can agree there :D Naming is so hard 😭

I attempted to overload with `@JvmName` but the concrete version always takes precedence so that doesn't work. Similar issue if we define both as extensions I think.

modules/benchmarks/arrow-benchmarks-fx/src/jmh/kotlin/arrow/benchmarks/Cancellable.kt
```diff
@@ -25,7 +27,7 @@ open class Cancellable {
   var size: Int = 0
 
   fun evalCancelable(n: Int): IO<Nothing, Int> =
-    IO.concurrent().cancelable<Int> { cb ->
+    IO.concurrent<Nothing>().cancelable<Int> { cb ->
```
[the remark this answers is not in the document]
nomisRev: Sadly yes.

It failed to add an overload `fun IO.Companion.concurrent(): Concurrent<IOPartialOf<Unit>>`, but should try to find a good solution again.
I haven't chased down nice APIs, only functionality and getting this compiling and testing again.

modules/core/arrow-test/src/main/kotlin/arrow/test/generators/Generators.kt
```diff
@@ -132,8 +132,8 @@ fun <E, A> Gen.Companion.validated(genE: Gen<E>, genA: Gen<A>): Gen<Validated<E,
 fun <A> Gen.Companion.`try`(genA: Gen<A>, genThrowable: Gen<Throwable> = throwable()): Gen<Try<A>> =
   Gen.either(genThrowable, genA).map { it.fold({ Failure(it) }, { Success(it) }) }
 
-fun <A> Gen.Companion.nonEmptyList(gen: Gen<A>): Gen<NonEmptyList<A>> =
-  gen.flatMap { head -> Gen.list(gen).map { NonEmptyList(head, it) } }
+fun <A> Gen.Companion.nonEmptyList(gen: Gen<A>, maxSize: Int = 100): Gen<NonEmptyList<A>> =
```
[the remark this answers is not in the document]
nomisRev: Yes, this was to speed up some RxJava tests. I think it performs similarly to a monad transformer atm, but that should be lifted with Arrow Meta.

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/SemaphoreTest.kt
```diff
@@ -138,7 +138,7 @@ class SemaphoreTest : UnitSpec() {
     tests("CancelableSemaphore") { Semaphore(it, IO.concurrent()) }
 
     "CancelableSemaphore - supports cancellation of acquire" {
-      Semaphore(0, IO.concurrent()).flatMap { s ->
+      Semaphore(0, IO.concurrent<Nothing>()).flatMap { s ->
```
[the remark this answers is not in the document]
nomisRev: See comment above. I failed to add this overload but haven't digged deep. Maybe by using `@JvmName` or similar we can.

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
```diff
@@ -106,115 +121,150 @@ class QueueTest : UnitSpec() {
...
+      label = "MaybeK: BoundedQueue",
+      queue = { capacity -> Queue.bounded(capacity, MaybeK.concurrent()) },
+      EQ = MaybeK.eq()
+    )
+  }
+}
+
+private fun <T> MaybeK.Companion.eq(): Eq<MaybeKOf<T>> = object : Eq<MaybeKOf<T>> {
+  override fun MaybeKOf<T>.eqv(b: MaybeKOf<T>): Boolean {
+    val res1 = arrow.core.Try { value().timeout(5, java.util.concurrent.TimeUnit.SECONDS).blockingGet() }
```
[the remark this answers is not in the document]
nomisRev: It'll be removed in `0.11.0` but removing `Try` is unrelated to this effort.
We should create a ticket with `0.11.0` deadline.

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/IOTest.kt
nomisRev: @pakoito I was just looking at these tests. None of these actually run, only 1 since only one of the `runAsync` tests calls `unsafeRunSync`.

I fixed and pushed them, but we should really get rid of these combinators. Super confusing and they don't do anything.

PR #1895, opened 2019-12-29
use GenK to generate testdata (ProfuctorLaws/CategoryLaws)

abendt:
use GenK to generate testdata. this is a part of (#1858)

modified laws are:
* ProfunctorLaws
* CategoryLaws


abendt:
is part of #1891  now

PR #1896, opened 2019-12-30
Make Kleisli.map stacksafe

1Jajen1:
From https://kotlinlang.slack.com/archives/C8UK6RTHU/p1577206350057000 .

- Made Kleisli.map and also some other combinators stacksafe using AndThen
- Deleted faulty CofreeTest. There is no point testing if something is not stacksafe, it is not a failure if it turns out to be stacksafe later
- Raised stacksafety test range to 20k. The tests were flaky with lower numbers but started to fail consistently at ~20k. ~~Not sure if we want to keep this because this is slow...~~ Fully build + test is pretty similar with other prs, so it is probably not that bad

Also another thing: StateT and Kleisli are both not stacksafe unless used with a stacksafe monad (like IO, Eval, Trampoline). Is that by design? Because we have State and Reader as aliases with Id which means they are not stacksafe.


pakoito:
>Also another thing: StateT and Kleisli are both not stacksafe unless used with a stacksafe monad (like IO, Eval, Trampoline). Is that by design?

ye, should be

pakoito:
>Deleted faulty CofreeTest. There is no point testing if something is not stacksafe, it is not a failure if it turns out to be stacksafe later

This is more like a smell test, if anything changes in the future we'll know. I'd keep it.

1Jajen1:
> This is more like a smell test, if anything changes in the future we'll know. I'd keep it.

I deleted it because @adomokos wrote in slack that it failed pretty consistently for him. For me it keeps passing, same as ci. But given that the MonadDefer stacksafety tests also pass on ci, but fail locally for me (after upping the number of repetitions to 20k), I'd like to avoid tests like this. We can try upping the number of repetetions here as well as it is just one test, but eventually if this keeps consistently failing we'll have to remove it.

For now I'll bring it back with 20k iterations instead of 10k

adomokos:
I've just checked your branch, ran the tests on it. This test is still failing:
```
arrow.free.CofreeTest > cata with an stack-unsafe monad should blow up the stack FAILED
    java.lang.AssertionError: Run should overflow on a stack-unsafe monad
        at arrow.free.CofreeTest$9$1.invokeSuspend(CofreeTest.kt:156)
        at arrow.free.CofreeTest$9$1.invoke(CofreeTest.kt)
```

Can it be isolated as flaky?

I am on Mac OS 10.14.6, Java version AdoptOpenJDK `8.0.232.j9-adpt`.

PR #1897, opened 2019-12-30
Add some convenience methods to alternative

1Jajen1:
While working on a parser combinator library (https://github.com/1Jajen1/Kparsec) I noticed that alternative could use some convenience methods. I also introduced the alternative equivalent of `lazyAp` because `asum` is very similar to `sequence` just that it shorts on success rather than failure. Having `lazyOrElse` gives the same benefits as `lazyAp`.


1Jajen1:
The added methods break the `@extension` code generator for some weird reason (only on lists and sequences), so I moved them out of the typeclass for now with a todo to move them back when arrow-meta comes around and hopefully does not have this issue ^^

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Alternative.kt
```diff
@@ -48,4 +54,35 @@ interface Alternative<F> : Applicative<F>, MonoidK<F> {
...
+}
+
+fun <T, F, A> Kind<T, A>.afold(AF: Alternative<F>, FT: Foldable<T>): Kind<F, A> = FT.run { toList().afromList(AF) }
+
+fun <F, A> List<A>.afromList(AF: Alternative<F>): Kind<F, A> = map { AF.just(it) }.k().asum(AF, object : Foldable<ForListK> {
+  override fun <A, B> Kind<ForListK, A>.foldLeft(b: B, f: (B, A) -> B): B = fix().foldLeft(b, f)
+  override fun <A, B> Kind<ForListK, A>.foldRight(lb: Eval<B>, f: (A, Eval<B>) -> Eval<B>): Eval<B> = fix().foldRight(lb, f)
+})
+
+fun <F, A> Option<A>.afromOption(AF: Alternative<F>): Kind<F, A> = fold({ AF.empty() }, { AF.just(it) })
```
pakoito: `altFromOption`?

1Jajen1:
The `StateT` test that is failing is flaky. It is similar to the Kleisli stackunsafety that was fixed in #1896 . 
I am pretty sure it is because:
https://github.com/arrow-kt/arrow/blob/a0d8c7043ca8b96729cbdef3be5cf7216ff27cec/modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/StateT.kt#L214-L220
Uses `andThen` which is not stacksafe, the fix should be easy, just wrap `sfsa` in `AndThen` and we should be fine (I hope ^^)

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Alternative.kt
```diff
@@ -70,19 +70,19 @@ interface Alternative<F> : Applicative<F>, MonoidK<F> {
 }
 
 // TODO move back to alternative once arrow-meta is used to handle @extension because currently this breaks for lists and sequences
-fun <T, F, A> Kind<T, Kind<F, A>>.asum(AF: Alternative<F>, FT: Foldable<T>): Kind<F, A> = AF.run {
+fun <T, F, A> Kind<T, Kind<F, A>>.altSum(AF: Alternative<F>, FT: Foldable<T>): Kind<F, A> = AF.run {
```
pakoito: Like it.

PR #1898, opened 2019-12-30
Change ap argument order

### LINKED ISSUE #1000: Flipped infix Applicative::ap

1Jajen1:
Closes #1000 and continues #1012 .

This is a breaking change, so it should probably wait for something like 0.11 or something. Although the change to `apPipe` is easy and most users probably use applicative through `map` or other methods anyway...

This does not change evaluation order, `ap` will still be left to right, same for `apPipe`.

~~`apPipe` should not get a default implementation `ff.ap(this)` because that would evaluate right to left, which especially when dealing with effects is very confusing imo.~~ `map(this, ff) { (a, f) -> f(a) }` will work fine

I started on arrow-core without having a default for `apPipe` which is why it's overriden there but not anywhere else. That should not matter.

All that is left now is a pass on the docs


1Jajen1:
This should be ready now. Should `ap` be infix as the issue and the previous pr noted?

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Alternative.kt
```diff
@@ -18,15 +20,15 @@ interface Alternative<F> : Applicative<F>, MonoidK<F> {
    * @receiver computation to repeat.
    * @returns the collection of results with at least 1 repetition.
    */
-  fun <A> Kind<F, A>.some(): Kind<F, SequenceK<A>> = lazyAp { many().map { { a: A -> (sequenceOf(a) + it).k() } } }
+  fun <A> Kind<F, A>.some(): Kind<F, NonEmptyList<A>> = map { a: A -> { xs: ListK<A> -> Nel(a, xs) } }.lazyAp { many() }
```
1Jajen1: These type signatures fit better, I ran into those because lazyAp changed and noticed that there is no point using `Sequence` (also using + there can lead to stackoverflows). This is potentially infinite, but you can't access the return value anyway in those cases because the effect of F is getting applied forever before you even get there.

modules/docs/arrow-docs/docs/docs/arrow-core/typeclasses/alternative/README.md
```diff
@@ -63,43 +63,33 @@ Option.monadCombine().run {
...
 #### Kind<F, A>.some
 
-`fun <A> Kind<F, A>.some(): Kind<F, SequenceK<A>>`
+`fun <A> Kind<F, A>.some(): Kind<F, NonEmptyList<A>>`
 
-Repeats the current computation, lazily collecting its results into a sequence, until it fails. It is required that the computation succeeds at least once.
+Repeats the current computation, collecting its results into a non-empty list, until it fails. This will loop forever if the computation cannot or does not fail.
 
-```kotlin:ank
+```kotlin
```
1Jajen1: This and the many example cannot be run with Option because Option is pure and thus never fails when retried. Something like OptionT<ForIO> or similar is better suited here. (Best would be a parser example)

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Applicative.kt
```diff
@@ -17,7 +17,7 @@ interface Applicative<F> : Apply<F> {
   fun unit(): Kind<F, Unit> = just(Unit)
 
   override fun <A, B> Kind<F, A>.map(f: (A) -> B): Kind<F, B> =
-    ap(just(f))
+    just(f).ap(this)
```
pakoito: Cute

1Jajen1:
Should we wait for a 0.11 or something release? Because this is quite the breaking-change. The migration is easy, but still ^-^ (just replace existing ap code with apPipe)

pakoito:
IO already forces an 0.11 release with what we have in master. Let's get someone else in here to double stamp, I'm g2g!

cc @raulraja @nomisRev

nomisRev:
@pakoito no related to the new IO changes is in master?

pakoito:
Oh, right. So let's make a call then.

1Jajen1:
Since we are now on 0.11 snapshot should we merge this?

1Jajen1:
Or nvm, should I retarget this to merge into 0.11 snapshot? That is what I wanted to ask :D

nomisRev:
@1Jajen1 we're not on `0.11.x` yet but we're preparing a `0.11.0` release branch as we speak.

So we can merge this code there, it'll also produce a `0.11.0-SNAPSHOT` if people want to try it already.

1Jajen1:
> @1Jajen1 we're not on 0.11.x yet but we're preparing a 0.11.0 release branch as we speak.

> So we can merge this code there, it'll also produce a 0.11.0-SNAPSHOT if people want to try it already.

Yup noticed that as well, but you were slightly quicker :D
That would make sense I guess

1Jajen1:
Let's hold this until we have a plan for deprecation from #1957 , otherwise this could (if we end up not wanting `apPipe`) create extra work.

1Jajen1:
I removed `apPipe` for now. This should be good to go, I'll do another pr targeted at current master that has the deprecation notice for `ap`, which explains this change, when this is approved/merged.

rachelcarmena:

PR #1900, opened 2019-12-30
Fix OptionT applicative to short-circuit properly

1Jajen1:
Currently `OptionT.ap` always evaluated both arguments before checking the `Option` in the first one. This leads to fun things like `listOf(1,2,3,4).traverse(OptionT.monad(IO.monad())) { OptionT(IO { println(it); None }) }` printing 1,2,3,4 instead of just 1.


pakoito:
I believe this behavior is correct, right? `ap` is independent operations, it's just the continuation that should never be called.

1Jajen1:
`ap` is independent operations but this short circuit behaviour is expected. This gets especially clear when you look at the default implementation in monad, which is `flatMap { a -> ff.map { it(a) } }` and the consistency law that it should be equal to `ap` in all cases. Both of those things are also present in haskell and scala, so that is what I'd expect.

For reference the haskell implementation of `MaybeT`: http://hackage.haskell.org/package/transformers-0.5.6.2/docs/src/Control.Monad.Trans.Maybe.html#line-133

PR #1902, opened 2020-01-01
Adds UnsafeRun instance for MaybeK

i-walker:
MaybeK's unsafeRun instance


i-walker:
I'll remove those instances which don't fulfill UnsafeRun

Issue #1827, opened 2019-11-28
Functor instance Queue


nomisRev:
In #1728 Queue was added which is missing a `Functor` instance

kioba:
I can take a look at this issue

PR #1904, opened 2020-01-03
Add queue functor

kioba:
This PR is based on #1827 .

Queue was renamed for ConcurrentQueue and Queue will be the base interface.
This solution extends the type arguments to provide the functor functionality without an inverse map function but still be able to provide the original queue offer.


nomisRev:
Okay, I see so you can queue items of type `A` and dequeue elements of type `B`.
That's why FS2 also works with multiple typeclasses there.

I'm a bit wary if this will introduce unnecessary complexity for now.

scott-christopher:
If the extra type argument is introduced then it could make sense to implement `Profunctor` too, which would allow for contravariantly mapping over calls to `offer`.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/Queue.kt
```diff
@@ -1,14 +1,38 @@
...
 
+class ForQueue private constructor() {
+  companion object
+}
+
+typealias QueueOf<F, A, B> = arrow.Kind3<ForQueue, F, A, B>
+typealias QueuePartialOf<F, A> = arrow.Kind2<ForQueue, F, A>
+typealias QueueKindedJ<F, A, B> = HkJ3<ForQueue, F, A, B>
+
+@Suppress("UNCHECKED_CAST", "NOTHING_TO_INLINE")
```
pakoito: This isn't inline. Should it be?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/QueueFunctor.kt
```diff
@@ -0,0 +1,27 @@
...
+import arrow.fx.QueuePartialOf
+import arrow.fx.fix
+import arrow.typeclasses.Functor
+import arrow.undocumented
+
+@extension
+@undocumented
+interface QueueFunctor<F, E> : Functor<QueuePartialOf<F, E>> {
+  fun functorF(): Functor<F>
+  override fun <B, C> QueueOf<F, E, B>.map(f: (B) -> C): Queue<F, E, C> =
```
pakoito: Nit: Why B to C instead of A to B?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/QueueFunctor.kt
```diff
@@ -0,0 +1,27 @@
...
+import arrow.fx.QueueOf
+import arrow.fx.QueuePartialOf
+import arrow.fx.fix
+import arrow.typeclasses.Functor
+import arrow.undocumented
+
+@extension
+@undocumented
+interface QueueFunctor<F, E> : Functor<QueuePartialOf<F, E>> {
+  fun functorF(): Functor<F>
```
pakoito: I believe we're using a different naming convention for this.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/QueueFunctor.kt
```diff
@@ -0,0 +1,27 @@
...
+  override fun <B, C> QueueOf<F, E, B>.map(f: (B) -> C): Queue<F, E, C> =
+    object : Queue<F, E, C> {
+      override fun offer(a: E): Kind<F, Unit> = this@map.fix().offer(a)
+      override fun take(): Kind<F, C> = with(functorF()) {
```
pakoito: style-wise, we prefer to use `run`. It'll make it easier in the refactor to 1.0

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/QueueFunctor.kt
```diff
@@ -0,0 +1,27 @@
...
+        this@map.fix().take().map(f)
+      }
+
+      override fun size(): Kind<F, Int> = this@map.fix().size()
+      override fun awaitShutdown(): Kind<F, Unit> = this@map.fix().awaitShutdown()
+      override fun shutdown(): Kind<F, Unit> = this@map.fix().shutdown()
```
pakoito: Nit: do `this@map.fix()` once at the beginning of the function to cache it.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/QueueFunctor.kt
kioba: I had some issues with the names (#1905)and wasn't sure about the right solution. I will fix it!

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/QueueFunctor.kt
kioba: 🤔 good point, thanks for flagging

kioba:
We had a discussion with @nomisRev , and the suggestion was to only do invariants for the queue. I didn't know until know how FS2 was able to implement the functor for Queue but now I see the differences thanks for Simon! 
I will revert the type arguments back to a single one and implement only invariant for the queue but keep the interfaces.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/QueueInvariant.kt
```diff
@@ -0,0 +1,29 @@
...
+import arrow.typeclasses.Invariant
+import arrow.undocumented
+
+@extension
+@undocumented
+interface QueueInvariant<F> : Invariant<QueuePartialOf<F>> {
+  fun FR(): Functor<F>
+
+  override fun <A, B> QueueOf<F, A>.imap(f: (A) -> B, g: (B) -> A): Queue<F, B> =
+    FR().run {
```
pakoito: Android had problems with this pattern and you had to put the `run` block as close as the usage as possible. Hopefully it's okay now.

pakoito:
To prod it goes!

nomisRev:
@kioba saw all the changes you made! Looks amazing 👏👏👏

PR #1906, opened 2020-01-03
Fix tests for `arrow-meta` subproject

Romastyi:
Hi!
After I've forked repo and tried to run tests, I've found that tests for `arrow-meta` subproject are not running.
This PR fix this and modifying some test according current code generation.


raulraja:
there should be no arrow meta subproject as a that is now in a different repo in the org

rachelcarmena:
> there should be no arrow meta subproject as a that is now in a different repo in the org

Hi @raulraja, we moved (see #1742):
* `modules/meta/arrow-meta-prototype/` to the new repository: **arrow-meta**
* `modules/meta/arrow-annotations/` content from Arrow Meta branch to `master` branch (**arrow** repository) 

So `modules/meta` in **arrow** repository includes now:

* `modules/meta/arrow-annotations`
* `modules/meta/arrow-meta`

Should we remove `modules/meta/arrow-meta`?

nomisRev:
@rachelcarmena No, we shouldn't remove `modules/meta/arrow-meta` that is our current solution based on `kapt`.

The `kapt` impl will get removed when we deprecate it and replace it with Arrow Meta 🎉

Thanks for the contribution @Romastyi but we no longer maintain that implementation since it'll get replaced with [Arrow Meta](https://github.com/arrow-kt/arrow-meta/) soon.

Issue #909, opened 2018-06-24
@product code generation fails for Data classes with generic fields


richard-gibson:
example, code generated bu product annotation processor won't compile
```
@product
data class Foo<T>(val bar: T) {
  companion object
}
```

Happy to take a look at this if its not on anyone's radar but may need some guidance

nomisRev:
That would be great. If you need guidance feel free to ask on [gitter](http://gitter.im/arrow-kt) or #arrow in [kotlinlang slack](https://kotlinlang.slack.com/messages/C5UPMM0A0/) .

pakoito:
`@product` needs static resolution to create the generics, so my first intuition is that it shouldn't to work.

richard-gibson:
I took the code currently generated by the product annotation processor for a data class with 2 generic fields and made modifications by hand before looking at generator updates
You can see the diff here 
https://github.com/richard-gibson/arrow/commit/88c0254e54ea56ea7ab632f3640d133cd7d9d637

Any feedback or corrections welcome
I think it looks like the only problems will be to bring semigroups and monoids for generic fields into scope.  Field semigroup instances are needed to override the plus operator so we would need to find a way to bring them into scope without adding as method params, any ideas?

raulraja:
@richard-gibson I invite you to give it a shot and try to contribute those changes to the `ProductProcessor` and `ProductFileGenerator`, only then we are going to discover what the limitations are. Deriving type classes will be soon moved to a la carte `@product(deriving = ...)` where users will list which type classes or generic functionalities to derive so it's not the end of the world if we try to bake type args support in.

richard-gibson:
Thanks @raulraja will look at the processor and file generator in the meantime.

raulraja:
Thanks to everyone that participated here; we no longer have this processor, so closing this outdated issue.

PR #1907, opened 2020-01-04
Add deriving option for `@product` annotation

Romastyi:
This PR can be a partial fix for #909. 

For it I've added option for customising of type class deriving with `@product` annotation as suggested by @raulraja [here](https://github.com/arrow-kt/arrow/issues/909#issuecomment-399795215).   

For correct test running PR #1906 need to be merged.


Romastyi:
Hi! I'm very exciting about `arrow-meta` too. And if it's need I can try to help you with this functionality in `arrow-meta`. But I supposed that this PR can be merged before `arrow-meta` release in April.

PR #1914, opened 2020-01-09
WriterT inconsistencies

1Jajen1:
WriterT currently produces the accumulated value backwards. This failed a law in `MonadWriterLaws`. 

The reason this was not caught in tests is because we used a commutative monoid (Int addition) which does not care about order. I changed the tests to run with `ListK` to produce the failed tests.

The first problem was in `flatMap` where the writer returned by `f` was the first argument to combine and the original writer was second.
For reference the semantics in this pr now match the ones from haskell mtl: https://hackage.haskell.org/package/transformers-0.5.6.2/docs/src/Control.Monad.Trans.Writer.Strict.html#line-200

The second problem was that this change broke monad-applicative-consistency, which revealed that `WriterT` also applied the underlying effects in reverse order. Changed that too.


nomisRev:
Good catch!

Issue #1873, opened 2019-12-20
Change error handling tutorial to use a model different than nukes/bombs


raulraja:
https://arrow-kt.io/docs/patterns/error_handling/

Places where this needs to change: 

https://github.com/arrow-kt/arrow/blob/04b28f462e5de576c029db8aa8cac35d6ac70c9e/modules/docs/arrow-docs/docs/_code/core-home-code.md
https://github.com/arrow-kt/arrow/blob/44bb3419febf2cafaae3812ffcd64a448b5c076f/modules/docs/arrow-docs/docs/docs/arrow-core/patterns/errorhandling/README.md
https://github.com/arrow-kt/arrow/blob/93091c50314b22b2589455bfa127b735ce2da7d2/modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadError.kt

rachelcarmena:
Closed via #1915

PR #1915, opened 2020-01-09
Replaced nuclear war examples with cooking examples

kylegoetz:
addresses #1873


kylegoetz:
A couple notes:

1. there are blocks of code without `:ank` that seem to be out of date with the documentation or at least wouldn't compile. For example, in one of the three files I updated, there is an extension function for MonadError that takes `ME` as a parameter. There is no type specified for this parameter, and the subsequent functions using this extension function don't pass anything as a parameter. I left as is but for changing the names of the labels away from war terminology to food terminology.

2. Try has been deprecated. The error handling tutorial still has Try in it. Didn't remove because I wanted to keep this PR to just the terminology changes.

3. After updating the docs, due to my unfamiliarity with Kotlin toolkits so far, I struggled to run dokka and got a bunch of errors and decided to just push my commits and let someone with more experiene decide if I created those errors or if they were pre-existing.

kylegoetz:
Also, I changed the destructuring and `!` operator usages in the `fx` examples to use `.bind()` instead, having seen chatter on slack that these might be deprecated in favor of `.bind()` due to some issues with the compiler optimizing things wrongly.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadError.kt
```diff
@@ -55,27 +55,28 @@ interface MonadError<F, E> : ApplicativeError<F, E>, Monad<F> {
...
+ * class InsufficientAmount(val quantityInGrams : Int) : Throwable("You need $quantityInGrams more grams of ingredient")
  *
- * fun <F> MonadThrow<F>.arm(): Kind<F, Nuke> = just(Nuke)
- * fun <F> MonadThrow<F>.aim(): Kind<F, Target> = just(Target)
- * fun <F> MonadThrow<F>.launchImpure(target: Target, nuke: Nuke): Impacted {
- *   throw MissedByMeters(5)
+ * fun <F> MonadThrow<F>.takeFoodFromRefrigerator(): Kind<F, Lettuce> = just(Lettuce)
+ * fun <F> MonadThrow<F>.getKnife(): Kind<F, Knife> = just(Knife)
+ * fun <F> MonadThrow<F>.launchImpure(tool: Knife, ingredient: Lettuce): Salad {
+ *   throw InsufficientAmount(5)
```
sraaphorst: What are 5 lettuces? Maybe a more discrete vegetable like a carrot might be a better example?

Edit: I noticed the quantity in grams. 5 g of lettuce seems like a pretty small quantity to me.

I really like the idea of replacing the nuclear war example with the cooking example.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadError.kt
```diff
@@ -103,29 +104,29 @@ interface MonadThrow<F> : MonadError<F, Throwable> {
...
    *
-   * typealias Impacted = Boolean
+   * typealias SaladPrepared = Boolean
    *
-   * object Nuke
-   * object Target
-   * class MissedByMeters(private val meters: Int) : Throwable("Missed by $meters meters")
+   * object Lettuce
+   * object Knife
+   * class InsufficientAmount(val quantityInGrams : Int) : Throwable("You need $quantityInGrams more grams of ingredient")
```
sraaphorst: Oh, here it is. Never mind. It's not clear from the above.

raulraja:
@kylegoetz looks like there is some type checking issues in the snippets. As soon as you address those we can merge this. Thank you!

modules/docs/arrow-docs/docs/docs/arrow-core/patterns/errorhandling/README.md
```diff
@@ -298,59 +298,59 @@ Try.monadError()
...
 Let's now rewrite our program as a polymorphic function that will work over any datatype for which a `MonadError` instance exists.
 Polymorphic code in Arrow is based on emulated `Higher Kinds`, as described in [Lightweight higher-kinded polymorphism](https://www.cl.cam.ac.uk/~jdy22/papers/lightweight-higher-kinded-polymorphism.pdf) and applied to Kotlin, a lang which does not yet support Higher Kinded Types.
 
 ```kotlin
-fun <f> arm(ME: MonadError<F, NukeException>): Kind<F, Nuke> = ME.just(Nuke)
-fun <f> aim(ME: MonadError<F, NukeException>): Kind<F, Target> = ME.just(Target)
-fun <f> launch(target: Target, nuke: Nuke, ME: MonadError<F, NukeException>):
-  Kind<F, Impacted> = ME.raiseError(MissedByMeters(5))
+fun <F> takeFoodFromRefrigerator(ME: MonadError<F, CookingException>): Kind<F, Lettuce> = ME.just(Lettuce)
+fun <F> getKnife(ME: MonadError<F, CookingException>): Kind<F, Knife> = ME.just(Knife)
```
pakoito: Someone pointed out that this version of `aim`/`getKnife` takes `ME` as parameter instead of an extfun, when in the snippet below for prepare it's used in that position.

pakoito: Oh I saw you mentioned it. Yeah make it an extfun.

modules/docs/arrow-docs/docs/docs/arrow-core/patterns/errorhandling/README.md
```diff
@@ -168,48 +168,48 @@ How would our example look when implemented with `Try`?
...
 Just like it does for `Option`, Arrow also provides `Monad` instances for `Try`, and we can use it exactly in the same way
 
-```kotlin
+```kotlin:ank
 import arrow.typeclasses.*
 import arrow.core.extensions.*
 
-fun attackTry(): Try<Impacted> =
+fun prepareTry(): Try<Salad> =
   fx.monad {
```
1Jajen1:
```suggestion
  fx {
```
Ank fails with: `error: function invocation 'fx(...)' expected`. So why not try it ^^ If this fails you may need to check the imports again

kylegoetz: I could try it, but the documentation this code comes with is specifically highlighting the `fx.monad` functionality, so the documentation would become wrong if I did that.

kylegoetz: It turns out the issue is that I added `:ank` to check this code. The original did not have it. If you put `:ank` around the code before my commit, it errors out.

I'm going to remove the :ank markup to my code and update.

1Jajen1: `fx {...}` is an alias for `monad().fx.monad {...}` which most datatypes provide. If you want to keep it replace it with `Try.monad().fx.monad {...}`. But you might need to add an import for `monad()` then.

kylegoetz:
OK so I will explain the typechecking issues and my fix for posterity's sake. My PR initially errored out while building the docs. I did some investigation and discovered that I had changed one documentation block from `\`\`\`kotlin` to `\`\`\`kotlin:ank`.

This block was what was erroring out. Even using the code before my commit, but adding the `:ank` part would cause the same error.

So now the ank doc builder skips that block just like it always had been doing before. No more documentation build error.

That being said, it does show that the existing sample code cannot compile. Over on #1917 @pakoito and I have been talking a little about some code blocks that aren't currently compilable.

All of this latest talk is around some code specifically about a `Try` example. But `Try` is deprecated. So rather than stressing about this code any further, I'd suggest just stripping it out, since otherwise it's using a deprecated class, `Try`, in a README. Perhaps replace it with an `Either` example since `Either<Throwable, A>` is advised as the default replacement for `Try<A>`.

raulraja:
Let's remove Try from the example if that is an issue. We will have to
remove it anyway and Try is already deprecated.

kylegoetz:
@raulraja I'm about to create a PR for the docs today that removes the Try examples from everywhere I found them in the documentation that seemed appropriate. For example, in this file, I removed it. But in the library code for Try applicative, for example, I didn't remove documentation explaining the functions there since I think those docs should remain until the code itself is removed).

Basically all the documentation README.md files I either stripped the Try examples out or replaced them with Either or Option in order to not reduce the number of samples.

raulraja:
@kylegoetz thanks for the new PR and taking care of cleaning up the docs to remove `Try`!
I will approve this now and merge as soon as it's ready. you can then merge origin/master into your new branch and it should be fine against master.

raulraja:
@kylegoetz thanks!

PR #1920, opened 2020-01-10
Move AndMonoid to core and redefine Ior.Left and Ior.Right

i-walker:
Changes:
- Moving AndMonoid to core
- Redefine Ior.Left as Ior<A, Nothing> and Ior.Right as Ior<Nothing, A>. Inference resolution takes care of the rest.


modules/optics/arrow-optics/src/test/kotlin/arrow/optics/AndMonoidTest.kt
```diff
@@ -1,14 +0,0 @@
-package arrow.optics
```
pakoito: And this

pakoito: Damn git I see it now

i-walker: No Problem :)

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -1018,7 +1018,7 @@ sealed class IO<out A> : IOOf<A> {
     override fun unsafeRunTimedTotal(limit: Duration): Option<A> = unsafeResync(this, limit)
   }
 
-  internal data class Bind<E, out A>(val cont: IO<E>, val g: (E) -> IO<A>) : IO<A>() {
+  data class Bind<E, out A>(val cont: IO<E>, val g: (E) -> IO<A>) : IO<A>() {
```
1Jajen1: Is this intended?

i-walker: Oh, that was laying around from the initial intent of the PR. Thank you @1Jajen1

i-walker:
```suggestion
  internal data class Bind<E, out A>(val cont: IO<E>, val g: (E) -> IO<A>) : IO<A>() {
```

pakoito: You can go to the file in the Files Changed tab, click on the `...` and select `Edit File`. AFAIK it should commit to this PR directly :D

PR #1921, opened 2020-01-10
Update Ank to 1.3.50

i-walker:
Though it is up-to-date memory consumption of the ReplCompiler 1.3.50 is higher than the one from 1.3.11. 
🥑 Happy to get that done


modules/ank/arrow-ank/build.gradle
```diff
@@ -2,11 +2,13 @@ apply plugin: 'kotlin'
 apply plugin: 'kotlin-kapt'
 
 dependencies {
-    compile project(":arrow-core")
     compile project(":arrow-fx")
 
-    compile "org.jetbrains.kotlin:kotlin-compiler:$ankKotlinVersion"
-    compile "org.jetbrains.kotlin:kotlin-script-util:$ankKotlinVersion"
+    runtime "org.jetbrains.kotlin:kotlin-compiler:$kotlinVersion"
```
i-walker: These dependencies come from https://github.com/JetBrains/kotlin/blob/01a05a5495049668c0255c9b6f42adb17245a32b/libraries/examples/kotlin-jsr223-local-example/build.gradle.kts#L16
in order to use the ScriptEngine. That was one of the reasons it didn't work in #1878

build.gradle
```diff
@@ -46,13 +45,14 @@ buildscript {
...
         maven { url "https://dl.bintray.com/arrow-kt/arrow-kt/" }
+        maven { url "https://oss.jfrog.org/artifactory/oss-snapshot-local/" }
     }
 
     dependencies {
         classpath "com.github.ben-manes:gradle-versions-plugin:$gradleVersionsPluginVersion"
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlinVersion"
         classpath 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.4'
-        classpath 'io.arrow-kt:ank-gradle-plugin:0.7.1'
+        classpath "io.arrow-kt:arrow-ank-gradle:0.10.5-SNAPSHOT"
```
rachelcarmena: Hi @i-walker ! Just a comment about this change. 

With this change, the build is getting the Arrow Ank Gradle artifact with 0.10.5-SNAPSHOT version from the OSS repository, which is still using Kotlin 1.3.11.

Besides, this can work now because there is already a published artifact with that version in OSS repository. However, `VERSION_NAME` cannot be used as an alternative because it will fail during release time (the source code cannot depend on an artifact that is deployed from that source code).

Maybe that's the reason why that version is fixed from `0.7.1` because the Ank Gradle Plugin hasn't evolved since then. Some alternatives:

- Don't apply Arrow Ank Gradle Plugin. Add the `runAnk` task in `arrow-docs/build.gradle`.
- Exclude Arrow Ank Gradle Plugin. Change build and publish checks to control the order.

i-walker: ;) I am aware of that- that is the same comment you made in the last PR. With the next PR it will use the one which is published now.

i-walker: > - Don't apply Arrow Ank Gradle Plugin. Add the runAnk task in arrow-docs/build.gradle

That suggestion doesn't make sense, because the only place runAnk is used is in the docs and it would lead to the same result.

> - Exclude Arrow Ank Gradle Plugin. Change build and publish checks to control the order.

The latter makes sense, but for this PR I solely focused on updating Ank. Maybe we can do that in another PR.

i-walker:
the error in mtl-data tests should be resolved after #1897

build.gradle
i-walker: Hi @rachelcarmena, I could not find something in the gradle docs, where we can set the build in a way where it uses the Gradle plugin from the project instead of a published one that has to exist in a repository. Do you know if there is an alternative? I am not very familiar with the current release cycle of arrow-ank-gradle.

rachelcarmena: Hi! I'm looking at this task now! Let's see what we can do...

rachelcarmena: Second alternative: #1925

rachelcarmena: I discarded the first alternative because it implies duplicated code (between Gradle Plugin and the created tasks in Gradle file directly) and maintenance would be harder.

i-walker: Done, I merged #1925.

i-walker:
Thanks @rachelcarmena :) for #1925, please take another look if we're missing something.

Issue #1882, opened 2019-12-22
[Bug] rx2 Traverse/Foldable instances are not lawful


1Jajen1:
Uncovered by #1877 

For `ObservableK` and `FlowableK` foldable is implemented using `blockingGet`. Not too familiar with those types but isn't that equal to `unsafeRunSync` in terms of semantics? Either way this fails when given errors and rightly so. We should remove those instances, they are not lawful. We could keep the combinators, but definitly not as part of `Foldable` or `Traversable`.

abendt:
i think `FluxK` Foldable is affected too. It uses `block`

PR #1922, opened 2020-01-11
add traverseIndexed

i-walker:
Targeted data types:
- Const
- Eval
- ListK
- MapK
- NonEmptyList
- SequenceK
- SortedMapK
- ObservableK
- FlowableK

And moved ListF and NonEmptyListF implementation of `traverse` to their DT

### DIFF
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/core/SortedMapK.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/core/SortedMapK.kt
@@ -48,6 +48,12 @@ data class SortedMapK<A : Comparable<A>, B>(private val map: SortedMap<A, B>) : 
     }.value()
   }
 
+  fun <G, C> traverseIndexed(GA: Applicative<G>, f: (Int, B) -> Kind<G, C>): Kind<G, SortedMapK<A, C>> = GA.run {
+    map.iterator().iterateRightIndexed(Eval.always { just(sortedMapOf<A, C>().k()) }) { index, kv, lbuf ->
+      Eval.later { f(index, kv.value).lazyAp { lbuf.value().map { xs -> { b: C -> (mapOf(kv.key to b).k() + xs).toSortedMap().k() } } } }
+    }.value()
+  }
+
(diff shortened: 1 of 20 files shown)


1Jajen1:
I think there is a more principled way of doing this because this feels like boilerplate code ^^.

There are different ways to tackle indexed structures, haskell has the keyed package which offers typeclasses for indexing with polymorphic indexes (Lists have Int, Map<K, V> has K, etc...).

Also if we ignore polymorphic indexes, there are more general ways to get `traverse` or `map` with an index. One of those ways is a simple `withIndex` method which pairs each element with its index. `Kind<F, A>.withIndex(): Kind<F, Tuple2<Int, A>>`. This has the same effect at the cost of another traversal (which a good compiler would fuse, but in it kotlin likely won't...). 

Also lastly, the traverse* methods on rx2 and reactor types should get slightly different names imo, this is related to: https://github.com/arrow-kt/arrow/issues/1882 Those methods are not lawful implementations of `traverse` so they should not be called like that. Not sure what a better name is tho, so if you have a good idea, great, if not leave it, just wanted to note this problem :)

modules/core/arrow-core-data/src/main/kotlin/arrow/core/SortedMapK.kt
```diff
@@ -48,6 +48,12 @@ data class SortedMapK<A : Comparable<A>, B>(private val map: SortedMap<A, B>) :
     }.value()
   }
 
+  fun <G, C> traverseIndexed(GA: Applicative<G>, f: (Int, B) -> Kind<G, C>): Kind<G, SortedMapK<A, C>> = GA.run {
```
pakoito: I would not use `run` scope for the whole function because we've seen the Android compiler choke on it. Instead, default to using it in the most localized spots you can, i.e. `GA.just` in the always and the `run` inside `later` to get `lazyAp`

i-walker:
Thanks for the feedback from both of you @pakoito and @1Jajen1. I will reiterate my attempt with consideration and your feedback in another PR - especially regarding Indexed structures.

Issue #1913, opened 2020-01-08
["Request"] Add monad-trans typeclass


1Jajen1:
MonadTrans is a typeclass that abstracts lifting arbitray monadic computations in another context. It is a necessity in mtl style programs as lifting becomes quite verbose if done manually for each type.

Note some instances might need to reorder the parameters of certain mtl types like `EitherT` otherwise it would make for a weird `MonadTrans` instance.

Basically port this + laws:
http://hackage.haskell.org/package/transformers-0.5.6.2/docs/Control-Monad-Trans-Class.html

abendt:
i had a look at this and have some questions now:

first issue is regarding the laws:
i didn't fully understand how to make the first law work (http://hackage.haskell.org/package/transformers-0.5.6.2/docs/Control-Monad-Trans-Class.html). Is it the same as basic monad associativity laws? I assumed that and implement laws based on that as it was easier to understand: 
https://github.com/abendt/arrow/blob/monad_trans/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadTransLaws.kt#L35

second i was not able to come up with a typeclass interface that can be used with `OptionT` and  `EitherT`. Issue here is that `EitherT` is a `Kind3` and `OptionT` just `Kind2` and i didn't find a common type. for now i have created two typeclasses  (https://github.com/abendt/arrow/blob/monad_trans/modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/typeclasses/MonadTrans.kt#L8) but i guess this is not the way to go.

1Jajen1:
For the first law imagine we have `MM = Monad<M>, MF = Monad<F>, MonadTrans<F>` then `MM.just(x).lift() == MF.just(x)`

`MonadTrans` takes a higher kinded type of arity 2 as a type. Those are, in arrow, the witness types `ForEither`, `ForOptionTPartial`, `EitherTPartial<L>` etc.

For example the lift instance for OptionT would be: `MonadTrans<ForOptionT>` and `fun <M, A> Kind<M, A>.lift(MM: Monad<M>): Kind3<ForOptionT, M, A> = OptionT.lift(MM, this)` or similar, not sure how the `lift` method  actually looks ^-^

[...]

PR #1924, opened 2020-01-13
MonadTrans

abendt:
this adds the `MonadTrans` typeclass together with an `OptionT` instance (#1913)


modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/typeclasses/MonadTrans.kt
```diff
@@ -0,0 +1,33 @@
...
+   *    // sampleStart
+   *    val result = OptionT.monadTrans().run {
+   *      lift(Id.monad(), Id.just("hello"))
+   *    }
+   *    // sampleEnd
+   *    println(result)
+   * }
+   * ```
+   */
+  fun <F, A> lift(MF: Monad<F>, fa: Kind<F, A>): Kind2<T, F, A>
```
1Jajen1: Can you define this as `fun <F, A> Kind<F, A>.lift(MF: Monad<F>): Kind2<T, F, A>`? Makes it easier to use imo.

modules/docs/arrow-docs/docs/docs/arrow-incubator/typeclasses/monadtrans/README.md
```diff
@@ -0,0 +1,48 @@
...
+import arrow.mtl.extensions.optiont.monadTrans.monadTrans
+import arrow.core.extensions.id.monad.monad
+import arrow.core.extensions.*
+import arrow.core.*
+import arrow.mtl.*
+
+val result = OptionT.monadTrans().run {
+  Id.just("hello").lift(Id.monad())
+}
+```
```
pakoito: Here `ank` won't inline any text, could you please either remove the `val` or hint the type yourself?

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/typeclasses/MonadTrans.kt
```diff
@@ -0,0 +1,34 @@
...
+import arrow.Kind
+import arrow.Kind2
+import arrow.typeclasses.Monad
+
+/**
+ * MonadTrans is a typeclass that abstracts lifting arbitray monadic computations in another context.
+ */
+interface MonadTrans<T> {
+  /**
+   * transforms a given monad `Kind<F, A>` to `Kind2<T, F, A>`
```
pakoito: Typo: capital T

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/typeclasses/MonadTrans.kt
```diff
@@ -0,0 +1,34 @@
+package arrow.mtl.typeclasses
+
+import arrow.Kind
+import arrow.Kind2
+import arrow.typeclasses.Monad
+
+/**
+ * MonadTrans is a typeclass that abstracts lifting arbitray monadic computations in another context.
+ */
+interface MonadTrans<T> {
```
pakoito: We normally use `F` and `G` for nested HKTs

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/typeclasses/MonadTrans.kt
```diff
@@ -0,0 +1,34 @@
...
+   *    // sampleStart
+   *    val result = OptionT.monadTrans().run {
+   *      Id.just("hello").lift(Id.monad())
+   *    }
+   *    // sampleEnd
+   *    println(result)
+   * }
+   * ```
+   */
+  fun <F, A> Kind<F, A>.lift(MF: Monad<F>): Kind2<T, F, A>
```
pakoito:
```
fun <G, A> Kind<G, A>.lift(MF: Monad<G>): Kind2<F, G, A>
```

pakoito: If we're using `liftF` for concrete instances and we have `lift` defined in `Functor`, maybe it'd be better to use `liftF` here too.

1Jajen1: I think we should do this on case by case basis, because most of the time the current `lift` is equal to the one from `MonadTrans`. I think the only exception is `Kleisli` because it can lift any `Kind` not just monads (although it needs to be a monad so that Kleisli is a monad), Not sure what name to give in what case, but all other mtl types afaik need at least `Functor` to lift. (And given the fact that the only useful types to lift are monads we might as well use monad everywhere)

pakoito: We have `liftF` for Free, and similar others. `liftM`? `liftT`?

1Jajen1: I completely forgot that we have `liftF` for `Free` but that is in an entirely different context so not sure it matters. Honestly I'd stay with just `lift` here, if that won't work `liftT` for "liftTransformer" sounds best.

abendt: i changed  it to liftT as this seemed to be the one both of you agreed on. if not, let me know its a quick change 😄

pakoito:
Merge tiiiiiime 🤩

PR #1925, opened 2020-01-14
Extract arrow-docs from build

rachelcarmena:
:warning: PR on `is-ank-update` branch.

It corresponds to the [second alternative](https://github.com/arrow-kt/arrow/pull/1921#discussion_r365356403) that was proposed in #1921 review.


i-walker:
:ta

i-walker:
Than

PR #1926, opened 2020-01-15
Small improvements to ank

1Jajen1:
Don't prepend a newline to the code and handle multiline output better.

Not sure why the arrow docs don't have a newline prepended, but I currently get that for all my snippets.


modules/ank/arrow-ank/src/main/kotlin/arrow/ank/interpreter.kt
```diff
@@ -150,7 +150,7 @@ val interpreter: AnkOps = object : AnkOps {
               } else { // accumulating code inside a fence
                 val modifiedSnippet = state.snippet.copy(
                   fence = state.snippet.fence + "\n" + line,
-                  code = state.snippet.code + "\n" + line
+                  code = state.snippet.code + (if (state.snippet.code.isEmpty()) "" else "\n") + line
```
1Jajen1: Does someone know why we don't have snippets in arrow with an extra newline at the start? This sure looks like there should be one. Is it because of the old ank gradle version (0.7.1)?

1Jajen1: On current master the output actually contains this one excessive newline, which is fixed with this change. Is there another processing step removing this somehow, because those don't show up on the docs site?

1Jajen1:
Here is an example for what ank produces currently for both multiline output and the prepended newline: https://1jajen1.github.io/kotlin-pretty/docs/api-reference/

modules/ank/arrow-ank/src/main/kotlin/arrow/ank/interpreter.kt
calvellido: I am not able to generate the site at the moment :disappointed:, so I can't really check what Ank produces, but bear in mind that code snippets in Arrow docs are processed by Arrow Playground (https://github.com/arrow-kt/arrow-playground/), so they might not be representative of what Ank generates (in terms of excessive new lines, that is).

1Jajen1: I checked both playground and non-playground snippets: https://next.arrow-kt.io/docs/arrow/core/listk/ Neither of them have a newline, however on a local run they had (the output in /docs/build/site that is).
I can take this out and open this in a new pr to get the other change in, because having multiline output rendered correcty is more important for me atm.

calvellido: Regarding https://next.arrow-kt.io/docs/quickstart/setup/, even if they are not executable, all Kotlin code (i.e. all code snippets marked as ```kotlin in the Markdown sources) in the Arrow docs is processed through Arrow Playground.

The only code snippets that Arrow Playground doesn't _stylize_ are other languages ones, e.g.:

https://next.arrow-kt.io/docs/quickstart/setup/

calvellido: While I can't check it locally, I think that what you said regarding the local output in `docs/build/site` having the newline at start falls in line with your fixes in the PR

1Jajen1:
Can we get this plus a rollback of #1921 in before the next release? #1921 currently breaks ank for some weird reason, see #1927 for more information. We can redo #1921 after that, if there is a solution to the problem described in #1927 that would also be good.

I'd be fine with just the second change (rendering multiline output with // after each new line) because that is currently what makes the docs at https://1jajen1.github.io/kotlin-pretty/docs/getting-started/ ugly.

Not sure when the next release is, but just so I don't forget ^^

@rachelcarmena @i-walker ideas/opinions?

i-walker:
From what I've seen in another project where I am using Ank's ScriptEngine it seems that the build process causes this. I am not an expert in Gradle, but I can try checking out your project at the weekend to push a PR to resolve it. Ank works in Arrow, maybe your missing something in your configuration. But this is just from an external perspective @1Jajen1.

1Jajen1:
> From what I've seen in another project where I am using Ank's ScriptEngine it seems that the build process causes this. I am not an expert in Gradle, but I can try checking out your project at the weekend to push a PR to resolve it.

That would be perfect, thank you :) I will also try around a bit more and write something in #1927 if I figure something out.

> Ank works in Arrow, maybe your missing something in your configuration. But this is just from an external perspective @1Jajen1.

I know that arrow comes first, however ank should still be usable from outside of arrow, it's alright if it's broken in a snapshot, but a release should have a working ank even for projects outside of arrow (otherwise whats the point of releasing it?!). Don't get me wrong, I don't think of this as super important or anything, just something to consider if no workaround/fix has been found before the next release.

1Jajen1:
The issue in #1927 has a workaround now. Thanks @i-walker 

Now back to this:
These changes seem logical to me, however because the result on next.arrow-kt.io does not match my expectation as to what should have happened I am unsure on how to proceed.

So for now I'd propose the following:
Lets merge this pr, see what happens on next.arrow-kt.io and just revert it if the snippets look fine. An alternative would be running them through the arrow playground and checking them manually, but I don't know how that works. @calvellido what do you think?

calvellido:
I'd love to give you more insight @1Jajen1, but I haven't found yet a way to build the project/web locally :disappointed:, so I'd say to follow your proposal.

1Jajen1:
> I'd love to give you more insight @1Jajen1, but I haven't found yet a way to build the project/web locally disappointed, so I'd say to follow your proposal.

Still no luck with the stackoverflow in arrow-meta? :/ 

Ok I'll try it, how long does it usually take for changes to appear on the docs?

calvellido:
Half an hour, more or less, changes should be up already:

https://next.arrow-kt.io/docs/core/

1Jajen1:
> Half an hour, more or less, changes should be up already:

From what I can tell it looks the same as before (comparing with arrow-kt.io) :tada: 

Also: https://1jajen1.github.io/kotlin-pretty/docs/getting-started/ Now has the nice multiline output and no extra newline at the start, so I'd call that a success ^^

calvellido:
Great!  :clap::clap:

PR #1930, opened 2020-01-15
Official Gradle Wrapper Validation GitHub Action

JLLeitschuh:
See: https://github.com/gradle/wrapper-validation-action


.github/workflows/gradle-wrapper-validation.yml
```diff
@@ -0,0 +1,10 @@
+name: "Validate Gradle Wrapper"
+on: [push, pull_request]
+
+jobs:
+  validation:
+    name: "Validation"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: gradle/wrapper-validation-action@v1
```
rachelcarmena: Thanks for this useful contribution @JLLeitschuh :tada: 

Please, instead of creating an individual check and having the check for every pull and pull_request, just add another step which uses `gradle/wrapper-validation-action@v1` before `Build with Gradle` step in `.github/workflows/build.yml`. 

Thanks again!

rachelcarmena: And the name of that step could be `Validate Gradle Wrapper` :+1:

JLLeitschuh: The design that we had in mind for this action was for it to be standalone so that it had a decreased chance of accidentally getting refactored out of an organization's CI pipeline.

I'm happy to contribute it to you in that other way if it's truly the way you want to consume it. But we recommend using this way instead.

rachelcarmena: Thanks @JLLeitschuh!! I think it's not possible to guarantee the order with separated files and I think it makes sense to do that check before building the project. 

The documentation of that action indicates a similar situation: 
> Simply add this action to your workflow after having checked out your source tree and before running any Gradle build:

Thanks again!!

JLLeitschuh:
Looks like this works now:
https://github.com/arrow-kt/arrow/pull/1930/checks?check_run_id=403751452#step:4:9

PR #1937, opened 2020-01-18
Prettier show instances

1Jajen1:
Added prettier show instances for arrow-core datatypes. Higherkind types now also use show instances instead of just `toString` to render content.

Lists render the haskell way: `[1, 2, 3]` instead of `List(1,2,3,4)`. I personally like that more and it makes defining `Sequence`, `Map` etc as `Sequence([1, 2, 3, 4])` much easier.

I also made the show instance for strings show certain escape characters differently (`"\n"` as `"\\n"` which renders as `"\n"`). Not so sure about that, but it does make text in some situations easier to read.


1Jajen1:
> Something I found really cool from Ocaml and the data implementation of toString is that you can copy it and use it as regular code because all the calls to the constructor had named parameters, for example.
>
> I'd repro that here, if possible.

Named parameters are not needed in most cases, so apart from tuples and collections this still works. (Unless you are in a sealed class in which the classes are inner classes, then this would not even have worked prior to these changes)

Also this is not even true for collections: `listOf(1,2,3,4)` produces `[1, 2, 3, 4]`.

So anyway here is a list of what cannot be copied to source with this pr:
- all collections (cannot be done with "vanilla" kotlin either)
- tuples (could bring that back by using `TupleN(...)` instead of `(...)`)
- Either (because Left and Right are nested classes, did not work before this)
- Ior (^^ same reason as either)
- Validated (^^ same)
- Strings (Escaped strings are escaped again when copied by some ide's but that is a decent tradeoff with readability imo. Newlines, tabs etc destroy the entire output structure when not escaped)

The datatypes on which it did work (and still does) are:
- Option
- Const
- Id

If Either, Ior, Validated were not inner classes they can also be copied as source.

I agree that keeping output close to source is a good thing, but I still would drop the named parameters on our types because they are noise, they contain no useful information and obstruct the output.

PR #1938, opened 2020-01-19
Fix #1936: update doc publication

rachelcarmena:
Current doc publication is failing because this update is missing.


rachelcarmena:
Thanks @pakoito :heart:

PR #1939, opened 2020-01-20
KotlinX Coroutines integration module

### LINKED ISSUE #1933: ["Request"] Integration with Kotlin Coroutines
Define some API’s to launch IO operations within a CoroutineScope in a way that in case of cancellation we avoid issues by making sure they get disposed.

aballano:
This PR adds a new module: arrow-integration-kotlin-coroutines. Fixes #1933

It includes 2 new extension functions for IO:
* `unsafeRunScoped(CoroutineScope, (Either<Throwable, A>) -> Unit)`: Which is an `unsafeRunAsyncCancellable` alike that runs until the given Scope is cancelled.
* `suspendCancellable`*: A `suspended` alike, but cancellable

*Note: Turns out cancelation is not included at the language level in Kotlin, therefore we need the kotlinx coroutines runtime to be able to convert IO to a cancellable coroutine. There've been different opinions on how to release this extension, one possibility is like it is in this PR: an extension explicitly included within the module (which already includes such dependency). Another suggested option is to use reflection within arrow-fx module and implicitly use the cancellable version if available in the classpath. Please note that this last way is still possible to build, although we opted for the former since testing was easier, but please feel free to discuss the topic on this PR 🙂


modules/fx/arrow-fx/src/test/kotlin/arrow/fx/IOTest.kt
```diff
@@ -460,14 +475,20 @@ class IOTest : UnitSpec() {
     "unsafeRunAsyncCancellable can cancel even for infinite asyncs" {
```
nomisRev: This changes the behavior of the test.
Now it's calling the `cb` in `guaranteeCase`, and before it was being called by `unsafeRunAsyncCancellable` when `OnCancel.ThrowCancellationException` occurs (by default this is `OnCancel.Silent`). Although it looks like it was not working because the new assertion is correct `shouldBe 1.some()`.

I need to double check this again, I remember looking at this a couple times. It should probably be rewritten to be a little simpler.

modules/integrations/arrow-integrations-kotlin-coroutines/src/test/kotlin/arrow/integrations/kotlinx/CoroutinesIntegrationTest.kt
```diff
@@ -0,0 +1,164 @@
...
+      val scope = TestCoroutineScope(TestCoroutineDispatcher())
+      val expected = 0
+      IO.just(expected).unsafeRunScoped(scope) { either ->
+        either.fold({ fail("") }, { it shouldBe expected })
+      }
+    }
+  }
+}
+
+fun <A> IOOf<A>.onCancel(token: IOOf<Unit>): IO<A> =
```
aballano: Btw I find this extension really convenient for newcomers and something not so straightforward to build, would be worth to consider if we should add it to the lib itself cc @arrow-kt/maintainers

nomisRev: Yes, we should add it to `IO`. There is a lot of combinators missing still, and a lot which are currently only defined as global extension functions due to the current kapt impl.

settings.gradle
```diff
@@ -75,6 +75,7 @@ include {
         integrations {
             _ 'integrations-retrofit-adapter'
             _ 'integrations-jackson-module'
+            _ 'integrations-kotlinx-coroutines'
```
nomisRev: I think this should belongs under `Arrow Fx KotlinX`

```
        fx {
            _ 'fx'
            _ 'fx-mtl'
            _ 'fx-rx2'
            _ 'fx-reactor'
            _ 'fx-kotlinx'
        }
```

aballano: oh, fair enough 👍

modules/fx/arrow-fx-kotlinx-coroutines/build.gradle
```diff
@@ -0,0 +1,14 @@
+apply plugin: 'kotlin'
+
+dependencies {
+    implementation project(':arrow-fx')
+    implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.3"
```
nomisRev: We should put these versions in `gradle.properties` like @rachelcarmena reorganized them

modules/fx/arrow-fx-kotlinx-coroutines/build.gradle
```diff
@@ -0,0 +1,14 @@
...
+
+    testCompile project(':arrow-test')
+    testImplementation project(':arrow-test')
+    testImplementation project(':arrow-fx')
+    testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:1.3.3"
```
nomisRev: Same

modules/fx/arrow-fx-kotlinx-coroutines/src/test/kotlin/arrow/integrations/kotlinx/CoroutinesIntegrationTest.kt
```diff
@@ -0,0 +1,166 @@
...
+    "suspendedCancellable should throw" {
+      forAll(Gen.throwable()) { e ->
+        val ceh = TestCoroutineExceptionHandler()
+        val scope = TestCoroutineScope(ceh + TestCoroutineDispatcher())
+
+        scope.launch {
+          IO { throw e }.suspendCancellable()
+        }
+        // suspendCancellableCoroutine re throws the exception so we need to compare the instance
+        ceh.uncaughtExceptions[0]::class == e::class
```
nomisRev: We should be able to perform instance checking here no?

Since the throwables are constructed by the `Gen` so the `stacktrace` etc is also constructed by the `Gen`, this happens at time of construction and not at the time of throwing IIRC so instance check should apply.

modules/fx/arrow-fx-kotlinx-coroutines/src/test/kotlin/arrow/integrations/kotlinx/CoroutinesIntegrationTest.kt
```diff
@@ -0,0 +1,166 @@
...
+          !promise.get()
+        }.unsafeRunSync() == i
+      }
+    }
+
+    "suspendCancellable can cancel even for infinite asyncs" {
+      IO.async { cb: (Either<Throwable, Int>) -> Unit ->
+        val scope = TestCoroutineScope(Job() + TestCoroutineDispatcher())
+        scope.launch {
+          IO.async<Int> { Thread.sleep(5000) }
```
nomisRev:
```suggestion
          IO.async<Int> { }
```

Shouldn't be a need to `Thread.sleep` here. `IO.never` or `IO.async<Int> { }` is the equivalent of a never finishing async task.

modules/fx/arrow-fx-kotlinx-coroutines/src/test/kotlin/arrow/integrations/kotlinx/CoroutinesIntegrationTest.kt
```diff
@@ -0,0 +1,166 @@
...
+            .onCancel(IO { cb(1.right()) })
+            .suspendCancellable().let {
```
nomisRev: This `let` will (can) never be called so can be removed.

modules/fx/arrow-fx-kotlinx-coroutines/src/test/kotlin/arrow/integrations/kotlinx/CoroutinesIntegrationTest.kt
```diff
@@ -0,0 +1,166 @@
...
+              cb(it.right())
+            }
+        }
+        IO(other) { Thread.sleep(500) }
```
nomisRev:
```suggestion
        IO.sleep(500.milliseconds)
```

Never use `Thread.sleep` always use `IO.sleep` or other semantically blocking calls.

Since coroutines simply wrap Java's thread pools using `Thread.sleep` can block multiple tasks from happening since it blocks the Thread. Since commonly threads are shared in work-stealing pools, this means you're breaking or violating the semantics of a work-stealing pool. That can drastically alter the behavior of your program and not realistically mimic the workings of an actual program working on a work-stealing thread.

=> `Thread.sleep` is the equivalent of running blocking Java tasks such as JBDC calls on a work-stealing pool which is a huge anti-pattern.

modules/fx/arrow-fx-kotlinx-coroutines/src/test/kotlin/arrow/integrations/kotlinx/CoroutinesIntegrationTest.kt
```diff
@@ -0,0 +1,166 @@
...
+      forAll(Gen.throwable()) { e ->
+        try {
+          val scope = TestCoroutineScope(TestCoroutineDispatcher())
+          val ioa = IO<Int> { throw e }
+          ioa.unsafeRunScoped(scope) { either ->
+            either.fold({ throw it }, { fail("") })
+          }
+          fail("Should rethrow the exception")
+        } catch (throwable: Throwable) {
+          throwable == e
```
nomisRev: 👍 See note other test about reference check, should be like this instead of `::class`.

modules/fx/arrow-fx-kotlinx-coroutines/src/test/kotlin/arrow/integrations/kotlinx/CoroutinesIntegrationTest.kt
```diff
@@ -0,0 +1,166 @@
...
+        }
+      }
+    }
+
+    "unsafeRunScoped should cancel correctly" {
+      forAll(1, Gen.int()) { i ->
+        IO.async<Int> { cb ->
+          val scope = TestCoroutineScope(Job() + TestCoroutineDispatcher())
+          IO(all) { }
+            .flatMap { IO.async<Int> { cb -> Thread.sleep(200); cb(i.right()) } }
```
nomisRev:
```suggestion
            .flatMap { IO.asyncF<Int> { cb -> IO.sleep(200.milliseconds).followedBy(IO { cb(i.right()) }) } }
```

You can use `asyncF` to write an `IO` program within `async`, i.e. useful in cases where you need parallel, or concurrent ops.

modules/fx/arrow-fx-kotlinx-coroutines/src/test/kotlin/arrow/integrations/kotlinx/CoroutinesIntegrationTest.kt
```diff
@@ -0,0 +1,166 @@
...
+            .unsafeRunAsync { scope.cancel() }
+        }.unsafeRunTimed(500.milliseconds) == None
+      }
+    }
+
+    "unsafeRunScoped can cancel even for infinite asyncs" {
+      IO.async { cb: (Either<Throwable, Int>) -> Unit ->
+        val scope = TestCoroutineScope(Job() + TestCoroutineDispatcher())
+        IO(all) { -1 }
+          .flatMap { IO.async<Int> { Thread.sleep(5000); } }
```
nomisRev:
```suggestion
          .flatMap { IO.async<Int> { } }
```

modules/fx/arrow-fx-kotlinx-coroutines/src/test/kotlin/arrow/integrations/kotlinx/CoroutinesIntegrationTest.kt
```diff
@@ -0,0 +1,166 @@
...
+          .onCancel(IO { cb(1.right()) })
+          .unsafeRunScoped(scope) {
+            cb(it)
+          }
+        IO(other) { Thread.sleep(500) }
```
nomisRev:
```suggestion
        IO.sleep(500.milliseconds)
```

modules/fx/arrow-fx-kotlinx-coroutines/src/test/kotlin/arrow/integrations/kotlinx/CoroutinesIntegrationTest.kt
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    "unsafeRunScoped can cancel even for infinite asyncs" {
+      IO.async { cb: (Either<Throwable, Int>) -> Unit ->
+        val scope = TestCoroutineScope(Job() + TestCoroutineDispatcher())
+        IO(all) { -1 }
+          .flatMap { IO.async<Int> { Thread.sleep(5000); } }
+          .onCancel(IO { cb(1.right()) })
+          .unsafeRunScoped(scope) {
+            cb(it)
```
nomisRev: Will never get called.

Nit: would `unsafeRunScoped(scope) { /* never called */ }` might make these kind of tests more explicit? Same for the current tests in Arrow Fx's IOTest.

modules/fx/arrow-fx-kotlinx-coroutines/src/test/kotlin/arrow/integrations/kotlinx/CoroutinesIntegrationTest.kt
```diff
@@ -0,0 +1,166 @@
...
+          IO.just(i).unsafeRunScoped(scope) { either ->
+            either.fold({ fail("") }, { cb(it.right()) })
+          }
+        }.unsafeRunSync() == i
+      }
+    }
+  }
+}
+
+fun <A> IOOf<A>.onCancel(token: IOOf<Unit>): IO<A> =
```
nomisRev: I am raising a PR where this is moved into `Bracket<F, E>` and enabled concretely for `IO` ;)

nomisRev: PR Raised #2017

modules/fx/arrow-fx-kotlinx-coroutines/src/main/kotlin/arrow/integrations/kotlinx/extensions.kt
```diff
@@ -0,0 +1,37 @@
...
+import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.Job
+import kotlinx.coroutines.newCoroutineContext
+import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.coroutines.resume
+import kotlin.coroutines.resumeWithException
+
+suspend fun <A> IOOf<A>.suspendCancellable() = suspendCancellableCoroutine<A> { cont ->
```
pakoito: NIT: Types, please. Even if it's Unit. I had to think about this one :D

modules/fx/arrow-fx-kotlinx-coroutines/src/main/kotlin/arrow/integrations/kotlinx/extensions.kt
```diff
@@ -0,0 +1,37 @@
...
+  val disposable = fix().unsafeRunAsyncCancellable(OnCancel.ThrowCancellationException) { result ->
+    result.fold(cont::resumeWithException) { cont.resume(it) }
```
pakoito: `result.fold(cont::resumeWithException, cont::resume)`

modules/fx/arrow-fx-kotlinx-coroutines/src/main/kotlin/arrow/integrations/kotlinx/extensions.kt
```diff
@@ -0,0 +1,37 @@
...
+): Unit {
+  val newContext = scope.newCoroutineContext(EmptyCoroutineContext)
+  val job = newContext[Job]
+
+  val disposable = fix().unsafeRunAsyncCancellable(cb = f)
+
+  job?.invokeOnCompletion { e ->
+    if (e is CancellationException) disposable()
+    else Unit
+  }
```
pakoito: Suggestion, add the opposite extfun:

```
fun CoroutineScope.unsafeRunIO(io: IOOf<A>, f: (Either<Throwable, A>) -> Unit)
```

that way it can be used directly from a CoroutineScope without the explicit argument.

modules/fx/arrow-fx-kotlinx-coroutines/src/test/kotlin/arrow/integrations/kotlinx/CoroutinesIntegrationTest.kt
aballano: That's what I though, but for some reason I'm getting the following:

> Expected --> java.lang.IllegalArgumentException#1073680203
Actual     --> java.lang.IllegalArgumentException#686013151

I just noticed that they're indeed copying the exception and rethrowing it, so I'll refer to this in the code 🤷‍♂️

modules/fx/arrow-fx-kotlinx-coroutines/src/test/kotlin/arrow/integrations/kotlinx/CoroutinesIntegrationTest.kt
aballano: Yes, agreed. I'm not sure how it ended like this, it might have been a copy+paste problem from another test or something we changed during the process 😅

modules/fx/arrow-fx-kotlinx-coroutines/src/main/kotlin/arrow/integrations/kotlinx/extensions.kt
aballano: Yep, already went through that one but the type system is failing here, had to specify the unit explicitly otherwise it won't compile 🤷‍♂ 

> Type inference failed: inline fun <C> fold(ifLeft: (Throwable) -> C, ifRight: (A) -> C): C
cannot be applied to
(KFunction1<@ParameterName Throwable, Unit>,KFunction1<@ParameterName A, Unit>)
FAILURE: Build failed with an exception.

modules/fx/arrow-fx-kotlinx-coroutines/src/main/kotlin/arrow/integrations/kotlinx/extensions.kt
aballano: You don't really get rid of the argument, but get the syntax the other way around:

```
IO.just(i).unsafeRunScoped(scope) { ... }

scope.unsafeRunIO(IO.just(i)) { ... }
```

But fine with me anyway 👍

modules/fx/arrow-fx-kotlinx-coroutines/src/test/kotlin/arrow/integrations/kotlinx/CoroutinesIntegrationTest.kt
aballano: I just changed this test to use fx with cancellable instead so is more clear

modules/fx/arrow-fx-kotlinx-coroutines/src/main/kotlin/arrow/integrations/kotlinx/extensions.kt
pakoito: Most `CoroutineScope`s I've seen in the wild are implemented by the class, so

```
class MyFragment: ScopedFragment {
  override fun onCreate() {
    runScopedIO(IO.just(1))
  }
}
```

aballano: Ahh I see, awesome 👍

aballano:
@pakoito @nomisRev Thanks for the great review! Please let me know if there are more issues left.

For the tickets suggested by Simon see https://github.com/arrow-kt/arrow/issues/2020 and https://github.com/arrow-kt/arrow/issues/2021

modules/fx/arrow-fx-kotlinx-coroutines/src/main/kotlin/arrow/integrations/kotlinx/extensions.kt
nomisRev: This is a conflict where they a `resume` with an additional param which has a default value. Due to that, it can't be used as a lambda reference.

Same issue as we have with the `dummy: Unit = Unit` overloads.

pakoito:
Up to @nomisRev to merge now :D

Issue #474, opened 2017-11-23
MonadIO/MonadBaseControl


pakoito:
Taken from: http://www.parsonsmatt.org/2017/11/21/monadbasecontrol_in_five_minutes.html

pakoito:
Stretch goal: `MonadBracket` for resource management https://www.fpcomplete.com/blog/2017/02/monadmask-vs-monadbracket

1Jajen1:
What exactly is needed for these two? Is it just a typeclass with a method each and instances for IO, mtl and the reactor/rx2 types? Because this would be quite useful in mtl style programs which use typeclasses and thus a polymorphic monad-stack . Currently you can do side-effects and almost all of `IO` in there by using `MonadDefer` and above, but you can't really interop with `IO` nicely.

pakoito:
`MonadIO` exists based on the core idea that everything in Haskell should boil down to IO. Our primitive is actually `suspend`, except that it doesn't have cancellation baked so we have to stick to `IO` for now.

1Jajen1:
Yes I agree suspend makes a good marker, but as you said it does not have the same capabilities, so I'd like to have some sort of `MonadIO` so that existing methods returning `IO` can be included in mtl style monad-stacks.

pakoito:
Yeah, both is good.

1Jajen1:
I'll see if I can do something there then :) `MonadIO` instances for mtl + IO and third party stuff should exists imo. There is also probably an easy implementation using the concurrent instances of those types, but we'll see :D

PR #1943, opened 2020-01-21
Add MonadIO to fx and instances to fx-mtl

1Jajen1:
Does the easy part of #474 

Motivation:
Most code in there will have `Kind<M, *>` as the return type and for effects one should always use `MonadDefer<M>` and above in the effect hierarchy. However there is one case where this does not work: Integrating concrete `IO` computations. If a library or arrow itself exposes methods that return `IO` directly, it cannot be run in a polymorphic environment, not even if we have something like `Concurrent` present.
I don't really like adding it, but also don't see another way to integrate with concrete types like `IO`.

Also at some point the third-party stuff from reactor, rx2 should get an equivalent (and if there is a way to go from `IO` -> any of their effect types that'd also be good. I am not using any of those, so I don't know how that would look ^^).


1Jajen1:
I was thinking about where in the docs to add this and now I think this should be moved to fx-mtl only! This typeclass is basically useless in fx, but then the question is: Where would I put in in the docs? Next to `MonadReader` etc?

1Jajen1:
> By constraining this typeclass you can run any F using IOs unsafe runners. That's a very powerful typeclass imo.

I am not sure what you mean by that, the only valid instance for `MonadIO` in fx is `IO` right?

Although on a second thought I think adding `MonadIO` for `Resource` might be worthwhile even though it's not quite the safest ^^

1Jajen1:
Added it to docs and also added Schedule to the entries to make it discoverable ^^

PR #1945, opened 2020-01-21
DOCS: replaced non-existent Either.just(a: A) with Either.right(a: A)…

kylegoetz:
While Try is deprecated, I don't know when the code will actually be removed. In the meantime, I noticed Try.just(a: A) recommends Either.just(a: A) as a replacement, but that function does not appear to exist. I changed this recommendation to Either.right(a: A), as Try.just(a) is meant to model a successful case, like Either.Right is by convention.


pakoito:
Good catch!

pakoito:
Rerunning a flaky test

nomisRev:
Thanks for & congrats with your first contribution @kylegoetz! 👏 👏  🎉

PR #1946, opened 2020-01-21
Adds z-index to slack channel button

AntonioMateoGomez:
We need to add z-index to the show slack channel button because it need it in the home page


calvellido:
It would also work if you remove _z-indices_ from the `slack-link-container` and the `show-slack-icon`, or if you set just `1` for the container and `2` for the icon, but whatever.

PR #1949, opened 2020-01-21
Add WriterT monad trans instance

### LINKED ISSUE #1913: ["Request"] Add monad-trans typeclass

1Jajen1:
This is the first of a few prs to resolve #1913  .

These prs are a follow up on #1767 and I am doing them one by one so that we can discuss each one.

The typeclass `MonadTrans<T>` abstracts away the inner monad of a monad transformer. This can only work if that monad is the second last type parameter in the datatype. With `conest` one could probably hack this together as well, but that is just an ugly hack and makes mtl much harder to use than it is already.

`cats-mtl` does this entirerly different and it might me a good idea to check that out at some time, but for now this is much easier to use and implement.

WriterT was by far the easiest one as it has no docs to convert ^^
I'll do the others over the next few days if this reordering is a good solution.


1Jajen1:
What are some thoughts on this? @arrow-kt/maintainers 

Also should this be un 0.11 because this is certainly breaking?

pakoito:
0.11 definitely

1Jajen1:
> 0.11 definitely

:+1:  Time to fix some conflicts ^^

1Jajen1:
@nomisRev your timing :see_no_evil: Anyway, this should work now if I didn't mess anything up :sweat_smile:

1Jajen1:
The ank failures are all fx and have nothing to do with this pr as far as I can tell. So this should be good to go. If this approach of reordering parameters to allow partial application (without conest) is good, I'll also do the other mtl types next.

nomisRev:
Yes, just needs a review and a green build step when targeting 0.11.0 for now.

The docs are currently broken, and if seemed futile to keep fixing them while we're still stream lining the api.

Not sure if we can temp disable the doc step for PRs towards release 0.11.0

1Jajen1:
> Yes, just needs a review and a green build step when targeting 0.11.0 for now.

Well the green build is there (just not for docs) :)

> Not sure if we can temp disable the doc step for PRs towards release 0.11.0

Github required actions can be setup in the repo settings to apply only to specific branches. So having a seperate action for pr's to master and pr's to release/0.11.0 should be possible.

1Jajen1:
I'll point this back to master because we are doing breaking changes in mtl anyway as @nomisRev noted here: https://github.com/arrow-kt/arrow/pull/2014#issuecomment-585293817

nomisRev:
Sorry, about the mix up there.

1Jajen1:
The failing test is from `rx2`... Since this doesn't do anything there I'd call that another flaky test. I'll try rerunning tomorrow...

rachelcarmena:

PR #1950, opened 2020-01-22
Rename Applicative#map to #mapN and tupled to tupledN

### LINKED ISSUE #1842: Rename Applicative#map to #mapN
We have `parMapN`, plus it's easy to confuse Applicative#map (constructor) with Functor#map (function)

danimontoya:
Fixes #1842 .


modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Apply.kt
```diff
@@ -44,13 +44,28 @@ interface Apply<F> : Functor<F> {
...
+  fun <A, B, Z> mapN(
+    a: Kind<F, A>,
+    b: Kind<F, B>,
+    lbd: (Tuple2<A, B>) -> Z
+  ): Kind<F, Z> =
+    a.product(b).map(lbd)
+
+  @Deprecated(
+    "map is being renamed to mapN",
+    ReplaceWith("mapN(a, b, c) { (a,b,c) -> z }")
```
nomisRev: Can this refactor the lambda? I've seen mixed results with `ReplaceWith` where I see it work well and very poorly.

danimontoya: No it does not, l'll fix it now!

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Apply.kt
```diff
@@ -78,6 +118,20 @@ interface Apply<F> : Functor<F> {
...
+    c: Kind<F, C>,
+    d: Kind<F, D>,
+    e: Kind<F, E>,
+    lbd: (Tuple5<A, B, C, D, E>) -> Z
+  ): Kind<F, Z> =
+    a.product(b).product(c).product(d).product(e).map(lbd)
+
+  @Deprecated(
+    "map is being renamed to mapN",
+    ReplaceWith("mapN(a, b, c, d, e, ff, lbd)")
```
1Jajen1: This and all below deprecation notices for `mapN` reference `ff`, but the variable is called `f` in all cases

PR #1951, opened 2020-01-22
Enable ParMapN for IO up to arity 9

nomisRev:
Small life of quality improvement for `IO` that enables `IO.parMapN` without an explicit `CoroutineContext` which will run on the default `IO` dispatchers.


modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IOParMap.kt
```diff
@@ -0,0 +1,315 @@
...
+
+  /**
+   * @see parMapN
+   */
+  fun <A, B, C, D, E, G, H, I, J, K> parMapN(fa: IOOf<A>, fb: IOOf<B>, fc: IOOf<C>, fd: IOOf<D>, fe: IOOf<E>, fg: IOOf<G>, fh: IOOf<H>, fi: IOOf<I>, fj: IOOf<J>, f: (A, B, C, D, E, G, H, I, J) -> K): IO<K> =
+    IO.parMapN(IODispatchers.CommonPool, fa, fb, fc, fd, fe, fg, fh, fi, fj, f)
+
+  fun <A, B, C> parMapN(ctx: CoroutineContext, fa: IOOf<A>, fb: IOOf<B>, f: (A, B) -> C): IO<C> = IO.Async(true) { conn, cb ->
+    // Used to store Throwable, Either<A, B> or empty (null). (No sealed class used for a slightly better performing ParMap2)
+    val state = AtomicRefW<Any?>(null)
```
1Jajen1: Out of curiosity: How would a sealed class negatively impact performance? Is it the extra wrapping of the throwable and either cases?

nomisRev: Very valid question! It's only slightly if simply used for `parMapN` but `parMapN` is what powers `traverse` so it can potentially be infinite wrappings if you `traverse` over `Sequence`.

The slightly more complex code is hidden & private.

1Jajen1: :+1: 

Btw: `traverse` over an infinite sequence with `parApplicative` won't even start any code because it will fold the entire sequence first, you need to have `lazyAp` to actually execute code before folding is complete :see_no_evil: 

This is not even true atm because it just stackoverflows, which is a bug (I'll open a ticket, because I have no idea how to fix it...). But still even without the stackoverflow `parApplicative` cannot be used on infinite sequences.

nomisRev: I made a fix for `IO`'s parallel ops not being stack-safe. So with this change, and your `lazyAp` change why wouldn't it work?

https://github.com/arrow-kt/arrow/pull/1928

1Jajen1: Because of this: https://github.com/arrow-kt/arrow/blob/3051b67bcc4298c5b1553be134bd4222da04f3c6/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/ParApplicative.kt#L23

Without properly implementing this method the applicative cannot short-circuit before the entire `IO` is constructed. Only after executing `ap` fully for every element it will start evaluating the `IO`. If you don't want this behaviour you need to defer calling the argument in `lazyAp` somehow. E.q. Option does this by first checking if it is Some or None and runs the function only if it is Some. Same for Either, Validated and the normal IO applicative

PR #1964, opened 2020-01-22
Remove ObservableK Traverse/MonadFilter instances as they're not lawful. 

### LINKED ISSUE #1881: [Bug] rx2 MonadFilter instances are not lawful in the presence of errors
Uncovered by #1877 

Specifically the right empty law does not hold because `flatMap` in the presence of errors will short-circuit. Not sure how this should be handled.

### LINKED ISSUE #1882: [Bug] rx2 Traverse/Foldable instances are not lawful

aballano:
As @1Jajen1 mentioned, ObservableK does foldable by using `blockingGet` which essentially is a `unsafeRunSync()`.

Fixes #1882.

--

Similarly for MonadFilter instances, Observable has the notion of errors, therefore the laws won't pass right now because an error + empty states will output an error, which is not considered in such laws.

Fixes #1881.

### DIFF
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/TraverseLaws.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/TraverseLaws.kt
@@ -47,17 +47,6 @@ class TIF {
 }
 
 object TraverseLaws {
-  // FIXME(paco): this implementation will crash the inliner. Wait for fix: https://youtrack.jetbrains.com/issue/KT-18660
-  /*
-  inline fun <F> laws(TF: Traverse<F>, AF: Applicative<F>, EQ: Eq<Kind<F, Int>>): List<Law> =
-      FoldableLaws.laws(TF, { AF.just(it) }, Eq.any()) + FunctorLaws.laws(AF, EQ) + listOf(
-              Law("Traverse Laws: Identity", { identityTraverse(TF, AF, { AF.just(it) }, EQ) }),
-              Law("Traverse Laws: Sequential composition", { sequentialComposition(TF, { AF.just(it) }, EQ) }),
-              Law("Traverse Laws: Parallel composition", { parallelComposition(TF, { AF.just(it) }, EQ) }),
-              Law("Traverse Laws: FoldMap derived", { foldMapDerived(TF, { AF.just(it) }) })
-      )
-  */
-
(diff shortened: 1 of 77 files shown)


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/TraverseLaws.kt
```diff
@@ -47,17 +47,6 @@ class TIF {
 }
 
 object TraverseLaws {
-  // FIXME(paco): this implementation will crash the inliner. Wait for fix: https://youtrack.jetbrains.com/issue/KT-18660
```
aballano: @pakoito hope is ok to remove this, or shall we refactor it somehow?

modules/fx/arrow-fx-rx2/src/main/kotlin/arrow/fx/rx2/extensions/observablek.kt
```diff
@@ -303,11 +286,9 @@ interface ObservableKFunctorFilter : FunctorFilter<ForObservableK>, ObservableKF
...
 
-@extension
-interface ObservableKMonadFilter : MonadFilter<ForObservableK>, ObservableKMonad {
-  override fun <A> empty(): ObservableK<A> =
-    Observable.empty<A>().k()
-
-  override fun <A, B> Kind<ForObservableK, A>.filterMap(f: (A) -> Option<B>): ObservableK<B> =
-    fix().filterMap(f)
-}
+// TODO Could be revisited in the future. Right now it fails to pass the laws due to the notion of errors.
```
raulraja: This should be just removed since we are not deprecating it. Observable will never be able to support this with the current encoding we have for MTL based on subtypes.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/TraverseLaws.kt
pakoito: The original bug is fixed, maybe it's worth bringing it back.

PR #1968, opened 2020-01-23
RaceN arity9

### LINKED ISSUE #1952: ["Request"] Enable IO racing up to arity 9
`IORacePair` & `IORaceTriple` should be merged so they can automatically derive and enable `raceN` up to arity 9.

Reference: PR that does this for `parMapN` #1951

danimontoya:
Fixes #1952 
Merged IORacePair & IORaceTriple into IORace and implemented `raceN` up until arity 9


modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORace.kt
```diff
@@ -0,0 +1,480 @@
...
+   *     }
+   *     //sampleEnd
+   *     return result
+   *   }
+   *
+   *   IO.concurrent().example().fix().unsafeRunSync().let(::println)
+   * }
+   * ```
+   *
+   * @param this@raceN [CoroutineContext] to execute the source [F] on.
```
aballano:
```suggestion
   * @param ctx [CoroutineContext] to execute the source [IO] on.
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORace.kt
```diff
@@ -0,0 +1,480 @@
...
+   * @param fa task to participate in the race
```
aballano:
```suggestion
   * @param ioA task to participate in the race
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORace.kt
```diff
@@ -0,0 +1,480 @@
...
+   * @param fb task to participate in the race
```
aballano:
```suggestion
   * @param ioB task to participate in the race
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORace.kt
```diff
@@ -0,0 +1,480 @@
...
+   * @return [F] either [Left] if [fa] won the race,
+   *   or [Right] if [fb] won the race.
```
aballano:
```
[IO] either [Left] if [ioA] won the race,
   *   or [Right] if [ioB] won the race.
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORace.kt
```diff
@@ -0,0 +1,480 @@
...
+   *
+   * @see racePair for a version that does not automatically cancel the loser.
```
aballano: I guess this doesn't apply anymore. BTW @nomisRev i guess this was referred to parMapN or something else?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORace.kt
```diff
@@ -0,0 +1,480 @@
...
+   * }
+   * ```
+   *
+   * @param ctx [CoroutineContext] to execute the source [IO] on.
+   * @param ioA task to participate in the race
+   * @param ioB task to participate in the race
+   * @param ioC task to participate in the race
+   * @return [RaceTriple]
+   *
+   * @see [arrow.fx.typeclasses.Concurrent.raceN] for a simpler version that cancels losers.
```
aballano: Same here, i guess this is not the case anymore, no? cc @nomisRev

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORace.kt
```diff
@@ -0,0 +1,480 @@
...
+   * }
+   * ```
+   *
+   * @param ctx [CoroutineContext] to execute the source [IO] on.
+   * @param ioA task to participate in the race
+   * @param ioB task to participate in the race
+   * @return [IO] either [Left] with product of the winner's result [ioA] and still running task [ioB],
+   *   or [Right] with product of running task [ioA] and the winner's result [ioB].
+   *
+   * @see [arrow.fx.typeclasses.Concurrent.raceN] for a simpler version that cancels loser.
```
aballano: and here

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORace.kt
```diff
@@ -0,0 +1,480 @@
...
+import arrow.fx.typeclasses.Fiber
+import kotlin.coroutines.CoroutineContext
+
+interface IORace {
+
+  fun <A, B> raceN(ioA: IOOf<A>, ioB: IOOf<B>): IO<RacePair<ForIO, A, B>> =
+    IO.racePair(IODispatchers.CommonPool, ioA, ioB)
+
+  fun <A, B, C> raceN(ioA: IOOf<A>, ioB: IOOf<B>, ioC: IOOf<C>): IO<RaceTriple<ForIO, A, B, C>> =
+    IO.raceTriple(IODispatchers.CommonPool, ioA, ioB, ioC)
```
aballano: I believe there might be a api nit here as when someone adds an extra IO to the race3 the result type will change from RaceTriple to Race4, which uses a different encoding for the fold, which is a bit annoying to be fixing, specially if you're trying out several approaches.

I suggest to forward to `racePair`/`raceTriple` if devs care about losers and just call the raceN here as well to return a Race3 (and Race2 above)

PR #1970, opened 2020-01-23
Fix recently added doc links

1Jajen1:
I totally did not mess them up the first time :see_no_evil: 

@calvellido are the links correct now?

Also is there a tool to check broken links at build time for docs? I've been using Hugo for the docs on my prettyprinter and it complains when I add broken links ^^


calvellido:
Is not as convenient as the one integrated in Hugo, but we usually use this:

```sh
wget --spider -r -nd -nv -l 5 http://127.0.0.1:4000/docs/
```

https://github.com/arrow-kt/arrow/tree/master/modules/docs/arrow-docs#how-to-test-links

PR #1978, opened 2020-01-24
Fix Async and Concurrent fx block's lazyness

### LINKED ISSUE #1977: Async and Concurrent Fx blocks aren't lazy
Similarly as for #1976, Async and Concurrent suffer from the same issue, which can potentially be avoided by deferring the coroutine evaluation. 

This would also allow removing specific defers on each instance for each data type as fixed in #1663

aballano:
- [x] Add Async law to check for fx lazyness
- [x] Remove defer from each Fx instance
- [x] Make tests fail
- [x] Fix fx eagerness globally
- [x] Remove old tests replaced by law

Fixes #1977

### DIFF
--- a/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/Concurrent.kt
+++ b/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/Concurrent.kt
@@ -974,14 +974,12 @@ interface Concurrent<F> : Async<F> {
 
 interface ConcurrentFx<F> : AsyncFx<F> {
   val concurrent: Concurrent<F>
-
-  override val async: Async<F>
-    get() = concurrent
-
-  fun <A> concurrent(c: suspend ConcurrentSyntax<F>.() -> A): Kind<F, A> {
+  override val async: Async<F> get() = concurrent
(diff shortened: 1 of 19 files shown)


modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/Concurrent.kt
```diff
@@ -973,15 +973,12 @@ interface Concurrent<F> : Async<F> {
...
 
 interface ConcurrentFx<F> : AsyncFx<F> {
-  val concurrent: Concurrent<F>
-
-  override val async: Async<F>
-    get() = concurrent
-
-  fun <A> concurrent(c: suspend ConcurrentSyntax<F>.() -> A): Kind<F, A> {
-    val continuation = ConcurrentContinuation<F, A>(concurrent)
+  override val M: Concurrent<F>
```
pakoito: If you're not overriding it, is it necessary?

aballano: It is in the implementation above, what i'm doing here is just narrowing the type to be concurrent

Issue #1916, opened 2020-01-09
["Request"] Remove Try examples from documentation


kylegoetz:
What version are you currently using? 0.10

What would you like to see?

The documentation includes `Try` examples. Since `Try` is deprecated, these examples should be removed (or transitioned to `Either` examples) to prevent confusion.

PR #1979, opened 2020-01-24
Removed (almost) all Try examples in documentation/comment blocks

kylegoetz:
As per #1916 

The only sample code/usages/tutorials in the codebase I can find now that uses Try is in `modules/docs/arrow-examples`. The code there is a port of a video that does not appear to exist at the provided URL anymore, so I don't even know an appropriate way to refactor the code (there's some references to stage 1, 2, etc.), and then some of it is just beyond my experience level right now.

I welcome any help so this can get ready.


kylegoetz:
It seems clear that the sample code, which involves hitting a repo (using `Try`), it should be refactored as `IO` but when the code starts refactoring them into `Kind`, I start running into issues because I'm not there yet skill-wise.

kylegoetz:
err, in `docs/arrow-examples/src/test/kotlin/arrow/typeclasses`. I do not know the pedagogy of the talk, if my refactoring would introduce something earlier than the speaker (whose video I can't view to confirm) intended, and at some point I hit a wall anyway. I think I refactored something as an `IO<Either<A, B>>` and then when having to refactor to `Kind` I ran into issues about the Kind-edness expected.

modules/docs/arrow-docs/docs/docs/arrow-optics/iso/README.md
```diff
@@ -60,16 +60,16 @@ We can do the same with a Functor mapping.
...
+pointIsoTuple.modifyF(Option.functor(), point) {
+    Some((tuple.a / 2) toT (tuple.b / 2))
 }
 ```
 
 ```kotlin:ank
-val liftF: (Point2D) -> TryOf<Point2D> = pointIsoTuple.liftF(Try.functor()) {
-    Try { (tuple.a / 2) toT (tuple.b / 0) }
+val liftF: (Point2D) -> OptionOf<Point2D> = pointIsoTuple.liftF(Option.functor()) {
+    try { Some((tuple.a / 2) toT (tuple.b / 0)) } catch(e:Throwable) { None }
```
pakoito: Option is lossy here, use Either instead.

kylegoetz: Done.

aballano:
This PR also fixes #2041

rachelcarmena:

PR #1981, opened 2020-01-25
Add MonadReader/MonadState/MonadWriter instances

1Jajen1:
Adds instances for those typeclasses for every mtl type that supports it.


modules/mtl/arrow-mtl-data/src/test/kotlin/arrow/mtl/WriterTTest.kt
```diff
@@ -108,12 +114,18 @@ class WriterTTest : UnitSpec() {
...
+        WriterT.monadState<StateTPartialOf<ForId, Int>, String, Int>(String.monoid(), StateT.monadState(Id.monad())),
+        WriterT.genK(StateT.genK(Id.genK(), Gen.int()), Gen.string()),
+        WriterT.eqK(StateT.eqK(Id.eqK(), Int.eq(), Id.monad(), 0), String.eq())
       )
     )
   }
 }
 
-private fun <F, W> WriterT.Companion.genK(
+internal fun <F, W> WriterT.Companion.genK(
```
nomisRev: These should publicly live in an `arrow-test`(-mtl) module?

1Jajen1: I think we could get away with putting them in `arrow-test` for now. That project depends on mtl-data, recursion-data, fx etc already anyway.

I think at some point that should be split apart, but that could be quite a big change as lots of tests could transiently depend on that behaviour...

1Jajen1: Moved them to `arrow-test` for now

1Jajen1:
Re-running tests because the fx tests continue to be flaky :/

1Jajen1:
This did some weird stuff during the last merge... Somehow committed a bunch of files that are not supposed to be there, will clean that up tomorrow. Please don't merge^^

1Jajen1:
Fixed the problems, this should be good to go now.

rachelcarmena:

PR #1985, opened 2020-01-27
Free should support fx syntax

PhBastiani:
This PR proposes to add the syntax fx to the Free monads.

### DIFF
--- a/modules/free/arrow-free-data/src/test/kotlin/arrow/free/TrampolineTest.kt
+++ b/modules/free/arrow-free-data/src/test/kotlin/arrow/free/TrampolineTest.kt
@@ -38,7 +38,7 @@ class TrampolineTest : UnitSpec() {
     }
   }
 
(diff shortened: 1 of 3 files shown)


modules/free/arrow-free-data/src/test/kotlin/arrow/free/TrampolineTest.kt
```diff
@@ -13,9 +14,13 @@ class TrampolineTest : UnitSpec() {
...
       odd(10001).runT() shouldBe true
     }
+
+    "trampoline should support fx syntax" {
+      tryfxsyntax(10000).runT() shouldBe true
+    }
   }
 
-  fun odd(n: Int): TrampolineF<Boolean> {
+  private fun odd(n: Int): TrampolineF<Boolean> {
```
pakoito: It's okay to have these public.

PR #1986, opened 2020-01-27
Race in fiber raises cancellation exception

### LINKED ISSUE #1971: [Bug] Race in fiber raises cancellation exception
```kotlin
IO.fx {
        dispatchers().io().raceN(timer().sleep(10.seconds).followedBy(effect { println("Hello") }), effect { println("Hello world") })
            .fork().bind().join().bind()
}.suspended()
// => Cancellation exception
```

This however works fine:
```kotlin
IO.fx {
        dispatchers().io().raceN(timer().sleep(10.seconds).followedBy(effect { println("Hello") }), effect { println("Hello world") })
            .bind()
}.suspended()
```

danimontoya:
* ~Removed OnCancel callback to throw the CancellationException since its not needed anymore~ (See discussion in #1971)
* Add explicit racePairCancellable as a lighter version of racePair, without promises
* Made all IO raceN functions depend on racePairCancellable
* Remove upstreamCancelToken from racePair and raceTriple, fixing the issue

Fixes #1971 .

### DIFF
--- a/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORace.kt
+++ b/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORace.kt
@@ -4,8 +4,6 @@ import arrow.core.Either
 import arrow.core.Left
 import arrow.core.Right
 import arrow.core.internal.AtomicBooleanW
-import arrow.fx.extensions.io.applicative.map
-import arrow.fx.extensions.io.monad.flatMap
 import arrow.fx.internal.IOFiber
 import arrow.fx.internal.IOForkedStart
 import arrow.fx.internal.Platform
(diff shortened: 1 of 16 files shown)


pakoito:
I may have missed something just reading the title. Isn't this throwing the baby with the bathwater? If what we want is to ignore CancelationException on races, why remove the choice of surfacing or ignoring it from the whole API. Until we have an API that's called on cancelation this is the only way of running things on it.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORace.kt
```diff
@@ -76,16 +339,10 @@ interface IORace {
     IO.Async(true) { conn, cb ->
       val active = AtomicBooleanW(true)
 
-      val upstreamCancelToken = IO.defer { if (conn.isCanceled()) IO.unit else conn.cancel() }
```
aballano: this is the actual fix for the issue, removing the upstream cancellation the exception is not thrown anymore

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORunLoop.kt
```diff
@@ -415,6 +415,14 @@ internal object IORunLoop {
...
 
     override operator fun invoke(either: Either<Throwable, Any?>) {
+      apply(either)
+    }
+
+    override fun resumeWith(result: Result<Any?>) {
+      invoke(result.fold(::Right, ::Left))
+    }
+
+    private inline fun apply(either: Either<Throwable, Any?>) {
```
pakoito: Why the additional function if it's only used once?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORunLoop.kt
```diff
@@ -415,6 +415,14 @@ internal object IORunLoop {
     }
 
     override operator fun invoke(either: Either<Throwable, Any?>) {
+      apply(either)
+    }
+
+    override fun resumeWith(result: Result<Any?>) {
+      invoke(result.fold(::Right, ::Left))
```
pakoito: `apply`?

pakoito: Then you're duplicating the bytecode to have a micromicromicrofemto perf improvement. Not worth it.

aballano: duh, good point 😅

pakoito: Wait, if both functions are the same you can have one single function call, make one call the other :D This was more about the inline hahaha

aballano: No prob, pushed into a different branch -> https://github.com/arrow-kt/arrow/pull/1993/commits/fefb020fd67e117d9fa6e1fff9ffc213b44d2ebd

PR #1988, opened 2020-01-27
Accum/AccumT

abendt:
For practice i did an implementation of Accum/AccumT (http://hackage.haskell.org/package/transformers-0.5.6.2/docs/Control-Monad-Trans-Accum.html) for Arrow.


abendt:
the check failed because of this test:
```
2020-01-27T20:14:17.6838459Z arrow.fx.MaybeKTests > Multi-thread Maybes should run on their required threads FAILED
2020-01-27T20:14:17.6843678Z     java.lang.AssertionError: "pool-4-thread-1 @coroutine#904" should not equal "pool-4-thread-1 @coroutine#904"
2020-01-27T20:14:17.6844078Z         at arrow.fx.MaybeKTests$3.invokeSuspend(MaybeKTests.kt:105)
2020-01-27T20:14:17.6844198Z         at arrow.fx.MaybeKTests$3.invoke(MaybeKTests.kt)
```

cannot reproduce this locally ...

modules/mtl/arrow-mtl-data/src/test/kotlin/arrow/mtl/AccumTTest.kt
```diff
@@ -0,0 +1,203 @@
...
+        Gen.string(),
+        AccumT.eqK(Either.monad(), Either.eqK(String.eq()), Int.eq(), 10)
+      ),
+
+      MonadErrorLaws.laws<AccumTPartialOf<Int, EitherPartialOf<String>>, String>(
+        AccumT.monadError(Int.monoid(), Either.monadError()),
+        AccumT.genK(Either.genK(Gen.string()), Gen.int()),
+        Gen.string(),
+        AccumT.eqK(Either.monad(), Either.eqK(String.eq()), Int.eq(), 10)
+      )
```
pakoito: Each of the Laws already includes the ones from the previous typeclasses in the hierarchy. In this case just by testing MonadError you already test the others.

1Jajen1: :point_up: Also please add the `functor` and `applicative` instance to that test (there should be an overload taking those together with the `Monad` instance, just use `Monad` directly for those that are missing). This then tests consistency between the implementations and the default implementations in `Monad`

1Jajen1:
Oh and don't forget to add `MonadState/MonadReader/MonadWriter` instances. These can just be direct pass-throughs like:
```kotlin
interface AccumTMonadState<M, W, S> : MonadState<AccumTPartialOf<W, M>, S>, AccumTMonad<M, W> {
  fun MS(): MonadState<M, S>
}
```
And then just lift every operations from the underlying monadstate to `AccumT`.

modules/mtl/arrow-mtl-data/src/test/kotlin/arrow/mtl/AccumTTest.kt
```diff
@@ -0,0 +1,173 @@
...
+        AccumT.functor(Either.functor()),
+        AccumT.monad(Int.monoid(), Either.monad()),
+        AccumT.monad(Int.monoid(), Either.monad()),
+        AccumT.genK(Either.genK(Gen.throwable()), Gen.int()),
+        AccumT.eqK(Either.monad(), Either.eqK(Eq.any()) as EqK<EitherPartialOf<Throwable>>, Int.eq(), 10)
+      )
+
+    )
+
+    "AccumT; flatMap combines State" {
```
pakoito: Nit: typo in the semicolon here

abendt:
> Oh and don't forget to add `MonadState/MonadReader/MonadWriter` instances. These can just be direct pass-throughs like:

@1Jajen1 tried to do that but i was not able to write the according tests, code did not typecheck. could you please have a look?

modules/mtl/arrow-mtl/src/main/kotlin/arrow/mtl/extensions/accumt.kt
```diff
@@ -0,0 +1,193 @@
...
+
+@extension
+interface AccumTMonadError<S, F, E> : MonadError<AccumTPartialOf<S, F>, E>, AccumTApplicativeError<S, F, E>, AccumTMonad<S, F> {
+  override fun MS(): Monoid<S>
+  override fun ME(): MonadError<F, E>
+  override fun MF(): Monad<F> = ME()
+}
+
+@extension
+interface AccumTMonadState<S, F> : MonadState<AccumTPartialOf<S, F>, S>, AccumTMonad<S, F> {
```
1Jajen1: This should be:
```
interface AccumTMonadState<S, W, F> : MonadState<AccumTPartialOf<W, F>, S>, AccumTMonad<W, F> {
```
The state is only lifted through and not the state of the `AccumT`.

modules/mtl/arrow-mtl/src/main/kotlin/arrow/mtl/extensions/accumt.kt
```diff
@@ -0,0 +1,193 @@
...
+
+  override fun get(): Kind<AccumTPartialOf<S, F>, S> =
+    AccumT.liftF(MSF(), MSF().get())
+
+  override fun set(s: S): Kind<AccumTPartialOf<S, F>, Unit> =
+    AccumT.liftF(MSF(), MSF().set(s))
+}
+
+@extension
+interface AccumTMonadReader<S, F> : MonadReader<AccumTPartialOf<S, F>, S>, AccumTMonad<S, F> {
```
1Jajen1: Same as above, the reader context is not the accumulating value of the `AccumT` it is just some reader value that gets lifted through

modules/mtl/arrow-mtl/src/main/kotlin/arrow/mtl/extensions/accumt.kt
```diff
@@ -0,0 +1,193 @@
...
+  override fun <A> Kind<AccumTPartialOf<S, F>, A>.local(f: (S) -> S): Kind<AccumTPartialOf<S, F>, A> =
+    MR().run {
+      AccumT(fix().accumT.map { accumT ->
+        AndThen(accumT).andThen { it.local(f) }
+      })
+    }
+}
+
+@extension
+interface AccumTMonadWriter<S, F> : MonadWriter<AccumTPartialOf<S, F>, S>, AccumTMonad<S, F> {
```
1Jajen1: Same^^

modules/mtl/arrow-mtl/src/main/kotlin/arrow/mtl/extensions/accumt.kt
abendt: here two tests fail:
```
Monad State Laws: set get
Monad State Laws: set twice eq to set once the last element
```

modules/mtl/arrow-mtl/src/main/kotlin/arrow/mtl/extensions/accumt.kt
abendt: could not make the types work for the test. maybe you could have another look.

abendt: got it now. this test fails:
```
java:test://arrow.mtl.AccumTTest.Monad Writer Laws: tell fusion
```

does this mean both instances are not lawful?

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/AccumT.kt
```diff
@@ -0,0 +1,211 @@
...
+
+/**
+ * Alias that represent stateful computation of the form `(S) -> Tuple2<S, A>` with a result in certain context `F`.
+ */
+typealias AccumTFun<S, F, A> = (S) -> Kind<F, Tuple2<S, A>>
+
+/**
+ * Alias that represents wrapped stateful computation in context `F`.
+ */
+typealias AccumTFunOf<S, F, A> = Kind<F, AccumTFun<S, F, A>>
```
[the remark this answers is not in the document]
1Jajen1: Oh and remember to wrap the correct places with AndThen when you invoke this function. (Unless you invoke it inside `F`, then you both cannot and need not use AndThen (because F is then responsible for stacksafety)

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/AccumT.kt
```diff
@@ -0,0 +1,211 @@
...
+        tailRecM(Tuple2(s, a)) { (s1, a0) ->
+          fa(a0).fix().runAccumT(MF, s1).map { (s2, ab) ->
+            ab.bimap({ a1 -> s2 toT a1 }, { b -> s2 toT b })
+          }
+        }
+      }
+    }
+
+    fun <F, S, A> liftF(AF: Applicative<F>, fa: Kind<F, A>): AccumT<S, F, A> = AF.run {
+      AccumT(AF) { s -> fa.map { a -> Tuple2(s, a) } }
```
1Jajen1: This differs from the MonadTrans definition and causes the tests to fail. The tuple should contain `mempty` and not `s`.

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/AccumT.kt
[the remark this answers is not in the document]
abendt: thanks, cool. i removed the wrapper & tests are green now! still need to look into the `AndThen`'s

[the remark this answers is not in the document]
1Jajen1: Yup, basically everywhere where you have a combinator (take in `AccumT` return `AccumT`, or similar) and you use `AccumT { s -> runAccumT(s) }` in any form without first `flatMapping` inside `F` needs to be wrapped in `AndThen`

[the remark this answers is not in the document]
abendt: i've added `AndThen´ in 3 places. i did not understand how to use it in the other places where `runAccumT` is used. think i  need some more direction here ...

modules/mtl/arrow-mtl/src/main/kotlin/arrow/mtl/extensions/accumt.kt
```diff
@@ -0,0 +1,186 @@
...
+interface AccumtTMonadTrans<S> : MonadTrans<Kind<ForAccumT, S>> {
+
+  fun MS(): Monoid<S>
+
+  override fun <G, A> Kind<G, A>.liftT(MG: Monad<G>): Kind2<Kind<ForAccumT, S>, G, A> =
+    AccumT { _: S ->
+      MG.run {
+        flatMap { a ->
+          MG.just(MS().empty() toT a)
+        }
```
1Jajen1: This can just be `map` instead of `flatMap { just(...) }`

1Jajen1:
If this is merged you can define all the missing combinators with `AndThen` as well: https://github.com/arrow-kt/arrow/pull/2005

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/AccumT.kt
```diff
@@ -0,0 +1,188 @@
...
+   * Convert an accumulation (append-only) computation into a fully stateful computation.
+   */
+  fun toStateT(MF: Monad<F>, MS: Monoid<S>): StateT<F, S, A> =
+    StateT(MF) { s: S ->
+      MF.run {
+        runAccumT(s).map { (s1, a) ->
+          MS.run { s.combine(s1) } toT a
+        }
+      }
+    }
```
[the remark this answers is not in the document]
abendt: i needed to access the outer 's' in the 'andThen' block here. please have a look if i am doing it the right way:

https://github.com/arrow-kt/arrow/pull/1988/commits/4da20729ee3e7fe655e50e7e2e20ccc67ca10a05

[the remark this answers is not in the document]
1Jajen1: The idea behind #2005 is that you can do: `StateT(MF, AndThen.id<S>().flatMap { s -> AndThen(runAccumT).andThen { it.map { (s1, a) -> MS.run { s + s1 } toT a } } }))`
And different from before #2005 the inner flatMap is correctly run through a loop and will not cause a stackoverflow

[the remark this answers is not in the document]
abendt: got that. i was just curious if what i did  here https://github.com/arrow-kt/arrow/commit/4da20729ee3e7fe655e50e7e2e20ccc67ca10a05 was correct too.

i will use the enhanced `AndThen` from #2005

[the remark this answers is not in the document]
1Jajen1: > got that. i was just curious if what i did here 4da2072 was correct too.

Sadly not: Imagine `runAccumT` being `AndThen.Single` with depth `x`, but when you create a new function with `AndThen { ... }` it will set the new depth to `1` but it is actually `x + 1`. My first attempt at solving this problem a while back was to pass this new depth to `AndThen.Single`, but one case was quite hard to solve: `x == maxStackDepth` because you are changing both input and output and keep the state, you to lay one more thing on the stack, so using the usualy `AndThen.Concat` won't work. It's much easier to just define `AndThen.join` instead, also leads to a nice implementation^^

abendt:
i messed up the merge :-(

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/AccumT.kt
```diff
@@ -0,0 +1,194 @@
...
+            AndThen(mv.accumT).andThen {
+              it.flatMap { (s1, v) ->
+                mf.accumT(MS.run { s.combine(s1) }).flatMap { (s2, f) ->
+                  MF.just(MS.run { s1.combine(s2) } toT f(v))
+                }
+              }
+            }
+          }
+        })
+    }
```
1Jajen1: Just use flatMap to define this instead. This is largely the same code anyway

modules/mtl/arrow-mtl/src/main/kotlin/arrow/mtl/extensions/accumt.kt
```diff
@@ -0,0 +1,186 @@
...
+  override fun MS(): Monoid<S>
+
+  override fun <A> Kind<AccumTPartialOf<S, F>, A>.orElse(b: Kind<AccumTPartialOf<S, F>, A>): Kind<AccumTPartialOf<S, F>, A> =
+    (this.fix() to b.fix()).let { (ls, rs) ->
+      AccumT { s: S ->
+        AF().run {
+          ls.runAccumT(s).orElse(rs.runAccumT(s))
+        }
+      }
+    }
```
1Jajen1: AndThen here ^^

modules/mtl/arrow-mtl/src/main/kotlin/arrow/mtl/extensions/accumt.kt
```diff
@@ -0,0 +1,186 @@
...
+  override fun <A> Kind<AccumTPartialOf<S, F>, A>.handleErrorWith(f: (E) -> Kind<AccumTPartialOf<S, F>, A>): Kind<AccumTPartialOf<S, F>, A> =
+    this.fix().let { accumT ->
+      AccumT { s: S ->
+        ME().run {
+          accumT.runAccumT(s).handleErrorWith { e ->
+            f(e).fix().runAccumT(s)
+          }
+        }
+      }
+    }
```
1Jajen1: AndThen

abendt:
flaky test:
```
2020-01-31T20:36:34.6684360Z arrow.fx.MaybeKTests > Multi-thread Maybes finish correctly FAILED
2020-01-31T20:36:34.6684583Z     java.lang.AssertionError: Subscriber still running! (latch = 1, values = 0, errors = 0, completions = 0, timeout!, disposed!)
2020-01-31T20:36:34.6684747Z         at io.reactivex.observers.BaseTestConsumer.fail(BaseTestConsumer.java:189)
2020-01-31T20:36:34.6685049Z         at io.reactivex.observers.BaseTestConsumer.assertTerminated(BaseTestConsumer.java:665)
2020-01-31T20:36:34.6685179Z         at arrow.fx.MaybeKTests$1.invokeSuspend(MaybeKTests.kt:75)
2020-01-31T20:36:34.6685298Z         at arrow.fx.MaybeKTests$1.invoke(MaybeKTests.kt)
```

PR #1992, opened 2020-01-28
Fix: change token to sync branches

rachelcarmena:
GitHub App cannot be added in **Restrict who can push to matching branches** so its provided token cannot be used to sync protected branches.

This PR changes that token by another one who has grants to do that sync.


rachelcarmena:
Thanks @calvellido :heart:

PR #1993, opened 2020-01-28
Cancellable version of unsafeRun doesn't throw within run block

aballano:
Excerpt of #1939

Fixes unsafeRunAsyncCancellable not rethrowing the exceptions within the run block due to a one shot prevention in `async`, so switching to `Async` fixes the problem

### DIFF
--- a/modules/fx/arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxKConnection.kt
+++ b/modules/fx/arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxKConnection.kt
@@ -5,9 +5,9 @@ import arrow.fx.KindConnection
 import arrow.fx.typeclasses.ExitCase
 import arrow.fx.typeclasses.MonadDefer
 
-@Deprecated("Cancelation should be done with the cancelable combinator")
+@Deprecated("Cancellation should be done with the cancelable combinator")
 typealias FluxKProc<A> = (KindConnection<ForFluxK>, (Either<Throwable, A>) -> Unit) -> Unit
-@Deprecated("Cancelation should be done with the cancelable combinator")
+@Deprecated("Cancellation should be done with the cancelable combinator")
 typealias FluxKProcF<A> = (KindConnection<ForFluxK>, (Either<Throwable, A>) -> Unit) -> FluxKOf<Unit>
 
 /**
--- a/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
+++ b/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
@@ -232,7 +232,7 @@ sealed class IO<out A> : IOOf<A> {
      * ```
      *
      * @param k an asynchronous computation that might fail typed as [IOProc].
-     * @see cancelable for an operator that supports cancelation.
+     * @see cancelable for an operator that supports cancellation.
      * @see asyncF for a version that can suspend side effects in the registration function.
      */
     fun <A> async(k: IOProc<A>): IO<A> =
@@ -286,7 +286,7 @@ sealed class IO<out A> : IOOf<A> {
      *
      * @param k a deferred asynchronous computation that might fail typed as [IOProcF].
      * @see async for a version that can suspend side effects in the registration function.
-     * @see cancelableF for an operator that supports cancelation.
+     * @see cancelableF for an operator that supports cancellation.
      */
     fun <A> asyncF(k: IOProcF<A>): IO<A> =
       Async { conn: IOConnection, ff: (Either<Throwable, A>) -> Unit ->
@@ -357,7 +357,7 @@ sealed class IO<out A> : IOOf<A> {
      * ```
      *
      * @param cb an asynchronous computation that might fail.
-     * @see async for wrapping impure APIs without cancelation
+     * @see async for wrapping impure APIs without cancellation
      */
     fun <A> cancelable(cb: ((Either<Throwable, A>) -> Unit) -> CancelToken<ForIO>): IO<A> =
       Async { conn: IOConnection, cbb: (Either<Throwable, A>) -> Unit ->
@@ -425,7 +425,7 @@ sealed class IO<out A> : IOOf<A> {
      * ```
      *
      * @param cb a deferred asynchronous computation that might fail.
-     * @see asyncF for wrapping impure APIs without cancelation
+     * @see asyncF for wrapping impure APIs without cancellation
      */
     fun <A> cancelableF(cb: ((Either<Throwable, A>) -> Unit) -> IOOf<CancelToken<ForIO>>): IO<A> =
       Async { conn: IOConnection, cbb: (Either<Throwable, A>) -> Unit ->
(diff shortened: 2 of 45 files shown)


modules/fx/arrow-fx/src/test/kotlin/arrow/fx/IOTest.kt
```diff
@@ -171,34 +173,57 @@ class IOTest : UnitSpec() {
...
+        }
+        fail("Should rethrow the exception")
+      } catch (myException: MyException) {
+        // Success
+      } catch (throwable: Throwable) {
+        fail("Should only throw MyException but was $throwable")
       }
     }
 
-    "should complete when running a return value with runAsync" {
```
aballano: @pakoito I'm not sure this tests are really doing anything, if I change the `shouldBe` part to expect something else nothing's gonna fail here, is that intended?

pakoito: Does it fail, or is the failure just swallowed?

In any case, don't leave dead code in. Remove it and we can return it later.

aballano: it's swallowed, yep

modules/fx/arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxKConnection.kt
```diff
@@ -5,9 +5,9 @@ import arrow.fx.KindConnection
 import arrow.fx.typeclasses.ExitCase
 import arrow.fx.typeclasses.MonadDefer
 
-@Deprecated("Cancelation should be done with the cancelable combinator")
+@Deprecated("Cancellation should be done with the cancelable combinator")
```
pakoito: I believe we discussed this and @nomisRev made the whole API cancelable, single l. Or the other way around hahaha in any case, we should make both uses here consistent.

pakoito:
Is the core change here? there are only changes to the tests.

modules/fx/arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxKConnection.kt
aballano: Yep, saw the commit somewhere, I referred to [this article ](https://www.grammarly.com/blog/canceled-vs-cancelled/)

> Canceled or cancelled is the past tense of the verb to cancel. Both spellings are correct; Americans favor canceled (one L), while cancelled (two Ls) is preferred in British English and other dialects. However, there is only one correct spelling of the word cancellation, no matter where you are.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -804,7 +804,7 @@ sealed class IO<out A> : IOOf<A> {
    * @see [unsafeRunAsyncCancellable] to run in a non-referential transparent manner.
    */
   fun runAsyncCancellable(onCancel: OnCancel = Silent, cb: (Either<Throwable, A>) -> IOOf<Unit>): IO<Disposable> =
-    async { ccb ->
+    Async { _, ccb ->
```
aballano: @pakoito this is the fix, sorry for the noise around 😅

aballano:
The flaky tests should hopefully be fixed within #1986, so waiting to merge that one first.

modules/fx/arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxKConnection.kt
pakoito: `Cancellation should be done with the cancel*l*able combinator` this is what I mean :D

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
pakoito: Gotcha. It may have not surfaced before we did `raceN`. Or maybe it's because before we could have the case of returning a value twice or something.

modules/fx/arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxKConnection.kt
aballano: But you can indeed spell it `cancelable`, I just fixed wrong usages of `cancellation`

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORunLoop.kt
```diff
@@ -432,20 +432,7 @@ internal object IORunLoop {
...
-        ).let { r ->
-          if (shouldTrampoline) {
-            this.value = r
-            Platform.trampoline { trampoline() }
-          } else {
-            signal(r)
-          }
-        }
-      }
+      invoke(result.fold(::Right, ::Left))
```
nomisRev: Can we refactor this in another manner?

If we split it in `fun finishPure` and `fun finishError` like we do in `parMap` & `race`? Then we can `fold` in both `resumeWith` and `invoke` and avoid the additional wrapping we introduced now.

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/IOTest.kt
nomisRev: I am still in favor of figuring out a solution for `runAsync` in `0.10.x` and deprecating it so we can remove it.
Operator makes no sense and it throws everyone off :D

In Cats-effect, it's being replaced by the following. The signature was meant to build edges of the world for `F`, and now `Effect` can be derived through `MonadIO` for `F ~> IO` which means you can build a polymorphic edge of the world. We can achieve the same with `suspend`, but for `ConcurrentEffect` (cancelation) we need support in the language. I would be in favor of doing both tho.

```
trait Effect[F[_]] {
   def to[G[_], A](fa: F[A]): G[A] =
}
```

modules/fx/arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxKConnection.kt
nomisRev: Oh, this again :D I care more about it being consistent throughout the APIs than one or the other.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORunLoop.kt
aballano: fine with me, i moved the let part to an inline function. cc @pakoito

modules/fx/arrow-fx-reactor/src/main/kotlin/arrow/fx/reactor/FluxKConnection.kt
aballano: Sure, this one is just about being correct, as `cancelation` doesn't exist. As I mentioned the other usages as `cancelable`, etc are kept

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/IOTest.kt
aballano: -> #2010

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORunLoop.kt
```diff
@@ -420,19 +420,27 @@ internal object IORunLoop {
...
+      if (canCall) {
+        canCall = false
+        result.fold(
+          { a -> IO.Pure(a) },
+          { e -> IO.RaiseError(e) }
+        ).let(::forward)
+      }
+    }
+
+    private inline fun forward(io: IO<Any?>) {
```
pakoito: `inline` for a function reference does not inline the content.

PR #1999, opened 2020-01-29
Fix guarantee/guaranteeCase not calling finalizer for error

### LINKED ISSUE #1996: guarantee/guaranteeCase does not call finalizer 0.104
Found by @ikwebdev in a workshop with @raulraja and @nomisRev, the title pretty much summarizes it:
```kotlin
import arrow.fx.IO
import arrow.fx.extensions.fx

class File(private val url: String) {
  fun open(): File = this.also { println("File with url $url opened") }
  fun close(): Unit = println("File with url $url closed")
  override fun toString(): String = "This file with url $url contains some interesting content!"
}

fun open(path: String): IO<File> =
  IO { File(path).open() }

fun release(file: File): IO<Unit> =
  IO { file.close() }

fun read(file: File): IO<Unit> =
  IO.effect { println(file) }

suspend fun main(): Unit = IO.fx {
  val file = !open("meh")
  !effect {
    throw Exception("bananas")
    println(file)
  // The release function is expected to be called, but it's not. Same thing with guarantee.
  }.guaranteeCase { release(file) }.attempt().unit()
}.suspended()
```

aballano:
Fixes two bugs within Bracket implementation:

* `IOBracket.guaranteeCase` was wrongly doing a flatMap between an IO and a recovery frame, which caused the frame not to be called -> Fixed by calling IO.Bind directly
* `Bracket.guaranteeCase` was calling the bracket function over itself and then again on the use function -> Fixed by executing the bracket over a unit IO for the acquire.

TODO:
* Discuss if we should update the Bracket docs and possibly some docs examples explaining the difference between `bracket(Case)` and `guarantee(Case)`.

Attempt to fix #1996

### DIFF
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
@@ -36,7 +36,8 @@ object BracketLaws {
       Law("Bracket: guaranteeCase is derived from bracketCase") { BF.guaranteeCaseIsDerivedFromBracketCase({ BF.just(Unit) }, EQ) },
       Law("Bracket: bracket propagates transformer effects") { BF.bracketPropagatesTransformerEffects(EQ) },
       Law("Bracket: bracket must run release task on use error") { BF.bracketMustRunReleaseTaskOnUseError(EQ) },
-      Law("Bracket: bracket must run release task on adquire error") { BF.bracketMustRunReleaseTaskOnAdquireError(EQ) }
+      Law("Bracket: bracket must not run release task on acquire error") { BF.bracketMustNotRunReleaseTaskOnAcquireError(EQ) },
+      Law("Bracket: guarantee must run finalizer task") { BF.guaranteeMustRunFinalizerOnError(EQ) }
     )
   }
 
@@ -133,7 +134,7 @@ object BracketLaws {
     forAll(Gen.int()) { i ->
       val msg = AtomicIntW(0)
       just(i).bracket<Int, Int>(
-        release = { ii -> msg.value = ii; unit() },
+        release = { ii -> unit().map { msg.value = ii } },
         use = { throw Throwable("Expected failure!") }
       )
         .attempt()
@@ -141,13 +142,25 @@ object BracketLaws {
         .equalUnderTheLaw(just(i), EQ)
     }
 
-  fun <F> Bracket<F, Throwable>.bracketMustRunReleaseTaskOnAdquireError(EQ: Eq<Kind<F, Int>>): Unit =
-    forAll(Gen.int()) { i ->
-      val msg = AtomicIntW(0)
+  fun <F> Bracket<F, Throwable>.bracketMustNotRunReleaseTaskOnAcquireError(EQ: Eq<Kind<F, Int>>): Unit =
+    forAll(Gen.int(), Gen.int()) { expected, other ->
+      val actual = AtomicIntW(expected)
       raiseError<Int>(Throwable("Expected failure!")).bracket(
-        release = { ii -> msg.value = ii; unit() },
+        release = { unit().map { actual.value = other } },
         use = { just(it) }
       )
+        .attempt()
+        .map { actual.value }
+        .equalUnderTheLaw(just(expected), EQ)
+    }
+
+  fun <F> Bracket<F, Throwable>.guaranteeMustRunFinalizerOnError(
+    EQ: Eq<Kind<F, Int>>
+  ): Unit =
+    forAll(Gen.int(), Gen.throwable()) { i, t ->
+      val msg = AtomicIntW(0)
+      raiseError<Int>(t)
+        .guarantee(unit().map { msg.value = i })
         .attempt()
         .map { msg.value }
         .equalUnderTheLaw(just(i), EQ)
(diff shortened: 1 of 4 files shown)


aballano:
Failed tests:

```
arrow.fx.IOTest > Bracket: guarantee must run finalizer task
arrow.fx.IOTest > guarantee should be called on finish with error
```

aballano:
Test failed due to bug in Bracket guarantee impl

```
arrow.fx.FluxKTest > Bracket: guarantee must run finalizer task
arrow.fx.MonoKTest > Bracket: guarantee must run finalizer task
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/Bracket.kt
```diff
@@ -145,5 +145,5 @@ interface Bracket<F, E> : MonadError<F, E> {
    *
    */
   fun <A> Kind<F, A>.guaranteeCase(finalizer: (ExitCase<E>) -> Kind<F, Unit>): Kind<F, A> =
-    bracketCase({ _, e -> finalizer(e) }, { this })
+    just<Unit>(Unit).bracketCase({ _, e -> finalizer(e) }, { this })
```
pakoito: I believe we have `IO.unit` for this.

aballano: Actually I was using `unit()` before which has a funny resolution. Anyway we're on Bracket, so I believe we need to stick with the explicit just like it's done above, no?

pakoito: I believe we have `lazy()` for this :trollface:

aballano: I cannot use it here unless you mean the kotlin one. Anyway is there a problem from using `just` here?

pakoito: Awww, it's above in the hierarchy 

https://github.com/arrow-kt/arrow/blob/master/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/MonadDefer.kt#L28

just works, ye

JYCabello: Just me trying to understand what's going on. (We just had a training and I'm beginning to grasp some things). The chaining of `just<Unit>(Unit)` first is meant to trigger a boundary?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/internal/IOBracket.kt
```diff
@@ -83,7 +83,7 @@ internal object IOBracket {
     IO.Async { conn, cb ->
       Platform.trampoline {
         val frame = EnsureReleaseFrame<A>(release)
-        val onNext = source.flatMap(frame)
+        val onNext = IO.Bind(source, frame)
```
JYCabello: Would you be so kind as to hint me on why this makes a difference? Not a full explanation, just something which scent I can track.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/Bracket.kt
nomisRev: If you remember we discussed the difference between `bracket` and `guarantee` is that `bracketCase` makes the `acquire` step uncancelable.

So you should use `guarantee` to schedule finalizers for your (cancelable) program, while you use `bracket` to guarantee that your `release` step is always called for your acquired resource. Doing so involves an uncancelable boundary between `acquire` and `use`.

In this case, we derive `guarantee` by providing a `Unit` resource, running our program in `use` which ensures that a finalizer is called for a `Unit` resource.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/internal/IOBracket.kt
nomisRev: - `IO.Bind(source, f)` is the ADT case that represents `source.flatMap(source)`.

- `IOFrame : (A) -> IO<A>` is mark functions as error handlers, which can be composed as `source.flatMap(IOFrame.ErrorHandler(...))`

- `flatMap` is optimized for `IO.raiseError` to not execute the `flatMap` since it short-circuits. The missing piece was that it should've scheduled it if it was an `IOFrame` since it cannot skip error handlers.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
```diff
@@ -128,15 +130,39 @@ object BracketLaws {
...
-        release = { ii -> msg.value = ii; unit() },
+        release = { ii -> unit().map { msg.value = ii } },
         use = { throw Throwable("Expected failure!") }
       )
         .attempt()
         .map { msg.value }
         .equalUnderTheLaw(just(i), EQ)
     }
+
+  fun <F> Bracket<F, Throwable>.bracketMustNotRunReleaseTaskOnAcquireError(EQ: Eq<Kind<F, Int>>): Unit =
```
nomisRev: 🥇

nomisRev:
> Discuss if we should update the Bracket docs and possibly some docs examples explaining the difference between bracket(Case) and guarantee(Case).

We absolutely should!!

aballano:
> > Discuss if we should update the Bracket docs and possibly some docs examples explaining the difference between bracket(Case) and guarantee(Case).
> 
> We absolutely should!!
> 
> EDIT: Approved & feel free to merge, we can also take this up in another ticket.

Created https://github.com/arrow-kt/arrow/issues/2007 out of this 👍

PR #2003, opened 2020-01-30
More meaningful mapping Either to 0.11.0 IO #2002

Hiosdra:
1) do we want to leave `toIO(f: (E) -> Throwable)` and deprecate it in favour of `toIOException`
2) will this be whole? Or add / change any other method?

PR #2004, opened 2020-01-31
Fix unsafeRunTimedTotal for know Errors

pakoito:
See https://github.com/arrow-kt/arrow/pull/2003#discussion_r373260146

### DIFF
--- a/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
+++ b/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
@@ -785,7 +785,7 @@ sealed class IO<out E, out A> : IOOf<E, A> {
     // Errors short-circuit
     override fun <B> map(f: (Nothing) -> B): IO<E, B> = this
 
-    override fun unsafeRunTimedTotal(limit: Duration): Option<Nothing> = None
+    override fun unsafeRunTimedTotal(limit: Duration): Option<Either<E, Nothing>> = Some(Left(error))
   }
 
   internal data class Delay<out A>(val thunk: () -> A) : IO<Nothing, A>() {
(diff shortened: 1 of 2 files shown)


modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IO.kt
```diff
@@ -785,7 +784,7 @@ sealed class IO<out E, out A> : IOOf<E, A> {
     // Errors short-circuit
     override fun <B> map(f: (Nothing) -> B): IO<E, B> = this
 
-    override fun unsafeRunTimedTotal(limit: Duration): Option<Nothing> = None
+    override fun unsafeRunTimedTotal(limit: Duration): Option<Either<E, Nothing>> = Some(Left(error))
```
JorgeCastilloPrz: Are we sure? `cats-bio` [throws an exception](https://github.com/LukaJCB/cats-bio/blob/2bcf0a496838f41c94683abd1d524876606bd17a/src/main/scala/cats/effect/bio/BIO.scala#L324( containing the error.toString in this case.

nomisRev: Why would you break type-safety and throw? If the user wants to do so he can still throw. That would also mean that `IO<E, A>.unsafeRunSync` returns `A` instead of `Either<E, A>`.

JorgeCastilloPrz: I guess I imagined Haskell. I'm not sure how it is because no experience, but I imagined the language and paradigm would push you to control all types of errors within your program, and otherwise assume they're unexpected and throw (logs error to repl when running the IO). We are kind of in a limbo in Kotlin since our program is still running outside of the pure boundaries and you can still add more logics on top of it.

PR #2005, opened 2020-01-31
Add a stacksafe variant of Join to AndThen

1Jajen1:
This allows fully a stacksafe version of `Kleisli.flatMap` and similar datatypes.
Currently even using a stacksafe monad for `Kleisli.flatMap` will cause a stackoverflow because it first builds up lots of `Kleisli { d -> run(d)... } ` functions which all collapse once evaluated (and overflow). Note that `Kleisli` will still be stackunsafe if used with a stackunsafe monad, that cannot ever be prevented.

Using `AndThen.flatMap` instead builds up a tower of `Join`'s which will all be collapsed later in the loop.

### DIFF
--- a/modules/core/arrow-core-data/src/main/kotlin/arrow/core/AndThen.kt
+++ b/modules/core/arrow-core-data/src/main/kotlin/arrow/core/AndThen.kt
@@ -40,6 +40,10 @@ sealed class AndThen<A, B> : (A) -> B, AndThenOf<A, B> {
 
   private data class Single<A, B>(val f: (A) -> B, val index: Int) : AndThen<A, B>()
 
+  private data class Join<A, B>(val fa: AndThen<A, AndThen<A, B>>) : AndThen<A, B>() {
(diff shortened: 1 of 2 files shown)


pakoito:
Great work this month, @1Jajen1. Seriously.

nomisRev:
@pakoito yes! @1Jajen1 is on 🔥this month with incredible work!!

1Jajen1:
> @pakoito yes! @1Jajen1 is on :fire:this month with incredible work!!

So many good ideas and interesting things came out of working with mtl and in general polishing up propCheck (now arrow-check^^) I wrote this initially a little different a while back so that `GenT` could be made stacksafe (at least on construction, not necessarily on evaluation ^^). And now I am stuck on the compiler running out of memory in the ci... Fun times :sweat_smile: 

I'll make another pr that uses this to actually make `Kleisli` and `StateT` stacksafe in those scenarios...

PR #2012, opened 2020-01-31
Add race 3 to guarantee order of scheduling

nomisRev:
Parallel operations should be scheduled in the order they appear in the parameter list.

Without having an arity-2 & arity-3 operator you cannot create N-arty operators that schedule the tasks in the same order as the parameter list.

### DIFF
--- a/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORace.kt
+++ b/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORace.kt
@@ -366,7 +366,7 @@ interface IORace {
 
       IORunLoop.startCancelable(IOForkedStart(ioC, ctx), connC) { result ->
         result.fold({
-          onError(active, cb, conn, connA, connC, it)
+          onError(active, cb, conn, connA, connB, it)
         }, {
           onSuccess(active, conn, connA, connB, cb, Race3.Third(it))
         })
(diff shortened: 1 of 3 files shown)


modules/fx/arrow-fx/src/main/kotlin/arrow/fx/IORace.kt
```diff
@@ -302,6 +293,89 @@ interface IORace {
...
+        result.fold({
+          onError(active, cb, conn, connA, connC, it)
+        }, {
+          onSuccess(active, conn, connA, connC, cb, Race3.Second(it))
+        })
+      }
+
+      IORunLoop.startCancelable(IOForkedStart(ioC, ctx), connC) { result ->
+        result.fold({
+          onError(active, cb, conn, connA, connC, it)
```
aballano:
```suggestion
          onError(active, cb, conn, connA, connB, it)
```

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/IOTest.kt
```diff
@@ -377,6 +377,32 @@ class IOTest : UnitSpec() {
       order.toList() shouldBe listOf(1L, 2, 3, 4, 5, 6)
     }
 
+    "Races are scheduled in the correct order" {
```
aballano: 🔝

PR #2014, opened 2020-01-31
Use AndThen to make Kleisli and StateT more stacksafe

1Jajen1:
This does not make them stacksafe in all cases, that is impossible in a polymorphic setting without the inner monad being stacksafe. For example flatMap-like cases where the initial value is need in the context F cannot be stacksafe without F being stacksafe)

This does however mean that in most concrete settings where the inner monad is not stacksafe (that usually implies you can extract the value from it) one can define `flatMap` fully stacksafe, as shown with the monad instance that now exists in `ReaderApi` and `StateApi`.


1Jajen1:
This is somewhat breaking as it removes a bunch of unnecessary constraints (and now unnecessary methods). Should this go into 0.11 or is this fine for now?

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/Kleisli.kt
```diff
@@ -23,55 +24,43 @@ fun <F, D, A> KleisliOf<F, D, A>.run(d: D): Kind<F, A> = fix().run(d)
  * @property run the arrow from [D] to `Kind<F, A>`.
  */
 @higherkind
-class Kleisli<F, D, A>(val run: (D) -> Kind<F, A>) : KleisliOf<F, D, A>, KleisliKindedJ<F, D, A> {
+class Kleisli<F, D, A>private constructor(val run: (D) -> Kind<F, A>) : KleisliOf<F, D, A>, KleisliKindedJ<F, D, A> {
```
nomisRev: Can we move this to `0.11.x` ?

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/StateT.kt
```diff
@@ -40,32 +33,16 @@ fun <F, S, A> StateTOf<F, S, A>.runM(MF: Monad<F>, initial: S): Kind<F, Tuple2<S
  * @param runF the stateful computation that is wrapped and managed by `StateT`
  */
 @higherkind
-class StateT<F, S, A>(
-  val runF: StateTFunOf<F, S, A>
+class StateT<F, S, A> private constructor(
```
nomisRev: Can we move this to `0.11.x`?

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/StateT.kt
```diff
@@ -17,18 +16,12 @@ import arrow.typeclasses.SemigroupK
...
-typealias StateTFunOf<F, S, A> = Kind<F, StateTFun<F, S, A>>
-
 /**
  * Run the stateful computation within the context `F`.
  *
- * @param MF [Monad] for the context [F]
  * @param initial state to start stateful computation
  */
-fun <F, S, A> StateTOf<F, S, A>.runM(MF: Monad<F>, initial: S): Kind<F, Tuple2<S, A>> = fix().run(MF, initial)
+fun <F, S, A> StateTOf<F, S, A>.run(initial: S): Kind<F, Tuple2<S, A>> = fix().runF(initial)
```
nomisRev: `0.11.x`

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/StateT.kt
```diff
@@ -40,32 +33,16 @@ fun <F, S, A> StateTOf<F, S, A>.runM(MF: Monad<F>, initial: S): Kind<F, Tuple2<S
...
-      StateT(AF) { s -> AF.just(s toT t) }
+    operator fun <F, S, A> invoke(f: (S) -> Kind<F, Tuple2<S, A>>): StateT<F, S, A> = StateT(f)
 
-    /**
-     * Constructor to create `StateT<F, S, A>` given a [StateTFun].
-     *
-     * @param AF [Applicative] for the context [F].
-     * @param run the stateful function to wrap with [StateT].
-     */
-    operator fun <F, S, A> invoke(AF: Applicative<F>, run: StateTFun<F, S, A>): StateT<F, S, A> = AF.run {
```
nomisRev: We should deprecate these changes in `0.10.x`.

modules/mtl/arrow-mtl-data/src/main/kotlin/arrow/mtl/Kleisli.kt
1Jajen1: This was done to prevent ambiguity between this and the other `run` function that no longer needs the monad parameter. Even if we were to keep that we'd need a replacement that has a different name than `run`. `runKleisli`?

nomisRev: I think this can be solved with either `@JvmName` or a dummy param `Unit = Unit`.

Is this change blocking for the fix? I haven't backtraced the changes yet, but I can look deeper into the code next week.

1Jajen1: That is a good idea. I'll readd the deleted and modified methods then and mark them as deprecated instead.

nomisRev:
As discussed on Slack here, https://kotlinlang.slack.com/archives/C8UK6RTHU/p1581515604055500.

`arrow-fx-mtl` will be renamed to `arrow-mtl-fx` to more clearly represent its dependencies. Since `arrow-mtl` is still an incubating project this changes can be merged and don't need a strict deprecation cycle such as a top-level module as Arrow Fx.

1Jajen1:
> Weird that it turned red like that.

The new `AccumT` borrows some stuff from `StateT` I think. Iirc that is only a typealias, so it should be easy to fix.

nomisRev:
@1Jajen1 yes, confirmed. Simply removed `MF,` and it's green again. Couldn't push it to your fork.

```
fun toStateT(MF: Monad<F>, MS: Monoid<S>): StateT<F, S, A> =
  StateT(MF,
    AndThen.id<S>().flatMap { s: S ->
      AndThen(accumT).andThen {
        MF.run {
          it.map { (s1, a) ->
            MS.run { s.combine(s1) } toT a
          }
        }
      }
    })
```

PR #2017, opened 2020-02-01
Add onCancel to Bracket

nomisRev:
This PR adds onCancel to `Bracket` which is a specialization over `guaranteeCase`.

It's used frequently in tests, and it can be useful for testing and building inter-op with other effect libraries. See #1939

### DIFF
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
@@ -65,6 +65,9 @@ object ConcurrentLaws {
       Law("Concurrent Laws: cancel on bracket releases") { CF.cancelOnBracketReleases(EQ, ctx) },
       Law("Concurrent Laws: acquire is not cancelable") { CF.acquireBracketIsNotCancelable(EQ, ctx) },
       Law("Concurrent Laws: release is not cancelable") { CF.releaseBracketIsNotCancelable(EQ, ctx) },
+      Law("Concurrent Laws: cancel on guarantee runs finalizer") { CF.guaranteeFinalizerOnCancel(EQ, ctx) },
+      Law("Concurrent Laws: release is not cancelable") { CF.guaranteeFinalizerIsNotCancelable(EQ, ctx) },
+      Law("Concurrent Laws: cancel on onCancel runs finalizer") { CF.onCancelFinalizerOnCancel(EQ, ctx) },
       Law("Concurrent Laws: async cancelable coherence") { CF.asyncCancelableCoherence(EQ) },
       Law("Concurrent Laws: cancelable cancelableF coherence") { CF.cancelableCancelableFCoherence(EQ) },
       Law("Concurrent Laws: cancelable should run CancelToken on cancel") { CF.cancelableReceivesCancelSignal(EQ, ctx) },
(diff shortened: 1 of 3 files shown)


1Jajen1:
Won't this swallow all cases apart from `ExitCase.cancel`? I mean if that is intended fine, but I think it would suprise me when using this.

pakoito:
I'm with @1Jajen1, write a test where two are installed, and an exception happens. See if it's still triggered.

nomisRev:
> Won't this swallow all cases apart from ExitCase.cancel? I mean if that is intended fine, but I think it would suprise me when using this.

That's the exact use-case for `onCancel` as an overload to `guaranteeCase`. It should only be called on cancelation.

> I'm with @1Jajen1, write a test where onCancel and another guaranteeCase are installed, and an exception happens. See if both are still triggered.

This test is already available in [BracketLaws](https://github.com/arrow-kt/arrow/blob/master/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt#L157), only `ConcurrentLaws` can test for cancelation cases for `Bracket`. 

@pakoito `doOnCancel` & `doOnError` are side-effecting in RxJava, implied by the `do` IMO,, which I did not want to follow.

1Jajen1:
What would be suprising is that `IO { TODO() }.onCancel(f).guaranteeCase(g)` won't call g with an exitcase that indicates failure right? The end result after `onCancel(f)` is always `Unit` or whatever `f` returned.

nomisRev:
> The end result after onCancel(f) is always Unit or whatever f returned.

No not at all, `onCancel` does *not* change the type of `IO`. It just registers a finaliser, same goes for `guaranteeCase` or `guarantee`.

`fun <A> IOOf<A>.onCancel(f: IO<Unit>): IO<A>`

1Jajen1:
> > The end result after onCancel(f) is always Unit or whatever f returned.
> 
> No not at all, `onCancel` does _not_ change the type of `IO`. It just registers a finaliser, same goes for `guaranteeCase` or `guarantee`.
> 
> `fun <A> IOOf<A>.onCancel(f: IO<Unit>): IO<A>`

Yup just noticed as well ^^ Was not on my laptop and the ank examples on the docs don't work :see_no_evil:

pakoito:
hmmmmm okay, so it works like handleError in that we take only the latest version. Could you please check what happens with nested ones tho?

```
IO.unit.flatMap { raiseException(a).handleError { Unit } }.handleErrorWith { IO { println("nope") }  }
```

shouldn't print anything. What does the guaranteeCase version do for errors?

nomisRev:
GuaranteeCase doesn't work like `HandleError`, it's implemented the same way tho.

`io.guaranteeCase(f) == IO.unit.bracketCase(use = { io }, release = { _, case -> f(case) }`

So it's like `bracketCase` with a `Unit` resource, which is a crucial difference.
Since `io.bracketCase(use = { a -> IO.just(a) }, release = { _, case -> f(case) }` would make your `io` program `uncancelable`. The `acquire` step in `bracket` is uncancelable.

`guarantee` & `guaranteeCase` can be used to register a `finalizer` to a cancelable program.
While `bracket` & `bracketCase` is used to compose an uncancelable acquire step, with a cancelable consumption step and a guaranteed resource release finalizer.

pakoito:
>would still print "nope" since both are guaranteed to be called.
That's good, that's what I wanted :D

nomisRev:
Alright, awesome! I'm going to merge this one already, I'll add a ticket for `onError` with laws and a reference to this PR.

pakoito:
@nomisRev just to reiterate what you already said,

```
IO.unit.flatMap { raiseException(a).guaranteeCase { IO { println("in") } } }.guaranteeCase { IO { println("out") } }
```

Would print both `in` and `out`, right?

nomisRev:
Yes, it will call both.

Issue #2018, opened 2020-02-01
["Request"] Add onError to Bracket 


nomisRev:
`Bracket` exposes several combinators to `guarantee` an `IO` being run no matter whether the `IO` failed, was cancelled or finished successfully.

An `onCancel` overload was added in #2017, this can be used as a reference to implement `onError` with the corresponding `BracketLaws` to ensure polymorphic behavior.

abendt:
i had a try on this: https://github.com/abendt/arrow/tree/issues/2018
is it how it should be?

aballano:
> i had a try on this: https://github.com/abendt/arrow/tree/issues/2018
> is it how it should be?

Yes indeed 👍 If you wanna raise a PR I can take a look

aballano:
Although @nomisRev do we really need the redeclaration for IO? I tested and one can use the generated extension over Bracket as it's suggested by the IDE

abendt:
> Yes indeed 👍 If you wanna raise a PR I can take a look

done

aballano:
Thank you @abendt for the contribution!! 🎉

aballano:
Hi folks, I'd like ask 2 things about this implementation towards the next release:

* Shouldn’t onError signature be like this instead `fun <A> Kind<F, A>.onError(finalizer: (E) -> Kind<F, Unit>): Kind<F, A>` (note the given E)? Otherwise we don't provide much over what was the failure.
* Also [my comment above](https://github.com/arrow-kt/arrow/issues/2018#issuecomment-581136362): do we need the redeclaration in IO of both functions? I don't see it's doing anything on top of the implementation on Bracket

1Jajen1:
As for the first: I noted that as well https://github.com/arrow-kt/arrow/pull/2022#discussion_r373851135 seemed to cause some kind of issue when compiling. Not sure this still persists but I agree, it should offer `E` as a paramter if possible.

abendt:
> * Shouldn’t onError signature be like this instead `fun <A> Kind<F, A>.onError(finalizer: (E) -> Kind<F, Unit>): Kind<F, A>` (note the given E)? Otherwise we don't provide much over what was the failure.

yes, i ran into some issues while trying that. didn't really understand what to do about it. i will prepare a branch later today so you can see

aballano:
[...]

PR #2022, opened 2020-02-02
add onError to Bracket

abendt:
see #2018

### DIFF
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
@@ -85,7 +85,8 @@ object ConcurrentLaws {
       Law("Concurrent Laws: parTraverse forks the effects") { CF.parTraverseForksTheEffects(EQ_UNIT) },
       Law("Concurrent Laws: parSequence forks the effects") { CF.parSequenceForksTheEffects(EQ_UNIT) },
       Law("Concurrent Laws: onError is run when error is raised") { CF.onErrorIsRunWhenErrorIsRaised(EQ, ctx) },
-      Law("Concurrent Laws: onError is not run when completes normally") { CF.onErrorIsNotRunByDefault(EQK.liftEq(Tuple2.eq(Int.eq(), Boolean.eq())), ctx) }
+      Law("Concurrent Laws: onError is not run when completes normally") { CF.onErrorIsNotRunByDefault(EQK.liftEq(Tuple2.eq(Int.eq(), Boolean.eq())), ctx) },
+      Law("Concurrent Laws: onError outer and inner finalizer is run when error is raised") { CF.outerAndInnerOnErrorIsRun(EQK.liftEq(Eq.any()), ctx) }
     )
   }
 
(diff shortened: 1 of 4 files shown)


modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/Bracket.kt
```diff
@@ -159,4 +159,15 @@ interface Bracket<F, E> : MonadError<F, E> {
         else -> just<Unit>(Unit)
       }
     }
+
+  /**
+   * Executes the given `finalizer` when the source is finished in error.
+   */
+  fun <A> Kind<F, A>.onError(finalizer: Kind<F, Unit>): Kind<F, A> =
```
1Jajen1: I think it might be better to pass the error `E` to the finalizer:
```
  fun <A> Kind<F, A>.onError(finalizer: (E) -> Kind<F, Unit>): Kind<F, A> =
```

abendt: with this signature i get a runtime error i cannot figure out:

```
arrow.fx.extensions.io.concurrent.IODefaultConcurrentKt$concurrent_singleton$1.onError2(Larrow/Kind;Lkotlin/jvm/functions/Function1;)Larrow/Kind;
java.lang.AbstractMethodError: arrow.fx.extensions.io.concurrent.IODefaultConcurrentKt$concurrent_singleton$1.onError2(Larrow/Kind;Lkotlin/jvm/functions/Function1;)Larrow/Kind;
	at arrow.fx.typeclasses.ConcurrentContinuation.onError2(ConcurrentContinuation.kt)
	at arrow.fx.IOTest$46$1.invokeSuspend(IOTest.kt:566)

```

note: the function is named `onError2` in the stacktrace.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
```diff
@@ -674,4 +659,41 @@ object ConcurrentLaws {
...
+
+        val (completed, _) = startLatch.complete(i)
+          .onError(onErrorRun.set(true))
+          .fork(ctx).bind()
+
+        completed.bind()
+
+        startLatch.get().bind() toT onErrorRun.get().bind()
+      }.equalUnderTheLaw(just(i toT false), EQ)
+    }
```
pakoito: Could you please add an additional law to check that

```
raiseError(e).flatMap { raiseError(e).onError(...) }.onError(...)
```

Triggers both. Maybe with an atomic counter.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
```diff
@@ -696,4 +697,19 @@ object ConcurrentLaws {
         startLatch.get().bind() toT onErrorRun.get().bind()
       }.equalUnderTheLaw(just(i toT false), EQ)
     }
+
+  fun <F> Concurrent<F>.outerAndInnerOnErrorIsRun(EQ: Eq<Kind<F, Unit>>, ctx: CoroutineContext) =
+    fx.concurrent {
+      val semaphore = Semaphore(0).bind()
```
pakoito: NIT: I'd rather it was an atomic count that we mutate, and check the count in the just.

pakoito: Up to you, otherwise tell me to merge and I'll do it.

abendt: @pakoito please have another look

pakoito: LGTM. Merging as soon as the build finishes.

pakoito: there it is!

pakoito:
It's in! Party hard :D

PR #2023, opened 2020-02-02
Fix MVar Async

nomisRev:
Due to time-outs some false positive turned up in `MVarAsync`, this improves the `MVar` test suite & also show 2 stack-overflows I was able to fix but need to investigate the impact with the work done on `Traverse` lately.


modules/fx/arrow-fx/src/test/kotlin/arrow/fx/predef.kt
```diff
@@ -13,9 +15,11 @@ import java.util.concurrent.ThreadFactory
...
 import java.util.concurrent.TimeUnit
 
-fun <A> EQ(EQA: Eq<A> = Eq.any(), timeout: Duration = 60.seconds): Eq<Kind<ForIO, A>> = Eq { a, b ->
-  arrow.core.Option.eq(arrow.core.Either.eq(Eq.any(), EQA)).run {
-    a.fix().attempt().unsafeRunTimed(timeout).eqv(b.fix().attempt().unsafeRunTimed(timeout))
+fun <A> EQ(EQA: Eq<A> = Eq.any(), timeout: Duration = 5.seconds): Eq<Kind<ForIO, A>> = Eq { a, b ->
+  arrow.core.Either.eq(Eq.any(), EQA).run {
+    IO.applicative().mapN(a.attempt(), b.attempt()) { (a, b) ->
+      a.eqv(b)
+    }.waitFor(timeout).unsafeRunSync()
```
pakoito: This `unsafeRunSync` is reentrant, right?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/internal/UncancelableMVar.kt
```diff
@@ -56,123 +57,131 @@ internal class UncancelableMVar<F, A> private constructor(initial: State<A>, pri
...
-            val rest = current.takes.drop(1)
-            if (rest.isEmpty()) State.empty()
-            else State.WaitForPut(emptyList(), rest)
-          }
+      is WaitForTake -> justFalse
+      is WaitForPut -> {
+        val reads = current.reads
+        val takes = current.takes
+        var first: Listener<A>? = null
+        val update: State<A> = if (takes.isEmpty()) WaitForTake(a, IQueue.empty()) else {
```
pakoito: NIT: formatting of this if-else

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/predef.kt
pakoito: 'pparently not hahaha Why 60s timeout, and not 5, or 10? When I was testing locally these tests took much longer than everything else on fx.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/typeclasses/Concurrent.kt
```diff
@@ -970,10 +971,10 @@ interface Concurrent<F> : Async<F> {
    * ```
    **/
   fun <A> Kind<F, A>.waitFor(duration: Duration): Kind<F, A> =
-    dispatchers().default().raceN(this, sleep(duration)).flatMap {
+    EmptyCoroutineContext.raceN(sleep(duration), this).flatMap {
```
nomisRev: Without this change the `Eq` instance below  makes the tests hang forever.

```
IO.applicative().mapN(a.attempt(), b.attempt()) { (a, b) -> a.eqv(b) }
  .waitFor(timeout)
  .unsafeRunSync()
```

This change removes a context switch, is guaranteed to fork due to `sleep` being the first argument, allowing `this` to run in its original context. So  `waitFor` now *does not* fork `this` on the `default` dispatcher anymore, which is more desirable IMO.

PR #2025, opened 2020-02-02
Deprecate Duration

i-walker:
Goal for this PR is to deprecate `arrow.fx.typeclasses.Duration` and replace it with `kotlin.time.Duration`. 

There are a few initial problems, which I encountered:
- ExtensionProcessor removes `@ExperimentalTime` [here](https://github.com/arrow-kt/arrow/blob/ebe3dac5aff0cd4b26c1c547070f17c61d226ed7/modules/meta/arrow-meta/src/main/java/arrow/core/extensions/ExtensionProcessor.kt#L203). Where is the right place to add that Annotation to generated Members? -- filtering it out from `removeConstrains` seems not to work properly. 
The issue here is that generated Members do not contain `@ExperimentalTime` and resolve in failed builds.
Ideally it generates: 
```
@JvmName("waitFor")
@ExperimentalTime // this is missing
@Suppress(
    "UNCHECKED_CAST",
    "USELESS_CAST",
    "EXTENSION_SHADOWED_BY_MEMBER",
    "UNUSED_PARAMETER"
)
fun <A> Kind<ForIO, A>.waitFor(duration: kotlin.time.Duration): IO<A> = arrow.fx.IO.concurrent().run {
  this@waitFor.waitFor<A>(duration) as arrow.fx.IO<A>
}
```
- Schedule relays on `arrow.fx.typeclasses.Duration` shall I change the type signature in this PR or create a duplicate with `kotlin.time.Duration`. 

Adding everywhere `@ExperimentalTime` seemed the better approach, because adding `-Xuse-experimental=kotlin.time.ExperimentalTime` to the free compiler Arguments breaks the build. One example:
```
@JvmName("sleep-LRDsOJo")
@Suppress(
    "UNCHECKED_CAST",
    "USELESS_CAST",
    "EXTENSION_SHADOWED_BY_MEMBER",
    "UNUSED_PARAMETER"
)
fun `sleep-LRDsOJo`(arg0: Double): FluxK<Unit> = arrow.fx.reactor.FluxK
   .timer()
   .sleep-LRDsOJo(arg0) as arrow.fx.reactor.FluxK<kotlin.Unit> // UnresolvedReference
```

### DIFF
--- a/modules/fx/arrow-fx/gradle.properties
+++ b/modules/fx/arrow-fx/gradle.properties
@@ -2,3 +2,4 @@
 POM_NAME=Arrow-Fx
 POM_ARTIFACT_ID=arrow-fx
 POM_PACKAGING=jar
+org.gradle.debug=true
--- a/modules/aql/arrow-aql/src/test/kotlin/arrow/aql/tests/IOExample.kt
+++ b/modules/aql/arrow-aql/src/test/kotlin/arrow/aql/tests/IOExample.kt
@@ -5,6 +5,7 @@ import arrow.core.left
 import arrow.core.right
 import arrow.fx.IO
 import arrow.fx.handleErrorWith
+import kotlin.time.ExperimentalTime
 
 data class UserId(val value: String)
 data class User(val userId: UserId)
@@ -69,6 +70,7 @@ class Module {
 }
 
 object test {
+  @ExperimentalTime
   @JvmStatic
   fun main(args: Array<String>) {
     val user1 = User(UserId("user1"))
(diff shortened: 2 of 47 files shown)


modules/fx/arrow-fx/gradle.properties
```diff
@@ -2,3 +2,4 @@
 POM_NAME=Arrow-Fx
 POM_ARTIFACT_ID=arrow-fx
 POM_PACKAGING=jar
+org.gradle.debug=true
```
i-walker: I am debugging the problem. This will be removed before the PR gets merged.

nomisRev:
@i-walker, I am sorry but looking up your issue I just saw it might still be too early to refactor to `Duration`.

cc\ @pakoito 

> Note that this API is in a preview state and has a very high chance of being changed in the future. Do not use it if you develop a library since your library will become binary incompatible with the future versions of the standard library.

[reference](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-use-experimental/index.html)

pakoito:
I don't see the need at the moment, let's wait it out IMO.

i-walker:
😄 No worries Simon. I am happy to contribute to something else if both of you have something in mind in ArrowFx.

PR #2028, opened 2020-02-03
Use Eval in lazyAp instead of a function

### LINKED ISSUE #1953: ["Bug"] Traverse on infinite sequences can stackoverflow
```kotlin
generateSequence { 0 }.k().parTraverse(SequenceK.traverse()) {
        IO { println(it) }
    }.unsafeRunSync()
// => stackoverflow
```
While this code is nonsense anyway because it will infinite loop without ever printing values, it should not stackoverflow and it shows that `traverse` is currently not stacksafe!

This is caused by poor use of `Eval`:
https://github.com/arrow-kt/arrow/blob/712d7cf917cbf60958d1918c113bbc24b22c70f8/modules/core/arrow-core-data/src/main/kotlin/arrow/core/SequenceK.kt#L27
And similar: 
https://github.com/arrow-kt/arrow/blob/712d7cf917cbf60958d1918c113bbc24b22c70f8/modules/core/arrow-core-data/src/main/kotlin/arrow/core/ListK.kt#L160

I honestly don't know how to fix this. This will only become a problem if used with a stackunsafe applicative (Either, Option e.g.) When used on something that is stacksafe and implements lazyAp properly like the normal IO applicative everything will work fine.

### LINKED ISSUE #2024: MVar StackOverflow
StackOverflow with `ListK` and `IO` as shown in the tests of #2023.

I was able to fix the SOE, but haven't investigated properly yet. I know @1Jajen1 was working on this and wanted to discuss again. Here is the diff of `ListK` I was able to fix the SOE with, full Arrow test suite still passed after this change.

```diff
fun <B> foldRight(lb: Eval<B>, f: (A, Eval<B>) -> Eval<B>): Eval<B> {
-  fun loop(fa_p: ListK<A>): Eval<B> = when {
-    fa_p.list.isEmpty() -> lb
-    else -> f(fa_p.fix().list.first(), Eval.defer { loop(fa_p.list.drop(1).k()) })
+  fun loop(fa_p: List<A>): Eval<B> = when {
+    fa_p.isEmpty() -> lb
+    else -> f(fa_p.first(), Eval.defer { loop(fa_p.drop(1)) })
  }

  return Eval.defer { loop(this) }
}
```

```diff
-fun <G, B> traverse(GA: Applicative<G>, f: (A) -> Kind<G, B>): Kind<G, ListK<B>> =
-  foldRight(Eval.now(GA.just(emptyList<B>().k()))) { a, eval ->
-    GA.run { Eval.later { f(a).lazyAp { eval.value().map { xs -> { a: B -> (listOf(a) + xs).k() } } } } }
-  }.value()
+fun <G, B> traverse(GA: Applicative<G>, f: (A) -> Kind<G, B>): Kind<G, ListK<B>> = GA.run {
+  foldRight(Eval.always { GA.just(emptyList<B>()) }) { a, lglb ->
+    f(a).map2Eval(lglb) { (a, aas) -> listOf(a) + aas }
+  }.value().map { it.k() }
+}
```

1Jajen1:
This should solve a few stacksafety problems:
Fixes #1953 
Probably also fixes #2024 ?

Wip because it is missing an implementation for StateT and some rx test failed locally


nomisRev:
Awesome work @1Jajen1! 👏 👏 That's a great fix to have for `0.10.5` 🎉 

If this fixes #1953 then it will also fix #2024.

If so then [this test](https://github.com/arrow-kt/arrow/blob/master/modules/fx/arrow-fx/src/test/kotlin/arrow/fx/MVarTest.kt#L187) should passed be enabled, and you may leave it enabled in this PR.

1Jajen1:
Not sure why, but the `parTraverse`- and `parSequence`-tests seem to hang forever... Also the `MVar`-test you linked needs a slightly higher timeout to pass. My best bet atm is that kotlintest is attempting to shrink a timeout failure... but I won't have the time to investigate further till wednesday.

nomisRev:
@1Jajen1 That's fine I can check that out! Thanks for the fix.

nomisRev:
Hey @1Jajen1, I checked out this branch and looked into the issue. `apLazy` was not accounted for in the `parApplicative` underneath `parTraverse` and it was making it hang.

Adding the following to `ParApplicative.kt` will fix the issue 👍

```
  override fun <A, B> Kind<F, A>.lazyAp(ff: Eval<Kind<F, (A) -> B>>): Eval<Kind<F, B>> =
    _ctx.parMapN(this@lazyAp, ff.value()) { a, f -> f(a) }.let(::Now)
```

1Jajen1:
The `defer` is a good idea :+1: Will add that in a sec, thanks

1Jajen1:
I still don't really get why the default implementation did not go for `parMap` through ap though ^^

rachelcarmena:

Issue #2029, opened 2020-02-04
Improve polymorphic test suite


nomisRev:
Currently the laws are not exhaustive on derived functionality. This can be breaking due to the fact that we can override derived implementations in an `interface`, which is something that is not possible for global derived functions like in Haskell.

`void :: Functor f => f a -> f ()` ~ `fun <F, A> Kind<F, A>.unit(FF: Functor<F>): Kind<F, Unit>`
Which is not override able, and this cannot be broken.

However since in Arrow they are derived within the interface, they can be overridden by more optimized concrete implementations. Which is desirable, but our test suite should make sure we keep the same guarantees when doing so.

- [ ] Functor
- [ ] Apply
- [ ] Applicative
- [ ] ApplicativeError
- [ ] Monad
- [ ] MonadError
- [ ] MonadThrow
- [ ] MonadDefer
- [ ] Bracket
- [ ] Async
- [ ] Concurrent
- [ ] Timer
- [ ] Semigroup
- [ ] Monoid

1Jajen1:
I'd go the opposite direction: Define everything for the highest constraint only, override whats necessary for performance and use those methods in the lower hierarchy. That guarantees consistency and makes it easier to maintain.

There is only one reason against it that I can think of: This will in same cases impose higher constraints than what is strictly necessary. But this happens mostly in mtl where for example, `lift` might only need a functor constraint. But usually one ends up needing the higher constraint (Monad) in mtl anyway so it would not make a big difference.

Won't this be default with arrow-meta anyway? Defining only for highest constraint and later getting that projected on all lower typeclasses. @raulraja

nomisRev:
@1Jajen1 I think you understood this ticket wrong. It's unrelated to what you propose here.

Currently you can do this and `FunctorLaws` would still pass, after this ticket that should no longer be possible anymore.

```
interface IOFunctor : Functor<ForIO> {
  override fun <A, B> IOOf<A>.map(f: (A) -> B): IO<B> =
    fix().map(f)

  override fun <A, B> lift(f: (A) -> B): (Kind<ForIO, A>) -> Kind<ForIO, B> =
    TODO()
}
```

1Jajen1:
Ah so you want to test every part of every typeclass and not just the core methods, got it. That still is an awful lot of tests to add...

I understood this as consistency checks in a typeclass hierarchy...

nomisRev:
[...]

PR #2031, opened 2020-02-04
Polymorphic Ref Test

nomisRev:
As part of the #2029 issue, I would also like to include the concurrency primitive tests into the `Arrow Fx` typeclass laws.

I.e. for any `MonadDefer` instance the `Ref` suite should also pass, this can also be parameterized like `testStackSafety` in case you want a leaner suite.

I'm not sure what the strategy is for `arrow-test` tho, are we planning to keep it a common module that depends on the whole Arrow ecosystem to provide laws for the whole ecosystem? If so shall I move these polymorphic tests there? Or we create a leaner `arrow-test` package, so each library can expose it's own laws in an `arrow-fx-test` module?

I just wanted to start the discussion here since @rachelcarmena is working on the repo organization, I am fine leaving it in Arrow Fx for now. We can also verify the other instances we currently own manually by depending on the modules through `testCompile`.

cc\ @raulraja @pakoito

### DIFF
--- a/modules/fx/arrow-fx/src/test/kotlin/arrow/fx/RefTest.kt
+++ b/modules/fx/arrow-fx/src/test/kotlin/arrow/fx/RefTest.kt
@@ -1,147 +1,162 @@
 package arrow.fx
 
-import arrow.core.extensions.list.traverse.sequence
-import arrow.fx.extensions.fx
-import arrow.fx.extensions.io.applicative.applicative
-import arrow.fx.extensions.io.monad.flatMap
-import arrow.fx.extensions.io.monad.flatten
-import arrow.fx.extensions.io.monad.map
+import arrow.Kind
+import arrow.core.extensions.eq
+import arrow.fx.extensions.io.concurrent.concurrent
 import arrow.fx.extensions.io.monadDefer.monadDefer
+import arrow.fx.typeclasses.Concurrent
+import arrow.fx.typeclasses.MonadDefer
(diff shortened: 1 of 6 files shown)


modules/fx/arrow-fx/src/test/kotlin/arrow/fx/RefTest.kt
```diff
@@ -1,147 +1,162 @@
...
+        val finalValue = 1000
+        RF.just(0).flatMap { r ->
+          (0 until finalValue)
+            .parTraverse { r.update { it + 1 } }
+            .flatMap { r.get() }
+        }.equalUnderTheLaw(just(finalValue), EQF.liftEq(Int.eq()))
       }
     }
+
+    IO.concurrent().tests(IO.eqK(), Ref.factory(IO.monadDefer()))
```
pakoito: ???

raulraja:
It will be less disruptive keeping in just Fx for now. The idea is that each module has its own tests and they all depend on `arrow-check` or `arrow-test` for utilities. Typeclasses in that module should provide their own laws and tests for those laws also will live in the individual module.

pakoito:
> I.e. for any MonadDefer instance the Ref suite should also pass, this can also be parameterized like testStackSafety in case you want a leaner suite.

I'd add these to the MonadDeferLaws, as a list same as we have BracketLaws. If that requires keeping it on arrow-test I'm okay with it.

PR #2034, opened 2020-02-05
Add stack-safety, parallel & timer tests to Concurrent

nomisRev:
Adds some laws to `ConcurrentLaws` and exposes the currently stack unsafe `parMapN` & `raceN` operations. These are currently disabled, and will be fixed before this PR is ready for review.


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
```diff
@@ -98,25 +98,26 @@ object ConcurrentLaws {
       Law("Concurrent Laws: onError outer and inner finalizer is run when error is raised") { CF.waitForTimesOutProgram(EQ) },
       Law("Concurrent Laws: onError outer and inner finalizer is run when error is raised") { CF.waitForTimesOutProgramWithDefault(EQ) }
     ) + (if (testStackSafety) {
+      val iterations = 20_000
```
pakoito: The reason why we had 5k was not to make the machine OOM, and keep the runtimes reasonable.

nomisRev: @pakoito alright, shall we make it configurable and default to `5_000`? It's currently also still fixed to `20_000` in `MonadDeferLaws`.

pakoito: Worth a try to keep it at 20k then. Configurable, for sure!

nomisRev:
@pakoito ready for re-review

modules/fx/arrow-fx-mtl/src/main/kotlin/arrow/fx/mtl/writert.kt
```diff
@@ -47,18 +47,17 @@ interface WriterTBracket<F, W> : Bracket<WriterTPartialOf<F, W>, Throwable>, Wri
...
-          WriterT(wa.just()).flatMap(use).value()
-        }, release = { wa, exitCase ->
-          val r = release(wa.b, exitCase).value()
-          when (exitCase) {
-            is ExitCase.Completed -> r.flatMap { (l, _) -> ref.set(l) }
-            else -> r.unit()
-          }
-        }).flatMap { (w, b) ->
-          ref.get().map { ww -> Tuple2(w.combine(ww), b) }
+      val atomic: AtomicRefW<W> = AtomicRefW(empty())
```
1Jajen1: The `MonadDefer` constraint and the fix to `Throwable` can be removed now right? This means more flexible instances^^

aballano: Just wondering, what's the reasoning to not use Ref here?

nomisRev: Yes to both.
`release` is guaranteed to be called lazily after `use` and before it continues, so removing the wrapping `later` shouldn't be necessary to `defer`.

nomisRev: I'm going to leave this for a separate ticket since this PR touches both Arrow Fx and Arrow Fx Mtl I'd rather close it now instead of re-targeting it after the re-org.

nomisRev:
Follow up for Arrow Mtl Fx. #2080

PR #2038, opened 2020-02-06
Update tests for ApplicativeLaws

danimontoya:
~[End goal: 100% coverage in the test suite]~

Update tests within ApplicativeLaws so we can prevent the user being able to break behaviour.

Original plan was also to extract mapN and tupleN with arity > 3 so these cannot be breaking behaviour. This causes breaking changes (imports) so this won't be done for 0.10.x


pakoito:
>so we can increase coverage of our test suite without too much boilerplate.

don't cheat the code coverage, guvna'

nomisRev:
@pakoito the goal wasn't to reach 100% coverage but to prevent the user being able to break behavior. Since we only need arity 2 & 3 to derive arity `N` I thought it would make sense to only include those 2 in the typeclasses.

But since we cannot do that without breaking, changing imports, we should probably avoid it for `0.10.x` in that case and just test arity 2 & 3 since we don't currently override higher arities in Arrow.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ApplicativeLaws.kt
```diff
@@ -50,12 +51,17 @@ object ApplicativeLaws {
...
-    forAll(Gen.intSmall(), Gen.intSmall(), Gen.intSmall(), Gen.intSmall(), Gen.intSmall(), Gen.intSmall()) { a: Int, b: Int, c: Int, d: Int, e: Int, f: Int ->
-      tupledN(just(a), just(b), just(c), just(d), just(e), just(f)).map { (x, y, z, u, v, w) -> x + y + z - u - v - w }.equalUnderTheLaw(just(a + b + c - d - e - f), EQ)
+  fun <F> Applicative<F>.cartesianBuilderTupled2(EQ: Eq<Kind<F, Int>>): Unit =
+    forAll(Gen.intSmall(), Gen.intSmall()) { a: Int, b: Int ->
+      tupledN(just(a), just(b)).map { (x, y) -> x + y }.equalUnderTheLaw(just(a + b), EQ)
+    }
+
+  fun <F> Applicative<F>.cartesianBuilderTupled3(EQ: Eq<Kind<F, Int>>): Unit =
+    forAll(Gen.intSmall(), Gen.intSmall(), Gen.intSmall()) { a: Int, b: Int, c: Int ->
+      tupledN(just(a), just(b), just(c)).map { (x, y, z) -> x + y + z }.equalUnderTheLaw(just(a + b + c), EQ)
```
nomisRev: We should use `Eq<Tuple3<Int, Int, Int>>` here because `Int` + `Int` + `Int` might result in weird results.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ApplicativeLaws.kt
```diff
@@ -50,12 +51,17 @@ object ApplicativeLaws {
...
+    forAll(Gen.intSmall(), Gen.intSmall(), Gen.intSmall()) { a: Int, b: Int, c: Int ->
+      mapN(just(a), just(b), just(c)) { (x, y, z) -> x + y + z }.equalUnderTheLaw(just(a + b + c), EQ)
     }
 
-  fun <F> Applicative<F>.cartesianBuilderTupled(EQ: Eq<Kind<F, Int>>): Unit =
-    forAll(Gen.intSmall(), Gen.intSmall(), Gen.intSmall(), Gen.intSmall(), Gen.intSmall(), Gen.intSmall()) { a: Int, b: Int, c: Int, d: Int, e: Int, f: Int ->
-      tupledN(just(a), just(b), just(c), just(d), just(e), just(f)).map { (x, y, z, u, v, w) -> x + y + z - u - v - w }.equalUnderTheLaw(just(a + b + c - d - e - f), EQ)
+  fun <F> Applicative<F>.cartesianBuilderTupled2(EQ: Eq<Kind<F, Int>>): Unit =
+    forAll(Gen.intSmall(), Gen.intSmall()) { a: Int, b: Int ->
+      tupledN(just(a), just(b)).map { (x, y) -> x + y }.equalUnderTheLaw(just(a + b), EQ)
```
nomisRev: `Eq<Tuple2<Int, Int>>`

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ApplicativeLaws.kt
```diff
@@ -50,12 +51,17 @@ object ApplicativeLaws {
     }
 
   fun <F> Applicative<F>.cartesianBuilderMap(EQ: Eq<Kind<F, Int>>): Unit =
-    forAll(Gen.intSmall(), Gen.intSmall(), Gen.intSmall(), Gen.intSmall(), Gen.intSmall(), Gen.intSmall()) { a: Int, b: Int, c: Int, d: Int, e: Int, f: Int ->
-      mapN(just(a), just(b), just(c), just(d), just(e), just(f)) { (x, y, z, u, v, w) -> x + y + z - u - v - w }.equalUnderTheLaw(just(a + b + c - d - e - f), EQ)
+    forAll(Gen.intSmall(), Gen.intSmall(), Gen.intSmall()) { a: Int, b: Int, c: Int ->
+      mapN(just(a), just(b), just(c)) { (x, y, z) -> x + y + z }.equalUnderTheLaw(just(a + b + c), EQ)
```
nomisRev: `Eq<Tuple3<Int, Int, Int>>`

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ApplicativeLaws.kt
1Jajen1: We should have a general rule that using commutative monoids should not be allowed at all in those kind of tests. It fails to uncover trivial ordering errors.

nomisRev: Yes! I've been guilty of this before too.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ApplicativeLaws.kt
```diff
@@ -49,13 +66,33 @@ object ApplicativeLaws {
...
+      tupledN(just(a), just(b)).equalUnderTheLaw(just(Tuple2(a, b)), EQ)
+    }
+
+  fun <F> Applicative<F>.cartesianBuilderTupled3(EQ: Eq<Kind<F, Tuple3<Int, Int, Int>>>): Unit =
+    forAll(Gen.intSmall(), Gen.intSmall(), Gen.intSmall()) { a: Int, b: Int, c: Int ->
+      tupledN(just(a), just(b), just(c)).equalUnderTheLaw(just(Tuple3(a, b, c)), EQ)
+    }
+
+  fun <F> Applicative<F>.replicateSize(EQ: Eq<Kind<F, Int>>): Unit =
+    forAll(Gen.intSmall()) { n ->
```
nomisRev:
```suggestion
    forAll(Gen.choose(0, 100)) { n ->
```

I looked into `Gen.intSmall` and it's defined for values that would still `StackOverflow` which is  something we do not protect from in `Aplicative` so we should use `Gen.choose(0, 100)` instead for `replicate`.

PR #2040, opened 2020-02-06
Nits oncancel onerror

aballano:
Because "code is worth a thousand words", following [this comment](https://github.com/arrow-kt/arrow/issues/2018#issuecomment-582947509):

* Add the error instance to Bracket's `onError`
* Remove redeclaration of `onCancel` and `onError` in IO

### DIFF
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
@@ -51,7 +51,6 @@ object ConcurrentLaws {
       Law("Concurrent Laws: release is not cancelable") { CF.releaseBracketIsNotCancelable(EQ, ctx) },
       Law("Concurrent Laws: cancel on guarantee runs finalizer") { CF.guaranteeFinalizerOnCancel(EQ, ctx) },
       Law("Concurrent Laws: release is not cancelable") { CF.guaranteeFinalizerIsNotCancelable(EQ, ctx) },
-      Law("Concurrent Laws: cancel on onCancel runs finalizer") { CF.onCancelFinalizerOnCancel(EQ, ctx) },
       Law("Concurrent Laws: async cancelable coherence") { CF.asyncCancelableCoherence(EQ) },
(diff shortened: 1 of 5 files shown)


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
```diff
@@ -670,7 +670,7 @@ object ConcurrentLaws {
         val errorLatch = Promise<F, Int>(this@onErrorIsRunWhenErrorIsRaised).bind()
```
nomisRev:
```suggestion
        val errorLatch = Promise<F, Throwable>(this@onErrorIsRunWhenErrorIsRaised).bind()
```

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
```diff
@@ -670,7 +670,7 @@ object ConcurrentLaws {
...
 
         startLatch.complete(Unit).flatMap { raiseError<Exception>(RuntimeException("Boom")) }
-          .onError(errorLatch.complete(i))
+          .onError { errorLatch.complete(i) }
```
nomisRev:
```suggestion
          .onError(errorLatch::complete)
```

This way you can use `Gen.throwable()` and assert that the correct pas used `!effect { waitExit shouldBe generatedThrowable }` and verify with `Eq<Kind<F, Unit>>` instead.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/ConcurrentLaws.kt
```diff
@@ -711,8 +711,8 @@ object ConcurrentLaws {
 
       just(Unit).flatMap {
         raiseError<Unit>(RuntimeException("failed"))
-          .onError(incrementCounter)
-      }.onError(incrementCounter)
+          .onError { incrementCounter }
```
arkivanov: What is the point of such a change? `incrementCounter` is just a `val`, looks like this makes no sense.

PR #2043, opened 2020-02-08
add a SemigroupK instance for IO

abendt:
this PR adds a SemigroupK instance for IO

### DIFF
--- a/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/io.kt
+++ b/modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/io.kt
@@ -4,7 +4,6 @@ import arrow.Kind
 import arrow.core.Either
 import arrow.core.identity
 import arrow.extension
-
 import arrow.fx.ForIO
 import arrow.fx.IO
 import arrow.fx.IODispatchers
@@ -43,6 +42,7 @@ import arrow.typeclasses.MonadError
 import arrow.typeclasses.MonadThrow
 import arrow.typeclasses.Monoid
 import arrow.typeclasses.Semigroup
+import arrow.typeclasses.SemigroupK
 import arrow.unsafe
 import kotlin.coroutines.CoroutineContext
 import arrow.fx.handleError as ioHandleError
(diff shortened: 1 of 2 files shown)


pakoito:
...this passes tests? nani?

1Jajen1:
It makes sense when you consider `SemigroupK` part of the alternative hierarchy and the only law required is associativity:  `(a + b) + c == a + (b + c)`.

These laws are tested with a generator that generates errors and pure values for io so the test is correct. The code is as well, it does hold the laws: Consider `a`, `b` and `c` binary cases of error or no-error and you have a boolean algebra which trivially passes the associativity law if `+ ~ or`. This also holds under side effects etc because in the case of `a + (b + c)` the `b + c` will only run if `a` has an error and `c` is only run if `a` and `b` error out. `(a + b) + c` also follows the same pattern.

The only thing I am not so sure about is if we want this instance, but that is something for @nomisRev to comment on I guess ^^
Oh and I do not know how well this plays with the bifunctor IO changes, so I'd definitly wait for simon to look at this.

There are probably also other `SemigroupK` instances regarding fire and forget (concurrent) computation etc.

Note: For this kind of hierarchy for error handling we cannot define `MonoidK` or `Alternative` because the empty element would have to be `IO.RaiseError ~ 0`. This fails the monoid law `x + empty == x` because if `x == IO.RaiseError` the error in `x` is thrown away and replaced by `empty` which will have a different error that is not `x`.

nomisRev:
@1Jajen1 thanks for the clear explanation! This is similar to 

`fun optionSemigroup(): Semigroup<Option<A>> = Semigroup { a, b -> a orElse b }`

but that's not what we currently provide in Arrow. All datatypes also use the underlying semigroup for `A` in case both cases are successful. Which is not the case now.

I.e. I'd expect `IO.just(1).combine(IO.just(2))` to return `IO.Pure(3)`

1Jajen1:
> @1Jajen1 thanks for the clear explanation! This is similar to
> 
> `fun optionSemigroup(): Semigroup<Option<A>> = Semigroup { a, b -> a orElse b }`
> 
> but that's not what we currently provide in Arrow. All datatypes also use the underlying semigroup for `A` in case both cases are successful. Which is not the case now.
> 
> I.e. I'd expect `IO.just(1).combine(IO.just(2))` to return `IO.Pure(3)`

Using or not using the `A` is the difference between `SemigroupK` and `Semigroup`.

Thing is: I really dislike the `Alternative` hierarchy in arrow:
- SemigroupK has very very few instances where there is no MonoidK
- MonoidK is almost always identical to Alternative
- MonadCombine is just useless

Imo we should clean this up and get rid of everything except MonadPlus #2047 and Alterntive

nomisRev:
@1Jajen1 if you have a proposal to how we can improve this hierarchy then I think we should because this has been a bit confusing to me and `MonadPlus` and `Alternative` make more sense to me after looking through the linked code.

The separation and resulting powers are much more straight forward and user-friendly.

rachelcarmena:

Issue #2033, opened 2020-02-05
["Request"] LogicT


pakoito:
https://hackage.haskell.org/package/logict

PR #2047, opened 2020-02-11
MonadPlus & MonadLogic

abendt:
part of #2033 

this PR adds MonadPlus & MonadLogic typeclass with some instances.

* still need to add some documentation 
* need a good name for the (>>-) operator. named it `fairConjunction` for now.
* i had some issues with the StateT tests, they were running very long. didn't understand why. i used less iterations


modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+  fun <A, B> Kind<F, A>.fairConjunction(ffa: (A) -> Kind<F, B>): Kind<F, B> =
+    msplit().flatMap { option ->
+      option.fold({ mzero<Tuple2<Kind<F, A>, A>>() }, { just(it) })
+    }.flatMap { (fa, a) ->
+      ffa(a).interleave(fa.flatMap(ffa))
+    }
+
+  fun <A, B> Kind<F, A>.ifte(fb: Kind<F, B>, ffa: (A) -> Kind<F, B>): Kind<F, B> =
```
pakoito: I'm okay with longer names. `ifThen` if that's what this does, or something more descriptive 🙏

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
```diff
@@ -0,0 +1,45 @@
...
+import arrow.Kind
+import arrow.core.ForOption
+import arrow.core.Option
+import arrow.core.Tuple2
+import arrow.core.fix
+import arrow.core.identity
+
+interface MonadLogic<F> : MonadPlus<F> {
+
+  fun <A> Kind<F, A>.msplit(): Kind<F, Option<Tuple2<Kind<F, A>, A>>>
```
pakoito: `splitM` if we're following Arrow's nomenclature

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadPlus.kt
```diff
@@ -0,0 +1,9 @@
+package arrow.typeclasses
+
+import arrow.Kind
+
+interface MonadPlus<F> : Monad<F>, Alternative<F> {
+  fun <A> mzero(): Kind<F, A> = empty()
```
pakoito: `zeroM`

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadPlus.kt
```diff
@@ -0,0 +1,9 @@
...
+
+  fun <A> Kind<F, A>.mplus(other: Kind<F, A>): Kind<F, A> = alt(other)
```
pakoito: `plusM`

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadLogicLaws.kt
```diff
@@ -0,0 +1,136 @@
...
+
+object MonadLogicLaws {
+
+  fun <F> laws(
+    ML: MonadLogic<F>,
+    GK: GenK<F>,
+    EQK: EqK<F>,
+    iterations: Int = 1000
+  ): List<Law> =
+    monadLogicLaws(ML, GK, EQK, iterations) +
```
pakoito: Nit: we never sort this list, so it may be better to do MonadPlusLaws first.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
```diff
@@ -0,0 +1,45 @@
...
+        ffa(a).mplus(fa.flatMap { ffa(it) })
+      }.fold({ fb }, ::identity)
+    }
+
+  fun <A> Kind<F, A>.once(): Kind<F, A> =
+    msplit().flatMap { option ->
+      option.fold({ mzero<A>() }, { just(it.b) })
+    }
+
+  fun <A> Kind<F, A>.lnot(): Kind<F, Unit> =
```
pakoito: We're calling this `void` in `Functor`. cc @1Jajen1 is it the same?

pakoito: EDIT: after checking with Haskell, it isn't

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
```diff
@@ -0,0 +1,45 @@
...
+  fun <A> Kind<F, A>.msplit(): Kind<F, Option<Tuple2<Kind<F, A>, A>>>
+
+  fun <A> Kind<F, A>.interleave(fa: Kind<F, A>): Kind<F, A> =
+    this.msplit().flatMap { option ->
+      option.map { (fa, a) ->
+        just(a).mplus(fa.interleave(fa))
+      }.fold({ fa }, ::identity)
+    }
+
+  fun <A, B> Kind<F, A>.fairConjunction(ffa: (A) -> Kind<F, B>): Kind<F, B> =
```
pakoito: https://www.powerthesaurus.org/interleave/antonyms

`unweave`

pakoito:
Buncha things around names. Feel free to ignore those you don't agree with.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
abendt: what do you think of `unitIfValue` here?

modules/mtl/arrow-mtl/src/main/kotlin/arrow/mtl/extensions/kleisli.kt
```diff
@@ -249,3 +254,30 @@ interface KleisliEqK<F, D> : EqK<KleisliPartialOf<F, D>> {
...
+  override fun MF(): Monad<F> = ML()
+  override fun AL(): Alternative<F> = ML()
+
+  override fun <A> Kind<KleisliPartialOf<F, D>, A>.splitM(): Kind<KleisliPartialOf<F, D>, Option<Tuple2<Kind<KleisliPartialOf<F, D>, A>, A>>> =
+    this.fix().let { fa ->
+      Kleisli(
+        AndThen(fa.run).andThen {
+          ML().run {
+            it.splitM().flatMap { option ->
+              option.fold({ just(Option.empty<Tuple2<Kind<KleisliPartialOf<F, D>, A>, A>>()) }, { (fa, a) -> just(Option.just(Kleisli.liftF<F, D, A>(fa) toT a)) })
```
1Jajen1: Please avoid choosing colliding names, makes this confusing to read

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
pakoito: We use `void` in Functor, better keep it consistent.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
```diff
@@ -0,0 +1,45 @@
...
+      option.map { (fa, a) ->
+        just(a).plusM(fa.interleave(fa))
+      }.fold({ fa }, ::identity)
+    }
+
+  fun <A, B> Kind<F, A>.unweave(ffa: (A) -> Kind<F, B>): Kind<F, B> =
+    splitM().flatMap { option ->
+      option.fold({ zeroM<Tuple2<Kind<F, A>, A>>() }, { just(it) })
+    }.flatMap { (fa, a) ->
+      ffa(a).interleave(fa.flatMap(ffa))
```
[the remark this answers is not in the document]
abendt: yes

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
```diff
@@ -0,0 +1,45 @@
...
+    splitM().flatMap { option ->
+      option.fold({ zeroM<A>() }, { just(it.b) })
+    }
+
+  fun <A> Kind<F, A>.unitIfValue(): Kind<F, Unit> =
+    once().ifThen(unit()) { zeroM() }
+}
+
+fun <F, A> Kind<ForOption, Tuple2<Kind<F, A>, A>>.reflect(ML: MonadLogic<F>): Kind<F, A> =
+  fix().fold({ ML.zeroM() }, { ML.run { just(it.b).plusM(it.a) } })
```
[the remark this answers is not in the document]
abendt: this didn't work. the generated code didn't compile.

modules/mtl/arrow-mtl/src/main/kotlin/arrow/mtl/extensions/kleisli.kt
abendt: yes

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
```diff
@@ -0,0 +1,45 @@
...
+import arrow.core.identity
+
+interface MonadLogic<F> : MonadPlus<F> {
+
+  fun <A> Kind<F, A>.splitM(): Kind<F, Option<Tuple2<Kind<F, A>, A>>>
+
+  fun <A> Kind<F, A>.interleave(fa: Kind<F, A>): Kind<F, A> =
+    this.splitM().flatMap { option ->
+      option.map { (fa, a) ->
+        just(a).plusM(fa.interleave(fa))
```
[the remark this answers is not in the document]
abendt: changed it

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadPlusLaws.kt
```diff
@@ -0,0 +1,67 @@
...
+    return MonadLaws.laws(MP, GK, EQK) +
+      AlternativeLaws.laws(MP, GK, EQK) +
+      monadPlusLaws(MP, G, EQ)
+  }
+
+  private fun <F> monadPlusLaws(MP: MonadPlus<F>, G: Gen<Kind<F, Int>>, EQ: Eq<Kind<F, Int>>): List<Law> =
+    listOf(
+      Law("MonadPlus Laws: Left identity") { MP.leftIdentity(G, EQ) },
+      Law("MonadPlus Laws: Right identity") { MP.rightIdentity(G, EQ) },
+      Law("MonadPlus Laws: associativity") { MP.associativity(G, EQ) },
```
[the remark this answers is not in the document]
abendt: don't understand why they are already tested? i mean the `plusM` function now is only tested from  within `MonadPlusLaws`?

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/sequenceK.kt
```diff
@@ -397,3 +399,32 @@ interface SequenceKEqK : EqK<ForSequenceK> {
...
+  override fun <A> Kind<ForSequenceK, A>.splitM(): Kind<ForSequenceK, Option<Tuple2<Kind<ForSequenceK, A>, A>>> {
+    val seqA = this.fix().sequence
+    val iterA = seqA.iterator()
+
+    return if (iterA.hasNext()) {
+      just(Option.just(seqA.drop(1).k() toT iterA.next()))
+    } else {
+      just(Option.empty())
+    }
+  }
```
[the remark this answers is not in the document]
abendt: > There are a few problems, especially around the sequenceK instance.

will have a look at it

> Also does WriterT have an instance for MonadLogic as well?

yes, added it. currently one test is failing. could you have a look please?
java:test://arrow.mtl.WriterTTest.MonadLogic Laws: reflect is inverse of msplit

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadPlusLaws.kt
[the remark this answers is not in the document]
1Jajen1: `plusM == orElse == alt == combineK` right? And those are all tested by `SemigroupK` for associativity and by `MonoidK` for the left and right identity.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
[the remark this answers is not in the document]
1Jajen1: Hmm, could you add a note there then, so that when we rework this with arrow-meta we'll not forget to move it?

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
```diff
@@ -0,0 +1,45 @@
...
+      }.fold({ fa }, ::identity)
+    }
+
+  fun <A, B> Kind<F, A>.unweave(ffa: (A) -> Kind<F, B>): Kind<F, B> =
+    splitM().flatMap { option ->
+      option.fold({ zeroM<Tuple2<Kind<F, A>, A>>() }, { just(it) })
```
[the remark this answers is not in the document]
1Jajen1: Turns out I was looking at some older version...

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/sequenceK.kt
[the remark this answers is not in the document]
1Jajen1: Turns out I was looking at an old version of MonadLogic and that had a WriterT instance which got removed because it was unlawful. Good that our tests catch that ^^ Sorry about that, we can just remove it I think.

1Jajen1:
Just finished reading the paper that the haskell package cites, the tic-tac-toe example it provides could be a nice one for the docs later on :)

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadPlusLaws.kt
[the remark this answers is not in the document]
abendt: true, `plusM` is just using `alt`. What i was trying to say:
In my understanding nothing would prevent me from implementing plusM like
`fun <A> Kind<F, A>.plusM(other) = TODO()` if there is not test that is actually using it.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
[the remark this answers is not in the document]
abendt: ok

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/sequenceK.kt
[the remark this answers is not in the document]
abendt: ok, i removed the WriterT instance

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadPlusLaws.kt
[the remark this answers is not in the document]
1Jajen1: These consistency issues are quite annoying and we don't really have a way to properly test this as of now. Not sure what the right approach is... Let's keep the duplicated laws for now as they don't really cause any problems.

modules/core/arrow-core/src/main/kotlin/arrow/core/extensions/sequenceK.kt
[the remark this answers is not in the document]
abendt: re the sequence: i don't think it's possible to copy an iterator without knowing its implementation. so i don't think this approach is possible here. do you see any other options?
i changed it to your first suggestion. i understand it has the same performance issues, however is way nicer.

abendt:
i have added some more docs now, so i think it can be reviewed now.

@1Jajen1 how should we proceed with the SequenceK instance now? i don't know how to do it without using `drop`.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
nomisRev: @pakoito we didn't settle on `void` yet did we? Last vote went to `devnull`? ;p

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
```diff
@@ -0,0 +1,45 @@
...
+  fun <A, B> Kind<F, A>.ifThen(fb: Kind<F, B>, ffa: (A) -> Kind<F, B>): Kind<F, B> =
+    splitM().flatMap { option ->
+      option.map { (fa, a) ->
+        ffa(a).plusM(fa.flatMap { ffa(it) })
+      }.fold({ fb }, ::identity)
+    }
+
+  fun <A> Kind<F, A>.once(): Kind<F, A> =
+    splitM().flatMap { option ->
+      option.fold({ zeroM<A>() }, { just(it.b) })
```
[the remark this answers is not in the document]
nomisRev: How come `None` is impossible here? Throwing seems strange here to me.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadPlus.kt
```diff
@@ -0,0 +1,9 @@
+package arrow.typeclasses
+
+import arrow.Kind
+
+interface MonadPlus<F> : Monad<F>, Alternative<F> {
```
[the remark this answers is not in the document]
nomisRev: There seems to be no reason for `MonadFilter` in `MonadCombine`. It relies on `empty` which is also already available through `MonoidK` from `Alternative`.

It seems we can just replace `MonadCombine` with `MonadPlus` IMO. It looks like we can implement its functions in this typeclass as well with less constraints.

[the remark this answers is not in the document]
nomisRev: Additionally, how would this impact `SemigroupK` or your comment [in #2047](https://github.com/arrow-kt/arrow/pull/2043#issuecomment-585203774)

1Jajen1:
Will take a closer look later today :+1: I think code-wise this is almost good to go. The only problem I have is that sequence is the only useful instance and yet is becomes very slow if it explores too many branches. The list instance is pointless as it'll just eat up memory in everything but trivial examples. The problem is that the continuation based `LogicT` is really hard to get right in kotlin.

If I have some time the next days I'll try to get the sequence instance to share the iterator somehow, but we can merge that before if the docs are good.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadLogic.kt
```diff
@@ -0,0 +1,188 @@
...
+
+import arrow.Kind
+import arrow.core.ForOption
+import arrow.core.Option
+import arrow.core.Tuple2
+import arrow.core.fix
+
+/**
+ * MonadLog is a typeclass that extends a MonadPlus. It provides functions to control
+ * when computations should be performed.
```
1Jajen1: This is a bit weak imo. A better description would be that it improves upon `MonadPlus` with fairness guarantees.

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadPlus.kt
```diff
@@ -0,0 +1,52 @@
+package arrow.typeclasses
+
+import arrow.Kind
+
+/**
+ *  MonadPlus is a typeclass that extends a Monad by supporting choice and failure also.
```
1Jajen1:
```suggestion
 *  MonadPlus is a typeclass that extends a Monad by supporting choice and failure.
 *  It is equal to [Alternative] in its api, but provides additional laws for how `flatMap` and `empty` interact.
```
Something like this might be a bit more descriptive

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadPlus.kt
```diff
@@ -0,0 +1,52 @@
...
+ */
+interface MonadPlus<F> : Monad<F>, Alternative<F> {
+  /**
+   * the identity of `mPlus`
```
1Jajen1:
```suggestion
   * Identity for `mPlus`. [MonadPlus] variant of [empty]
```

modules/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/MonadPlus.kt
```diff
@@ -0,0 +1,52 @@
...
+   *   }
+   *   //sampleEnd
+   *   println(result)
+   * }
+   * ```
+   */
+  fun <A> zeroM(): Kind<F, A> = empty()
+
+  /**
+   * an associative operation to combine two structures.
```
1Jajen1:
```suggestion
   * Associative operation to combine two structures. [MonadPlus] variant of [orElse].
```

modules/core/arrow-core-data/src/test/kotlin/arrow/core/SequenceKTest.kt
```diff
@@ -52,6 +56,11 @@ class SequenceKTest : UnitSpec() {
     val EQ = SequenceK.eq(Int.eq())
 
     testLaws(
+      MonadPlusLaws.laws(
```
1Jajen1: What is the point testing `MonadPlus` and `MonadLogic` seperately? `MonadLogic` alone should test both right?

modules/docs/arrow-docs/docs/docs/arrow-core/typeclasses/monadlogic/README.md
```diff
@@ -0,0 +1,70 @@
...
+layout: docs-core
+title: MonadLogic
+permalink: /docs/arrow/typeclasses/monadlogic/
+redirect_from:
+  - /docs/typeclasses/monadlogic/
+---
+
+## MonadLogic
+
+The `MonadLogic` typeclass provides some functions to control when computations should be performed.
```
1Jajen1: A much better explanation would be to show why and where [MonadPlus] is unfair (left-biased usually) and how [MonadLogic] improves that. In general mentioning fairness is always good here.

modules/docs/arrow-docs/docs/docs/arrow-core/typeclasses/monadlogic/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+#### Kind<F, A>.splitM(): Kind<F, Option<Tuple2<Kind<F, A>, A>>>
+
+Splits a computations into its first result and another computation that computes the rest.
+
+#### <F, A>.interleave(other: Kind<F, A>): Kind<F, A>
+
+Some computations can have an infinite number of potential results. Such computations can cause problems in 
+some circumstances as combined computations may never terminate. `Interleave` allows to combine computations
+in a way that fair consideration of branches is ensured.
```
1Jajen1: An example using [MonadPlus] here to show unfairness would be good. The paper that the haskell package listed has some good toy examples.

modules/docs/arrow-docs/docs/docs/arrow-core/typeclasses/monadlogic/README.md
```diff
@@ -0,0 +1,70 @@
...
+#### Kind<F, A>.unweave(ffa: (A) -> Kind<F, B>): Kind<F, B>
+
+#### Kind<F, A>.ifThen(fb: Kind<F, B>, ffa: (A) -> Kind<F, B>): Kind<F, B>
+
+Logical conditional. The equivalent of Prolog's soft-cut. If its first argument succeeds at all, 
+then the results will be fed into the success branch. Otherwise, the failure branch is taken.
+
+#### Kind<F, A>.once(): Kind<F, A>
+
+Pruning. Selects one result out of many. Useful for when multiple results of a computation will be equivalent, or should be treated as such.
```
1Jajen1: Quick note here to show how this improves performance as well.

modules/docs/arrow-docs/docs/docs/arrow-core/typeclasses/monadplus/README.md
```diff
@@ -0,0 +1,71 @@
...
+title: MonadPlus
+permalink: /docs/arrow/typeclasses/monadplus/
+redirect_from:
+  - /docs/typeclasses/monadplus/
+---
+
+## MonadPlus
+
+The `MonadPlus` typeclass lets us combine two structures of type `Kind<F, A>` with a associative function `plusM`
+into a `Kind<F, A>`. It also provides the function `zeroM` which acts as a neutral argument to the `plusM` function.
```
1Jajen1: Some comment on the relation between this and Alternative would be good. Just note how this is basically Alternative with a Monad constraint and additional laws.

rachelcarmena:

PR #2058, opened 2020-02-14
Increase test suite for MonadLaws

danimontoya:
Covers small part of #2029 .

Increase test suite for MonadLaws so we can prevent the user being able to break behaviour.


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadLaws.kt
```diff
@@ -117,4 +133,39 @@ object MonadLaws {
...
+
+  fun <F> Monad<F>.derivedMProductConsistent(GK: GenK<F>, EQ: Eq<Kind<F, Tuple2<Int, Int>>>): Unit =
+    forAll(GK.genK(Gen.int()), Gen.functionAToB<Int, Kind<F, Int>>(Gen.int().applicative(this))) { fa, fb: (Int) -> Kind<F, Int> ->
+      fa.mproduct(fb).equalUnderTheLaw(fa.flatMap { a -> fb(a).map { Tuple2(a, it) } }, EQ)
+    }
+
+  fun <F> Monad<F>.derivedIfMConsistent(GK: GenK<F>, EQ: Eq<Kind<F, Int>>): Unit =
+    forAll(GK.genK(Gen.bool()), Gen.functionToA(Gen.int().applicative(this)), Gen.functionToA(Gen.int().applicative(this))) { fa, fTrue, fFalse ->
+      fa.ifM(fTrue, fFalse).equalUnderTheLaw(fa.flatMap { if (it) fTrue() else fFalse() }, EQ)
+    }
```
1Jajen1: Please don't use `Gen<A>.applicative()` unless you really want a type to only be `just(a)`. This will for example never generate the `None` case of `Option` or `Left` of `Either`. Prefer the `GenK.genK(genA)` version as that properly handles those cases.

1Jajen1: Oh and the same applies to the `Gen.eval` generator you added, please use `GenK` instead, it provides much better test-data.

nomisRev: Should we get rid of `Gen<A>.applicative()` in that case? Or deprecate it for ourselves?

aballano: @nomisRev I think it depends, you might need gens of pure liftings in certain situations. In a recent PR i'm actually adding one explicit for raising errors only

1Jajen1: That behaviour is indeed useful in some situations, but I'd prefer to get rid of it and write an explicit `genA.map { just(it) }` instead. Since most tests already have the instance in scope that should be as easy as writing `genA.applicative(this)`.
Another option is to rename the generator because `applicative` is easy to misunderstand. Something like `Gen.pure`, `Gen.just` etc might be better to explicitly show what this generator produces. Plus a note that points out the weaknesses and refers to `GenK`.

danimontoya: Thank you @1Jajen1 for the explanation!! I've removed eval and replaced applicative in the tests

danimontoya:
> I am still not quite sure I like the duplication of logic to test this, but it'll work 😅

AFAIK this prevents the user to break behaviour.. meaning, if any API gets overridden we still want to be sure that behaviour is the same. Same approach is followed for some other API laws.

danimontoya:
Merging this, if there is any comment about it feel free to shout and we'll fix it

PR #2059, opened 2020-02-14
Increase test suite for MonadDeferLaws

aballano:
Covers small part of #2029

Added derived test for `later(fa: Kind<F, A>)` and `lazy()` to check right implementation


aballano:
This test failed, let's see after re-running

>arrow.fx.MVarTest > CancelableMVar - concurrent take and put FAILED
        arrow.fx.internal.TimeoutException: Duration(amount=5, timeUnit=SECONDS)

PR #2060, opened 2020-02-14
Add MonadThrow laws

danimontoya:
Covers small part of #2029 .

Create test suite for MonadThrowLaws.

### DIFF
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadThrowLaws.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadThrowLaws.kt
@@ -1,0 +1,57 @@
+package arrow.test.laws
+
+import arrow.Kind
+import arrow.core.extensions.eq
+import arrow.test.generators.GenK
+import arrow.test.generators.fatalThrowable
+import arrow.test.generators.throwable
+import arrow.typeclasses.Apply
+import arrow.typeclasses.Eq
+import arrow.typeclasses.EqK
+import arrow.typeclasses.Functor
+import arrow.typeclasses.MonadThrow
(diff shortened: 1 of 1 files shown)


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/MonadThrowLaws.kt
```diff
@@ -0,0 +1,57 @@
...
+import arrow.typeclasses.EqK
+import arrow.typeclasses.Functor
+import arrow.typeclasses.MonadThrow
+import arrow.typeclasses.Selective
+import io.kotlintest.fail
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.forAll
+import io.kotlintest.shouldThrowAny
+
+object MonadThrowLaws {
```
aballano: I believe you should probably hook this up in `MonadDeferLaws`'s laws functions as:

```
BracketLaws.laws(...) +
  MonadThrowLaws.laws(...) +
      monadDeferLaws(...)
``` 

Although I'm unsure if this won't run all MonadError tests twice as both Bracket and MonadThrow come from MonadError, but independently

danimontoya: yeah, I guess that should be aligned so it doesn't execute the tests more times than required

aballano: Ah nevermind, I just noticed that in `testLaws` function there's a distinct! So good to go with the suggested change

danimontoya: perfect then

PR #2064, opened 2020-02-15
Increase test suite for TimerLaws

aballano:
Covers small part of #2029

Add TimerLaws as part of ConcurrentLaws


nomisRev:
Looks good 👏 👏 👏

PR #2065, opened 2020-02-15
Conf: extract versions

rachelcarmena:
## Changes

- Extract versions from modules.
- Order versions (reason: I've seen several duplicated versions because maybe it's not easy to find them if they don't have an order)

## Question

About JVM 1.7 or 1.8:

- In the main `build.gradle`:
```
    sourceCompatibility = 1.7
    targetCompatibility = 1.7
```
- In `modules/benchmarks/arrow-kio-benchmarks/build.gradle`:
```
compileKotlin {
    kotlinOptions.jvmTarget = "1.8"
}
compileTestKotlin {
    kotlinOptions.jvmTarget = "1.8"
}
```
- On the other hand:
```
modules/meta/arrow-meta/build.gradle
modules/meta/arrow-meta/models/build.gradle
modules/benchmarks/arrow-kio-benchmarks/build.gradle
modules/ank/arrow-ank-gradle/build.gradle
```
include:
```
org.jetbrains.kotlin:kotlin-stdlib-jdk8:$KOTLIN_VERSION
```
However, the rest of `build.gradle` files include:
```
org.jetbrains.kotlin:kotlin-stdlib-jdk7:$KOTLIN_VERSION
```

Does it make sense if I try to unify them with `1.8`?


rachelcarmena:
Build error:

```
arrow.fx.MVarTest > CancelableMVar - concurrent take and put FAILED
    arrow.fx.internal.TimeoutException: Duration(amount=5, timeUnit=SECONDS)
```

rachelcarmena:
Thanks @pakoito !!! :kissing_heart:

rachelcarmena:
BTW @pakoito  :pray: , what do you think about the question I included in the description? Does it make sense if I try to unify them with `1.8`? Thanks!!

pakoito:
7 is a requirement for old versions of Android, and KIO may only work on newer ones. Better keep them separated.

rachelcarmena:
> 7 is a requirement for old versions of Android, and KIO may only work on newer ones. Better keep them separated.

Thank you so much for the info, @pakoito !!!

PR #2076, opened 2020-02-17
Increase test suite for MonoidLaws and SemiringLaws

aballano:
Covers small part of #2029

* Adds derived and empty test for Monoid's combineAll
* Removed Semiring test for combine as it's testing Semigroup's
* Add MonoidLaws to SemiringLaws
* Made Semiring laws use Gen for values instead
* Add extra tests for SemiringLaws
* Removed duplicated tests from NumberSemiringTest
* **Changed Schedule impl** for Semiring's combineMultiplicate function to perform the multiplication like combine does


modules/core/arrow-core-data/src/test/kotlin/arrow/core/extensions/NumberSemiringTest.kt
```diff
@@ -1,80 +1,31 @@
 package arrow.core.extensions
 
 import arrow.test.UnitSpec
+import arrow.test.generators.byte
+import arrow.test.generators.short
 import arrow.test.laws.SemiringLaws
+import io.kotlintest.properties.Gen
 import io.kotlintest.shouldBe
 
 class NumberSemiringTest : UnitSpec() {
```
danimontoya: Is this needed? isn't it covered by
>   SemiringLaws.laws(
        Schedule.semiring<ForId, Any?, Int>(Int.monoid(), Id.monad()),
        Schedule.genK<ForId, Int>(Id.monad()),
        eqK(Id.eqK(), Id.monad(), 0 as Any?).liftEq(Int.eq())
      ),

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/SemiringLaws.kt
```diff
@@ -2,42 +2,110 @@ package arrow.test.laws
...
-      Law("Semiring: Additive commutativity") { SG.semiringAdditiveCommutativity(A, B, EQ) },
-      Law("Semiring: Multiplicative commutativity") { SG.semiringMultiplicativeCommutativity(A, B, EQ) },
-      Law("Semiring: Right distributivity") { SG.semiringRightDistributivity(A, B, C, EQ) },
-      Law("Semiring: Left distributivity") { SG.semiringLeftDistributivity(A, B, C, EQ) },
-      Law("Semiring: Multiplicative left identity") { SG.semiringMultiplicativeLeftIdentity(A, C, EQ) },
-      Law("Semiring: Multiplicative right identity") { SG.semiringMultiplicativeRightIdentity(A, EQ) },
-      Law("Semiring: Multiplicative left absorption") { SG.semiringMultiplicativeLeftAbsorption(A, C, EQ) },
-      Law("Semiring: Multiplicative right absorption") { SG.semiringMultiplicativeRightAbsorption(A, C, EQ) }
-    )
+  fun <F> laws(SG: Semiring<F>, GEN: Gen<F>, EQ: Eq<F>): List<Law> =
```
danimontoya: Nice, lots of laws covered!

modules/core/arrow-core-data/src/test/kotlin/arrow/core/extensions/NumberSemiringTest.kt
aballano: Not exactly, it's covering a Schedule with Int specification, but not the laws for each number type as here

aballano:
Got several failures:

> arrow.core.MapKTest > Monoid Laws: combineAll should be derived FAILED
    java.lang.OutOfMemoryError: Java heap space

-> Adding a commit for this to reduce memory footprint

> arrow.fx.ScheduleTest > Semiring: empty is derived FAILED
    org.opentest4j.AssertionFailedError: expected: true but was: false

-> I believe this is caused due to Semiring overriding `empty()` as `zero()` instead of the other way around. This can be noticed in the Semiring instance for Schedule as it provides different impls for both functions.

The following are under investigation:
> arrow.fx.ScheduleTest > Semiring: Multiplicative commutativity FAILED
>     java.lang.AssertionError: Property failed for
>     Arg 0: arrow.fx.Schedule$ScheduleImpl@2953a40a
>     Arg 1: arrow.fx.Schedule$ScheduleImpl@6cb59b5c
>     after 2 attempts
>     Caused by: expected: true but was: false
> 
>         Caused by:
>         org.opentest4j.AssertionFailedError: expected: true but was: false
> 
> arrow.fx.ScheduleTest > Semiring: Left distributivity FAILED
>     java.lang.AssertionError: Property failed for
>     Arg 0: arrow.fx.Schedule$ScheduleImpl@1fdf0a99
>     Arg 1: arrow.fx.Schedule$ScheduleImpl@34779754
>     Arg 2: arrow.fx.Schedule$ScheduleImpl@b5dc622
>     after 4 attempts
>     Caused by: expected: true but was: false
> 
>         Caused by:
>         org.opentest4j.AssertionFailedError: expected: true but was: false
> 
> arrow.fx.ScheduleTest > Semiring: Multiplicative left identity FAILED
>     java.lang.AssertionError: Property failed for
>     Arg 0: arrow.fx.Schedule$ScheduleImpl@44236fe
>     after 1 attempts
>     Caused by: expected: true but was: false
> 
>         Caused by:
>         org.opentest4j.AssertionFailedError: expected: true but was: false
> 
> arrow.fx.ScheduleTest > Semiring: Multiplicative left absorption FAILED
>     java.lang.AssertionError: Property failed for
>     Arg 0: arrow.fx.Schedule$ScheduleImpl@5856fb39
>     after 1 attempts
>     Caused by: expected: true but was: false
> 
>         Caused by:
>         org.opentest4j.AssertionFailedError: expected: true but was: false
> 
> arrow.fx.ScheduleTest > Semiring: Multiplicative right absorption FAILED
>     java.lang.AssertionError: Property failed for
>     Arg 0: arrow.fx.Schedule$ScheduleImpl@65207730
>     after 1 attempts
>     Caused by: expected: true but was: false
> 
>         Caused by:
>         org.opentest4j.AssertionFailedError: expected: true but was: false
>

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/schedule.kt
```diff
@@ -43,45 +43,39 @@ interface ScheduleApplicative<F, Input> : Applicative<SchedulePartialOf<F, Input
...
-    Schedule.forever<F, Input>(MF()).const(MA().empty())
-
-  override fun zero(): Schedule<F, Input, Output> =
-    Schedule.never<F, Input>(MF()).const(MA().empty())
+  override fun one(): Schedule<F, Input, Output> =
+    Schedule.once<F, Input>(MF()).const(OI().one())
 
   override fun Schedule<F, Input, Output>.combineMultiplicate(b: Schedule<F, Input, Output>): Schedule<F, Input, Output> =
-    andThen(b).map { it.fold(::identity, ::identity) }
+    and(b).map { (a, b) -> OI().run { a * b } }
```
aballano: 👀

nomisRev: cc\ @1Jajen1

1Jajen1: That depends on what ring instance you want. Schedule has a lot of them ^^ I'll take a look

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/schedule.kt
```diff
@@ -43,45 +43,39 @@ interface ScheduleApplicative<F, Input> : Applicative<SchedulePartialOf<F, Input
   fun MF(): Monad<F>
 
   override fun <A> just(a: A): Kind<SchedulePartialOf<F, Input>, A> =
-    Schedule.forever<F, A>(MF()).const(a) as Schedule<F, Input, A>
+    Schedule.once<F, A>(MF()).const(a) as Schedule<F, Input, A>
```
1Jajen1: This should break applicative laws. Schedule has an applicative instance with `const . forever` as `just` and `and` as `ap`.
`and` combines two schedules by combining their decision with `and` for the decision to continue and `max` for the delay. This means that the neutral element produced by `just` must always continue and have the minimum delay, which is why it used forever.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/schedule.kt
```diff
@@ -43,45 +43,39 @@ interface ScheduleApplicative<F, Input> : Applicative<SchedulePartialOf<F, Input
...
 @extension
 interface ScheduleMonoid<F, Input, Output> : Monoid<Schedule<F, Input, Output>>, ScheduleSemigroup<F, Input, Output> {
-  override fun SA(): Semigroup<Output> = MA()
-  fun MA(): Monoid<Output>
+  override fun OI(): Monoid<Output>
   fun MF(): Monad<F>
 
   override fun empty(): Schedule<F, Input, Output> =
-    Schedule.forever<F, Input>(MF()).const(MA().empty())
+    Schedule.once<F, Input>(MF()).const(OI().empty())
```
1Jajen1: Same here as with the applicative one. This cannot be lawful as `once` is not an identity element. If our tests cannot uncover this, something is wrong. Likely with Schedules eqK or genK instances.

modules/core/arrow-core-data/src/test/kotlin/arrow/core/MapKTest.kt
```diff
@@ -36,51 +36,51 @@ import io.kotlintest.properties.forAll
...
-    val testLaws = testLaws(
-      ShowLaws.laws(MapK.show(String.show(), Int.show()), EQ_TC, Gen.mapK(Gen.string(), Gen.int())),
-      TraverseLaws.laws(MapK.traverse(), MapK.genK(Gen.string()), MapK.eqK(String.eq())),
-      MonoidLaws.laws(MapK.monoid<String, Int>(Int.semigroup()), Gen.mapK(Gen.string(), Gen.int()), EQ),
-      FoldableLaws.laws(MapK.foldable(), MapK.genK(Gen.string())),
-      EqLaws.laws(MapK.eq(String.eq(), Int.eq()), Gen.mapK(Gen.string(), Gen.int())),
-      FunctorFilterLaws.laws(MapK.functorFilter(), MapK.genK(Gen.string()), MapK.eqK(String.eq())),
-      HashLaws.laws(MapK.hash(String.hash(), Int.hash()), EQ_TC, Gen.mapK(Gen.string(), Gen.int())),
+    testLaws(
+      ShowLaws.laws(MapK.show(Long.show(), Int.show()), EQ_TC, Gen.mapK(Gen.long(), Gen.int())),
```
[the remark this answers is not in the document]
aballano: No strong reason, I was getting OutOfMemory errors in the MonoidLaws and though about using a more lightweight key for the map to do the sums 😅

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/schedule.kt
1Jajen1: The idea behind this ring instance was:
- addition is parallel composition such that every added schedule runs in parallel and is combined via `and + plus` (similar to the applicative instance which combines with `and + function application`)
- multiplication is sequential composition such that every added schedule runs after the other ones, which means combination via `andThen`

I think we can choose other instances, but since schedule has so many viable ones we should choose one which actually makes sense to use.

Here are a few monoid instances to mix and match from: ^^
- `forever().const(mempty)` + `and + plus`
- `never()` + `or + plus`
- `never()` + `andThen`
... and any other combination of the decision and the result. The decision is just a tuple of boolean and delay, so the full result that needs to be combined somehow is a triple. `and = and (on the boolean) + max (on the delay)` etc.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/io.kt
```diff
@@ -238,20 +238,17 @@ fun IO.Companion.concurrentEffect(dispatchers: Dispatchers<ForIO>): ConcurrentEf
 @extension
 interface IOSemigroup<A> : Semigroup<IO<A>> {
 
-  fun SG(): Semigroup<A>
+  fun AI(): Semigroup<A>
```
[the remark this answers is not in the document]
aballano: I opted to narrow the type on instances extending this one, so went for "A Instance" name if it makes any sense. I'm fine leaving it as SG, although it'll be a monoid somewhere else 🤷‍♂

modules/core/arrow-core-data/src/test/kotlin/arrow/core/extensions/NumberSemiringTest.kt
```diff
@@ -1,80 +1,23 @@
...
-            }
-        }
-    }
+  init {
+    testLaws(SemiringLaws.laws(Byte.semiring(), Gen.byteSmall(), Byte.eq()))
+    testLaws(SemiringLaws.laws(Double.semiring(), Gen.doubleSmall(), Double.eq()))
+    testLaws(SemiringLaws.laws(Int.semiring(), Gen.intSmall(), Int.eq()))
+    testLaws(SemiringLaws.laws(Short.semiring(), Gen.shortSmall(), Short.eq()))
+    testLaws(SemiringLaws.laws(Float.semiring(), Gen.floatSmall(), Float.eq()))
+    testLaws(SemiringLaws.laws(Long.semiring(), Gen.longSmall(), Long.eq()))
```
1Jajen1: Hmm can you double check if all these are run? Iirc kotlintest filters out tests with same names (or was that just on our side?)

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/schedule.kt
aballano: the problem I had with `andThen` is that, as you said, it runs one and then the other, therefore we would need to accumulate them before multiplying the outputs.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/schedule.kt
aballano: I see what you mean and makes sense. I believe the tests passed anyway so I'll try to check them out as well 👍

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/extensions/schedule.kt
1Jajen1: We should not need to multiply the outputs when using `andThen` (that'd be impossible anyway). I think this should also be lawful, but not entirely sure.

modules/core/arrow-core-data/src/test/kotlin/arrow/core/extensions/NumberSemiringTest.kt
aballano: Yep they are, because each `testLaws` call is doing the filtering

1Jajen1:
So after talking to @aballano for the `Schedule` instance, I played around with it a bit and it is indeed not lawful. I removed it as we don't really have an interesting `Semiring` instance for Schedule...

I also changed `Semiring` to not inherit from `Monoid` because that is ambiguos because a semiring is basically two monoids combined with a bunch of extra laws. I also added the law that the additive part has to be commutative. This ruled out the `Schedule` instance. I also removed the `Option` instance because it failed the laws as well.

modules/core/arrow-core-data/src/test/kotlin/arrow/core/extensions/NumberInstancesTest.kt
```diff
@@ -2,12 +2,40 @@ package arrow.core.extensions
...
 import io.kotlintest.properties.forAll
 
-class NumberMonoidTest : UnitSpec() {
+class NumberInstancesTest : UnitSpec() {
+
+  fun <F> testAllLaws(SG: Semiring<F>, M: Monoid<F>, HF: Hash<F>, GEN: Gen<F>, EQ: Eq<F>) {
+    testLaws(SemiringLaws.laws(SG, GEN, EQ))
+    testLaws(MonoidLaws.laws(M, GEN, EQ))
+    testLaws(HashLaws.laws(HF, GEN, EQ))
+  }
```
1Jajen1: :+1:

modules/core/arrow-test/src/main/kotlin/arrow/test/generators/Generators.kt
```diff
@@ -65,8 +68,27 @@ fun Gen.Companion.throwable(): Gen<Throwable> = Gen.from(listOf(RuntimeException
...
+}
+
 fun Gen.Companion.intSmall(): Gen<Int> = Gen.oneOf(Gen.choose(Int.MIN_VALUE / 10000, -1), Gen.choose(0, Int.MAX_VALUE / 10000))
 
+fun Gen.Companion.byteSmall(): Gen<Byte> = Gen.oneOf(Gen.choose(Byte.MIN_VALUE / 10, -1), Gen.choose(0, Byte.MAX_VALUE / 10)).map { it.toByte() }
+
+fun Gen.Companion.shortSmall(): Gen<Short> = Gen.oneOf(Gen.choose(Short.MIN_VALUE / 1000, -1), Gen.choose(0, Short.MAX_VALUE / 1000)).map { it.toShort() }
+
+fun Gen.Companion.longSmall(): Gen<Long> = Gen.oneOf(Gen.choose(Long.MIN_VALUE / 100000L, -1L), Gen.choose(0L, Long.MAX_VALUE / 100000L))
+
```
1Jajen1: God this is awful ^^ Anyway, nothing we can do on our side really

PR #2077, opened 2020-02-17
Increase test suite for AsyncLaws

danimontoya:
Covers small part of #2029 .

Increase test suite for AsyncLaws so we can prevent the user being able to break behaviour.

### DIFF
--- a/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/AsyncLaws.kt
+++ b/modules/core/arrow-test/src/main/kotlin/arrow/test/laws/AsyncLaws.kt
@@ -13,6 +13,7 @@ import arrow.test.generators.GenK
 import arrow.test.generators.applicativeError
 import arrow.test.generators.either
 import arrow.test.generators.functionAToB
+import arrow.test.generators.functionToA
 import arrow.test.generators.intSmall
 import arrow.test.generators.throwable
 import arrow.typeclasses.Apply
@@ -29,9 +30,10 @@ object AsyncLaws {
   private val one = newSingleThreadContext("1")
   private val two = newSingleThreadContext("2")
 
-  private fun <F> asyncLaws(AC: Async<F>, EQK: EqK<F>): List<Law> {
+  private fun <F> asyncLaws(AC: Async<F>, GENK: GenK<F>, EQK: EqK<F>): List<Law> {
     val EQ = EQK.liftEq(Int.eq())
     val EQB = EQK.liftEq(Boolean.eq())
+    val EQUnit: Eq<Kind<F, Unit>> = EQK.liftEq(Eq.any())
 
     return listOf(
       Law("Async Laws: success equivalence") { AC.asyncSuccess(EQ) },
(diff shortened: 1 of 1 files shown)


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/AsyncLaws.kt
```diff
@@ -163,6 +170,31 @@ object AsyncLaws {
...
+    }
+
+  fun <F> Async<F>.derivedContinueOn(EQK: Eq<Unit>) {
+    fx.async {
+      continueOn(one).equalUnderTheLaw(one.shift().bind(), EQK)
+    }
+  }
+
+  fun <F> Async<F>.derivedShift(EQK: Eq<Kind<F, Unit>>) {
+    one.shift().equalUnderTheLaw(one.run { effect(this) { Unit } }, EQK)
```
nomisRev: Can we also verify the actual context shift done by `shift` and `continueOn`?
You can check `IOTest` for some unit tests `Async` tests that are missing here.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/AsyncLaws.kt
```diff
@@ -163,6 +170,31 @@ object AsyncLaws {
...
+
+  fun <F> Async<F>.derivedLaterOrRaise(EQK: Eq<Kind<F, Int>>): Unit =
+    forAll(Gen.functionToA(Gen.either(Gen.throwable(), Gen.int()))) { f: () -> Either<Throwable, Int> ->
+      laterOrRaise(one, f).equalUnderTheLaw(defer(one) { f().fold({ raiseError<Int>(it) }, { just(it) }) }, EQK)
+    }
+
+  fun <F> Async<F>.derivedContinueOn(EQK: Eq<Unit>) {
+    fx.async {
+      continueOn(one).equalUnderTheLaw(one.shift().bind(), EQK)
+    }
```
aballano: here nothing is ever executed, you need to make sure you either assert something or call equalUnderTheLaw as the last statement:

```
fun <F> Async<F>.derivedContinueOn(EQK: Eq<Kind<F, Unit>>) {
  fx.async { continueOn(one) }.equalUnderTheLaw(fx.async { one.shift().bind() }, EQK)
}
```

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/AsyncLaws.kt
```diff
@@ -163,6 +169,40 @@ object AsyncLaws {
         p.equalUnderTheLaw(just(true), EQK)
     }
 
+  fun <F> Async<F>.derivedDefer(GK: GenK<F>, EQK: Eq<Kind<F, Int>>): Unit =
+    forAll(Gen.functionToA(GK.genK(Gen.int()))) { f: () -> Kind<F, Int> ->
```
nomisRev: I'd use `val f: () -> Kind<F, String> = effect { Thread.currentThread().name }` instead and use `Eq<Kind<F, String>>` this should verify that they behave the same, if the behavior is verified in one of the two.

PR #2081, opened 2020-02-18
Refactor ConcurrentQueue internals

nomisRev:
The current ConcurrentQueue implementation based on ZIO was based on `Ref` and using high-level `F` constructs which were quite slow for writing such a low level primitive like `Queue`.

As an experiment I re-wrote `Queue` while traveling, and ended up with this implementation based on `MVar`.

The benchmarks results are quite significant (Throughput, higher is better):

```
Benchmark              (size)   Mode  Cnt    Score    Error  Units
Queue.cancelableQueue    1000  thrpt   10  339.690 ± 23.860  ops/s
Queue.concurrentQueue    1000  thrpt   10   41.941 ±  5.508  ops/s
```

However this new implementation breaks 2 tests, due to the new restriction of `capacity > 0` for all `Queue`s except for `unbounded`. This removes the ability to create `Dropping` and `Bounded` Queues with capacity `0`. I haven't attempted to build in support for these 2 use-cases.
It however does add the ability to `read` or `peek` which is currently not possible in the Queue impl.

cc\ @richard-gibson @scott-christopher @kioba


modules/benchmarks/arrow-benchmarks-fx/build.gradle
```diff
@@ -23,24 +23,25 @@ apply plugin: "io.morethan.jmhreport"
 
 jmh {
     include = [
-            'arrow.benchmarks.Async',
-            'arrow.benchmarks.AttemptNonRaised',
-            'arrow.benchmarks.AttemptRaisedError',
+//            'arrow.benchmarks.Async',
```
raulraja: Should these be reenabled?

nomisRev: We're currently not using these atm in CI with `Hood` so this does not impact anything. I can revert it.

pakoito: I'd say remove, don't delete.

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
```diff
@@ -185,22 +184,11 @@ class QueueTest : UnitSpec() {
...
-          val start = !effect { System.currentTimeMillis() }
-          val wontComplete = q.offer(2)
-          val received = !wontComplete.map { Some(it) }
-            .waitFor(100.milliseconds, default = just(None))
-          val elapsed = !effect { System.currentTimeMillis() - start }
-          !effect { received shouldBe None }
-          !effect { (elapsed >= 100) shouldBe true }
-        }.unsafeRunSync()
+      "$label - capacity must be a positive integer" {
+        queue(0).attempt().suspended().fold(
```
pakoito: Can we used a refined tipe for positive integers?

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/internal/CancelableQueue.kt
```diff
@@ -0,0 +1,366 @@
...
+          else unsafeTryTake()
+        }
+      }
+      is State.Deficit -> just(None)
+      is State.Shutdown -> just(None)
+    }
+
+  private tailrec fun unsafeTake(onTake: (Either<Throwable, A>) -> Unit): Kind<F, CancelToken<F>> =
+    when (val current = state.value) {
+      is State.Surplus -> {
```
pakoito: jesus take the wheel

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/internal/CancelableQueue.kt
```diff
@@ -0,0 +1,366 @@
...
+      override fun unsafeTryHandleSurplus(state: AtomicRefW<State<F, A>>, a: A, surplus: State.Surplus<F, A>): Kind<F, Boolean>? {
+        val nextQueue = if (surplus.value.length() < capacity) surplus.value.enqueue(a) else surplus.value.dequeue().b.enqueue(a)
+        return if (!state.compareAndSet(surplus, surplus.copy(value = nextQueue))) null
+        else AP.just(true)
+      }
+    }
+
+    data class Dropping<F, A>(val capacity: Int, val AP: Applicative<F>) : SurplusStrategy<F, A>() {
+      override fun unsafeTryHandleSurplus(state: AtomicRefW<State<F, A>>, a: A, surplus: State.Surplus<F, A>): Kind<F, Boolean>? {
+        val nextQueue = if (surplus.value.length() < capacity) surplus.value.enqueue(a) else surplus.value
```
pakoito: NIT: brackets on ifs

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/internal/CancelableQueue.kt
```diff
@@ -0,0 +1,366 @@
...
+    this@map.map(f)
+  }
+
+  internal sealed class SurplusStrategy<F, A> {
+    abstract fun unsafeTryHandleSurplus(state: AtomicRefW<State<F, A>>, a: A, surplus: State.Surplus<F, A>): Kind<F, Boolean>?
+
+    data class Bounded<F, A>(val capacity: Int, val AP: Applicative<F>) : SurplusStrategy<F, A>() {
+      override fun unsafeTryHandleSurplus(state: AtomicRefW<State<F, A>>, a: A, surplus: State.Surplus<F, A>): Kind<F, Boolean>? = when {
+        surplus.value.length() >= capacity -> AP.just(false)
+        state.compareAndSet(surplus, surplus.copy(surplus.value.enqueue(a))) -> AP.just(true)
```
pakoito: side-effect in the when branch? cheeky lol

nomisRev: That's the price to pay for writing low level concurrency primitives, about as dirty as it gets 😁 `tailrec` over `when` and `atomic#compareAndSet`

nomisRev: EDIT: I was looking for a nicer signature here, but I haven't spend much time making the code look nicer.

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/internal/CancelableQueue.kt
nomisRev: Is this something we can configure in KtLint?
I'm used to using expressions for `if/else`, and probably in more places than I should...

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
nomisRev: Not without having a proper refinement system? Or did you have something in mind?

pakoito: Some Q&D wrapper without getting too much into refined types.

```
class Size private (val s: Int) {
 co {
  of invoke(s: Int): Option<Size>
 }
}
```

modules/fx/arrow-fx/src/main/kotlin/arrow/fx/internal/CancelableQueue.kt
pakoito: My point is that this will happen even if the value doesn't pass out. Is that okay?

modules/fx/arrow-fx/src/test/kotlin/arrow/fx/QueueTest.kt
nomisRev: Wouldn't take be as much of a hassle as what the current API have without compile time literal verification?

We could provide alternative factory methods which return `IO<Option<Queue<ForIO, A>>>`?

I'm opening another PR asap which adds the `peek` and `try` methods to `Queue` and removes the other implementation. I will also add a `QueueFactory` like we have for `MVar` and `Ref`. We can continue this discussion there.

Issue #2073, opened 2020-02-17
Fx: enable "MVarTest - concurrent take and put"


rachelcarmena:
It was disabled with #2072 because it was failing with several pushes on `master` branch and blocking the artifacts publication on OSS.

PR #2086, opened 2020-02-19
Re-enabled MVar test

danimontoya:
Re-enabled MVar test _concurrent take and put_

This test was timing out because the number of iterations. Since it's not a stackoverflow test, 5K iterations for a stress test seems sufficient.

### DIFF
--- a/modules/fx/arrow-fx/src/test/kotlin/arrow/fx/MVarTest.kt
+++ b/modules/fx/arrow-fx/src/test/kotlin/arrow/fx/MVarTest.kt
@@ -276,20 +276,20 @@ class MVarTest : UnitSpec() {
         }.equalUnderTheLaw(IO.unit, EQ())
       }
 
-      // "$label - concurrent take and put" {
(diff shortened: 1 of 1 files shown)


modules/fx/arrow-fx/src/test/kotlin/arrow/fx/MVarTest.kt
```diff
@@ -276,20 +276,20 @@ class MVarTest : UnitSpec() {
...
-      //     val puts = (0 until count).map { mvar.put(1) }.parSequence()
-      //     val f1 = !takes.fork()
-      //     val f2 = !puts.fork()
-      //     !f1.join()
-      //     !f2.join()
-      //     !ref.get()
-      //   }.equalUnderTheLaw(IO.just(count), EQ())
-      // }
+      "$label - concurrent take and put" {
+        val count = 5000
```
aballano: 👍

aballano:
Closes #2073

PR #2087, opened 2020-02-19
Increase test suite for BracketLaws

aballano:
Covers small part of #2029


modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
```diff
@@ -29,15 +34,17 @@ object BracketLaws {
...
       Law("Bracket: bracket propagates transformer effects") { BF.bracketPropagatesTransformerEffects(EQ) },
       Law("Bracket: bracket must run release task on use error") { BF.bracketMustRunReleaseTaskOnUseError(EQ) },
       Law("Bracket: bracket must not run release task on acquire error") { BF.bracketMustNotRunReleaseTaskOnAcquireError(EQ) },
-      Law("Bracket: guarantee must run finalizer task") { BF.guaranteeMustRunFinalizerOnError(EQ) }
+      Law("Bracket: guaranteeCase must run finalizer task") { BF.guaranteeCaseMustRunFinalizerOnError(EQ) },
+      Law("Bracket: bracket is derived from bracketCase") { BF.bracketIsDerivedFromBracketCase(EQ) },
+      Law("Bracket: guarantee is derived from bracket") { BF.guaranteeIsDerivedFromBracket(BF.just(Unit), EQ) },
+      Law("Bracket: guaranteeCase is derived from bracketCase") { BF.guaranteeCaseIsDerivedFromBracketCase({ BF.just(Unit) }, EQ) },
+      Law("Bracket: onCancel is derived from guaranteeCase") { BF.onCancelIsDerivedFromGuaranteeCase(GENK, EQ) },
+      Law("Bracket: onError is derived from guaranteeCase") { BF.onErrorIsDerivedFromGuaranteeCase(GENK, EQ) }
```
aballano: This 2 are the new added tests

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
```diff
@@ -154,15 +137,64 @@ object BracketLaws {
...
+
+  fun <F> Bracket<F, Throwable>.guaranteeCaseIsDerivedFromBracketCase(
+    finalizer: (ExitCase<Throwable>) -> Kind<F, Unit>,
+    EQ: Eq<Kind<F, Int>>
+  ): Unit =
+    forAll(Gen.int().applicativeError(this)) { fa: Kind<F, Int> ->
+      fa.guaranteeCase(finalizer).equalUnderTheLaw(just(Unit).bracketCase({ _, e -> finalizer(e) }) { fa }, EQ)
+    }
+
+  fun <F> Bracket<F, Throwable>.onCancelIsDerivedFromGuaranteeCase(
```
nomisRev: This can never test anything since `cancel` is never triggered.

modules/core/arrow-test/src/main/kotlin/arrow/test/laws/BracketLaws.kt
```diff
@@ -154,15 +137,64 @@ object BracketLaws {
...
+          else -> unit()
+        }
+      }, EQ)
+    }
+
+  fun <F> Bracket<F, Throwable>.onErrorIsDerivedFromGuaranteeCase(
+    GENK: GenK<F>,
+    EQ: Eq<Kind<F, Int>>
+  ): Unit =
+    forAll(Gen.throwable().raiseError<F, Int, Throwable>(this), Gen.functionAToB<Throwable, Kind<F, Unit>>(GENK.genK(Gen.unit()))) { fe: Kind<F, Int>, ff: (Throwable) -> Kind<F, Unit> ->
```
nomisRev: You're always comparing the result of `{ _ -> unit() }` with each-other.
I think here you'd want a test similar to `guaranteeCaseMustRunFinalizerOnError` but for `onError` rather then checking if `onError` was dereived from `guaranteeCaseMustRunFinalizerOnError`

aballano: not in this case, because `Gen.throwable().raiseError` will always emit an error, therefore the function is called. If i replace the `ff(it.e)` in the error case for something else the test will actually fail

rachelcarmena:

PR #2079, opened 2020-02-18
Change docs urls

AntonioMateoGomez:
Modifies urls and make them relatives (relative_url). Specifically it removes the `docs/` part from urls since from now it´ll be part from the base url along the version.

Also modifies the `doc-version.yml` file to keep the same url structure when build the sidebar doc version selector. Once we are ready to split the arrow modules, **this file should be moved to the root path** and copied trhough the jekyll build into its `_data` folder in every version folder in order to maintain just a single version file.

Closes #2069

PR #2091, opened 2020-02-20
Doc: fix versions menu

rachelcarmena:
PR on branch for #2079


modules/docs/arrow-docs/docs/_data/commons.yml
```diff
@@ -6,3 +6,4 @@ google_analytics: "UA-18433785-18"
 github_repository: "https://github.com/arrow-kt/arrow"
 stable_version: "https://arrow-kt.io/"
 next_version: "https://arrow-kt.io/docs/next/"
+previous_version: "https://arrow-kt.io/docs/0.9/"
```
AntonioMateoGomez: This line is not needed, `previous_version` is declared in the `_head-docs.html` file, and it takes it from the previous title of the `stable_version` from `doc-versions.yml` file, so we can remove this line.
Doing this way when we do the changes to a new version we only need to change the `doc-versions.yml` file

Captura de pantalla 2020-02-20 a las 10 38 17

modules/docs/arrow-docs/docs/_includes/_sidebar-doc-versions.html
```diff
@@ -32,10 +32,9 @@
       </li>
 
       <li>
-        <a href="{{ previous_version.url | append: 'docs/' }}">
+        <a href="{{ previous_version.url | append: doc-link }}">
```
AntonioMateoGomez: We should remove the append here since the url that is going to build would be `https://arrow-kt.io/docs/0.9/core/` in the case of core, and that url doesn´t exist in that version.

modules/docs/arrow-docs/docs/_includes/_sidebar-doc-versions.html
```diff
@@ -45,7 +44,7 @@
             and item.url != next_version.url
             and item.url != previous_version.url %}
       <li>
-        <a href="{{ item.url | append: 'docs/' }}">
+        <a href="{{ item.url | append: doc-link }}">
```
AntonioMateoGomez: The same here for even older version. We should remove the append here.

PR #2103, opened 2020-03-05
Add sync with release/0.11.0 branch

rachelcarmena:
To avoid similar cases like https://github.com/arrow-kt/arrow-ank/issues/17


rachelcarmena:
Thanks for the review @aballano !!! :raised_hands:

PR #2107, opened 2020-03-11
Add kotlin binary compatibility version

danimontoya:
Adding version of the binary compatibility tool.
> The tool allows to dump binary API of a Kotlin library that is public in sense of Kotlin visibilities and ensures that the public binary API wasn't changed in a way that make this change binary incompatible.


gradle.properties
```diff
@@ -29,6 +29,7 @@ KOTLIN_TEST_VERSION=3.3.3
 KOTLIN_VERSION=1.3.50
 KOTLINX_COLLECTIONS_IMMUTABLE_VERSION=0.1
 KOTLINX_COROUTINES_VERSION=1.3.3
+KOTLINX_VALIDATOR_VERSION = "0.2.2"
```
rachelcarmena:
```suggestion
KOTLINX_VALIDATOR_VERSION=0.2.2
```

danimontoya: Oops! thanks @rachelcarmena .. I just copy paste it

rachelcarmena: No worries!! :raised_hands:

PR #2108, opened 2020-03-16
Bump version of KOTLIN_VERSION

danimontoya:
Bump version of KOTLIN_VERSION=1.3.70


gradle.properties
```diff
@@ -25,8 +25,8 @@ JUNIT_VINTAGE_VERSION=5.5.2
 KINDEDJ_VERSION=1.1.0
 KOTLIN_METADATA_VERSION=1.4.0
 KOTLIN_POET_VERSION=1.1.0
-KOTLIN_TEST_VERSION=3.3.3
-KOTLIN_VERSION=1.3.50
+KOTLIN_TEST_VERSION=3.4.2
```
rachelcarmena: Why are we changing this version?

danimontoya: KoTest is still in beta version.. so thats the latest version we can bump it up. 
Also I thought that version would be using kotlin 1.3.70.. but there are some other dependencies that uses still 1.3.50.. so, as we discussed in slack, we can exclude those dependencies we don't want.. and add them with the version we want separately.

Issue #2129, opened 2020-04-11
["Request"] Consider creating a BOM


ps-feng:
### What version are you currently using?
0.10.5

### What would you like to see?
I wonder if you've considered providing a BOM so that Arrow's version numbers only need to appear once in the dependencies. Usage in Gradle is then simplified a bit:
```kotlin
dependencies {
    implementation(platform("io.arrow-kt:arrow-stack:$arrowVersion"))

    // versions no longer required here
    implementation("io.arrow-kt:arrow-core")
    implementation("io.arrow-kt:arrow-fx")
    implementation("io.arrow-kt:arrow-mtl")
    implementation("io.arrow-kt:arrow-syntax")
}
```
It also opens the possibility for diverging versions in modules without increasing the dependency burden on the user's side, even though I guess that's generally not desirable.

pakoito:
@rachelcarmena I believe we had this at some point, I don't know where it ended up with the repo split.

rachelcarmena:
Thank you so much @ps-feng :raised_hands: 

> @rachelcarmena I believe we had this at some point, I don't know where it ended up with the repo split.

@pakoito , please, could you provide me more details? I didn't find anything about it:

* Just JAR artifacts with the correspondent POM file in [OSS](https://oss.jfrog.org/artifactory/oss-snapshot-local/io/arrow-kt/). 
* Nothing about publishing a BOM file in `arrow` repository for old versions.

Don't worry if nothing exists. Let's create it!

rachelcarmena:
It will be solved via #2131

pakoito:
Maybe it was `arrow-shadow` or `arrow-bom`, I don't remember.

rachelcarmena:
> Maybe it was `arrow-shadow` or `arrow-bom`, I don't remember.

I didn't find any BOM file in OSS repository. `arrow-bom` doesn't exist. `arrow-shadow` is a shadowed JAR which is not published since 0.9x.

Don't worry! `arrow-stack` already exists! :raised_hands: https://oss.jfrog.org/artifactory/oss-snapshot-local/io/arrow-kt/arrow-stack/ and it will be published in Maven Central for the next release.

I created an example of use: https://github.com/arrow-kt/arrow-examples/pull/4

Thanks again @ps-feng! Closing the issue. Please, reopen it in case of a missing feature :raised_hands:

PR #2131, opened 2020-04-13
Publish BOM file

rachelcarmena:
It solves #2129 

I'll update `arrow-examples` with the use of the BOM file then.


rachelcarmena:
Thanks for the review @pakoito :raised_hands:

PR #2155, opened 2020-06-09
Remove duplicated task

rachelcarmena:
`check` is already included when running `build`

### DIFF
--- a/.github/workflows/check_doc_integration.yml
+++ b/.github/workflows/check_doc_integration.yml
@@ -12,7 +12,7 @@ jobs:
 
     env:
       BASEDIR: ${{github.workspace}}/..
-      JAVA_OPTS: -Xms512m -Xmx1024m
+      JAVA_OPTS: -Xms3g
 
     runs-on: ubuntu-16.04
 
--- a/scripts/project-build.sh
+++ b/scripts/project-build.sh
@@ -3,4 +3,3 @@
 set -e
 cd $BASEDIR/$1
 ./gradlew clean build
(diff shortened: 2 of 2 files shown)


.github/workflows/check_doc_integration.yml
```diff
@@ -12,7 +12,7 @@ jobs:
 
     env:
       BASEDIR: ${{github.workspace}}/..
-      JAVA_OPTS: -Xms512m -Xmx1024m
+      JAVA_OPTS: -Xms3g
```
JorgeCastilloPrz, line 15: How much can we afford on CI?

rachelcarmena: Of course! It's a temporary change to fix the current error. It must be improved for all the libraries because JVM args are being set in `gradle.properties` and as an environment variable at the same with different values. I removed the max heap size because it's set in `gradle.properties` and I increased the initial memory according to https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources. However, my final idea is to remove the values from `gradle.properties` because they depend on the environment.

rachelcarmena: I changed my mind @JorgeCastilloPrz :wink: They are necessary in `gradle.properties`  because those properties control the VM running the build. It defaults to -Xmx512m "-XX:MaxMetaspaceSize=256m". JAVA_OPTS just controls the command line client, which is only used to display console output. It defaults to -Xmx64m.

PR #2170, opened 2020-07-06
Add previous global checks

rachelcarmena:
It will allow to add optional global checks execution during pull requests.

If several pull requests use the same branch name, they will be considered together during those checks execution.


rachelcarmena:
Merging! (the same current `arrow-core` master branch errors are happening)

PR #2180, opened 2020-07-15
Bump kotest version

aballano:


gradle.properties
```diff
@@ -25,7 +25,7 @@ JUNIT_VINTAGE_VERSION=5.5.2
 KINDEDJ_VERSION=1.1.0
 KOTEST_VERSION=4.1.0
 KOTLIN_METADATA_VERSION=1.4.0
-KOTLIN_TEST_VERSION=3.3.3
+KOTLIN_TEST_VERSION=4.1.1
```
rachelcarmena, line 28: Thanks @aballano !! `KOTEST_VERSION` already exists because it's being used for `arrow-fx-coroutines` so far.

aballano: Thanks! I'll change it 🤦

PR #2182, opened 2020-07-20
Add automodule version

pablisco:
For https://github.com/arrow-kt/arrow-incubator/pull/122


pablisco:
I've moved the automodule set up to [`automodule-conf.gradle `](https://github.com/arrow-kt/arrow/pull/2182/commits/bbd28e9ad4fdfcda4161a3fe272c2ff5f4685212) 
So, now we can import it on [`arrow-incubator`](https://github.com/arrow-kt/arrow-incubator/pull/122/commits/1af76430793a0455ff7ea6cac324157395f1f158) which will fail until we merge this one to master :)

rachelcarmena:
> I've moved the automodule set up to [`automodule-conf.gradle `](https://github.com/arrow-kt/arrow/pull/2182/commits/bbd28e9ad4fdfcda4161a3fe272c2ff5f4685212)
> So, now we can import it on [`arrow-incubator`](https://github.com/arrow-kt/arrow-incubator/pull/122/commits/1af76430793a0455ff7ea6cac324157395f1f158) which will fail until we merge this one to master :)

Good idea @pablisco :raised_hands: It should work without merging this change into master branch. I created #2184 to fix it.

I think we could use that file to add more things for `settings.gradle` so maybe the filename could be related to `settings` instead of `automodule`. What do you think about it? Thanks again!!

pablisco:
I was thinking about that actually. I'll change the name. We can also apply the setup-conf.gradle from settings via `gradle.rootProject` if we want to remove more set up steps :) what do you think @rachelcarmena ?

rachelcarmena:
> I was thinking about that actually. I'll change the name. We can also apply the setup-conf.gradle from settings via `gradle.rootProject` if we want to remove more set up steps :) what do you think @rachelcarmena ?

Good idea!! Thanks @pablisco !!

PR #2191, opened 2020-08-10
Documentation: How to fix documentation for latest release

rachelcarmena:
Please @aballano , @nomisRev , let me know if the steps to fix the documentation for the latest release version is clear enough. Thanks!!


rachelcarmena:
> I'm a bit confused by the last two steps. What if my changes shouldn't land in master?
> 
> I.e. I want to clarify something on the deprecation page of `Option` which won't be there anymore when we're preparing the 1.0 release.
> 
> So should we then always manuaally publish the site as mentioned in step 8? Or do we also hacve to do that regardless if we need the change in master?

Of course! Just to create the pull request if changes should exist for the next version as well. I'll clarify it!

> > if branch is changed by a non master branch, it won't work
> 
> thinking Can you clarify with an example?

That's for the "publish site" action because GitHub allows selecting the branch by default:

Screenshot from 2020-08-11 14-07-26

However, it shouldn't be executed from a different branch. So if someone choose a different branch, it will be skipped. Maybe it's better "it will be skipped" rather than "it won't work".

Thanks for the feedback!!

rachelcarmena:
Done @nomisRev !! I think I clarified the text.

rachelcarmena:
Thanks @aballano , @nomisRev !!!!

PR #2200, opened 2020-08-17
Unify logo

rachelcarmena:


README.md
```diff
@@ -1,4 +1,4 @@
-<a href="https://arrow-kt.io" title="Arrow website"><img src="https://github.com/arrow-kt/arrow-site/blob/master/docs/img/home/header-image.png" width="200" alt=""></a>
```
calvellido, line 1: Not sure if this continue to be the case, but I think this way of writing the full URL for the logo was because this README could be used outside of GitHub/GitHub pages (on the Arrow site/docs), so in that case the new relative URL might fail, be careful with that :roll_eyes:

rachelcarmena: Thank you so much for the warning @calvellido !! I think that's QuickStart section: https://github.com/arrow-kt/arrow-core/tree/master/arrow-docs/docs/quickstart which is planned to extract to `arrow-site` repository :raised_hands:

rachelcarmena:
Thanks for the review @calvellido :raised_hands:

PR #2206, opened 2020-08-25
Configuration: Remove maven plugin in favor of maven-publish

rachelcarmena:
# Context

* `maven` plugin has been deprecated
* `uploadArchives` from `maven` plugin  doesn't allow using the new configurations for dependencies. For instance, the use of `api` doesn't include the transitive dependency in the POM file as expected into OSS.
* `publishToMavenLocal` from `maven-publish` allows the new configurations.
* `publishToMavenLocal` is being used for integration tests
* Android Gradle plugin includes support for `maven-publish` plugin
* Gradle Build Scans recommends this change:
> The maven plugin has been deprecated. 
> This is scheduled to be removed in Gradle 7.0.
> Please use the maven-publish plugin instead.

# Reason of this PR

* Unify the publication into local repository and external artifact repositories

# Changes

* Replace `uploadArchives` from `maven` plugin by `publish` from `maven-publish` plugin
* Related configuration


nomisRev (approved):
Thank you so much @rachelcarmena for looking into this 👏 👏 👏

rachelcarmena:
Thanks for the review @nomisRev :raised_hands:

PR #2209, opened 2020-08-26
Fix archive classifier deprecation

tschuchortdev:
Something I came across: `classifier` seems to have been deprecated since Gradle 5.0 and has been replaced with `archiveClassifier` which ironically also produces a warning because the member is private (unless you use `set` o_O). 

See https://github.com/gradle/gradle/issues/8257

@rachelcarmena

### DIFF
--- a/gradle/android-module.gradle
+++ b/gradle/android-module.gradle
@@ -93,11 +93,13 @@ task androidJavadocs(type: Javadoc) {
 }
 
 task javadocJar(type: Jar, dependsOn: androidJavadocs) {
-    classifier = 'javadoc'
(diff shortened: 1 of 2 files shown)


rachelcarmena (commented):
Super useful, thank you so much!! I didn't realize about it: [classifier](https://docs.gradle.org/current/dsl/org.gradle.api.tasks.bundling.Jar.html#org.gradle.api.tasks.bundling.Jar:classifier). And it seems it's not a `String` but `Property<String>`: [archives](https://docs.gradle.org/current/userguide/working_with_files.html#sec:archives). However, the examples show the same assignment. Thanks for the comment!!

PR #2225, opened 2020-10-02
Kotlin metadata version wasn't correct

pablisco:
The version was updated to 1.4.10, but there is only 1.4.0 on maven central:
https://mvnrepository.com/artifact/me.eugeniomarletti.kotlin.metadata/kotlin-metadata


pablisco:
Ignore this, was using an outdated version of core :)

PR #2251, opened 2021-02-24
Remove Deprecated code from Arrow Fx Coroutines

nomisRev:


arrow-libs/fx/arrow-fx-coroutines/src/main/kotlin/arrow/fx/coroutines/CircuitBreaker.kt
```diff
@@ -388,26 +347,13 @@ private constructor(
...
-      val resetTimeout: FxDuration = resetTimeoutNanos.toLong().oldNanoseconds
-
       /** The timestamp in milliseconds since the epoch, specifying
        * when the `Open` state is to transition to [HalfOpen].
        *
        * It is calculated as:
        * `startedAt + resetTimeout`
        */
-      val expiresAt: Long = startedAt + resetTimeout.millis
+      val expiresAt: Long = startedAt + (resetTimeoutNanos.toLong() / 1_000_000)
```
i-walker, line 356: can we use `toLongMilliseconds()` ?

arrow-libs/fx/arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/RaceNTest.kt
```diff
@@ -211,29 +215,31 @@ class RaceNTest : ArrowFxSpec(
 
     "Cancelling race 3 cancels all participants" {
       checkAll(Arb.int(), Arb.int(), Arb.int()) { a, b, c ->
-        val s = Semaphore(0L)
```
i-walker, line 214: Does ths mean we will deprecate Semaphore for Channel?

arrow-libs/fx/arrow-fx-stm/src/main/kotlin/arrow/fx/stm/STM.kt
```diff
@@ -119,26 +118,24 @@ import arrow.fx.stm.internal.lookupHamtWithHash
...
- *       atomically { acc1.write(100_000_000) }
- *     }
- *     println("Performing transaction")
- *     atomically {
- *       println("Trying to transfer")
- *       transfer(acc1, acc2, 50)
- *     }
- *     println("Balance account 1: ${acc1.unsafeRead()}")
- *     println("Balance account 2: ${acc2.unsafeRead()}")
+ * fun main(): Unit = runBlocking {
```
i-walker, line 121: Does it make a difference here to use the suspended main function?

arrow-libs/fx/arrow-fx-stm/src/main/kotlin/arrow/fx/stm/STM.kt
```diff
@@ -168,18 +165,16 @@ import arrow.fx.stm.internal.lookupHamtWithHash
...
- *     println("Value is ${v.unsafeRead()}")
- *     atomically { transaction(v) }
- *       .also { println("Transaction returned $it") }
- *     println("Set value to 5")
- *     println("Value is ${v.unsafeRead()}")
- *     atomically { v.write(5) }
- *     atomically { transaction(v) }
- *       .also { println("Transaction returned $it") }
- *   }
+ * fun main(): Unit = runBlocking {
```
i-walker, line 168: Same here

arrow-libs/fx/arrow-fx-stm/src/test/kotlin/arrow/fx/stm/STMTest.kt
```diff
@@ -103,7 +103,7 @@ class STMTest : ArrowFxSpec(
       val v1 = TVar.new(0)
       val v2 = TVar.new(0)
       val v3 = TVar.new(0)
-      val f = ForkConnected {
+      val f = async {
```
i-walker, line 106: A more theoretical question why we can substitute `ForkConnected` with `async`?

i-walker (commented):
🙌 Really was looking for this PR 💚

i-walker (approved):

arrow-libs/fx/arrow-fx-coroutines/src/test/kotlin/arrow/fx/coroutines/RaceNTest.kt
nomisRev: No, we deprecate `Semaphore` for `Semaphore` and `ConcurrentVar` for `Channel`.

However, `Channel` of type `RENDEZ_VOUS` (or channel size = 0) can work in a similar way as a `Semaphore` with `0` permits. Whilst a `Semaphore` with `0` permits is not supported by the KotlinX Semaphore.

So this is use-case specific but since is quite low-level concurrency code to test other concurrent operators we have a bit of a special case here.

arrow-libs/fx/arrow-fx-stm/src/main/kotlin/arrow/fx/stm/STM.kt
nomisRev: Yes, `runBlocking` spins up an `EventLoop` `CoroutineContext` which allows you to go back to the original `CoroutineContext`.

For example:

```
suspend fun printThread() : Unit = println(Thread.currentThread().name)

suspend fun main() {
    printThread() //main
    withContext(Dispatchers.IO) {
        printThread() // DefaultDispatcher-worker-1
    }
    printThread() // DefaultDispatcher-worker-1
}
```

VS

```
suspend fun main() = runBlocking<Unit> {
    printThread() //main
    withContext(Dispatchers.IO) {
        printThread() // DefaultDispatcher-worker-1
    }
    printThread() // main
}
```

arrow-libs/fx/arrow-fx-stm/src/test/kotlin/arrow/fx/stm/STMTest.kt
nomisRev: Both launch a coroutine on a  `CoroutineContext`, here the one inherited from the parent or `Dispatchers.Default` by default and expose the result with a `await/cancel` contract.

So they model the exact same use-case. Which is also the reason why `Fiber` and `ForkXXX` is deprecated in favor of `kotlinx.coroutines.async`.

arrow-libs/fx/arrow-fx-stm/src/main/kotlin/arrow/fx/stm/STM.kt
i-walker: Thank you

arrow-libs/fx/arrow-fx-stm/src/test/kotlin/arrow/fx/stm/STMTest.kt
i-walker: Thanks 🙌

PR #2253, opened 2021-02-25
CU-ev1jrd Fix Monoid for Sequence and Set

nomisRev:
The `Monoid` instance for `Sequence` and `Set` were resulting in `StackOverflow` since they were delegating to their `Monoid#plus` function instead of the `plus` function from the standard library.


franciscodr:
Task linked: [CU-ev1jrd Fix Set/Sequence Monoid instance](https://app.clickup.com/t/ev1jrd)

danimontoya (approved):

aballano (approved):

i-walker (approved):

PR #2254, opened 2021-02-25
CU-et6b7a Deprecate ConcurrentVar

nomisRev:
`ConcurrentVar` has the same semantics as `Channel(1)`, which is a `Channel` with the size for a single element. So it can also be `empty` or `filled` and it provides suspending semantics upon `send` and `receive`, it covers more use-cases as `ConcurrentVar` does.


franciscodr:
Task linked: [CU-et6b7a Deprecate `ConcurrentVar` in favor of Channel(1)](https://app.clickup.com/t/et6b7a)

rachelcarmena (approved):
:tada:

danimontoya (approved):

i-walker (approved):

PR #2260, opened 2021-02-25
CU-ev2g8f Fix Sequence#travserse, Option#apEval & Either#apEval

nomisRev:
This PR fixes two issues for traversing an `Sequence`.

- Sequence#traverse was implemented using `foldRight` without `Eval` which first materialises the whole list in memory. This _always_ results in an `OOM` even if the transformation of `traverse` results in a _short-circuit_.
- 
- Fix `apEval` for `Either` and `Option` to properly _short-circuit_.

### DIFF
--- a/arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
@@ -1588,7 +1588,7 @@ fun <A, B, C> EitherOf<A, B>.ap(ff: EitherOf<A, (B) -> C>): Either<A, C> =
   flatMap { a -> ff.fix().map { f -> f(a) } }
 
 fun <A, B, C> Either<A, B>.apEval(ff: Eval<Either<A, (B) -> C>>): Eval<Either<A, C>> =
-  ff.map { this.ap(it) }
+  fold({ l -> Eval.now(l.left()) }, { r -> ff.map { it.map { f -> f(r) } } })
 
 fun <A, B> EitherOf<A, B>.combineK(y: EitherOf<A, B>): Either<A, B> =
   when (this) {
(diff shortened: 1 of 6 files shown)


franciscodr:
Task linked: [CU-ev2g8f Fix Sequence#traverse](https://app.clickup.com/t/ev2g8f)

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -1588,7 +1588,7 @@ fun <A, B, C> EitherOf<A, B>.ap(ff: EitherOf<A, (B) -> C>): Either<A, C> =
   flatMap { a -> ff.fix().map { f -> f(a) } }
 
 fun <A, B, C> Either<A, B>.apEval(ff: Eval<Either<A, (B) -> C>>): Eval<Either<A, C>> =
-  ff.map { this.ap(it) }
+  fold({ l -> Eval.now(l.left()) }, { r -> ff.map { it.map { f -> f(r) } } })
```
aballano, line 1591: is there or should be a specific test for this change?

aballano (approved):

i-walker (approved):
Thank you

raulraja (approved):
thanks @nomisRev

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
nomisRev: We should add many tests.. 😭😅 This code was taken from the `EitherApply.kt` code, and is verified in `TraverseTest`. By reverting this change it makes `TraverseTest` fail.

aballano: > By reverting this change it makes TraverseTest fail.

That's what I meant, perfect then! 👌

PR #2267, opened 2021-03-01
Remove redundant SuspendConnection

nomisRev:
`SuspendConnection` is no longer used since KotlinX Coroutines now drives the cancellation system of Arrow Fx Coroutines.


aballano (approved):

rachelcarmena (approved):
:fire:

PR #2270, opened 2021-03-01
Remove deprecate typeclasses

nomisRev:
This PR removes all deprecated typeclasses, and removes the usage from the code.

The only update/refactor is `Eq<A>` parameters in Laws to `(A, A) -> Boolean`

TODO:
 - [x] Fix docs


Daniel-Jacob (approved):

raulraja (approved):
Thanks @nomisRev !

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Hashed.kt
```diff
@@ -35,9 +33,4 @@ inline fun <A> HashedOf<A>.fix(): Hashed<A> =
  *
  * Provides a fast inequality check with its [Eq] instance and its [Hash] instance will use the cached hash.
  */
-@Deprecated(HashDeprecation)
-data class Hashed<A>(val hash: Int, val value: A) : HashedOf<A> {
-  companion object {
-    fun <A> A.fromHash(HA: Hash<A>): Hashed<A> = Hashed(HA.run { this@fromHash.hash() }, this)
-  }
-}
+data class Hashed<A>(val hash: Int, val value: A) : HashedOf<A>
```
aballano, line 36: is this now undeprecated?

arrow-libs/core/arrow-core-test/src/main/kotlin/arrow/core/test/laws/FxLaws.kt
```diff
@@ -23,38 +22,38 @@ object FxLaws {
   fun <Eff : Effect<*>, F, A> suspended(
     pureGen: Gen<F>,
     G: Gen<F>,
-    EQ: Eq<F>,
+    f: (F, F) -> Boolean,
```
aballano, line 25: I'm wondering, wouldn't be easier replacing Eq as a typeclass for an internal typealias? That way you wouln't need to replace this everywhere. 

nit: Also we could name it the same way as before to indicate the behaviour hasn't change, currently is mixed, some places is `f`, some others is `eq` or `EQ`

arrow-libs/core/arrow-core-test/src/main/kotlin/arrow/core/test/laws/MonoidLaws.kt
```diff
@@ -1,37 +1,36 @@
...
 import arrow.typeclasses.Monoid
 import io.kotlintest.properties.Gen
 import io.kotlintest.properties.forAll
 import io.kotlintest.shouldBe
 
 object MonoidLaws {
 
-  fun <F> laws(M: Monoid<F>, GEN: Gen<F>, EQ: Eq<F>): List<Law> =
-    SemigroupLaws.laws(M, GEN, EQ) +
+  fun <F> laws(M: Monoid<F>, GEN: Gen<F>, f: (F, F) -> Boolean = { a, b -> a?.equals(b) == true }): List<Law> =
```
aballano, line 10: this default value could be a `Eq.default()` that can be used in multiple places if you go for the typealias

arrow-libs/optics/arrow-optics-test/src/main/kotlin/arrow/optics/test/laws/OptionalLaws.kt
```diff
@@ -50,40 +49,41 @@ object OptionalLaws {
...
     aGen: Gen<A>,
-    bGen: Gen<B>
+    bGen: Gen<B>,
+    EQB: (B?, B?) -> Boolean
   ): Unit =
     forAll(optionalGen, aGen, bGen) { optional, a, b ->
       optional.run {
         getOrNull(set(a, b))
-          .equalUnderTheLaw(getOrNull(a)?.let { b }, Eq.any())
+          .equalUnderTheLaw(getOrNull(a)?.let { b }) { a, b -> EQB(a, b) }
```
aballano, line 58:
```suggestion
          .equalUnderTheLaw(getOrNull(a)?.let { b }, EQB)
```

arrow-libs/optics/arrow-optics/src/main/kotlin/arrow/optics/Prism.kt
```diff
@@ -128,8 +127,8 @@ interface PPrism<S, T, A, B> : POptional<S, T, A, B>, PSetter<S, T, A, B>, Fold<
     /**
      * A [PPrism] that checks for equality with a given value [a]
      */
-    fun <A> only(a: A, EQA: Eq<A>): Prism<A, Unit> = Prism(
-      getOrModify = { a2 -> (if (EQA.run { a.eqv(a2) }) Either.Left(a) else Either.Right(Unit)) },
+    fun <A> only(a: A, equals: (A, A) -> Boolean): Prism<A, Unit> = Prism(
```
aballano, line 130: nit: name consistency

arrow-libs/optics/arrow-optics/src/test/kotlin/arrow/optics/std/OptionTest.kt
```diff
@@ -31,40 +23,29 @@ class OptionTest : UnitSpec() {
...
     testLaws(IsoLaws.laws(
       iso = Iso.optionToNullable<Int>().reverse(),
       aGen = Gen.int().orNull(),
       bGen = Gen.option(Gen.int()),
-      EQA = Eq.any(),
-      EQB = Eq.any(),
       funcGen = Gen.functionAToB(Gen.option(Gen.int())),
-      bMonoid = Monoid.option(Semigroup.int())
+      EQA = { a, b -> a == b },
+      EQB = { a, b -> a == b }
```
aballano, line 41: cannot we use the defaults here?

arrow-libs/core/arrow-core-test/src/main/kotlin/arrow/core/test/laws/FxLaws.kt
nomisRev: Yes, a lamda for it named `eq` would be a lot better. We can't use an internal typealias since we're exposing these laws to users.

arrow-libs/core/arrow-core-test/src/main/kotlin/arrow/core/test/laws/MonoidLaws.kt
nomisRev: `Eq.default` or `Eq.any` like we had before can still be used like this but be replaced with a simple lambda.

arrow-libs/optics/arrow-optics/src/test/kotlin/arrow/optics/std/OptionTest.kt
nomisRev: The default lambda `{ a, b -> a?.equals(b) == true }` was incorrect so I replaced it with `{ a, b -> a == b }` now, which is equivalent to `a?.equals(b) ?: (b === null)`.

This makes this fix redundant.

danimontoya (approved):
good job!!! 🙌

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Hashed.kt
nomisRev: No, it's not 😅 Thank you 🕵️!

PR #2274, opened 2021-03-02
Undeprecate Option data type

franciscodr:
The `Option` datatype was deprecated in favor of `Nullable` types because the latter is faster than boxed types like `Option`. However,  it will be supported by Arrow again to interop with Java-based libraries that use `null` as signal or interruption value like [ReactiveX RxJava](https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#nulls).


aballano (approved):
yass 🙌

nomisRev (approved):
👍

nomisRev:
@franciscodr I think these PRs should still be going into `master` atm. I'll recreate the `release/0.12.0` when we're ready for reviewing the APIs

franciscodr:
> @franciscodr These PRs should still be going into `master` atm. I'll recreate the `release/0.12.0` when we're ready for reviewing the APIs

Ok, I'll close this PR and create a new one against `master`

franciscodr:
Closing it to create a new one against `master` branch

PR #2275, opened 2021-03-02
Remove Higher Kinded Types

nomisRev:
This PR removes `Kind` from the Arrow codebase.


raulraja (approved):
@nomisRev 😘 👏 🥳 💌

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Const.kt
```diff
@@ -1,41 +1,9 @@
...
-)inline
-fun <A, T> ConstOf<A, T>.fix(): Const<A, T> =
-  this as Const<A, T>
-
-@Deprecated(
-  message = KindDeprecation,
-  level = DeprecationLevel.WARNING
-) fun <A, T> ConstOf<A, T>.value(): A = this.fix().value()
-
-data class Const<A, out T>(private val value: A) : ConstOf<A, T> {
```
franciscodr, line 38: I think we shouldn't remove this part, should we? Just the `ConstOf<A, T>` extension

arrow-libs/core/arrow-annotations/src/main/java/arrow/extension.kt
```diff
@@ -7,7 +7,6 @@ package arrow
   AnnotationTarget.FUNCTION
 )
 @MustBeDocumented
-@Deprecated(KindDeprecation)
 annotation class extension
```
aballano, line 10: This and `higherkind` got undeprecated?

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Const.kt
nomisRev: Only the `ConstOf<A, T>` extensions has been removed? 🤔 Am I missunderstanding your comment?

aballano (approved):
just one nit, looks good!

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Const.kt
franciscodr: Nevermind, I didn't see you were adding 
```
data class Const<A, out T>(private val value: A) {
```

arrow-libs/core/arrow-annotations/src/main/java/arrow/extension.kt
nomisRev: Twice in a row 🤦 3 times and I owe you a beer!

franciscodr (approved):
Thanks, @nomisRev!

arrow-libs/core/arrow-annotations/src/main/java/arrow/extension.kt
aballano: 😘

PR #2276, opened 2021-03-02
Undeprecate Option data type

franciscodr:
The `Option` datatype was deprecated in favor of `Nullable` types because the latter is faster than boxed types like `Option`. However,  it will be supported by Arrow again to interop with Java-based libraries that use `null` as signal or interruption value like [ReactiveX RxJava](https://github.com/ReactiveX/RxJava/wiki/What's-different-in-2.0#nulls).


aballano (approved):
still "yass" 😄

Daniel-Jacob (approved):

nomisRev (approved):
👌

PR #2281, opened 2021-03-02
CU-9rr4w5 remove K Wrappers 

nomisRev:
This PR removes the K Wrappers `ListK`, `MapK`, `SetK` and `SequenceK`.

It leaves behind the type as `object` to expose `mapN`, but that needs to `@Deprecated` and needs to become top-level.
This required a 2 step deprecation since it conflicts with the top-level `mapN` for `Nullable`. This will be step 2, so in the follow-up PR we also need to remove the top-level functionality for nullable.


franciscodr:
Task linked: [CU-9rr4w5 Remove all the K-wrappers from Arrow Core (ListK, MapK, SetK)](https://app.clickup.com/t/9rr4w5)

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
```diff
@@ -1,217 +1,219 @@
...
+      }?.let { l -> destination.put(key, l) }
+    }
+    return destination
   }
-
-@Deprecated(MapKDeprecation, ReplaceWith("this.map { it.key to it.value }.toMap()"))
-fun <K, A> List<Map.Entry<K, A>>.k(): MapK<K, A> = this.map { it.key to it.value }.toMap().k()
-
-@Deprecated("Deprecated, use nullable instead", ReplaceWith("map[k]?.let { ... }"))
-fun <K, A> Map<K, A>.getOption(k: K): Option<A> = Option.fromNullable(this[k])
```
aballano, line 193: I think this might have been prematurely deprecated, since Option is still available

arrow-libs/optics/arrow-optics/src/main/kotlin/arrow/optics/std/map.kt
```diff
@@ -46,7 +45,7 @@ fun <K, V> PTraversal.Companion.map(): Traversal<Map<K, V>, V> =
 fun <K, V> At.Companion.map(): At<Map<K, V>, K, Option<V>> =
   At { i ->
     PLens(
-      get = { it.getOption(i) },
+      get = { Option.fromNullable(it[i]) },
```
aballano, line 48: I liked more the previous version 😆

aballano (approved):
🔥

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
nomisRev: I was wondering that myself, but that would mean we should duplicate all APIs for `null` and `Option` which I also don't think is great.
Especially since the idea of `Option` is only to use it in certain use-cases, and you can always lift nullable to `Option` whenever you need to like you have to do with Kotlin's Std today.

arrow-libs/optics/arrow-optics/src/main/kotlin/arrow/optics/std/map.kt
nomisRev: I'm actually thinking if this should be `At<Map<K, V>, K, V?>` but the `V` should be restricted to `Any`.

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/MapK.kt
aballano: having that in the balance I think we could all live without this tiny extension 😂

nomisRev: We do need to fix the incorrect `ReplaceWith` tho 🙈

pakoito (approved):
Wooo new encoding!

franciscodr (approved):
Thanks, @nomisRev!

PR #2288, opened 2021-03-03
CU-f31wbh Deprecate top-level Left & Right Either constructors

nomisRev:
These constructors duplicate the behavior of the actual constructors except they're imported with `import arrow.core.Left` instead of `arrow.core.Either.Left` so they're redundant.


franciscodr:
Task linked: [CU-f31wbh Deprecate top-level Left & Right constructor for Either](https://app.clickup.com/t/f31wbh)

aballano (approved):

pakoito (approved):

PR #2292, opened 2021-03-03
CU-f34dk7 remove old mapN and expose new APIs

nomisRev:
Second, and last, step of the deprecation/removacl top-level `mapN` for nullable. So now we can expose top-level `mapN` APIs for Kotlin Std types that don't have a `companion object` to expose `mapN` upon.


franciscodr:
Task linked: [CU-f34dk7 2nd step in top-level Nullable.mapN deprecation](https://app.clickup.com/t/f34dk7)

aballano (approved):

pakoito:
This is a hammer to perf, all the nested flatMaps create a new list.

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/List.kt
```diff
@@ -4,6 +4,125 @@ import arrow.typeclasses.Monoid
...
+    c.flatMap { cc ->
+      d.flatMap { dd ->
+        e.flatMap { ee ->
+          f.flatMap { ff ->
+            g.flatMap { gg ->
+              h.flatMap { hh ->
+                i.flatMap { ii ->
+                  j.flatMap { jj ->
+                    k.map { kk ->
+                      map(bb, cc, dd, ee, ff, gg, hh, ii, jj, kk)
```
raulraja, line 114: ayuken

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Sequence.kt
```diff
@@ -3,6 +3,202 @@ package arrow.core
...
 
+fun <B, C, D> mapN(
+  b: Sequence<B>,
+  c: Sequence<C>,
+  map: (B, C) -> D
+): Sequence<D> =
+  mapN(
+    b,
+    c,
+    sequenceOf(Unit),
```
raulraja, line 14: Does `sequence(Unit)` allocate new sequences or is it backed by a singleton value? If not we could optimize this.

raulraja (approved):
Thanks @nomisRev !

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Sequence.kt
nomisRev: It's probably not backed by a singleton value.

nomisRev:
> This is a hammer to perf, all the nested flatMaps create a new list even when zipping just 2. I'd specialise every implementation.

@pakoito  For all implementations or just `list`?

pakoito:
Just `List` for now.

PR #2299, opened 2021-03-04
CU-f33z0z Move Arrow Syntax to Arrow Core

nomisRev:
This moves the `syntax` package to `Arrow Core`, and removes the `Arrow Syntax` module.
This also deprecates the duplicated functions in `arrow.core.*` to prefer the functions of `arrow.syntax.*`


franciscodr:
Task linked: [CU-f33z0z Move arrow-syntax into arrow-core](https://app.clickup.com/t/f33z0z)

arrow-site/docs/docs/quickstart/libraries/README.md
```diff
@@ -102,22 +101,6 @@ Dependency: `arrow-fx-coroutines`
...
-dependencies {
-    implementation "io.arrow-kt:arrow-syntax:$arrow_version"
-}
-```
-
-Multiple extensions functions to work better with function objects and collections.
-
-For function objects, the library provides composition, currying, partial application, memoization, pipe operator, complement for predicates, and several more helpers.
-
-For collections, arrow-syntax provides `firstOption`, tail, basic list traversal, and tuple addition.
```
aballano, line 117: shouldn't this information be included somewhere else instead?

aballano (approved):

rachelcarmena:
Hi @nomisRev :raised_hands:  

I think it's being moved to `arrow-core-data` instead of `arrow-core` in this pull request. 

As far as I know, both `arrow-syntax` and `arrow-core-data` should move to `arrow-core`.

And just one thing else:



:point_up_2: that folder appeared and it's not inside a module.

Thank you so much!!

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/predef.kt
```diff
@@ -1,14 +1,40 @@
 package arrow.core
 
+import arrow.syntax.function.curried
```
raulraja, line 3: I think the `syntax` namespace is no longer needed. Can these functions live directly in the `arrow.core` namespace?

arrow-site/docs/docs/quickstart/libraries/README.md
raulraja: It should show up in the List menu automatically as `List` extensions if we move these with the rest of extensions to the `arrow.core` namespace. I personally think we should get rid of the name `syntax` all together as this came from cats to differentiate type class members from actual extensions but in Kotlin it's just extensions and these should be IMO in the same package as the rest.

raulraja (commented):
Everything looks great but I think we should get rid of the `syntax` name all together. Thanks @nomisRev !

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/predef.kt
nomisRev: It can but then we need to update the deprcation on `0.12.0`, I can take care of that first in another PR then :)

nomisRev:
> I think it's being moved to arrow-core-data instead of arrow-core in this pull request.
> As far as I know, both arrow-syntax and arrow-core-data should move to arrow-core.

Okay, then I understood incorrectly from the ticket but that makes a lot more sense IMO!
I'll close this PR, and open another one where things are moved to `arrow-core` instead.

PR #2302, opened 2021-03-04
Cu-f726n2 Remove extensions objects

nomisRev:


franciscodr:
Task linked: [CU-f726n2 Remove ListExtensions, SetExtensions, SortedSetInstances, MapInstances & SortedMapInstances](https://app.clickup.com/t/f726n2)

aballano (approved):

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/set.kt
```diff
@@ -4,10 +4,6 @@ import arrow.typeclasses.Monoid
 import arrow.typeclasses.Semigroup
 import kotlin.collections.plus as _plus
 
-object SetExtensions
-
-object SortedSetInstances
-
 fun <A> Semigroup.Companion.set(): Semigroup<Set<A>> = object : Semigroup<Set<A>> {
```
raulraja, line 7: This is probably unsafe since while `setOf(1, 2, 3)` guarantees iteration order and in most examples this would work other `Set` instances at runtime may not guarantee order and this may break associativity in the semigroup. I think all direct extensions over `Set` should be discarded. Users can always choose to use `Set<A>::plus` once Semigroup is a fun interface if they choose to do so without us providing explicit instances and behaviors for it like Set.combine that may lead to non deterministic programs. IMO Set.combine dos not bring anything to the table since to combine the two sets we still need to iterate them and the order there is not guaranteed, at most this should maybe just appear over SortedSet. Thoughts?

raulraja (approved):
thanks @nomisRev !

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/set.kt
nomisRev: Looks at Haskell and Cats it seems `Set` indeed has no instance for `Monoid` unless it's an `OrderedSet`.

Cats however has a `MonoidK` instance for `union` which makes me wonder a bit. It makes me wonder tho which law is broken for this instance because as far as I know, it's a lawful instance.

Actually, the order is irrelevant in `Set#equals`. Right?

The only reason I added this instance is that we currently have it in `0.11.0`, so if we want to remove it we need to add the proper deprecations in `0.12.0` which I can take care of in a separate PR.

nomisRev: Will be removed in this PR ;) https://github.com/arrow-kt/arrow/pull/2318
Going to go ahead and merge this one. The Monoid instance for Set will be removed in the update from master.

nomisRev:
Closed due to already in release/0.1.30 cc/ @rachelcarmena

PR #2308, opened 2021-03-08
CU-fb0myr Remove deprecated code

nomisRev:
This PR removes all the remaining `@Deprecated` code, beside the one introduced in 0.13.0 for making `mapN` top-level.


franciscodr:
Task linked: [CU-fb0myr Remove remaining deprecated code](https://app.clickup.com/t/fb0myr)

i-walker (approved):
Thanks !

arrow-libs/core/arrow-meta/src/main/java/arrow/common/utils/AbstractProcessor.kt
```diff
@@ -92,13 +91,7 @@ abstract class AbstractProcessor : KotlinAbstractProcessor(), ProcessorUtils, Ko
     else -> knownError("Unsupported @documented $kind")
   }
 
-  private fun processDocs(roundEnv: RoundEnvironment): Unit =
```
aballano, line 95: is this on purpose? If so, wouldn't be possible to delete the entire class instead?

aballano (approved):

arrow-libs/core/arrow-meta/src/main/java/arrow/common/utils/AbstractProcessor.kt
nomisRev: This is on purpose. The `Docs` feature is no longer supported, but the `AbstractProcessor` is still in use until `kapt` for Optics is replaced by Arrow Meta.

PR #2313, opened 2021-03-09
Update typeclasses docs

aballano:
Adapt Monoid, Semigroup and Semiring docs to the new typeclass syntax

### DIFF
--- a/arrow-docs/docs/arrow/typeclasses/monoid/README.md
+++ b/arrow-docs/docs/arrow/typeclasses/monoid/README.md
@@ -9,37 +9,39 @@ permalink: /arrow/typeclasses/monoid/
 
 
 
-`Monoid` extends the `Semigroup` type class, adding an `empty` method to semigroup's `combine`. The empty method must return a value that, when combined with any other instance of that type, returns the other instance, i.e.,
+`Monoid` extends the `Semigroup` type class, adding an `empty` function to semigroup's `combine`. The empty method must return a value that, when combined with any other instance of that type, returns the other instance, i.e.,
 
 ```kotlin
-(combine(x, empty) == combine(empty, x) == x)
+(a.combine(empty()) == empty().combine(a) == a)
 ```
 
-For example, if we have a `Monoid<String>` with `combine` defined as string concatenation, then `empty = ""`.
+For example, if we have a `Monoid<String>` with `combine` defined as string concatenation, then `empty() = ""`.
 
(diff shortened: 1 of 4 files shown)


arrow-docs/docs/arrow/typeclasses/monoid/README.md
```diff
@@ -9,77 +9,92 @@ permalink: /arrow/typeclasses/monoid/
...
 
 ```kotlin
-(combine(x, empty) == combine(empty, x) == x)
+(a.combine(empty()) == empty().combine(a) == a)
 ```
 
-For example, if we have a `Monoid<String>` with `combine` defined as string concatenation, then `empty = ""`.
+For example, if we have a `Monoid<String>` with `combine` defined as string concatenation, then `empty() = ""`.
 
 Having an empty defined allows us to combine all the elements of some potentially empty collection of `T` for which a `Monoid<T>` is defined and return a `T`, rather than an `Option<T>` as we have a sensible default to fall back to.
```
raulraja, line 20:
```suggestion
Having an empty identity allows us to combine all the elements of some potentially empty collection of `T` for which a `Monoid<T>` is defined and return a `T`, rather than an `Option<T>` as we have a sensible default to fall back to.
```

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/typeclasses/Semigroup.kt
```diff
@@ -11,7 +9,8 @@ interface Semigroup<A> {
   operator fun A.plus(b: A): A =
     this.combine(b)
 
-  fun A.maybeCombine(b: A?): A = Option.fromNullable(b).fold({ this }, { combine(it) })
+  fun A.maybeCombine(b: A?): A =
```
raulraja, line 12: 👍

raulraja (approved):
thanks @aballano !

arrow-docs/docs/arrow/typeclasses/monoid/README.md
aballano: I think it refers to having `empty()` defined

aballano: how about:

```suggestion
Having `empty` defined allows us to combine all the elements of some potentially empty collection of `T` for which a `Monoid<T>` is defined and return a `T`, rather than an `Option<T>` as we have a sensible default to fall back to.
```

raulraja: Sounds good!

i-walker (approved):
Thanks @aballano 🙌🏾

PR #2315, opened 2021-03-09
Replace mapN with zip

nomisRev:
Currently, in Arrow we have `mapN` from `Apply`. This PR proposes to replace it in Arrow with `zip`, which is more in line with the Kotlin Std.

For example the Kotlin Std exposes `zip` arity 2 for `Iterable` and instead of duplicating `zip` with `mapN` and providing up to arity 10 we can now just extend the Kotlin Std API and expose arity 3-10 for `Iterable#zip`. Doing so we clearly complement the Kotlin Std instead of proposing alternative APIs.

It also prevents us from having to do 2 deprecation cycles, and will result in a `@Deprecated` free library @ `0.13.0`.

This PR also removes all `N` based APIs by also replacing `parMapN` in Arrow Fx Coroutines with `parZip`.


arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -1703,6 +1584,110 @@ fun <A, B, C> Either<A, B>.zip(fb: Either<A, C>): Either<A, Pair<B, C>> =
...
+  c: Either<A, C>,
+  d: Either<A, D>,
+  e: Either<A, E>,
+  f: Either<A, F>,
+  g: Either<A, G>,
+  h: Either<A, H>,
+  i: Either<A, I>,
+  map: (B, C, D, E, F, G, H, I) -> J
+): Either<A, J> =
+  zip(c, d, e, f, g, h, i, Either.unit, Either.unit) { b, c, d, e, f, g, h, i, _, _ -> map(b, c, d, e, f, g, h, i) }
```
raulraja, line 1642: just realized we have `Either.unit` vs `Eval.Unit`

raulraja (approved):
Thanks @nomisRev !

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Either.kt
nomisRev: `Either.unit` is `internal` and `Eval.Unit` is deprecated ;)

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Validated.kt
```diff
@@ -233,25 +233,41 @@ fun <E, A> ValidatedOf<E, A>.fix(): Validated<E, A> =
...
+ * import arrow.core.Validated
+ * import arrow.core.validNel
+ * import arrow.core.zip
+ *
+ * //sampleStart
+ * val parallelValidate =
+ *   1.validNel().zip(2.validNel())
+ *     { a, b -> /* combine the result */ }
+ * //sampleEnd
+ * ```
```
nomisRev, line 269: @JorgeCastilloPrz ;)

i-walker (approved):
Thanksk @nomisRev 🙌🏾

PR #2318, opened 2021-03-10
Remove Set Monoid and update deprecations

nomisRev:
This PR removes Set Monoid from Arrow, as discussed [here](https://github.com/arrow-kt/arrow/pull/2302).


raulraja (approved):
thanks @nomisRev !

i-walker (approved):

PR #2324, opened 2021-03-11
CU-fh2vcf - Deprecate mapN and fix zip for collections

danimontoya:
This PR updates/fixes some `zip` implementations and adds tests for it.
It also deprecates `mapN` for collections, which were cartesian builders. For collections, this easily results in really big lists that make the JVM OOM.

Example of dangerous usage:

```kotlin
val x: List<Int> = listOf(1, 2, 3, 4, 5, 6)

// OOMs
val result = mapN(x, x, x, x, x, x, x, x, x, x, ::Tuple10)
```

This tries to create a list with 6^10 (60466176) elements and as a result OOMs. As you can see this easily results in bad code since here we're only using a list with 6 elements and it already results in problematic code.

Since this pattern isn't so common and is only safe with low arities we will recommend to simply use `flatMap/map` chains instead.

### DIFF
--- a/arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Iterable.kt
+++ b/arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Iterable.kt
@@ -9,26 +9,72 @@ import kotlin.collections.foldRight as _foldRight
 inline fun <B, C, D, E> Iterable<B>.zip(
   c: Iterable<C>,
   d: Iterable<D>,
-  map: (B, C, D) -> E
-): List<E> =
-  zip(c, d, unit, unit, unit, unit, unit, unit, unit) { b, c, d, _, _, _, _, _, _, _ -> map(b, c, d) }
+  transform: (B, C, D) -> E
+): List<E> {
+  val bb = iterator()
+  val cc = c.iterator()
+  val dd = d.iterator()
+
+  val size = minOf(
+    collectionSizeOrDefault(10),
+    c.collectionSizeOrDefault(10),
+    d.collectionSizeOrDefault(10)
+  )
+  val list = ArrayList<E>(size)
+  while (bb.hasNext() && cc.hasNext() && dd.hasNext()) {
+    list.add(transform(bb.next(), cc.next(), dd.next()))
+  }
+  return list
+}
 
 inline fun <B, C, D, E, F> Iterable<B>.zip(
   c: Iterable<C>,
   d: Iterable<D>,
   e: Iterable<E>,
-  map: (B, C, D, E) -> F
-): List<F> =
-  zip(c, d, e, unit, unit, unit, unit, unit, unit) { b, c, d, e, _, _, _, _, _, _ -> map(b, c, d, e) }
+  transform: (B, C, D, E) -> F
+): List<F> {
+  val bb = iterator()
+  val cc = c.iterator()
+  val dd = d.iterator()
+  val ee = e.iterator()
+  val size = minOf(
+    collectionSizeOrDefault(10),
+    c.collectionSizeOrDefault(10),
+    d.collectionSizeOrDefault(10),
+    e.collectionSizeOrDefault(10)
+  )
+  val list = ArrayList<F>(size)
+  while (bb.hasNext() && cc.hasNext() && dd.hasNext() && ee.hasNext()) {
+    list.add(transform(bb.next(), cc.next(), dd.next(), ee.next()))
+  }
+  return list
+}
 
 inline fun <B, C, D, E, F, G> Iterable<B>.zip(
   c: Iterable<C>,
   d: Iterable<D>,
   e: Iterable<E>,
   f: Iterable<F>,
-  map: (B, C, D, E, F) -> G
-): List<G> =
-  zip(c, d, e, f, unit, unit, unit, unit, unit) { b, c, d, e, f, _, _, _, _, _ -> map(b, c, d, e, f) }
+  transform: (B, C, D, E, F) -> G
+): List<G> {
+  val bb = iterator()
+  val cc = c.iterator()
+  val dd = d.iterator()
+  val ee = e.iterator()
+  val ff = f.iterator()
+  val size = minOf(
+    collectionSizeOrDefault(10),
+    c.collectionSizeOrDefault(10),
+    d.collectionSizeOrDefault(10),
+    e.collectionSizeOrDefault(10),
+    f.collectionSizeOrDefault(10)
+  )
+  val list = ArrayList<G>(size)
+  while (bb.hasNext() && cc.hasNext() && dd.hasNext() && ee.hasNext() && ff.hasNext()) {
+    list.add(transform(bb.next(), cc.next(), dd.next(), ee.next(), ff.next()))
+  }
+  return list
+}
 
 inline fun <B, C, D, E, F, G, H> Iterable<B>.zip(
   c: Iterable<C>,
@@ -36,9 +82,28 @@ inline fun <B, C, D, E, F, G, H> Iterable<B>.zip(
   e: Iterable<E>,
   f: Iterable<F>,
   g: Iterable<G>,
-  map: (B, C, D, E, F, G) -> H
-): List<H> =
-  zip(c, d, e, f, g, unit, unit, unit, unit) { b, c, d, e, f, g, _, _, _, _ -> map(b, c, d, e, f, g) }
+  transform: (B, C, D, E, F, G) -> H
+): List<H> {
+  val bb = iterator()
+  val cc = c.iterator()
+  val dd = d.iterator()
+  val ee = e.iterator()
+  val ff = f.iterator()
+  val gg = g.iterator()
+  val size = minOf(
+    collectionSizeOrDefault(10),
+    c.collectionSizeOrDefault(10),
+    d.collectionSizeOrDefault(10),
+    e.collectionSizeOrDefault(10),
+    f.collectionSizeOrDefault(10),
+    g.collectionSizeOrDefault(10)
+  )
+  val list = ArrayList<H>(size)
+  while (bb.hasNext() && cc.hasNext() && dd.hasNext() && ee.hasNext() && ff.hasNext() && gg.hasNext()) {
+    list.add(transform(bb.next(), cc.next(), dd.next(), ee.next(), ff.next(), gg.next()))
+  }
+  return list
+}
 
 inline fun <B, C, D, E, F, G, H, I> Iterable<B>.zip(
   c: Iterable<C>,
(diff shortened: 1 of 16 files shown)


franciscodr:
Task linked: [CU-fh2vcf Replace mapN with zip (NonEmptyList & NonEmptyListApply)](https://app.clickup.com/t/fh2vcf)

nomisRev (approved):
LGTM thanks @danimontoya ❤️

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -550,8 +429,185 @@ object NonEmptyListSemigroup : Semigroup<NonEmptyList<Any?>> {
...
+  f: NonEmptyList<F>,
+  g: NonEmptyList<G>,
+  h: NonEmptyList<H>,
+  i: NonEmptyList<I>,
+  j: NonEmptyList<J>,
+  map: (A, B, C, D, E, F, G, H, I, J) -> Z
+): NonEmptyList<Z> =
+  NonEmptyList(
+    map(head, b.head, c.head, d.head, e.head, f.head, g.head, h.head, i.head, j.head),
+    tail.flatMap { aa ->
```
aballano, line 592: I believe [this comment](https://github.com/arrow-kt/arrow/pull/2292#issuecomment-789809329) would apply here as well

aballano (approved):

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt
danimontoya: it does!
shall I fixed like done [here](https://github.com/arrow-kt/arrow/pull/2292/files#diff-69aba3b9460908255c0db2f50cc13506aada48c4e1a1d3d8f7bc9db8ff1f8616R104)?

danimontoya: something like this:
```
inline fun <A, B, C, D, E, F, G, H, I, J, Z> NonEmptyList<A>.zip(
  b: NonEmptyList<B>,
  c: NonEmptyList<C>,
  d: NonEmptyList<D>,
  e: NonEmptyList<E>,
  f: NonEmptyList<F>,
  g: NonEmptyList<G>,
  h: NonEmptyList<H>,
  i: NonEmptyList<I>,
  j: NonEmptyList<J>,
  map: (A, B, C, D, E, F, G, H, I, J) -> Z
): NonEmptyList<Z> {
  val buffer = ArrayList<Z>()
  for (aa in tail) {
    for (bb in b.tail) {
      for (cc in c.tail) {
        for (dd in d.tail) {
          for (ee in e.tail) {
            for (ff in f.tail) {
              for (gg in g.tail) {
                for (hh in h.tail) {
                  for (ii in i.tail) {
                    for (jj in j.tail) {
                      buffer.add(map(aa, bb, cc, dd, ee, ff, gg, hh, ii, jj))
                    }
                  }
                }
              }
            }
          }
        }
      }
    }
  }
  return NonEmptyList(
    map(head, b.head, c.head, d.head, e.head, f.head, g.head, h.head, i.head, j.head),
    buffer
  )
}
```

danimontoya: cc @nomisRev

aballano: Actually we should:

I just tested and It seems the `unit` variable used for calling the higher arity function [has a bug](https://github.com/arrow-kt/arrow/blob/master/arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt#L331) because it creates a `NonEmptyList(Unit, emptyList())` wich will do nothing for when we do `tail.flatMap` [here](https://github.com/arrow-kt/arrow/blob/master/arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt#L435).

At some point we should add some tests as well

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -550,8 +429,188 @@ object NonEmptyListSemigroup : Semigroup<NonEmptyList<Any?>> {
...
+  e: NonEmptyList<E>,
+  f: NonEmptyList<F>,
+  g: NonEmptyList<G>,
+  h: NonEmptyList<H>,
+  i: NonEmptyList<I>,
+  j: NonEmptyList<J>,
+  map: (A, B, C, D, E, F, G, H, I, J) -> Z
+): NonEmptyList<Z> {
+  val buffer = ArrayList<Z>()
+  for (aa in tail) {
```
aballano, line 591: this should iterate over this and the below over the nonemptylist entirely, not just the tail

```suggestion
  for (aa in this) {
```

Although now that I'm thinking of it, is there a reason why we need mapN here at all? We could simply rely on the Iterable version as it would do the same, no?

nomisRev: Yes, we actually just use `NonEmptyList.unsafeFromList(all.zip(other.all, f))`

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -550,8 +429,188 @@ object NonEmptyListSemigroup : Semigroup<NonEmptyList<Any?>> {
...
+          }
+        }
+      }
+    }
+  }
+  return NonEmptyList(
+    map(head, b.head, c.head, d.head, e.head, f.head, g.head, h.head, i.head, j.head),
+    buffer
+  )
+}
```
nomisRev, line 616:
```suggestion
): NonEmptyList<Z> =
  NonEmptyList(
    map(b.head, c.head, d.head, e.head, f.head, g.head, h.head, i.head, j.head, k.head),
    b.tail.zip(c.tail, d.tail, e.tail, f.tail, e.tail, f.tail, g.tail, h.tail, i.tail, j.tail, k.tail, map)
  )
```

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt
nomisRev: What is the bug @aballano? If the `tail == emptyList` then there is nothing to `zip`/`mapN` there, right?

aballano: @nomisRev Indeed, so for this:

    NonEmptyList.mapN(
        nonEmptyListOf(1, 2, 3),
        nonEmptyListOf(4, 5),
    ) { one, two ->
        println("->$one $two")
        ""
    }

there's just 1 iteration and prints "->1 4" and nothing else, because it just maps the head, not the tails.

aballano: @nomisRev And BTW I know we are changing map to zip because of the stdlib, but we're doing it differently: the stdlib's zip combines 2 lists using the same index, meaning it does return a list with the shortest length of both:

listOf(1, 2).zip(listOf(4, 5, 6)) { one, two ->
        println("->$one $two")
    }

// ->1 4
// ->2 5

while our zip does a series of flatmaps and maps each value to each value:

// ->1 4
// ->1 5
// ->1 6
// ->2 4
// ->2 5
// ->2 6

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt
aballano: I think this isn't right, you're missing the value for `this`

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -550,8 +429,165 @@ object NonEmptyListSemigroup : Semigroup<NonEmptyList<Any?>> {
...
+  f: NonEmptyList<F>,
+  g: NonEmptyList<G>,
+  h: NonEmptyList<H>,
+  i: NonEmptyList<I>,
+  j: NonEmptyList<J>,
+  map: (A, B, C, D, E, F, G, H, I, J) -> Z
+): NonEmptyList<Z> =
+  NonEmptyList(
+    map(head, b.head, c.head, d.head, e.head, f.head, g.head, h.head, i.head, j.head),
+    tail.zip(b.tail, c.tail, d.tail, e.tail, f.tail, g.tail, h.tail, i.tail, j.tail, map)
```
aballano, line 592: I think this is right, but will probably not work still, we might need to update the `unit` variable

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/NonEmptyList.kt
nomisRev: > there's just 1 iteration and prints "->1 4" and nothing else, because it just maps the head, not the tails.

Since `for(xx in emptyList<A>())` never triggers, right.

> but we're doing it differently: the stdlib's zip combines 2 lists using the same index, meaning it does return a list with the shortest length of bot

Right, `zip` is not a cartesian builder which is problematic for collections. 😐
I still think it's a better fit for `Either`, `Validated`, etc but that leaves open the question for cartesian builders for collections. Do we replace them, or remove them altogether.

I just ran into the same issue with `Map`, but there `mapN == zip` since `flatMap` works over the `union` of the `keys` but deriving with `emptyMap` is invalid.

nomisRev (dismissed):
We need to figure out what to do with `zip` for collections versus cartesian builders.

arrow-libs/core/arrow-core-data/src/main/kotlin/arrow/core/Iterable.kt
```diff
@@ -9,36 +9,101 @@ import kotlin.collections.foldRight as _foldRight
...
+  val size = minOf(
+    collectionSizeOrDefault(10),
+    c.collectionSizeOrDefault(10),
+    d.collectionSizeOrDefault(10)
+  )
+  val list = ArrayList<E>(size)
+  while (bb.hasNext() && cc.hasNext() && dd.hasNext()) {
+    list.add(transform(bb.next(), cc.next(), dd.next()))
+  }
+  return list
```
nomisRev, line 27: This impl was inspired by the Kotlin Std implementation

arrow-libs/core/arrow-core-data/src/test/kotlin/arrow/core/NonEmptyListTest.kt
```diff
@@ -90,5 +90,137 @@ class NonEmptyListTest : UnitSpec() {
         }
       }
     }
+
+    "zip2" {
+      forAll(Gen.nonEmptyList(Gen.int()), Gen.nonEmptyList(Gen.int())) { a, b ->
+        val result = a.zip(b)
+        val expected = a.all.zip(b.all).let(NonEmptyList.Companion::fromListUnsafe)
```
aballano, line 97: aren't we testing against our own zip version here?

aballano (approved):
looks good!

arrow-libs/core/arrow-core-data/src/test/kotlin/arrow/core/NonEmptyListTest.kt
nomisRev: Yes, but that `zip` is tested against Kotlin Std `zip` in `IterableTest` so it assumes our `zip` is correct/tested. Which is added in this PR also.

raulraja (approved):
thanks folks!

i-walker (approved):
🙌🏾

PR #2327, opened 2021-03-16
Fix Const & Either mapN deprecation

nomisRev:
Fixes missing import for Either#mapN deprcation and removes unneeded import for Const.


raulraja (approved):
thanks @nomisRev

i-walker (approved):
Thanks Simon :D

PR #2328, opened 2021-03-17
CU-fv6xq2 Clean up Arrow Syntax based on API Review

nomisRev:
This PR cleans up the API for Arrow Syntax for Arrow 0.13.0 (1.0.0). The API review can be found [here](https://github.com/nomisRev/arrow-api-review/pull/2)

- Reduce to arity-5 instead of 22
- Remove arity post-fix from Partials.kt
- Deprecate Complement
- Deprecate pipe
- Deprecate reverse
- Deprecate tupling.kt
- Deprecate pairing.kt

PR #2331, opened 2021-03-22
Revert partially without explicit arity

nomisRev:
In https://github.com/arrow-kt/arrow/pull/2328, we renamed `partiallyN` to `partially` but this conflicts for functions that have the same type. This needs to be reverted, so the user an explicitly choose which parameter that needs to be partially applied.

Such as the following test:

```kotlin
"partially" {
  val sum5ints = { a: Int, b: Int, c: Int, d: Int, e: Int -> a + b + c + d + e }
  val sum4intsTo10 = sum5ints.partially5(10)
  val sum3intsTo15 = sum4intsTo10.partially4(5)
  val sum2intsTo17 = sum3intsTo15.partially3(2)
  sum2intsTo17(1, 2) shouldBe 20
  val prefixAndPostfix = { prefix: String, x: String, postfix: String -> "$prefix$x$postfix" }
  val helloX = prefixAndPostfix.partially1("Hello, ").partially2("!")
  helloX("Arrow") shouldBe "Hello, Arrow!"
}
```

### DIFF
--- a/arrow-libs/core/arrow-syntax/src/main/kotlin/arrow/core/partials.kt
+++ b/arrow-libs/core/arrow-syntax/src/main/kotlin/arrow/core/partials.kt
@@ -1,60 +1,46 @@
 package arrow.core
 
-fun <P1, R> ((P1) -> R).partially(p1: P1): () -> R =
+fun <P1, R> ((P1) -> R).partially1(p1: P1): () -> R =
   { this(p1) }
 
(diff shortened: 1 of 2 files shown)


raulraja (approved):
thanks @nomisRev !

rachelcarmena (approved):
:rocket:

i-walker (approved):

PR #2334, opened 2021-03-23
Fix Java API for typeclasses in Arrow Core

nomisRev:
Currently the Java API for consuming `Monoid`, `Semigroup` or `Semigroup` APIs are pretty bad and in some cases even unusable. This PR proposes a fix for it.

Currently:

```
List<String> list = Arrays.asList("1", "2", "3", "4", "5");
String res = IterableKt.combineAll(
        list,
        StringKt.string(Monoid.Companion) // Monoid.string() in Kotlin
);

List<Integer> listB = Arrays.asList(1, 2, 3, 4, 5);
Integer resB = IterableKt.combineAll(
        listB,
        // Broken since `int` is a reserved keyword
        NumberKt.int(Monoid.Companion) // Monoid.int() in Kotlin
);
```

New Syntax:

```
List<String> list = Arrays.asList("1", "2", "3", "4", "5");
String res = IterableKt.combineAll(
        list,
        Monoid.string() // Monoid.string() in Kotlin
);

List<Integer> listB = Arrays.asList(1, 2, 3, 4, 5);
Integer resB = IterableKt.combineAll(
        listB,
        // Primitives are exposed by uppercase names from Java
        Monoid.Integer() // Monoid.int() in Kotlin
);
```

As you can see we can now simply call it in the same (or almost same) manner from Java as we do from Kotlin.
Additionally this change removes the need in Kotlin to add imports for the instanes. It's no longer needed to import `arrow.core.int` to call `Monoid.int()` since now they're concrete methods on the actual `companion object` of `Monoid`.


i-walker (approved):
🙌🏾 thanks @nomisRev

raulraja (approved):
nice improvement! thanks @nomisRev

1Jajen1 (approved):
Changing `String.monoid()` to `Monoid.string()` does seem to solve a lot of api issues ^^
Time to check out what else changed over the last months... :)

PR #2335, opened 2021-03-23
@JvmStatic companion object functions & rename new Arrow Syntax files

nomisRev:
Follow-up PR of #2334 which fixed using the Arrow Core typeclasses from Java.
This PR improves calling the functions defined on the `companion object` from Java by annotating them with `JvmStatic`.

It also fixes a conflict with a Java reserved keyword `catch`, by renaming it to `tryCatch` for `Option`, `Either` & `Validated`.

Additionally, it also improves the generated names for the `Arrow Syntax` files.


arrow-libs/core/arrow-core-data/src/test/kotlin/arrow/core/EitherUsage.java
```diff
@@ -0,0 +1,23 @@
+package arrow.core;
+
+import kotlin.Unit;
+import kotlin.jvm.functions.Function1;
+
+public class EitherUsage {
+
+    public void testUsage() {
+        Either<Throwable, Integer> throwableIntegerEither = Either.tryCatch(() -> 1);
```
raulraja, line 9: Arrow on Java nice!

raulraja (approved):
thanks @nomisRev !

i-walker (approved):

PR #2336, opened 2021-03-23
CU-g52kvp - Fix Java API for typeclasses in Arrow Optics

danimontoya:
Fixes https://app.clickup.com/t/g52kvp
Move arrow.optics.std.* functions into each type class companion object with JvmStatic annotation and remove imports from deprecations:
- list
- either
- map
- nonemptylist
- option
- sequence
- set
- string
- validated
- tuple

PR #2341, opened 2021-03-29
Fix redundant tuple naming

nomisRev:
Small follow up for https://github.com/arrow-kt/arrow/pull/2336


arrow-libs/optics/arrow-optics/src/main/kotlin/arrow/optics/std/tuple.kt
```diff
@@ -75,7 +75,7 @@ fun <A, B> Tuple2.Companion.second(): Lens<Tuple2<A, B>, B> = pSecond()
 @Deprecated(
   "Use the pairPTraversal function exposed in the Traversal' companion object",
```
raulraja, line 76: Should all the deprecated messages get updated with? :
```suggestion
  "Use the `pPair` function exposed in the `Traversal` companion object",
```

raulraja (approved):
It looks great @nomisRev! Let a comment regarding updating the names in the deprecation messages too.

danimontoya (approved):
thanks @nomisRev 👍

PR #2355, opened 2021-03-31
Fix Iso generation using letters instead of named params

nomisRev:
`Iso` generation is still using `a`, `b`, `c`, etc instead of `first`, `second` and so on.
I think the meta tests are currently disabled, but this is okay since we're moving to Arrow Meta as soon as Koltin 1.5 is released which has much better testing facilities.

Reported on Kotlin Slack: https://kotlinlang.slack.com/archives/C5UPMM0A0/p1617199987116700


rachelcarmena (approved):
:raised_hands:

danimontoya (approved):
🚢

PR #2368, opened 2021-04-05
Change traverseEither/Validated to use buildList/Map

kyay10:
This also uses the list/map's size when available to prevent unnecessary array resizes in the middle of traversing. I also added 10 as the default for collectionSizeOrDefault because it seemed to be the de-facto all throughout the code.

(I just realized that I accidentally used Ctrl+Alt+L on the code which I think expanded some unrelated calls but oh well)


kyay10:
It might be beneficial btw to add the tests from the Nel PR that make sure the head isn't traversed twice here just in case, but I'm not sure tbh.

1Jajen1:
I'll use this, if we end up being fine with `buildList`, thanks!
However the diff looks quite weird, is this from the right branch?

> It might be beneficial btw to add the tests from the Nel PR that make sure the head isn't traversed twice here just in case, but I'm not sure tbh.

I believe the new tests there cover those well enough as there are tests verifying the result lists including all elements. (They match against every traversed element and the initial lists)

nomisRev:
I don't think we shouldn't add `@OptIn(ExperimentalStdlibApi::class)` to our APIs.
Only in some exceptional cases where we for example add overloads for `kotlin.time.*`, so that it doesn't hinder the user from using them.

I'm not sure when this is becoming stable, but I think we should post-pone this enhancement until then.

PR #2369, opened 2021-04-05
Setup contracts for Sumtypes to improve compiler autocasts.

1Jajen1:
Not sure this is wanted, but this may help the compiler provide a few obvious invariants to the user:
- `zip(nullableA, nullableB, f)?.let { // now nullableA and nullableB should be proven not null }`
- `if (either.exists(f)) /** { either should be proven Right here } */ else /** { either is either left or right here } */`
- `if (ior.isEmpty()) /** { ior is Left } */ else /** { ior is either Right or Both } */`

All of this is annoying to do manually, by using contracts we can prove it automatically. This can be extended quite a bit more by adding `callsInPlace(f, InvocationKind)` in modules, which allows more flexible non-local returns (a return from a function with `InvocationKind` `EXACTLY_ONCE` or `AT_LEAST_ONCE` tells the compiler no other branch outside the lambda needs to be checked for returns. Also some var assignment stuff becomes possible with those contracts, but imo the most important bits are non-local returns in inline functions and sumtype assertions/autocasts.

This does require the flag `ExperimentalContractApi` tho, so not sure what to do about that. Should it be `OptIn` or per method? Or can we enable this globally somehow?

### DIFF
--- a/arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
@@ -5,6 +5,9 @@ import arrow.core.Either.Left
 import arrow.core.Either.Right
 import arrow.typeclasses.Monoid
 import arrow.typeclasses.Semigroup
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.InvocationKind
+import kotlin.contracts.contract
 
 /**
  *
@@ -719,6 +722,7 @@ import arrow.typeclasses.Semigroup
  * Option does not require a type parameter with the following functions, but it is specifically used for Either.Left
  *
  */
+@OptIn(ExperimentalContracts::class)
 sealed class Either<out A, out B> {
 
   /**
@@ -733,9 +737,21 @@ sealed class Either<out A, out B> {
    */
   internal abstract val isLeft: Boolean
 
-  fun isLeft(): Boolean = isLeft
+  fun isLeft(): Boolean {
+    contract {
+      returns(true) implies (this@Either is Either.Left)
+      returns(false) implies (this@Either is Either.Right)
+    }
+    return isLeft
+  }
 
-  fun isRight(): Boolean = isRight
+  fun isRight(): Boolean {
+    contract {
+      returns(true) implies (this@Either is Either.Right)
+      returns(false) implies (this@Either is Either.Left)
+    }
+    return isRight
+  }
 
   /**
    * Applies `ifLeft` if this is a [Left] or `ifRight` if this is a [Right].
@@ -770,7 +786,7 @@ sealed class Either<out A, out B> {
       is Left -> initial
     }
 
-  fun <C> foldMap(MN: Monoid<C>, f: (B) -> C): C = MN.run {
+  inline fun <C> foldMap(MN: Monoid<C>, f: (B) -> C): C = MN.run {
     foldLeft(MN.empty()) { b, a -> b.combine(f(a)) }
   }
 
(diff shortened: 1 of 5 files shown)


arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -733,9 +737,21 @@ sealed class Either<out A, out B> {
    */
   internal abstract val isLeft: Boolean
 
-  fun isLeft(): Boolean = isLeft
+  fun isLeft(): Boolean {
+    contract {
+      returns(true) implies (this@Either is Either.Left)
+      returns(false) implies (this@Either is Either.Right)
+    }
```
nomisRev, line 744: YES!

arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
```diff
@@ -1211,8 +1256,13 @@ inline fun <A, B> Either<A, B?>.leftIfNull(default: () -> A): Either<A, B> =
  * @param elem the element to test.
  * @return `true` if the option has an element that is equal (as determined by `==`) to `elem`, `false` otherwise.
  */
-fun <A, B> Either<A, B>.contains(elem: B): Boolean =
-  exists { it == elem }
+@OptIn(ExperimentalContracts::class)
```
nomisRev, line 1259: Why is this needed?

arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -325,6 +328,7 @@ import arrow.typeclasses.Semigroup
  * Contents partially adapted from [Scala Exercises Option Tutorial](https://www.scala-exercises.org/std_lib/options)
  * Originally based on the Scala Koans.
  */
+@OptIn(ExperimentalContracts::class)
```
nomisRev, line 331: We shouldn't make `Option` `Experimental`. Are contracts still experimental?

nomisRev (commented):
Looks amazing! 😍 Only concerned about `ExperimentalContracts`

arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
1Jajen1: Because the top level one from Either does not apply to extensions :/

arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Option.kt
1Jajen1: I really don't want this as well. I'll test adding the ExperimentalContracts annotation to each method instead and check from a different lib like fx if the compiler still complains there. But it won't work without the annotation, intellij and the compiler both complain.

1Jajen1: Per method annotations force me to add this to every method that calls a annotated one. This isn't nice, but yeah I kind of think they should not be experimental as well. Maybe a old kotlin version? Will check more tomorrow.

nomisRev:
@1Jajen1 it seems not all kind of contracts are stable yet. https://kotlinlang.org/docs/whatsnew13.html#contracts

For example `contract { callsInPlace(block, EXACTLY_ONCE) }` doesn't seem to be a problem, but tho other ones which have to evaluate expressions to check values are currently still experimental. I don't we should add those add the time yet :(

1Jajen1:
> @1Jajen1 it seems not all kind of contracts are stable yet. https://kotlinlang.org/docs/whatsnew13.html#contracts
> 
> For example `contract { callsInPlace(block, EXACTLY_ONCE) }` doesn't seem to be a problem, but tho other ones which have to evaluate expressions to check values are currently still experimental. I don't we should add those add the time yet :(

Damn, those are the more interesting ones :/ I'll close this for now, this isn't too much work too add once its stable.

PR #2370, opened 2021-04-06
Deprecate foldRight and all lazy folds

1Jajen1:
FoldRight and similar are useless in most cases in arrow atm.
- No kinds, without kinds we know that all implementations of foldRight atm are on non-lazy types that call the fold in place
- A consequence of the above is that all of these folds are `inline` and can thus short circuit using non-local returns.

A quick overview of all the types of folds present in arrow atm:
- `foldLeft = fold` This iterates over the elements in order left to right and accumulates along the way
- `foldRight` (without `Eval`) This iterates over the elements in order right to left. I'd argue that this is hardly ever what anyone wants or needs and writing `reverse().fold` isn't that annoying or common to warrant a shortcut (especially not one named `foldRight`)
- `foldRight` (with `Eval`): This returns a lazy computation that is right nested. This means it only ever iterates the list as far as necessary to calculate a result. Thus it is also safe for infinite structures!
> Not in arrow as this is almost never useful
- `foldLeft` (with `Eval`) This builds a lazy computation that is left nested, which means to produce the final lazy computation we must traverse the entire list first. Highly inefficient when we want to short circuit or when we actually need every result applied.

So out of the 4 only the strict `fold` and `foldRight` with `Eval` seem useful. `fold` for obvious reasons, and lazy `foldRight` because it can short circuit at any point by simply not forcing more of the accumulator.

The stdlib includes the strict left fold and the strict right fold, so why no lazy folds? Well all of the folds inside the stdlib are inline and thus non-local returns can handle the use-case of short-circuiting way better. It is cleaner, faster and easier to understand to return from a local inline fold function than to explicitly not-bind the lazy accumulator.

The other useful part of lazy folds is that their results themselves are lazy. Top level laziness can be  achieved by wrapping them in `Eval.later { .. }` instead.

Here is an example how a lazy right fold can be replaced by a normal fold for exactly the same functionality:
```kotlin
fun <A> List<A>.find(p: Predicat<A>): A? = foldRight(Eval.now(null)) { a, acc -> 
  if (p(a))
    Eval.now(a) // short circuit since we don't evaluate acc anymore
  else
    acc // bind acc and thus force one more step of evaluation
}.value() // evaluate computation
// This now becomes
fun <A> List<A>.find(p: Predicat<A>): A? = fold(null) { acc, a ->
  if (p(a)) return@find a else acc
}
// This can be done without a fold as well, but lets ignore that for the sake of the example.
```

### DIFF
--- a/arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
+++ b/arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Either.kt
@@ -764,6 +764,7 @@ sealed class Either<out A, out B> {
       is Left -> initial
     }
 
+  @Deprecated(FoldRightDeprecation)
   inline fun <C> foldRight(initial: Eval<C>, crossinline rightOperation: (B, Eval<C>) -> Eval<C>): Eval<C> =
     when (this) {
       is Right -> Eval.defer { rightOperation(value, initial) }
@@ -777,6 +778,7 @@ sealed class Either<out A, out B> {
   inline fun <C> bifoldLeft(c: C, f: (C, A) -> C, g: (C, B) -> C): C =
     fold({ f(c, it) }, { g(c, it) })
 
+  @Deprecated(FoldRightDeprecation)
   inline fun <C> bifoldRight(c: Eval<C>, f: (A, Eval<C>) -> Eval<C>, g: (B, Eval<C>) -> Eval<C>): Eval<C> =
     fold({ f(it, c) }, { g(it, c) })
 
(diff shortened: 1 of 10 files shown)


nomisRev (approved):
Great work @1Jajen1, and great explentation! 👏 👏 👏 

I'd actually like to add this in the docs somewhere since short-circuiting fold instead of traversing the whole `fold` seems to be a common question in FP.

raulraja (approved):
thanks @1Jajen1 great description! 👏

PR #2373, opened 2021-04-08
Documentation: update Arrow Meta example

rachelcarmena:
NOTE: It cannot be checked with Ank because the Kotlin version doesn't match with Arrow now. It will be checked as soon as they match.


raulraja (approved):
Thanks @rachelcarmena

Issue #2375, opened 2021-04-16
["Bug"] Deadlock when trying to create Either.Left(Unit) and Either.Right(null) concurrently


ilya40umov:
Arrow version: 0.13.1
Coroutines version: 1.4.3

The following code hangs and never finished on 0.13.1 (I discovered this while trying to update Arrow from 0.11.0 and one of my tests that involves concurrency failed).

```
fun main() {
    runBlocking {
        (0..10).map { i ->
            GlobalScope.launch {
                println("Before")
                if (i % 2 == 0) {
                    Either.Left(Unit)
                } else {
                    Either.Right(null)
                }
                println("After") // never gets printed
            }
        }.joinAll()
    }
}
```

Here is a thread dump from some of the threads:
```
"DefaultDispatcher-worker-7" #19 daemon prio=5 os_prio=0 cpu=1.03ms elapsed=282.59s tid=0x00007f7b1c5ce800 nid=0x1a37da in Object.wait()  [0x00007f7acea01000]
   java.lang.Thread.State: RUNNABLE
	at arrow.core.Either.<clinit>(Either.kt:945)
	at MainKt$main$1$1$1.invokeSuspend(main.kt:13)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
	at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
	at kotlinx.coroutines.scheduling.CoroutineScheduler.runSafely(CoroutineScheduler.kt:571)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.executeTask(CoroutineScheduler.kt:750)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.runWorker(CoroutineScheduler.kt:678)
	at kotlinx.coroutines.scheduling.CoroutineScheduler$Worker.run(CoroutineScheduler.kt:665)

"DefaultDispatcher-worker-1" #13 daemon prio=5 os_prio=0 cpu=1.05ms elapsed=282.59s tid=0x00007f7b1c543000 nid=0x1a37d4 in Object.wait()  [0x00007f7acf008000]
   java.lang.Thread.State: RUNNABLE
	at MainKt$main$1$1$1.invokeSuspend(main.kt:13)
	at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
[...]

PR #2377, opened 2021-04-21
Fix potential classloader deadlock in Either, Validated, Option, Ior, and Eval (#2375)

dberg:


nomisRev:
Looks great @dberg! 👏 
We need to do the same for `Eval.kt`, `Ior.kt`, `Option.kt`, and then I think this is good to go 👍

dberg:
@nomisRev, I've applied the changes to Eval, Ior, and Option as well. They should all follow the same pattern now. I was also able to reproduce the deadlock prior to the changes in all classes on my local machine.

rachelcarmena:
Welcome to Arrow @dberg !! :raised_hands:

nomisRev (approved):
Looks perfect @dberg!! Thank you for your contribution, and congrats on your first contribution in Arrow 🥳 Welcome to the Arrow 😁

danimontoya (approved):
Awesome! thanks @dberg for your contribution 🙌 🚀

PR #2387, opened 2021-04-28
Configuration: disable parallel execution when publishing

rachelcarmena:
It's the reason why multiple staging repositories are created when publishing.

It also reverts #2385.


47degdev:
Task linked: [CU-hz3z3q Fix the multiple staging repositories creation when releasing](https://app.clickup.com/t/hz3z3q)

PR #2392, opened 2021-05-05
Remove Gradle error from newer JDK versions

### LINKED ISSUE #2391: [Investigation] Allow to use JDK 9 and later
I'm not entirely sure of what is required but limiting the JDK to 9 may come with problems in the near future. Jetpack Compose and AGP 7.0.0 require JDK 11, so it's not possible to have a project using Compose at the same time as this project on the same IDE.

pablisco:
Closes #2391 

This PR potentially removes the requirement to have JDK 8 installed in our machines 🤞


rachelcarmena:
Thanks @pablisco !! It should be checked with newer JDK versions because it's being checked with JDK 8 here. I'll review it asap!

arrow-libs/core/arrow-meta/build.gradle
```diff
@@ -30,8 +30,7 @@ dependencies {
     testImplementation "io.kotlintest:kotlintest-runner-junit4:$KOTLIN_TEST_VERSION", excludeArrow
     def toolsJar = Jvm.current().getToolsJar()
 
-    if (!toolsJar)
-        throw new GradleException("tools.jar not found at your JAVA_HOME dir ${Jvm.current().getJavaHome().getAbsolutePath()}.\n" +
-                "Building with a JRE or JDK9 is currently not supported.")
-    testCompileOnly files(toolsJar)
+    if (toolsJar) {
+        testCompileOnly files(toolsJar)
```
rachelcarmena, line 34: Meanwhile you could check if it's possible to remove this dependency for testing to have the same behavior with JDK8 or higher.

pablisco: Just tested `clean build` with JDK 8 and it doesn't seem to need it :) Removing it: a353f73

rachelcarmena: Great, thanks @pablisco !! :clap: I'd miss 2 tasks:

- Just JDK8 is checked here so I miss adding checks for other JDKs.
- Update the documentation.

I'll help you asap!!

.github/workflows/build.yml
```diff
@@ -12,8 +12,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
-        java: [8, 8.0.192, 11, 11.0.2, 13, 13.0.4, 15, 16-ea]
+        os: [macos-latest]
+        java: [8, 11, 13, 15, 16-ea]
```
magneticflux-, line 16: Java 16 is GA now IIRC, so I think this should be updated to remove the `-ea` part. Also, `actions/setup-java@v2` is [available now](https://github.com/actions/setup-java/blob/main/docs/switching-to-v2.md).

pablisco: Good shout :bow:  Updating.

arrow-libs/core/arrow-meta/src/main/java/arrow/meta/encoder/jvm/TypeElementEncoder.kt
```diff
@@ -430,6 +430,7 @@ interface TypeElementEncoder : KotlinMetatadataEncoder, KotlinPoetEncoder, Proce
       Modifier.SYNCHRONIZED -> null
       Modifier.NATIVE -> null
       Modifier.STRICTFP -> null
+      else -> null
```
pablisco, line 433: This could be handled move gracefully depending on the version but it's out of the scope of this ticket. Happy to add a new one to look at ways to support newer modifiers.
cc: @raulraja

pablisco: This is regarding SEALED and NON\_SEALED modifiers introduced in JDK 15

arrow-libs/core/arrow-meta/build.gradle
pablisco: @rachelcarmena I've updated the documentation (with the JDK 16 caveat) and updated the build tests to test on multiple JDK versions

.github/workflows/build.yml
```diff
@@ -9,17 +9,23 @@ jobs:
       BASEDIR: ${{github.workspace}}/arrow-libs
       JAVA_OPTS: -Xms512m -Xmx1024m
 
-    runs-on: macos-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest]
+        java: [8, 11, 13, 15]
+    name: Build libraries with JDK ${{ matrix.java }} on ${{ matrix.os }}
```
nomisRev, line 17: Wondering if we want to build on all these versions. I was thinking about splitting github actions to build/test jvm and javascript in parallel. Since the build went up considerable now we added MPP.

Before 4-5 minutes for a build, now ~20 minuntes.
But for every MPP library we generate 4 JS binaries  + JVM, and soon native which has a lot of targets as well.

How does this impact CI time? How many Github Actions can we run in parallel? 🤔

pablisco: I think they are run in Parallel as far as I can tell, so we should be safe :crossed\_fingers:

pablisco: Just had to check and it seems that way :partying\_face:

nomisRev:
@pablisco looking at the times every task takes we're running JS tests on all 4 of those actions.
Can we split off JS, so it runs in a separate action? I think that will bring down the time of every action back to ~10 minutes.

pablisco:
> @pablisco looking at the times every task takes we're running JS tests on all 4 of those actions.
> Can we split off JS, so it runs in a separate action? I think that will bring down the time of every action back to ~10 minutes.

This should do the trick 🤞 
https://github.com/arrow-kt/arrow/pull/2392/commits/0129c2697d9833b6c5f4e68740628d52820bba4a

nomisRev (approved):
LGTM. Thanks for fixing this issue @pablisco. I was running  into  this every time I open Arrow since I work with JDK 11 😅

arrow-libs/core/arrow-meta/src/main/java/arrow/meta/encoder/jvm/TypeElementEncoder.kt
raulraja: whatever works, most of this code in the old meta part is going away soon and the new meta does not depend on the TypeeElement java apis.

raulraja (approved):
thanks @pablisco !

PR #2397, opened 2021-05-09
add common extensions for option

### LINKED ISSUE #2374: ["Request"] Either.toOption() is gone so code completion sugests A?.toOption()
In the latest Arrow release it seems [`Either.toOption()`](https://github.com/arrow-kt/arrow-core/blob/master/arrow-core-data/src/main/kotlin/arrow/core/Either.kt#L945) was deprecated and removed, so after upgrading the version, the only function suggested is [`A?.toOption()`](https://github.com/arrow-kt/arrow-core/blob/master/arrow-core-data/src/main/kotlin/arrow/core/Option.kt#L998).

Before the problem **was** that the two functions can be easily mixed up as the option one is available for all types, including Either. The problem now is that there's no handy way anymore of converting an Either to an Option.

My suggestion would be to bring back `Either.toOption()` and rename the option one to `A?.asOption()` or something similar, to avoid confusions due to an import.

myuwono:
### In this PR
- added strong `Option<A>.pairLeft(l: L): Option<Pair<L, A>>` and `Option<A>.pairRight(r: R): Option<Pair<A, R>>`
- added `List<Option<A>>.flattenOption(): List<A>`
- added `Map<K, Option<V>>.filterOption(): Map<K, V>` 
- added `Sequence<Option<A>>.filterOption(): Sequence<A>` 
- added `F<A>.traverseOption(f: (A) -> Option<B>): Option<F<B>>` and `F<Option<T>>.sequence(): Option<F<T>>` for either, option, validated, ior, lists, nonemptylists, sequence, and maps.
- added some missing tests for Traverse

fixes #2374

### DIFF
--- a/arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Ior.kt
+++ b/arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Ior.kt
@@ -516,7 +516,11 @@ sealed class Ior<out A, out B> {
     )
 
   inline fun <C> traverseOption(fa: (B) -> Option<C>): Option<Ior<A, C>> =
-    traverseEither { fa(it).toEither { Unit } }.orNone()
+    fold(
+      { a -> Some(Left(a)) },
+      { b -> fa(b).map { Right(it) } },
+      { a, b -> fa(b).map { Both(a, it) } }
+    )
 
   inline fun <AA, C> traverseValidated(fa: (B) -> Validated<AA, C>): Validated<AA, Ior<A, C>> =
     fold(
--- a/arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Ior.kt
+++ b/arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Ior.kt
@@ -428,6 +428,15 @@ sealed class Ior<out A, out B> {
       { a, b -> fa(a).zip(fb(b)) { aa, c -> Both(aa, c) } }
     )
 
+  inline fun <C, D> bitraverseOption(
+    fa: (A) -> Option<C>,
+    fb: (B) -> Option<D>
+  ): Option<Ior<C, D>> =
+    bitraverseEither(
+      { fa(it).toEither { Unit } },
+      { fb(it).toEither { Unit } }
+    ).orNone()
+
   inline fun <AA, C, D> bitraverseValidated(
     SA: Semigroup<AA>,
     fa: (A) -> Validated<AA, C>,
@@ -505,6 +514,9 @@ sealed class Ior<out A, out B> {
       { a, b -> fa(b).map { Both(a, it) } }
     )
 
+  inline fun <C> traverseOption(fa: (B) -> Option<C>): Option<Ior<A, C>> =
+    traverseEither { fa(it).toEither { Unit } }.orNone()
+
   inline fun <AA, C> traverseValidated(fa: (B) -> Validated<AA, C>): Validated<AA, Ior<A, C>> =
     fold(
       { a -> Valid(Left(a)) },
@@ -548,6 +560,9 @@ fun <A, B> Ior<Iterable<A>, Iterable<B>>.bisequence(): List<Ior<A, B>> =
 
 fun <A, B, C> Ior<Either<A, B>, Either<A, C>>.bisequenceEither(): Either<A, Ior<B, C>> =
   bitraverseEither(::identity, ::identity)
+
+fun <B, C> Ior<Option<B>, Option<C>>.bisequenceOption(): Option<Ior<B, C>> =
+  bitraverseOption(::identity, ::identity)
 
 fun <A, B, C> Ior<Validated<A, B>, Validated<A, C>>.bisequenceValidated(SA: Semigroup<A>): Validated<A, Ior<B, C>> =
   bitraverseValidated(SA, ::identity, ::identity)
@@ -606,6 +621,9 @@ fun <A, B> Ior<A, Iterable<B>>.sequence(): List<Ior<A, B>> =
 
 fun <A, B, C> Ior<A, Either<B, C>>.sequenceEither(): Either<B, Ior<A, C>> =
   traverseEither(::identity)
+
+fun <A, B> Ior<A, Option<B>>.sequenceOption(): Option<Ior<A, B>> =
+  traverseOption(::identity)
 
 fun <A, B, C> Ior<A, Validated<B, C>>.sequenceValidated(): Validated<B, Ior<A, C>> =
   traverseValidated(::identity)
--- a/arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Ior.kt
+++ b/arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Ior.kt
@@ -432,10 +432,11 @@ sealed class Ior<out A, out B> {
     fa: (A) -> Option<C>,
     fb: (B) -> Option<D>
   ): Option<Ior<C, D>> =
-    bitraverseEither(
-      { fa(it).toEither { Unit } },
-      { fb(it).toEither { Unit } }
-    ).orNone()
+    fold(
+      { fa(it).map { Left(it) } },
+      { fb(it).map { Right(it) } },
+      { a, b -> fa(a).zip(fb(b)) { aa, c -> Both(aa, c) } }
+    )
 
   inline fun <AA, C, D> bitraverseValidated(
     SA: Semigroup<AA>,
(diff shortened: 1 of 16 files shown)


arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Option.kt
```diff
@@ -357,22 +357,52 @@ sealed class Option<out A> {
...
+      Some.unit,
+      Some.unit
+    ) { b, c, _, _, _, _, _, _, _, _ -> map(b, c) }
 
   inline fun <B, C, D> zip(
     b: Option<B>,
     c: Option<C>,
     map: (A, B, C) -> D
   ): Option<D> =
-    zip(b, c, Some.unit, Some.unit, Some.unit, Some.unit, Some.unit, Some.unit, Some.unit) { b, c, d, _, _, _, _, _, _, _ -> map(b, c, d) }
```
raulraja, line 367: Looks like these changes are just new formatting? I don't have a preference but just thought I'd mention it

raulraja (approved):
It looks great, @myuwono! Thanks so much for taking care of this and making it consistent. I've noticed we added the bifunctor methods like bitraverse, but they have no test associated; perhaps adding a couple more tests for those will help as we are about to turn codecov. Someone else would eventually have to add it.

Thanks again!

myuwono:
thanks @raulraja yeah indeed it did look like those codes were added without tests. I've taken care of those as well.

arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Option.kt
myuwono: yeah, it's only reformatting. it kept doing that everytime i did a reformat code through intellij. My intellij seems to have picked up everything in editor config but the line length settings..

arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Ior.kt
```diff
@@ -428,6 +428,15 @@ sealed class Ior<out A, out B> {
...
     )
 
+  inline fun <C, D> bitraverseOption(
+    fa: (A) -> Option<C>,
+    fb: (B) -> Option<D>
+  ): Option<Ior<C, D>> =
+    bitraverseEither(
+      { fa(it).toEither { Unit } },
+      { fb(it).toEither { Unit } }
+    ).orNone()
```
nomisRev, line 438:
```suggestion
    fold(
      { fa(it).map { Left(it) } },
      { fb(it).map { Right(it) } },
      { a, b -> fa(a).zip(fb(b)) { aa, c -> Both(aa, c) } }
    )
```

Nit: a direct implementation reads better IMO, and is (slightly) more performant.

arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Ior.kt
```diff
@@ -505,6 +514,9 @@ sealed class Ior<out A, out B> {
       { a, b -> fa(b).map { Both(a, it) } }
     )
 
+  inline fun <C> traverseOption(fa: (B) -> Option<C>): Option<Ior<A, C>> =
+    traverseEither { fa(it).toEither { Unit } }.orNone()
```
nomisRev, line 518:
```suggestion
    fold(
      { a -> Some(Left(a)) },
      { b -> fa(b).map { Right(it) } },
      { a, b -> fa(b).map { Both(a, it) } }
    )
```

Nit: a direct implementation reads better IMO, and is (slightly) more performant.

arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Iterable.kt
```diff
@@ -329,6 +329,12 @@ fun <E, A> Iterable<Validated<E, A>>.sequenceValidated(semigroup: Semigroup<E>):
 fun <E, A> Iterable<ValidatedNel<E, A>>.sequenceValidated(): ValidatedNel<E, List<A>> =
   traverseValidated(Semigroup.nonEmptyList(), ::identity)
 
+inline fun <A, B> Iterable<A>.traverseOption(f: (A) -> Option<B>): Option<List<B>> =
+  traverseEither { f(it).toEither { Unit } }.orNone()
```
nomisRev, line 333:
```suggestion
inline fun <A, B> Iterable<A>.traverseOption(f: (A) -> Option<B>): Option<List<B>> {
  val acc = mutableListOf<B>()
  forEach { a ->
    when (val res = f(a)) {
      is Some -> acc.add(res.value)
      None -> return@traverseOption res
    }
  }
  return acc.some()
}
```

Nit: a direct implementation reads better IMO, and is (slightly) more performant.
That'll save 1 allocation per element in the `Iterable`.

arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -447,3 +447,9 @@ inline fun <E, A, B> NonEmptyList<A>.traverseValidated(
 
 fun <E, A> NonEmptyList<Validated<E, A>>.sequenceValidated(semigroup: Semigroup<E>): Validated<E, NonEmptyList<A>> =
   traverseValidated(semigroup, ::identity)
+
+inline fun <A, B> NonEmptyList<A>.traverseOption(f: (A) -> Option<B>): Option<NonEmptyList<B>> =
+  traverseEither { f(it).toEither { Unit } }.orNone()
```
nomisRev, line 452:
```suggestion
inline fun <A, B> NonEmptyList<A>.traverseOption(f: (A) -> Option<B>): Option<NonEmptyList<B>> {
  val acc = mutableListOf<B>()
  forEach { a ->
    when (val res = f(a)) {
      is Right -> acc.add(res.value)
      is Left -> return@traverseOption res
    }
  }
  // Safe due to traverse laws
  return NonEmptyList.fromListUnsafe(acc).right()
}
```

Nit: a direct implementation reads better IMO, and is (slightly) more performant.
That'll save 1 allocation per element in the NonEmptyList.

arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/Sequence.kt
```diff
@@ -659,6 +662,9 @@ fun <E, A, B> Sequence<A>.traverseEither(f: (A) -> Either<E, B>): Either<E, Sequ
   return acc.asSequence().right()
 }
 
+fun <A, B> Sequence<A>.traverseOption(f: (A) -> Option<B>): Option<Sequence<B>> =
+  traverseEither { f(it).toEither { Unit } }.orNone()
```
nomisRev, line 666:
```suggestion
fun <A, B> Sequence<A>.traverseOption(f: (A) -> Option<B>): Option<Sequence<B>> {
  // Note: Using a mutable list here avoids the stackoverflows one can accidentally create when using
  //  Sequence.plus instead. But we don't convert the sequence to a list beforehand to avoid
  //  forcing too much of the sequence to be evaluated.
  val acc = mutableListOf<B>()
  forEach { a ->
    when (val res = f(a)) {
      is Some -> acc.add(res.value)
      None -> return@traverseOption res
    }
  }
  return acc.asSequence().some()
}
```

Nit: a direct implementation reads better IMO, and is (slightly) more performant.
That'll save 1 allocation per element in the Sequence.

arrow-libs/core/arrow-core/src/main/kotlin/arrow/core/map.kt
```diff
@@ -224,6 +224,12 @@ inline fun <K, E, A, B> Map<K, A>.traverseValidated(
 fun <K, E, A> Map<K, Validated<E, A>>.sequenceValidated(semigroup: Semigroup<E>): Validated<E, Map<K, A>> =
   traverseValidated(semigroup, ::identity)
 
+inline fun <K, A, B> Map<K, A>.traverseOption(f: (A) -> Option<B>): Option<Map<K, B>> =
+  traverseEither { f(it).toEither { Unit } }.orNone()
```
nomisRev, line 228:
```suggestion
inline fun <K, A, B> Map<K, A>.traverseOption(f: (A) -> Option<B>): Option<Map<K, B>> {
  val acc = mutableMapOf<K, B>()
  forEach { (k, v) ->
    when (val res = f(v)) {
      is Some -> acc[k] = res.value
      None -> return@traverseOption res
    }
  }
  return acc.some()
}
```


Nit: a direct implementation reads better IMO, and is (slightly) more performant.
That'll save 1 allocation per value pair in the map.

myuwono:
@nomisRev I've used direct implementation for `traverseOption` as suggested, it's definitely much nicer.

rachelcarmena:
Welcome to this repository as well @myuwono and thanks!! :tada:

PR #2402, opened 2021-05-19
Initial step for Arrow multiplatform

rachelcarmena:
- Transform some libraries to use `org.jetbrains.kotlin.multiplatform`
- First step: moving all the source code to the `jvm` target
- Include configuration for publishing the transformed libraries


raulraja (approved):
🚀 thanks @rachelcarmena !

PR #2404, opened 2021-05-20
Review on sv-arrow-mpp branch

rachelcarmena:
@nomisRev, it includes my suggestions and the fix for the found issue with `kotlin.stdlib.default.dependency` and `js` target.


nomisRev (approved):
Thank you so much for looking into this @rachelcarmena !! 👏 👏 👏 🙌

PR #2407, opened 2021-05-29
Profunctor optics

1Jajen1:
Optics with a different likely more powerful encoding.

This is pretty complete feature wise, however it is not well tested, documented and benchmarked. My thoughts behind opening this pr now are to gather feedback on the api + its internals before I get onto further testing/benchmarking and documenting,

As for how it all works: There is quite a bit of documentation on the `Optic` interface and the `Profunctor` typeclass which should hopefully help to understand the inner workings.

I also had to introduce `combineLazy` to monoids simply because I have some monoid instances that benefit from laziness (such as `Monoid.first`. Laziness is decided at the point of invoking a terminal operator and isn't usually a user decision, for example `firstOrNull` uses eval to avoid too much work on traversals, but `sum` will use a non-eval traversal which should be faster. This boils down to `traverseLazy/traverse` and `foldLazy/fold` (plus their indexed variants).

This implementation comes from the haskell `optics` package, tho has seen some changes, since kotlin makes variance a lot easier.

The `ixcompose` method for `Optic` also has 13 type parameters which I believe could be a new arrow record ignoring tuple/zipN definitions! :laughing: 

See https://github.com/arrow-kt/arrow/blob/jo-profunctor-optics/arrow-libs/optics/arrow-optics-profunctor/src/jvmMain/kotlin/test.kt for some examples of the api.


serras:
It's been more than a year since the last commit here; is it OK to close this PR?

1Jajen1:
> It's been more than a year since the last commit here; is it OK to close this PR?

Yeah, afaik we have a different optics implementation anyway. Imo profunctor optics are really really nice to work with, but that's in haskell, where they play nice with type inference and have decent performance.

arrow-libs/optics/arrow-optics-profunctor/src/commonMain/kotlin/arrow/optics/Optic.kt
```diff
@@ -0,0 +1,218 @@
...
+  override fun <P, J> Profunctor<P>.transform(focus: Pro<P, J, C, D>): Pro<P, (I1) -> J, S, T> {
+    val pab = other.run { transform(focus) }.ixMap<J, (I2) -> J, A2, B2> { j -> { j } }
+    return this@compose.run { transform(pab) }
+  }
+}
+
+fun <K1, K2 : K1, I1, I2, I3, S, T, A1 : A2, A2, B1, B2 : B1, C, D> Optic<K1, I1, S, T, A1, B1>.ixCompose(
+  other: Optic<K2, I2, A2, B2, C, D>,
+  f: (I1, I2) -> I3
+): Optic<K1, I3, S, T, C, D> = object : Optic<K1, I3, S, T, C, D> {
```
1Jajen1, line 186: Tho I do miss seeing these type signatures :D

PR #2408, opened 2021-05-30
Add partiallyX for arity over 5 and suspend functions

dnowak:
Added missing `partiallyX` for arity over 5 and suspending functions.

It is hard to define a limit for partial application as a use case may require several parameters -> I decided to add all possibilities.


raulraja (approved):
Thanks @dnowak ! Looks great!

nomisRev (approved):
Thank you @dnowak!

PR #2415, opened 2021-06-05
Specify type variance for `compose` and `plus` methods across the lenses

### LINKED ISSUE #2414: ["Request"] Variance for generic types in compositions

Arrow Optics `0.13.2`

What would you like to see?

Variant generic types in `compose` functions. Due to limitations of the Kotlin language, these type variants can't be applied to the type as a whole. This is a compromise that should cover the main use-case.

Test case:

```kotlin
interface MyInterface {
    val myValue: String
}

// 2 or more subclasses
class Implentation1(override val myValue: String) : MyInterface
class Implentation2(override val myValue: String) : MyInterface

val getter = Getter<MyInterface, String> { it.myValue }

val myOtherLens: Lens<String, Implentation1> = TODO("external to the problem")

val composition = myOtherLens + getter // Error!
```

magneticflux-:
These all must be specified, otherwise the compiler is unable to determine which function to call when there are multiple (e.g. when using `Getter compose Getter` to create a `Getter` instead of a `Fold`).

Closes #2414

### DIFF
--- a/arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Every.kt
+++ b/arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Every.kt
@@ -33,7 +33,7 @@ interface PEvery<S, T, A, B> : PTraversal<S, T, A, B>, Fold<S, A>, PSetter<S, T,
   /**
    * Compose a [PEvery] with a [PEvery]
    */
-  infix fun <C, D> compose(other: PEvery<A, B, C, D>): PEvery<S, T, C, D> =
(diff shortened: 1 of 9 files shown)


magneticflux-:
I was able to build the documentation on my own machine and nothing in the docs was changed by this PR, so this seems like a spurious failure.

nomisRev:
There seems to be something randomly failing after bumping to Kotlin 1.5.10 at the start-up of the documentation checking tools. Re-running the CI task seems to work...

nomisRev (approved):
Looks great to me!! Thanks @magneticflux- 👏 👏

nomisRev:
Thanks for your first contribution @magneticflux- 🥳

PR #2436, opened 2021-06-25
Change implementation of `Option`-related ext. functions

AzimMuradov:
# Changes have been made to achieve consistency with the Iterable<T?> types.

Prior to this, `firstOrNone` and other modified functions returned `None` if the element was `null`.

### For example:

```kotlin
import arrow.core.elementAtOrNone
import arrow.core.firstOrNone
import arrow.core.lastOrNone

fun main() {
    val list1 = listOf(1, 2, 3, null, 5)
    val list2 = listOf(null, 2, 3, 4, 5)
    val list3 = listOf(1, 2, 3, 4, null)

    println(list1.elementAtOrNone(3))
    println(list1.firstOrNone { it == null })
    println(list2.firstOrNone())
    println(list3.lastOrNone())
}
```

##### Before:

```
Option.None
Option.None
Option.None
Option.None
```

##### Now:

```
Option.Some(null)
Option.Some(null)
Option.Some(null)
Option.Some(null)
```

### Changed functions:

- `fun <T> Iterable<T>.firstOrNone(): Option<T>`
- `inline fun <T> Iterable<T>.firstOrNone(predicate: (T) -> Boolean): Option<T>`
- `fun <T> Iterable<T>.singleOrNone(): Option<T>`
- `inline fun <T> Iterable<T>.singleOrNone(predicate: (T) -> Boolean): Option<T>`
- `fun <T> Iterable<T>.lastOrNone(): Option<T>`
- `fun <T> Iterable<T>.lastOrNone(predicate: (T) -> Boolean): Option<T>`
- `fun <T> Iterable<T>.elementAtOrNone(index: Int): Option<T>`

## EDIT

New commits summary:
 - rewrite functions, so they conform to style guides
 - add docs and tests to them
 - found 2 more functions (`align` and `filterIsInstance`) that have the same issues and fix them


nomisRev:
Hey @AzimMuradov,

You're correct in these cases it should be `Option.Some(null)` since there is a value present, and you don't want it to get flattened into `?` which is the whole point of using `Option`. 👍

Nice catch, and thank you for the contribution! 🙌

Would be great if we could also add some tests for this, so we never fall into any accidental regression.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -812,19 +812,92 @@ fun <A> Iterable<Option<A>>.combineAll(MA: Monoid<A>): Option<A> =
...
+  if (this is List) return if (index in 0..lastIndex) Some(get(index)) else None
+  if (index < 0) return None
+  val iterator = iterator()
+  var count = 0
+  while (iterator.hasNext()) {
+    val element = iterator.next()
+    if (index == count++) return Some(element)
+  }
+  return None
+}
```
pablisco, line 900:
```suggestion
fun <T> Iterable<T>.elementAtOrNone(index: Int): Option<T> =
  when(this) {
    is Collection -> elementAtOrNone(index)
    else -> iterator().skip(index - 1).nextOrNone()
  }

fun <T> Collection<T>.elementAtOrNone(index: Int): Option<T> =
  when {
    size >= index -> Some(elementAt(index))
    else -> None
  }

tailrec fun <T> Iterator<T>.skip(count: Int): Iterator<T> =
  when {
    count > 0 && hasNext() -> {
      next()
      skip(count - 1)
    }
    else -> this
  }
```

pablisco (commented):
Great idea to improve the contracts for nullable iterables 💯 

The suggested solutions however may not be following the style we have elsewhere in the project. Generally, we try to:

- Aim for small functions that can be composed together
- Use expression bodies instead of mid function returns

I've added some suggestions as a starting point. They are not tested and potentially not 100% correct but should get you in the right direction after some testing is in place 😃 

Thank you for the contribution, this is super helpful 🙂

AzimMuradov:
> Great idea to improve the contracts for nullable iterables 
> 
> The suggested solutions however may not be following the style we have elsewhere in the project. Generally, we try to:
> 
> * Aim for small functions that can be composed together
> * Use expression bodies instead of mid function returns
> 
> I've added some suggestions as a starting point. They are not tested and potentially not 100% correct but should get you in the right direction after some testing is in place 
> 
> Thank you for the contribution, this is super helpful

Thanks for support and helpful advices!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -812,19 +812,92 @@ fun <A> Iterable<Option<A>>.combineAll(MA: Monoid<A>): Option<A> =
...
+      return if (isEmpty()) None else Some(this[0])
+    }
+    else -> {
+      val iterator = iterator()
+      if (!iterator.hasNext())
+        return None
+      return Some(iterator.next())
+    }
+  }
+}
```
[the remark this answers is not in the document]
AzimMuradov: I think this solution is much better, thanks!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -812,19 +812,92 @@ fun <A> Iterable<Option<A>>.combineAll(MA: Monoid<A>): Option<A> =
...
 
-inline fun <T> Iterable<T>.firstOrNone(predicate: (T) -> Boolean): Option<T> = this.firstOrNull(predicate).toOption()
+inline fun <T> Iterable<T>.firstOrNone(predicate: (T) -> Boolean): Option<T> {
+  for (element in this) if (predicate(element)) return Some(element)
+  return None
+}
```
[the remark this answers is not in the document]
AzimMuradov: I agree about `tailrec`, but I think using `nextOrNone()` in a loop would be too much overhead, so I tried to rewrite it a bit differently using `hasNext()` function

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -812,19 +812,92 @@ fun <A> Iterable<Option<A>>.combineAll(MA: Monoid<A>): Option<A> =
...
+      val iterator = iterator()
+      if (!iterator.hasNext())
+        return None
+      var last = iterator.next()
+      while (iterator.hasNext())
+        last = iterator.next()
+      return Some(last)
+    }
+  }
+}
```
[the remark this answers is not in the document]
AzimMuradov: I think that solution might be too slow.

How about solution below?

```kotlin
fun <T> Iterable<T>.lastOrNone(): Option<T> =
  when (this) {
    is Collection -> lastOrNone()
    else -> iterator().run {
      if (hasNext()) {
        var last: T
        do last = next() while (hasNext())
        Some(last)
      } else {
        None
      }
    }
  }
```

It's still using expression body, and it's quite compact.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -812,19 +812,92 @@ fun <A> Iterable<Option<A>>.combineAll(MA: Monoid<A>): Option<A> =
...
+  var found = false
+  val list = mutableListOf<T>()
+  for (element in this) {
+    if (predicate(element)) {
+      if (found) list[0] = element else list.add(element)
+      found = true
+    }
+  }
+  return if (found) Some(list[0]) else None
+}
```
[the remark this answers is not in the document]
AzimMuradov: It's a good solution, but I think it could potentially be slow. Besides, this function has only one return, and I think it doesn't need so much simplification and separation.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
AzimMuradov: Thanks for help!

AzimMuradov:
> Hey @AzimMuradov,
> 
> You're correct in these cases it should be `Option.Some(null)` since there is a value present, and you don't want it to get flattened into `?` which is the whole point of using `Option`. 
> 
> Nice catch, and thank you for the contribution! 
> 
> Would be great if we could also add some tests for this, so we never fall into any accidental regression.

Thanks for support!
In the new commits, I've also added tests for these functions (especially considering `T?` cases).

raulraja:
If possible these functions regardless of impl should be `inline` so they work inside `suspend fun`, thanks @AzimMuradov !

AzimMuradov:
> If possible these functions regardless of impl should be `inline` so they work inside `suspend fun`, thanks @AzimMuradov !

Can I remove the recursion and make the functions a little non-idiomatic?

For example:

```kotlin
inline fun <T> Iterable<T>.firstOrNone(predicate: (T) -> Boolean): Option<T> {
  for (element in this) if (predicate(element)) return Some(element)
  return None
}
```

nomisRev:
Hey @AzimMuradov,

The function implementation you shared is not considered `non-idiomatic`. I'd say it's considered idiomatic in Kotlin, and thus it's also idiomatic for implementation in Arrow.

Whilst in user code we typically prefer using functions such as `firstOrNone` rather than writing loops, writing the base implementation in Arrow using loops is encouraged rather than discouraged.
It leads to simple, Kotlin idiomatic code that everyone can understand if they try to read the implementation which makes it easy to understand and maintain. In addition to it being very efficient.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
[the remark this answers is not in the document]
pablisco: Yeah, you are right. Until we can do Union types or generic value classes using Some comes with some overhead :100:

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/OptionTest.kt
```diff
@@ -143,10 +178,141 @@ class OptionTest : UnitSpec() {
...
+        override fun iterator(): Iterator<Int> = list.iterator()
+      }
+      iterable.firstOrNone() shouldBe Some(1)
+      iterable.firstOrNone { it > 2 } shouldBe Some(3)
+      iterable.firstOrNone { it > 7 } shouldBe None
+
+      val emptyIterable = object : Iterable<Int> {
+        private val emptyList = emptyList<Int>()
+        override fun iterator(): Iterator<Int> = emptyList.iterator()
+      }
```
pablisco, line 193:
```suggestion
      val emptyIterable = Iterable { iterator { yieldAll(listOf(1, 2, 3, 4, 5, 6)) } }
```

These and other interators can be defined like this :point\_up\_2:   
  
Alternatively, we can also define a file function like this:

```kotlin
fun <T> nonCollectionOf(vararg items: T): Iterable<T> = Iterable { iterator { yieldAll(items.toList()) } }
```

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/OptionTest.kt
```diff
@@ -143,10 +178,141 @@ class OptionTest : UnitSpec() {
       none.toList() shouldBe listOf()
     }
 
-    "firstOption" {
-      val l = listOf(1, 2, 3, 4, 5, 6)
-      l.firstOrNone() shouldBe Some(1)
-      l.firstOrNone { it > 2 } shouldBe Some(3)
+    "Iterable.firstOrNone" {
```
pablisco, line 181: Amazing work with the tests btw :slightly\_smiling\_face: Really appreciate it :pray:

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -812,19 +828,149 @@ fun <A> Iterable<Option<A>>.combineAll(MA: Monoid<A>): Option<A> =
...
+ * Returns an element as [Some(element)][Some] at the given [index] or [None] if the [index] is out of bounds of this collection.
+ */
+fun <T> Collection<T>.elementAtOrNone(index: Int): Option<T> =
+  when (index) {
+    in 0 until size -> Some(elementAt(index))
+    else -> None
+  }
 
-fun <T> Iterable<T>.elementAtOrNone(index: Int): Option<T> = this.elementAtOrNull(index).toOption()
+private tailrec fun <T> Iterator<T>.skip(count: Int): Iterator<T> =
```
pablisco, line 966: Not a blocker, but is it worth making skip public? Could be handy for someone. :slightly\_smiling\_face:

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/OptionTest.kt
AzimMuradov: Definitely going to use this, really helpful :+1:

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
AzimMuradov: I think it looks more like a utility function. Also, I don't think it's a good idea to add extension functions to the `Iterator` (as a public API), at least I haven't seen cases like this. :thinking:

pablisco: Actually, if we call it `drop(n)` it would match the one for Iterables in `stdlib`:  
[https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/drop.html](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/drop.html)

I think there is a use case to do things like: `iterator().drop(3).nextOrNone()` but it's a new function so probably out of the scope of this PR :)

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -812,19 +828,149 @@ fun <A> Iterable<Option<A>>.combineAll(MA: Monoid<A>): Option<A> =
     acc.combine(MA, a)
   }
 
-fun <T> Iterable<T>.firstOrNone(): Option<T> = this.firstOrNull().toOption()
+/**
+ * Returns the first element as [Some(element)][Some], or [None] if the iterable is empty.
+ */
+fun <T> Iterable<T>.firstOrNone(): Option<T> =
```
nomisRev, line 834: All extensions on `Iterable` are in `Iterable.kt` so they're all grouped under `arrow.core.IterableKt` from Java.

nomisRev (commented):
This PR is looking great @AzimMuradov! Thank you so much for the contribution 🙌 🥳

If we move the extension on `Iterabe.kt` I think this is ready to be merged.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -812,19 +828,149 @@ fun <A> Iterable<Option<A>>.combineAll(MA: Monoid<A>): Option<A> =
...
+fun <T> Iterable<T>.singleOrNone(): Option<T> =
+  when (this) {
+    is Collection -> singleOrNone()
+    else -> iterator().run { nextOrNone().filter { !hasNext() } }
+  }
+
+/**
+ * Returns single element as [Some(element)][Some], or [None] if the collection is empty or has more than one element.
+ */
+fun <T> Collection<T>.singleOrNone(): Option<T> =
```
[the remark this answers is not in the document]
pablisco: I suggested the smaller functions to avoid functions being too complex but guess it may make sense to have a single one of both show up for collections on autocomplete.

AzimMuradov:
> This PR is looking great @AzimMuradov! Thank you so much for the contribution  
> 
> If we move the extension on `Iterabe.kt` I think this is ready to be merged.

Thanks for support! :+1: 

When I was moving changed functions, I noticed that there are still one Iterable's ext function left:
`fun <A> Iterable<Option<A>>.combineAll(MA: Monoid<A>): Option<A>`. I didn't touch it because it was not part of the pr, and I don't know if it was intentional or not. So I decided to just tell :smile:

AzimMuradov:
I'm not sure, but I think there was a problem connecting to the server during the execution of the workflow. If this is really the problem, please restart it.

nomisRev:
@AzimMuradov I restarted the build. It was just an issue with Github Action not starting correctly.

> fun <A> Iterable<Option<A>>.combineAll(MA: Monoid<A>): Option<A>. I didn't touch it because it was not part of the pr, and I don't know if it was intentional or not. So I decided to just tell 😄

Oh, thanks for raising this 🙌 I think we can move this to `Iterable.kt` as well.

We should annotate the current `Iterable<Option<A>>.combineAll` function with:
`@Deprecated("This method has been moved to IterableKt. It's hidden for binary compatibility reasons", level= DeprecationLevel.HIDDEN)` and then we can move it to `Iterable.kt`.

AzimMuradov:
> @AzimMuradov I restarted the build. It was just an issue with Github Action not starting correctly.
> It's green already 
> 
> > fun  Iterable<Option>.combineAll(MA: Monoid): Option. I didn't touch it because it was not part of the pr, and I don't know if it was intentional or not. So I decided to just tell smile
> 
> Oh, thanks for raising this  I think we can move this to `Iterable.kt` as well.
> 
> We should annotate the current `Iterable<Option<A>>.combineAll` function with:
> `@Deprecated("This method has been moved to IterableKt. It's hidden for binary compatibility reasons", level= DeprecationLevel.HIDDEN)` and then we can move it to `Iterable.kt`.

Thanks for fast response! :raised_hands:

nomisRev:
@AzimMuradov we enabled explicit API mode. So you need to explicitly add `public` to the new methods to make your PR green. Sorry for the inconvenience.

nomisRev (approved):
Thanks for the awesome contribution @AzimMuradov !! 👏 👏 🥳

pablisco (approved):
Amazing stuff 💯  thanks for the contribution 🙂

PR #2440, opened 2021-07-05
Deprecate Float & Double comparison operators

nomisRev:
Due to the way that floating numbers work in Kotlin MPP, we cannot support comparison operators for them (`Float`, `Double`).

https://kotlinlang.org/docs/equality.html#floating-point-numbers-equality


i-walker (approved):
👍🏾

raulraja (approved):
Thanks @nomisRev !

Issue #2311, opened 2021-03-09
Add require() and error() functions


CLOVIS-AI:
Standard Kotlin makes heavy use of the `require`, `check` and `error` functions to shortcut functions.

These could be brought to the `either` block.

Where the idea was first proposed:
https://kotlinlang.slack.com/archives/C5UPMM0A0/p1615240882105000?thread_ts=1615240882.105000&cid=C5UPMM0A0

nomisRev:
> Thinking about it more, the arrow version should definitely not be called require, because otherwise, it's way too easy to import the wrong version (and good luck finding that...).

Actually, this should not cause any issues. If they're defined on `EitherEffect` directly inside Arrow then there would be no way to accidentally import the wrong `require` unless you explicitly use `kotlin.require(..) { }` inside the `either { }`.
Otherwise, it would require you to rename `import kotlin.require as KotlinRequire`.

CLOVIS-AI:
Another thought: this would look like:

```kotlin
either {
  check(something) { MyErrorType }
  foo(something).bind()
}
```

Currently it's clear what can shortcut and what can't (stuff that shortcuts always has `.bind()` or previously `!`). Maybe `check` is not explicit enough about this?

nomisRev:
Then I think exposing something like `fun <E> Validated.Companion.check(boolean: Boolean, f: () -> E): Validated<E, Unit>` is also enough since then you can do.

```kotlin
either {
  Validated.check(something) { MyErrorType }.bind()
  foo(something).bind()
}
```

raulraja:
@CLOVIS-AI closing this as it seems the main use case was addressed in #2449 , thanks!

CLOVIS-AI:
Yep it does, thanks a lot! That's going to make imperative-style code even more compatible with the benefits of pure functions

PR #2449, opened 2021-07-19
Add ensure, and ensureNotNull in computation DSL

nomisRev:
This PR adds some convenient syntax for mixing the `nullable` effect/monad with other effect/monads such as `either`, `option` and `nullable` itself.

It also deprecates the currently gotcha behavior of nullable `bind` syntax in `option { }`.
Example of gotcha behavior:

```kotlin
either<Int, String> {
  option<Int> {
     Either.Left(1).bind() // Expected to short-circuit but uses `fun <A> A?.bind(): A` instead and *doesnt* short-circuit,
     
  }.fold({ "0" }, Int::toString)
}
```

This was originally proposed in #2311.
The proposed `require`/`requireNotNull` syntax conflicts with with `kotlin.require` and `kotlin.requireNotNull` which is auto-imported and it can be easily confused. If it's mistake it leads to an exception instead of participating in the computation block.

### DIFF
--- a/arrow-libs/core/arrow-continuations/src/commonTest/kotlin/generic/SuspendingComputationTest.kt
+++ b/arrow-libs/core/arrow-continuations/src/commonTest/kotlin/generic/SuspendingComputationTest.kt
@@ -143,11 +143,13 @@ class SuspendingComputationTest : StringSpec({
     fun square(i: Int): Int = i * i
 
     checkAll(Arb.int().orNull(), Arb.string()) { i: Int?, lValue: String ->
-      either<String, Int> {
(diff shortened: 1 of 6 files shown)


franciscodr:
Task linked: [CU-p9935d Add special validation syntax like kotlin.require and kotlin.requireNotNull](https://app.clickup.com/t/p9935d)

nomisRev:
Closes https://github.com/arrow-kt/arrow/issues/2311

raulraja (approved):
Looks great @nomisRev !

i-walker (approved):
🙌🏾 Thanks Simon

PR #2453, opened 2021-07-25
Flow parMap & parMapUnordered

nomisRev:


arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/flow.kt
```diff
@@ -57,3 +71,91 @@ public fun <A, B> Flow<A>.retry(schedule: Schedule<Throwable, B>): Flow<A> = flo
     emit(it)
   }
 }
+
+/**
```
raulraja, line 75: Maybe we should add a menu entry under "types" or somewhere that points to the `Flow` extensions

raulraja (approved):
thanks @nomisRev !

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/flow.kt
nomisRev: I'm going to make some improvements to the Fx docs before M1.
I'm taking this together with brining back the intro doc we had when we originally released Fx.

PR #2455, opened 2021-07-25
Fix test impl atomic on JS

nomisRev:
There was a failing test on JS for values `x, y, z => 1f, 1f, 0f`.

It does a reference check in `compareAndSet` underneath, and on JS this made the test fail for `x` and `y` of the same value.


raulraja (approved):
thanks @nomisRev !

i-walker (approved):

Issue #2416, opened 2021-06-07
["Request"] tap / effectful extensions for Either / Validated


codymikol:
What version are you currently using?

0.13.2

What would you like to see? 

Currently I have written a few extensions written for handling effectful things like logging.

`tap` `tapLeft` `tapFold`

Essentially these are called in the same way that `map` `mapLeft` and `fold` are called, but their return values are unused. These are primarily useful for things like logging.

for example

```kotlin
doEitherFn()
  .tap { RedisService.commit("foo", it.message) }
  .tapLeft { logger.error("Failed to get foo: ${it.error}") }
  .map { transformFoo(it) }
  .tapFold({ logger.error("Failed Transformation! ${it.error}") }, { logger.info("Did Transformation! ${it.message}) })
```

Admittedly `tapFold` might not be the greatest nomenclature, but I think it gets the idea across 😉. I have these written out internally, and if the arrow team is interested I'd be willing to make a PR.

pakoito:
This was on the repo at some point, I don't know if it accidentally went away when typeclasses got removed.

nomisRev:
They were removed, but they're welcome to come back since they seem to be very commonly used.
And they make sense in combination with `suspend` for convenient logging etc like shown above.

A PR would be great @codymikol!!

nomisRev:
This has been closed https://github.com/arrow-kt/arrow/pull/2459.

Issue #2410, opened 2021-06-02
["Request"] Option.forEach


gabrieljones:
What version are you currently using?
0.13.2

What would you like to see?
Option.forEach

Unless Arrow is strictly attempting to avoid use in non pure code.  But then one might be attempted to abuse map and put side effects in a map call.

Or maybe there is already an Option.forEach under some other name?

gabrieljones:
See
https://www.scala-lang.org/api/2.13.6/scala/Option.html#foreach[U](f:A=%3EU):Unit
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/for-each.html

pakoito:
`map` is inline so it supports suspension, which means that yes, mapping or tapping (#2416) to side-effect is legitimate.

nomisRev:
This has been closed with https://github.com/arrow-kt/arrow/pull/2459

PR #2459, opened 2021-07-29
CU-m3dyef Either, Option and Validated tap effect

raulraja:
Addresses https://github.com/arrow-kt/arrow/issues/2416 and https://github.com/arrow-kt/arrow/issues/2410
In the second case in https://github.com/arrow-kt/arrow/issues/2410 `tap` is the same as `foreach` and there is also `tapNone`

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
@@ -821,11 +821,53 @@ public sealed class Either<out A, out B> {
    * Example:
    * ```
    * Right(12).mapLeft { "flower" } // Result: Right(12)
-   * Left(12).mapLeft { "flower" }  // Result: Left("flower)
+   * Left(12).mapLeft { "flower" }  // Result: Left("flower")
    * ```
    */
   public inline fun <C> mapLeft(f: (A) -> C): Either<C, B> =
     fold({ Left(f(it)) }, { Right(it) })
+
+  /**
+   * The given function is applied as a fire and forget effect
+   * if this is a `Left`.
+   * When applied the result is ignored and the original
+   * Either value is returned
+   *
+   * Example:
+   * ```
+   * Right(12).tapLeft { println("flower") } // Result: Right(12)
+   * Left(12).tapLeft { println("flower") }  // Result: prints "flower" and returns: Left(12)
+   * ```
+   */
+  public inline fun tapLeft(f: (A) -> Unit): Either<A, B> =
+    when (this) {
+      is Left -> {
+        f(this.value)
+        this
+      }
+      is Right -> this
+    }
+
+  /**
+   * The given function is applied as a fire and forget effect
+   * if this is a `Right`.
+   * When applied the result is ignored and the original
+   * Either value is returned
+   *
+   * Example:
+   * ```
+   * Right(12).tap { println("flower") } // Result: prints "flower" and returns: Right(12)
+   * Left(12).tap { println("flower") }  // Result: Left(12)
+   * ```
+   */
+  public inline fun tap(f: (B) -> Unit): Either<A, B> =
+    when (this) {
+      is Left -> this
+      is Right -> {
+        f(this.value)
+        this
+      }
+    }
 
   /**
    * Map over Left and Right of this Either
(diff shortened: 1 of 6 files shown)


franciscodr:
Task linked: [CU-m3dyef Review User requested functionality](https://app.clickup.com/t/m3dyef)

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -821,12 +821,54 @@ public sealed class Either<out A, out B> {
...
+   * ```
+   */
+  public inline fun tapLeft(f: (A) -> Unit): Either<A, B> =
+    when (this) {
+      is Left -> {
+        f(this.value)
+        this
+      }
+      is Right -> this
+    }
```
pablisco, line 849: We can also express this with a scoping function:
```suggestion
  public inline fun tapLeft(f: (A) -> Unit): Either<A, B> = apply {
    if(this is Left) {
       f(value)
    }
  }
```
Is it easier to read? 🤔

pablisco: If this feels better it would apply to the other taps 🙂 Otherwise happy to be ignored 😅 #notablocker

pablisco (approved):
Just one potential suggestion. Otherwise looks good 👍

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
raulraja: @pablisco apply it's fine since I suppose the lambda has no cost, and its body is inlined since it's flagged as invoked once in its contract. I speculate it does not need to capture the state for the lambda. I'd be curious to see the resulting bytecode because the `when` statement I wrote has no allocations and most likely results in a table-switch indexed lookup.
In whatever case, It's not important to me. If you want to suggest some changes to the impl in the same branch, we can apply them before merging; I'm sure whatever users write in the lambda is gonna be more costly than any optimizations :)

pablisco:
Ah, I got something else. :sweat\_smile:  Would it make sense to call these `onLeft`, `onRight`, etc? To follow the pattern that other components use like `onEach` on the stdlib and `onNext` on things like Rx.

raulraja:
@pablisco I have no preference for the name of these functions and @nomisRev brought up to me he saw them useful in some context but I'm also not familiar with `onX` to refer to them. If the std lib already has a convention and is `onX` then we should probably use that but not sure.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
pablisco: @raulraja Actually, now I think of it we could add the option to add effects to both sides of the bifunctor and make use of existing operations :slightly\_smiling\_face: 

```kotlin
  public inline fun onLeft(f: (A) -> Unit): Either<A, B> =
    onEither(left = f, right = {})

  public inline fun onRight(f: (B) -> Unit): Either<A, B> =
    onEither(left = {}, right = f)

  public inline fun onEither(left: (A) -> Unit, right: (B) -> Unit): Either<A, B> = 
    apply { fold(left, right) }
```

raulraja: that's probably allocating for each one of those lambdas? If that is the case, I'd rather have an alloc free impl that is less readable as we have no context as to how these functions will be used.

pablisco: That is probably right. One option may be to use a decoy for the empty lambda if we don't have something like this already. Need to check the byte code :)   
As I said, unless someone else sees this useful, we can explore it later on

nomisRev: I think `apply` is confusing here since `apply` is typically for mutable instantiation logic, like builder syntax.
Which this definitely doesn't do.

I think the `when` code Raul has is the simplest 👍

nomisRev (approved):
I think the convention we've been using is `ifLeft`, `ifRight`, `ifNone`, etc.

`onX` seems to imply executing a side-effect, which is actually why this naming was chosen in ReactiveX IIRC.

PR #2462, opened 2021-08-01
Improvements Flow.parMap

nomisRev:
KotlinX Flow prefers compositional operator for `ctx` selection, so parallel operators for `Flow` don't define a `context: CoroutineContext` params but prefers composition by `flowOn`.

This PR changes this to be in line with KotlinX Flow.

It also improves the doc a little bit on the behavior of `parMap`.


1Jajen1 (approved):
Looks good to me :+1:

raulraja (approved):
thanks @nomisRev !

PR #2465, opened 2021-08-03
WIP: Use patched version of Dokka's Jekyll Renderer

nomisRev:
The Dokka Gfm/Jekyll render doesn't correctly work with code fences, and uses `//n` for new lines instead of `/n`.
This PR fixes both issues and would allow us to upgrade Dokka to 1.5.0 without having to wait for upstream to fix these things.

Continuting using the older Dokka version 0.10.x has other issues, like not having support for the new MPP functionality.


nomisRev:
Closing in favor of: https://github.com/arrow-kt/arrow/tree/testing-web-update

PR #2469, opened 2021-08-16
Expose CoroutineScope in Fx

nomisRev:
This an binary breaking change, and this is a "semi" API breaking change.

Semi-breaking in that it only breaks for method references since the Kotlin compiler doesn't automatically ignore the unused receiver scope. Only `parSequence` can be overloaded without causing type ambuigity.

```kotlin
suspend fun add(i: Int): Int = i + 1
listOf(1, 2, 3).parTraverse(::add) // expected CoroutineScope.() -> Int
```

- [ ] We should fix binary break using visbility HIDDEN for `0.13.2` -> `0.13.3`


raulraja (approved):
Thanks, @nomisRev! 🙏

i-walker (approved):

PR #2478, opened 2021-09-06
Add Arrow Core & Fx syntax for Kotlin's Result type

nomisRev:


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Result.kt
```diff
@@ -0,0 +1,211 @@
...
+  var a: Throwable? = null
+  other.forEach { b ->
+    when {
+      a == null -> a = b
+      b != null -> a?.addSuppressed(b)
+      else -> Unit
+    }
+  }
+  return a
+}
```
pablisco, line 211:
```suggestion
internal fun composeErrors(vararg other: Throwable?): Throwable? =
    other.reduceOrNull { a, b ->
        zip(a, b, Throwable::addSuppressed)
        a ?: b
    }
```

Not sure if it reads better but here is a suggestion :) #notABlocker

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/computations/result.kt
```diff
@@ -0,0 +1,11 @@
...
+
+public object ResultEffect {
+  public fun <A> Result<A>.bind(): A = getOrThrow()
+}
+
+@Suppress("ClassName")
+public object result {
+  public inline operator fun <A> invoke(c: ResultEffect.() -> A): Result<A> =
+    kotlin.runCatching { c(ResultEffect) }
+}
```
pablisco, line 11: Wow! This is so simple. Love it :)

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
```diff
@@ -48,6 +48,40 @@ class IterableTest : UnitSpec() {
...
+        val evens = ints.traverseResult {
+          if (it % 2 == 0) {
+            acc.add(it)
+            Result.success(it)
+          } else Result.failure(RuntimeException())
+        }
+        acc shouldBe ints.takeWhile { it % 2 == 0 }
+        evens.fold(
+          { it shouldBe ints },
+          { }
```
pablisco, line 74: Should we add a check on the other side? If we make the exception in L69 ew can then make sure it arrives here :)

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/computations/result.kt
```diff
@@ -0,0 +1,11 @@
+package arrow.core.computations
+
+public object ResultEffect {
+  public fun <A> Result<A>.bind(): A = getOrThrow()
```
pablisco, line 4: Would it be in the spirit of this Effect to have cross bind like:

```
fun <A> A?.bindOr(f: () -> Throwable): A = this ?: throw f()
```

Happy to do this on a separate PR :slightly\_smiling\_face:

pablisco (approved):
Amazing stuff! Thank you for adding this :heart\_eyes:

raulraja (approved):
Looks great, @nomisRev. I wonder if we should have `Result.bind()` inside the `either` block too when `E` is `Throwable` and in the generic block perhaps `Result.bind { throwable -> E }`.

arrow-libs/core/arrow-core-test/src/commonMain/kotlin/arrow/core/test/generators/utils.kt
```diff
@@ -0,0 +1,30 @@
...
+    suspend { throw this }.startCoroutine(
+      Continuation(Dispatchers.Default) {
+        cont.intercepted().resumeWith(it)
+      }
+    )
+
+    COROUTINE_SUSPENDED
+  }
+
+public suspend fun <A> A.suspend(): A =
```
pakoito, line 21: How is this disambiguated from the specialization?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Result.kt
```diff
@@ -0,0 +1,211 @@
...
+public inline fun <A, B> Result<A>.flatMap(transform: (value: A) -> Result<B>): Result<B> =
+  fold({ a ->
+    try {
+      transform(a)
+    } catch (e: Throwable) {
+      failure(e)
+    }
+  }) { throwable ->
+    failure(throwable)
+  }
```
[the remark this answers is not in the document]
nomisRev: Well, I guess `mapCatching { transform(it).getOrThrow() }` is actually a better implementation here since that preserves all the catching capabilities which were not covered here in the `failure` case.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
nomisRev: Actually, this case should be covered. This is the verbose version.

```kotlin
        evens.fold(
          {
             acc shouldBe ints.takeWhile { it % 2 == 0 }
             it shouldBe ints
           },
          { acc shouldBe ints.takeWhile { it % 2 == 0 } }
        )
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/computations/result.kt
nomisRev: Actually, there is no need for this. You can just throw inside `result { }` since it's just a simple DSL on top of `runCatching`.

arrow-libs/core/arrow-core-test/src/commonMain/kotlin/arrow/core/test/generators/utils.kt
nomisRev: This is disambiguated by the type. The concrete signature takes precedence for top-level functions.

```
fun <A> suspend(a: A): A = ..
fun suspend(throwable: Throwable): Nothing = ...
```

nomisRev:
> Looks great, @nomisRev. I wonder if we should have Result.bind() inside the either block too when E is Throwable and in the generic block perhaps Result.bind { throwable -> E }.

Sounds great, and this is actually very easy to add.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/computations/result.kt
pablisco: Very true, we can do:

```
val profile = nullableProfile ?: error("MissingProfile")
```

or

```
val profile = checkNotNull(nullableProfile)
```

Maybe we should add this to the documentation :thinking:

i-walker (approved):
👏🏾

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Result.kt
```diff
@@ -0,0 +1,211 @@
...
+  b.exceptionOrNull(),
+  c.exceptionOrNull(),
+  d.exceptionOrNull(),
+  e.exceptionOrNull(),
+  f.exceptionOrNull(),
+  g.exceptionOrNull(),
+  h.exceptionOrNull(),
+  i.exceptionOrNull(),
+  j.exceptionOrNull(),
+)!!.let(::failure)
```
pablisco, line 198: I had a random thought, not blocking of course. I was wondering if it would make it easier to read these zip implementations if we use computations instead of delegate to higher arity functions,

nomisRev:
> I wonder if we should have Result.bind() inside the either block too when E is Throwable

@raulraja, to fully supoort this we need multiple receivers. I added the generic variant.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Result.kt
nomisRev: I choose the most effiecient implementation that was still somewhat readablee and short here. Writing this as a `nullable { }` will suffer from the same problem. The only way to write this properly is to be fully exhaustive with `fold` but that exploses quickly.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -306,6 +306,21 @@ public inline fun <E, A, B> Iterable<A>.traverseEither(f: (A) -> Either<E, B>):
...
+public inline fun <A, B> Iterable<A>.traverseResult(f: (A) -> Result<B>): Result<List<B>> {
+  val acc = mutableListOf<B>()
+  forEach { a ->
+    val res = f(a)
+    res.fold(
+      { acc.add(it) },
+      { return@traverseResult Result.failure(it) }
+    )
+  }
+  return Result.success(acc)
```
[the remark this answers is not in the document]
nomisRev: I applied this to all traverse operations! Thanks for the great suggestion.

nomisRev:
Thanks for the great reviews, I updated with all suggested changes if you want to take another look. @pablisco @i-walker @raulraja

arrow-libs/fx/arrow-fx-coroutines/api/arrow-fx-coroutines.api
```diff
@@ -182,7 +182,6 @@ public final class arrow/fx/coroutines/ParTraverse {
 	public static final fun parSequence (Ljava/lang/Iterable;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static final fun parSequence (Ljava/lang/Iterable;Lkotlin/coroutines/CoroutineContext;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static synthetic fun parSequence$default (Ljava/lang/Iterable;Lkotlin/coroutines/CoroutineContext;Lkotlin/coroutines/Continuation;ILjava/lang/Object;)Ljava/lang/Object;
-	public static final fun parSequenceEither (Ljava/lang/Iterable;Larrow/typeclasses/Semigroup;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
```
nomisRev, line 185: These were incorrectly named inside `ParTraverseValidated.kt`.. Should we deprecate them for `0.13.3`?

raulraja (approved):
thanks @nomisRev!

PR #2485, opened 2021-09-11
publish missing targets

i-walker:


nomisRev:
Can we combine all files into a single file? That way it runs as a single Github action, meaning the jobs can short-circuit each other if they fail.

.github/workflows/publish.yml
```diff
@@ -1,71 +1,387 @@
...
+        working-directory: arrow-libs
+        run: ./gradlew publishMingwX64PublicationToDeployRepository
+  publish-arrow-stack:
     env:
       BASEDIR: ${{github.workspace}}/arrow-libs
-      JAVA_OPTS: -Xms512m -Xmx1024m
       SONATYPE_USER: ${{ secrets.SONATYPE_USER }}
       SONATYPE_PWD: ${{ secrets.SONATYPE_PWD }}
       ORG_GRADLE_PROJECT_signingKey: ${{ secrets.ORG_GRADLE_PROJECT_signingKey }}
       ORG_GRADLE_PROJECT_signingPassword: ${{ secrets.ORG_GRADLE_PROJECT_signingPassword }}
```
nomisRev, line 199: I don't think we need to repeat this here if we have it at the very top.

nomisRev (approved):
LGTM. We should see SNAPSHOT appear for all binaries now.

.github/workflows/publish.yml
```diff
@@ -1,71 +1,387 @@
...
+      - main
+      - release/**
     paths:
-    - 'arrow-libs/**'
-    - '.github/**'
+      - 'arrow-libs/**'
+      - '.github/**'
+
+env:
+  BASEDIR: ${{github.workspace}}/arrow-libs
```
nomisRev, line 13: This is not an env property shared for all jobs, is it?

.github/workflows/publish.yml
```diff
@@ -1,71 +1,387 @@
...
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
+
+      - name: Publish MingwX64
+        working-directory: arrow-libs
+        run: ./gradlew publishMingwX64PublicationToDeployRepository
+  publish-arrow-stack:
     env:
       BASEDIR: ${{github.workspace}}/arrow-libs
```
nomisRev, line 195: Is this used? We changed to `working-directory: arrow-stack` in the `run` step.

.github/workflows/publish.yml
```diff
@@ -1,71 +1,387 @@
...
+          github-token: ${{github.token}}
+          script: |
+            await github.issues.create({...context.repo,
+             title: "[PUBLISH STACK] Error when merging '${{ steps.create-issue.outputs.title }}' on main branch",
+             body: "Please, check https://github.com/arrow-kt/arrow/actions/runs/${{ github.run_id }}"});
+
+  publish_doc:
+
+    env:
+      BASEDIR: ${{github.workspace}}/arrow-libs
```
nomisRev, line 244: Think this is also unused here.

nomisRev (approved):
Looks great, I think there is a couple of redundant configurations.

.github/workflows/publish.yml
i-walker: Yes it's used as an env variable for a script

.github/workflows/publish.yml
i-walker: the env variable is not affecting this step `working-directory: arrow-stack`

franciscodr (approved):
Thanks, @nomisRev!

franciscodr:
And thanks to @i-walker, who has created this pull request 😂

PR #2490, opened 2021-09-13
Fix dependencies and run native tests

i-walker:
removed watchosArm64 target due to #2491


.github/workflows/publish.yml
```diff
@@ -154,6 +154,27 @@ jobs:
         working-directory: arrow-libs
         run: ./gradlew publishTvosSimulatorArm64PublicationToMavenRepository
 
+  publish_watchos:
```
nomisRev, line 157: Why did you split these? We can optimise this later, but we want to find a sweet spot where all jobs take aabout the same time, and we also want to cut back as much as possible on `macos-latest` because those are limited in how much can run in parallel. Whilst `ubuntu` we can run a lot of them for free in parallel.

So by optimising this (not needed in this PR) we can make more PRs, or branch run their action in parallel reducing wait time.

i-walker: it got cancled because it took to long

.github/workflows/pull_request.yml
```diff
@@ -155,7 +157,7 @@ jobs:
 
       - name: Build
         working-directory: arrow-libs
-        run: ./gradlew macosX64Test iosX64Test
+        run: ./gradlew macosX64Test iosX64Test macosArm64Test iosArm64TestKlibrary iosArm64TestKlibrary iosArm32TestBinaries iosArm32TestKlibrary iosSimulatorArm64Test
```
nomisRev, line 160: Do these tasks not have a `xxxTest` variant i.g is `iosArm64Test` not a task? And why do we need both `iosArm64TestKlibrary` and `iosArm64TestKlibrary`?

.github/workflows/publish.yml
i-walker: Will reduce the duration once we have a example time frame

.github/workflows/pull_request.yml
```diff
@@ -212,6 +215,48 @@ jobs:
...
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+            ~/.konan
+          key: watchos-step-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            watchos-step-gradle-
+
+      - name: Build
+        working-directory: arrow-libs
+        run: ./gradlew watchosX64Test watchosX86Test watchosSimulatorArm64Test watchosArm32TestBinaries watchosArm32TestKlibrary watchosArm64TestBinaries watchosArm64TestKlibrary
```
nomisRev, line 242: Idem: seems we have a duplications between `TestKlibrary` and `TestBinaries`.

nomisRev (commented):
Looks good, couple questions about the tasks. The tasks generated for the new targets seem to have other naming conventions as the others.

.github/workflows/pull_request.yml
i-walker: Thanks for that catch. 

```
Do these tasks not have a xxxTest variant i.g is iosArm64Test not a task?
```
Yes, there aren't yet gradle test task for those, so we do this instead to run binary tests and library tests

.github/workflows/pull_request.yml
```diff
@@ -172,7 +174,8 @@ jobs:
         with:
           name: test-reports
           path: test-reports
-  tvos_and_watchos:
+
+  tvos:
```
nomisRev, line 178: Why are we splitting this up? Is this taking so much longer than `macos + ios` ?
Adding more `macos-latest` jobs will really slow down the build, since there is a low limit on how much we can run in parallel.

nomisRev (commented):
How come we're deleting `getTimeMillis` for all native targets?

.github/workflows/pull_request.yml
i-walker: yes, either that or I up the wait time

nomisRev: Yes, let's combine them. Feel free to also move JS job into the Linux job at the same time.

nomisRev (approved):

PR #2524, opened 2021-09-21
Deduplicate binary creation

nomisRev:
This PR deduplicates the publishing of binaries.

MacOS is able to produce all binaries except Windows, so we have a seperate windows job that only produces the windows artifacts.


raulraja (approved):
thanks @nomisRev !

PR #2529, opened 2021-09-22
Introduce Kotlin Gradle Script

nomisRev:
Introduces Kotlin Gradle Script into Arrow's Gradle configuration.

It creates a Kotlin Gradle Plugin that simplifies how we setup Dokka and Publishing in Arrow Org projects. We can share this setup by publishing the Gradle plugin so we can consume it from Arrow Meta, Arrow Endpoints and other projects.


arrow-libs/ank/build.gradle
```diff
@@ -12,9 +12,9 @@ buildscript {
 
 plugins {
     id "org.jetbrains.kotlin.jvm" version "$KOTLIN_VERSION"
-    id "org.jlleitschuh.gradle.ktlint" version "$KTLINT_GRADLE_VERSION" apply false
```
raulraja, line 15: are we dropping the properties for these versions or does this module need special overrides?

arrow-libs/build.gradle.kts
```diff
@@ -0,0 +1,114 @@
+import io.github.nomisrev.setupDokka
+import io.github.nomisrev.setupPublishing
```
raulraja, line 2: Should these functions live in Arrow?

arrow-libs/build.gradle.kts
```diff
@@ -0,0 +1,114 @@
...
+  apply(from = "gradle/setup.gradle")
+  repositories {
+    mavenCentral()
+  }
+}
+
+plugins {
+  id("org.jlleitschuh.gradle.ktlint") version "10.1.0" apply false
+  id("ru.vyarus.animalsniffer") version "1.5.0" apply false
+  id("io.kotest.multiplatform") version "5.0.0.5" apply false
```
raulraja, line 14: same question as above regarding versions and property files.

raulraja (approved):
thanks @nomisRev !

arrow-libs/build.gradle.kts
nomisRev: Oh right. Yes I need to rename this.

arrow-libs/ank/build.gradle
nomisRev: You cannot access `gradle.properties` from `gradle.kts` but you can from `.gradle`...
However, for us to be able to leverage dependabot for automatic version updating we're going to abandon the pattern of version in `gradle.properties` and we'll use regular `groupId:artifactid:version` for dependencies.

This way we dependabot will be automatically able to create PRs for all version updates, this will happen asap but over multiple PRs. We also want to fully migrate to `gradle.kts` but in small increments, and without blocking any releases.

JavierSegoviaCordoba: You should be able to access the properties in `gradle.properties` via `properties["some.property"]` (and there are more ways, but I don't remember them now).

There are a few more options:

1. refreshVersions plugin
2. Version Catalogs + migrate from dependabot to Renovate (I was using dependabot, I migrated recently to Renovate and it is not only more powerful, but the support is also just really amazing).

I haven't used refreshVersion plugin, but it has a task to automigrate everything, maybe you can try it. For the second option, I am using it in a lot of projects and I can help you with the implementation.

nomisRev: Hey @JavierSegoviaCordoba,

Thanks for commenting! I'm really not so good with this stuff 😅
Version Catalogs looks great I hadn't heard of it before.

If you're interested in contributing Arrow could definitely use some improvements in Gradle. There are no tickets for it but feel free to make some, but we're trying to get maintenance down as much as possible whilst still being able to do frequent releases with important version updates. So if we can get powerful automatic version updating that would be awesome!

I'll try your suggestion, thank you!

JavierSegoviaCordoba: Yeah, I can contribute @nomisRev. If you like the Version Catalog + renovate approach, I can create a PR adding them and tell you how to activate renovate, what do you think if we talk in kotlinlang slack to organize it?

nomisRev: Yes, a contribution would of course be super appreciated 🙏

arrow-libs/plugins/src/main/kotlin/arrow/gradle/MppPublishingPlugin.kt
```diff
@@ -0,0 +1,159 @@
...
+import org.gradle.kotlin.dsl.register
+import org.gradle.kotlin.dsl.withType
+import org.gradle.plugins.signing.SigningExtension
+
+/**
+ * An empty [Plugin] that exposes a set of utility functions to simplify setting up MPP publishing.
+ */
+class MppPublishingPlugin : Plugin<Project> {
+  override fun apply(target: Project) {}
+}
```
pablisco, line 22: In AutoModule, I discovered a way to add a build source code without a plugin:  
[https://github.com/pablisco/auto-module/blob/master/plugin/src/main/kotlin/com/pablisco/gradle/automodule/AutoModulePlugin.kt#L62](https://github.com/pablisco/auto-module/blob/master/plugin/src/main/kotlin/com/pablisco/gradle/automodule/AutoModulePlugin.kt#L62)

However, we may want to keep it like this if we want to publish the plugin :thinking:

arrow-libs/plugins/src/main/kotlin/arrow/gradle/MppPublishingPlugin.kt
```diff
@@ -0,0 +1,159 @@
...
+  licenseDist: String,
+  scmConnection: String,
+  scmDevConnection: String,
+) {
+  pom {
+    if (!name.isPresent) {
+      name by artifactId
+    }
+    this@pom.description.set(description)
+    this@pom.url.set(url)
```
pablisco, line 122: I think we may not need the tags on all the `this` in this function

arrow-libs/plugins/src/main/kotlin/arrow/gradle/MppPublishingPlugin.kt
```diff
@@ -0,0 +1,159 @@
...
+        )
+      }
+      repositories {
+        maven(
+          if (version.toString().endsWith("SNAPSHOT")) snapshotRepo else releaseRepo,
+          sonatypeUsername,
+          sonatypePassword
+        )
+      }
+      Nullable.zip(key, pass) { key, pass -> signPublications(key, pass) }
```
pablisco, line 65: I love the `zip` function, however, I wonder if we could do this too:

```kotlin
if (key != null && pass != null) signPublications(key, pass)
```

They are, to all effects, the same :thinking: 

Or add Arrow itself to the plugin classpath :joy:   
PS: Not a blocker

pablisco (commented):
Amazing stuff!! thanks for starting this. I've been wanting to do this for ages hehe.
I have a couple of comments, but non of them are blockers 🙂

arrow-libs/plugins/src/main/kotlin/arrow/gradle/MppPublishingPlugin.kt
nomisRev: Yes, I think we want to publish this as a plugin so we can share it amongst other projects like Arrow Meta, and soon Arrow Endpoints.

arrow-libs/plugins/src/main/kotlin/arrow/gradle/MppPublishingPlugin.kt
nomisRev: Ye, I was thinking that too. Using `!= null` is probably easier here.

Let's not put Arrow on the classpath 😂 It's hard enough to work with without more new things.

arrow-libs/plugins/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.1-bin.zip
```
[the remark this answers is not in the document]
nomisRev: I guess this showed up due to the empty `settings.gradle.kts` I added? I added that because otherwise, I couldn't open the `plugins` folder in IntelliJ to get tooling to work.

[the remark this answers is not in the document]
nomisRev: If these duplicated wrappers files can be removed that'd be great! There is a couple more in nested folders in `arrow-libs` I think :/

arrow-libs/plugins/src/main/kotlin/arrow/gradle/MppPublishingPlugin.kt
```diff
@@ -0,0 +1,159 @@
...
+    .getByType(PublishingExtension::class.java)
+    .publications
+
+  project
+    .extensions
+    .getByType(SigningExtension::class.java)
+    .apply {
+      useInMemoryPgpKeys(key, pass)
+    }.sign(publications)
+}
```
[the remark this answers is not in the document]
nomisRev: 🤯

arrow-libs/plugins/src/main/kotlin/arrow/gradle/MppPublishingPlugin.kt
pablisco: I've just remembered that we may be able to use gradle extensions instead and kts will generate accessor functions. This way it's compatible with both groovy and Kotlin scripts. I'll create a separate ticket to investigate this :slightly\_smiling\_face:

arrow-libs/plugins/gradle/wrapper/gradle-wrapper.properties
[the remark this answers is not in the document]
pablisco: Ok, we can look at this on a separate ticket then. It's interesting that IntelliJ is required to add it. 

I've created a [ticket](https://github.com/arrow-kt/arrow/issues/2531) to investigate this separately :slightly\_smiling\_face:

arrow-libs/ank/build.gradle
pablisco: I've created [this one]( https://github.com/arrow-kt/arrow/issues/2532) :eyes:

arrow-libs/build.gradle.kts
```diff
@@ -0,0 +1,86 @@
+import arrow.gradle.setupDokka
+import arrow.gradle.setupPublishing
```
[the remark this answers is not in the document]
nomisRev: Is there any best pratices around this?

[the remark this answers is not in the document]
pablisco: Not with extension function. However, packaging in this case doesn't add much value. However, if we change the extension functions for Gradle extensions instead and let gradle generate the accessors (see [here](https://github.com/arrow-kt/arrow/issues/2530)) then this problem will go away. I guess that's the closest to a best practice. But again, we can do it on a separate PR.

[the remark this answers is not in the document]
nomisRev: I don't know what accessors are ... 😅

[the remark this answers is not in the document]
pablisco: Gradle generates type-safe accessors that are normally done JIT with groovy :)  
https://docs.gradle.org/current/userguide/migrating_from_groovy_to_kotlin_dsl.html

arrow-libs/plugins/src/main/kotlin/arrow/gradle/MppPublishingPlugin.kt
nomisRev: We do because it conflicts with the params of the function... Better/different paramnames for the functions? :/

.github/workflows/pull_request.yml
```diff
@@ -136,7 +136,7 @@ jobs:
 
   tv_and_watchos:
     runs-on: macos-latest
-    timeout-minutes: 30
+    timeout-minutes: 40
```
pakoito, line 139: F

nomisRev: F

nomisRev: Seriously though, not sure why it sometimes flukes. It runs on avarage in about 20 minutes, but I guess ocassionally it runs on a machine with less resources?

PR #2539, opened 2021-09-27
Adds a andThen extension method to Validated to chain Validated

magott:
As discussed on slack [https://kotlinlang.slack.com/archives/C5UPMM0A0/p1632399994123800] proposing a andThen method for `Validated` to easier chain Validations without going through `withEither`

### DIFF
--- a/arrow-libs/core/arrow-core/api/arrow-core.api
+++ b/arrow-libs/core/arrow-core/api/arrow-core.api
@@ -2331,6 +2331,7 @@ public final class arrow/core/Validated$Valid$Companion {
 }
 
 public final class arrow/core/ValidatedKt {
+	public static final fun andThen (Larrow/core/Validated;Lkotlin/jvm/functions/Function1;)Larrow/core/Validated;
 	public static final fun attempt (Larrow/core/Validated;)Larrow/core/Validated;
 	public static final fun bisequence (Larrow/core/Validated;)Ljava/util/List;
 	public static final fun bisequenceEither (Larrow/core/Validated;)Larrow/core/Either;
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Validated.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Validated.kt
@@ -364,7 +364,71 @@ public typealias Invalid<E> = Validated.Invalid<E>
  * ## Sequential Validation
  *
  * If you do want error accumulation, but occasionally run into places where sequential validation is needed,
- * then Validated provides a `withEither` method to allow you to temporarily turn a Validated
+ * then Validated provides a couple of methods that can be used.
+ *
+ * ### `andThen`
+ *
+ * The `andThen` method is similar to `flatMap`. In case of a valid instance it will pass the valid value into
+ * the supplied function that in turn returns a `Validated` instance
+ *
+ * ```kotlin:ank:playground
+ * import arrow.core.Validated
+ * import arrow.core.computations.either
+ * import arrow.core.valid
+ * import arrow.core.invalid
+ *
+ * abstract class Read<A> {
+ *  abstract fun read(s: String): A?
+ *
+ *  companion object {
+ *
+ *   val stringRead: Read<String> =
+ *    object : Read<String>() {
+ *     override fun read(s: String): String? = s
+ *    }
+ *
+ *   val intRead: Read<Int> =
+ *    object : Read<Int>() {
+ *     override fun read(s: String): Int? =
+ *      if (s.matches(Regex("-?[0-9]+"))) s.toInt() else null
+ *    }
+ *  }
+ * }
+ *
+ * data class Config(val map: Map<String, String>) {
+ *   suspend fun <A> parse(read: Read<A>, key: String) = either<ConfigError, A> {
+ *     val value = Validated.fromNullable(map[key]) {
+ *       ConfigError.MissingConfig(key)
+ *     }.bind()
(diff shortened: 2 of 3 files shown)


arrow-libs/core/arrow-core/api/arrow-core.api
```diff
@@ -2331,6 +2331,7 @@ public final class arrow/core/Validated$Valid$Companion {
 }
 
 public final class arrow/core/ValidatedKt {
+	public static final fun andThen (Larrow/core/Validated;Lkotlin/jvm/functions/Function1;)Larrow/core/Validated;
```
nomisRev, line 2334: 👌

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Validated.kt
```diff
@@ -1152,6 +1216,23 @@ public inline fun <E, A> Validated<E, A>.findValid(SE: Semigroup<E>, that: () ->
...
+ * Valid(5).andThen { Invalid(10) } // Result: Invalid(10)
+ * Invalid(5).andThen { Valid(10) } // Result: Invalid(5)
+ * ```
+ */
+public inline fun <E, A, B> Validated<E, A>.andThen(f: (A) -> Validated<E, B>): Validated<E, B> {
+  return when (this) {
+    is Validated.Valid -> f(value)
+    is Validated.Invalid -> this
+  }
+}
```
nomisRev, line 1234: Nit: Arrow codebase prefers expression bodies where easily possible.

```suggestion
public inline fun <E, A, B> Validated<E, A>.andThen(f: (A) -> Validated<E, B>): Validated<E, B> =
  when (this) {
    is Validated.Valid -> f(value)
    is Validated.Invalid -> this
  }
```

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/ValidatedTest.kt
```diff
@@ -498,5 +498,17 @@ class ValidatedTest : UnitSpec() {
...
+      Valid(10).andThen { Valid(it + 5) } shouldBe Valid(15)
+    }
+
+    "andThen should only run f on valid instances " {
+      Invalid(10).andThen { Valid(5) } shouldBe Invalid(10)
+    }
+
+    "andThen should return Invalid(result) if f return Invalid " {
+      Valid(10).andThen { Invalid(5) } shouldBe Invalid(5)
+    }
```
nomisRev, line 512: Tip:
If you'd want to turn this into property based tests you would simply abstract the two input values:

```
    "andThen should return Valid(result) if f return Valid" {
      checkAll(Arb.int(), Arb.int()) { x, y ->
        Valid(x).andThen { Valid(it + y) } shouldBe Valid(x + y)
      }
    }
```

nomisRev (approved):
Looks awesome! Thank you @magott!! 🙏 👏 🥳 
I left a nit, and a tip but this can also be merged as is 👍

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/ValidatedTest.kt
magott: Good suggestions, I'll fix them and push an update

raulraja (approved):

magott:
Updated the PR with the suggestions implemented

nomisRev:
Thanks thee contribution @magott !! 🙌

nomisRev:
Should become available in 1.0.1 with Kotlin 1.5.31 soon ;)

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Validated.kt
```diff
@@ -364,7 +364,71 @@ public typealias Invalid<E> = Validated.Invalid<E>
...
+ *  data class MissingConfig(val field: String) : ConfigError()
+ *  data class ParseConfig(val field: String) : ConfigError()
+ * }
+ *
+ * //sampleStart
+ * val config = Config(mapOf("house_number" to "-42"))
+ *
+ * suspend fun main() {
+ *   val houseNumber = config.parse(Read.intRead, "house_number").andThen { number ->
+ *     if (number >= 0) Valid(0)
```
froden, line 420: @magott should it rather be `Valid(number)` here?

nomisRev: 🕵️

magott: haha, surprised there weren't more errors in code written in doc. Do you want a new PR @nomisRev ? Or will you just fix it in main?

froden: Looks great! Good job 👏 Looking forward to delete the extension method from our code ❤️

nomisRev: I'll fix it directly in main.

This is actually why I'm working on this Dokka Plugin: https://github.com/nomisRev/AnkDokkaPlugin/pull/7/files
I think we can already do it today too with our current setup, but if we wrote tests instead of examples than we could catch these mistakes on CI as well.

magott: Yeah, I think the docs compile as part of the build, so this isn't an error as such. Just a logical bug.

PR #2540, opened 2021-09-28
Add bind with error mapping for Either

pakoito:
Add a helper function that allows working on any `either` within a comprehension, or map the errors of an existing one.

Done from github directly, pls don't judge hahaha

```
suspend fun leftThrow(): Either<Throwable, Int>

suspend fun leftString(): Either<String, Int> = either {
  // current
  leftThrow().mapLeft { it.message }.bind()

  // new
  leftThrow().bind { it.message }
}
```

Maybe it needs another name such as `bindOr` but I suspect that'd make it look like the recovery value instead.

### DIFF
--- a/arrow-libs/core/arrow-core/api/arrow-core.api
+++ b/arrow-libs/core/arrow-core/api/arrow-core.api
@@ -2386,14 +2386,18 @@ public final class arrow/core/ValidatedKt {
 
 public abstract interface class arrow/core/computations/EitherEffect : arrow/continuations/Effect {
 	public abstract fun bind (Larrow/core/Either;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public abstract fun bind (Larrow/core/Either;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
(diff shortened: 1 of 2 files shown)


nomisRev:
This can already be done today with `getOrElse`, or `getOrHandle`.

```
suspend fun leftThrow(): Either<Throwable, Int>

suspend fun leftString(): Either<String, Int> = either {
  leftThrow().mapLeft { it.message }.bind()
  leftThrow().getOrHandle { it.message }
  leftThrow().getOrElse { "default" }
  1
}
```

raulraja:
@pakoito discussing with @nomisRev and @i-walker  I recall we used to have this or the same with a different name before in fx but when we went over APIs for the following series we dropped it to simplify the overall Arrow APIs. The reason to drop them was not that they were not useful but because users came frequently confused with several ways to do the same thing. I think in this case we should stick with `mapLeft` and keep `bind` without overloads. In general, we had dropped all overloads as a rule of thumb from Arrow unless they bring some new functionality. I think here it applies in the sense that this is not new just saving the user from a call to `bind`

raulraja:
I discussed this over slack with @pakoito and others. I am closing this as it's preferred to have this as an external ext fun in the user codebase to avoid increasing the size of the arrow-core API.

pakoito:
I knew it rang a bell!

PR #2543, opened 2021-09-29
Add Arrow Gradle config

JavierSegoviaCordoba:
It compiles, all tests are passing, and all publications to the maven local repository seem right.

Tasks:

- [x] Change Gradle config Java compatibility to 8 instead of 11
- [x] Publish Gradle config to MavenCentral and Gradle Plugin Portal
- [ ] Try publications in remote repositories


arrow-libs/ank/arrow-ank-gradle/build.gradle.kts
```diff
@@ -0,0 +1,44 @@
...
+gradlePlugin {
+  plugins {
+    create("ank") {
+      id = "ank-gradle-plugin"
+      implementationClass = "arrow.ank.AnkPlugin"
+    }
+  }
+}
+
+tasks.whenTaskAdded {
```
nomisRev, line 36: Duplicated with line 40?

arrow-libs/core/arrow-core-test/build.gradle.kts
```diff
@@ -0,0 +1,36 @@
...
+    jvmTest {
+      dependencies {
+        runtimeOnly(libs.kotest.runnerJUnit5)
+      }
+    }
+    jsMain {
+      dependencies {
+        implementation(libs.kotlin.stdlibJS)
+      }
+    }
```
nomisRev, line 34: Is this actually needed? 🤔 

We don't rely on `stdlibJS` or `stdlibJDK8` anywhere directly, and we don't specify it for native either.

I think specifying `runnerJUnit5` is no longer needed for Kotest since M2 it resolves `framework-engine` to JVM also cc\\ @sksamuel

arrow-libs/core/arrow-meta/arrow-meta-test-models/src/main/java/arrow/ap/objects/iso/IsoSealed.kt
```diff
@@ -6,7 +6,11 @@ import arrow.optics.OpticsTarget
 import arrow.optics.optics
 
 @optics([OpticsTarget.ISO])
-sealed class IsoSealed(val field: String, val nullable: String?, val option: Option<String>) {
-  data class IsoSealed2(val a: String?) : IsoSealed("", null, None)
-  companion object
+public sealed class IsoSealed(
+  public val field: String,
+  public val nullable: String?,
```
nomisRev, line 11: These cannot be changed, and should be excluded from `explicitApi()`.
It's meant for testing kapt, and the tests rely on `public` not being there...

Changing this means we also have to update the tests, but kapt is actually deprecated but blocking by Arrow Meta not being released yet.

arrow-libs/gradle/libs.versions.toml
```diff
@@ -0,0 +1,59 @@
+[versions]
+animalSniffer = "1.5.0"
+arrowGradleConfig = "0.0.1-SNAPSHOT"
+autoService = "1.0-rc7"
+compileTesting = "0.18"
+coroutines = "1.5.2"
+dokka = "1.5.30"
```
nomisRev, line 7: I'm not sure we can bump to `1.5.30` already.
The fixes we have in the jekyll module was build for `1.5.0`, if we want to bump the version we need to do manual testing of the website to verify nothing broke.

If a fix landed in 1.5.30, then we might be duplicating the fix. I.e. wrapping code blocks twice in ```lang

cc\\ @raulraja @franciscodr unless this version bump gets deployed only to the next version, then we can let this version bump go through and we an do manual testing on `next`. ?

arrow-libs/gradle/libs.versions.toml
```diff
@@ -0,0 +1,59 @@
+[versions]
```
nomisRev, line 1: Will this work with renovate like this? 😍

nomisRev (commented):
Looks awesome @JavierSegoviaCordoba. I have a couple of questions, and I think perhaps there is some opportunity to remove some redundant config we had before.

arrow-libs/core/arrow-core-test/build.gradle.kts
JavierSegoviaCordoba: I am not sure why but they are necessary, without them, some projects don't compile.

About the Kotest runner, I have no idea, just copied from the previous `build.gradle`

.github/workflows/pull_request.yml
```diff
@@ -157,7 +157,7 @@ jobs:
 
       - name: Build
         working-directory: arrow-libs
-        run: ./gradlew watchosX64Test watchosX86Test watchosSimulatorArm64Test watchosArm32TestBinaries watchosArm32TestKlibrary tvosX64Test tvosSimulatorArm64Test tvosArm64TestBinaries tvosArm64TestKlibrary
+        run: ./gradlew allTests
```
nomisRev, line 160: Now we're running all Apple tasks twice, no? 
Before we tried splitting all targets across different actions to achieve the shortest total CI time as possible.

arrow-libs/core/arrow-meta/arrow-meta-test-models/src/main/java/arrow/ap/objects/iso/IsoSealed.kt
JavierSegoviaCordoba: This is the reason I enabled it: https://kotlinlang.slack.com/archives/C0B8MA7FA/p1632946470226500

I have to remove it from the published shared config until that is fixed.

arrow-libs/gradle/libs.versions.toml
JavierSegoviaCordoba: I downgraded it, no problem with that :)

arrow-libs/gradle/libs.versions.toml
JavierSegoviaCordoba: Yeah, it will work, I have to cleanup how we are using all dependencies yet, but when I can delete all gradle.properties versions and move to the toml file, I will add the Renovate file :)

.github/workflows/pull_request.yml
JavierSegoviaCordoba: Yeah the CI will be larger because it will run jvm tests and so on, I can set the concrete Gradle tests but it is easy to forget to add new targets in the future, what do you think?

Is there a lot of difference?

arrow-libs/gradle/libs.versions.toml
nomisRev: Awesome. Really looking forward to seeing that in action.

.github/workflows/pull_request.yml
nomisRev: Yes, that's a good point. We can try to use `allTests`, but then we should remove one of the two macos jobs. Let's see how long it takes to run all Apple targets on a single machine.

arrow-libs/core/arrow-core-test/build.gradle.kts
nomisRev: Yes, it's a bit strange to me too but doesn't matter for this PR. We can make any further improvements later.

arrow-libs/core/arrow-meta/arrow-meta-test-models/build.gradle.kts
```diff
@@ -0,0 +1,14 @@
+import org.jetbrains.kotlin.gradle.dsl.ExplicitApiMode
+
+plugins {
+  id(libs.plugins.arrowGradleConfig.jvm.get().pluginId)
```
nomisRev, line 4: Do we need this here?

Before it was just using, which bypasses the configuration from Arrow and just turns this into a vanilla JVM module.

```
id "org.jetbrains.kotlin.jvm"
id "org.jetbrains.kotlin.kapt"
```

arrow-libs/ank/arrow-ank-gradle/build.gradle.kts
JavierSegoviaCordoba: Yep, removed :)

arrow-libs/core/arrow-meta/arrow-meta-test-models/src/main/java/arrow/ap/objects/iso/IsoSealed.kt
JavierSegoviaCordoba: Fixed with your tip, thank you :)

arrow-libs/core/arrow-meta/arrow-meta-test-models/build.gradle.kts
JavierSegoviaCordoba: Thank you for this one 💯

arrow-libs/core/arrow-core-test/build.gradle.kts
sksamuel: You need junit5 for JVM because the Kotlin multi platform plugin isn't responsible for tests unlike in native and js

arrow-libs/gradle/libs.versions.toml
```diff
@@ -0,0 +1,59 @@
...
+kotlinTest = "3.3.3"
+kotlin = "1.5.31"
+kotlinBinaryCompatibilityValidator = "0.7.1"
+kotlinPoet = "1.5.0"
+ktlint = "10.1.0"
+mockWebServer = "3.13.1"
+retrofit = "2.8.1"
+
+[libraries]
+arrow-kotlinMetadata = { module = "io.arrow-kt:kotlin-metadata", version.ref = "kotlin" }
```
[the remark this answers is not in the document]
JavierSegoviaCordoba: Yes, version can be referenced so and it will work too

nomisRev (approved):

PR #2544, opened 2021-10-03
Change more Gradle config and GitHub Actions workflows

JavierSegoviaCordoba:
- Migrate more files to `kts`
- Use Version Catalogs in all projects
- Rename some Gradle properties in multiple `gradle.properties` files
- Change and improve GitHub Actions workflows


.github/workflows/publish.yml
```diff
@@ -97,7 +100,10 @@ jobs:
         working-directory: arrow-libs
         run: |
           ./gradlew publishToSonatype
-          echo "$(cat gradle.properties | grep VERSION_NAME | cut -d'=' -f2) MingwX64 deployed!"
+          echo "$(cat gradle.properties | grep projects.version | cut -d'=' -f2) MingwX64 deployed!"
+
+      - name: Stop Gradle daemons
+        run: ./gradlew --stop
```
nomisRev, line 106: This was to fix the cache issue? Is there any link you can reference? Curious about the technical details :p

arrow-libs/settings.gradle.kts
```diff
@@ -26,6 +26,12 @@ pluginManagement {
 }
 
 dependencyResolutionManagement {
+  versionCatalogs {
```
nomisRev, line 29: 🥳

nomisRev:
@JavierSegoviaCordoba looking at the Github Actions times, it looks like we can collapse the split between `macos_and_ios` and `tv_and_watchos`. I think all those targets will still be faster than ktlint & Dokka. The limit for free macos jobs is 5, to this would allow more PRs to run in parallel.

Screenshot 2021-10-03 at 11 57 49

raulraja (approved):
thanks @JavierSegoviaCordoba , looks great!

.github/workflows/publish.yml
JavierSegoviaCordoba: I think it was a workaround in the issue filed in the Cache repo, this one https://github.com/actions/cache/issues/454.

Not sure why, I guess the daemon blocks the access to those folders and then it fails, similar what you get in local when you run a lot of Gradle tasks over the same project.

JavierSegoviaCordoba:
> @JavierSegoviaCordoba looking at the Github Actions times, it looks like we can collapse the split between `macos_and_ios` and `tv_and_watchos`. I think all those targets will still be faster than ktlint & Dokka. The limit for free macos jobs is 5, so this would allow more PRs to run in parallel.
> 
> Screenshot 2021-10-03 at 11 57 49

Changed to run all apple targets in the same job

.github/workflows/publish.yml
nomisRev: Thanks for the update @JavierSegoviaCordoba!

PR #2546, opened 2021-10-04
Improve Gradle config

JavierSegoviaCordoba:
- Bump `arrow-gradle-config` plugins to 0.3.2
- Set `group` and `version` in `arrow-libs`
- Disable Dokka in some projects
- Improve publish workflow

### DIFF
--- a/arrow-libs/core/arrow-meta/build.gradle.kts
+++ b/arrow-libs/core/arrow-meta/build.gradle.kts
@@ -4,6 +4,10 @@ plugins {
   id("org.jetbrains.kotlin.kapt")
 }
 
+publishJVM {
+  isDokkaEnabled = false
+}
+
 dependencies {
   api(projects.arrowAnnotations)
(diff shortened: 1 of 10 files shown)


raulraja (approved):
thanks @JavierSegoviaCordoba !

arrow-libs/core/arrow-meta/build.gradle.kts
```diff
@@ -4,6 +4,10 @@ plugins {
   id("org.jetbrains.kotlin.kapt")
 }
 
+publishJVM {
+  isDokkaEnabled = false
```
nomisRev, line 8: 👌 So by default it sets up dokka, but you disable it it only uses Dokka for the MPP doc jars?

nomisRev (approved):
Looks great!! Thank you @JavierSegoviaCordoba ❤️‍🔥

PR #2548, opened 2021-10-05
docs(Either): correct error in `getOrElse`

codymikol:
this fixes the example of `getOrElse` to use a
lambda function rather than just passing a value.

Fixes N/A


nomisRev:
@codymikol, thanks for these contributions!

If you're interested, you could also make these examples compile safe.
Which means we can check on CI if the examples compile, and they have the correct output.

By changing the code to ```kotlin:ank, and adding the imports.
Then you could also use assertions to replace `// Result: 12`.

I think kotest assertions are on the classpath where we run the examples as scripts on CI.

PR #2551, opened 2021-10-05
docs(Either): describe the `all` method

codymikol:
this adds documentation to the `all` method, and
moves it closer to the very similar, but subtly
different `exists` function.

Fixes N/A


raulraja (approved):
thank you @codymikol !

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -881,6 +881,22 @@ public sealed class Either<out A, out B> {
...
+   * Returns `true` if [Left] or returns the result of the application of
+   * the given predicate to the [Right] value.
+   *
+   * Example:
+   * ```
+   * Right(12).all { it > 10 } // Result: true
+   * Right(7).all { it > 10 }  // Result: false
+   *
+   * val left: Either<Int, Int> = Left(12)
+   * left.all { it > 10 }      // Result: false
```
i-walker, line 894:
```suggestion
   * left.all { it > 10 }      // Result: true
```

i-walker:
Thanks @codymikol 👍🏾

i-walker:
Thanks @codymikol 👍🏾

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
codymikol: 🤦‍♂️  good catch!

PR #2572, opened 2021-11-06
Add Cont<R, A>

nomisRev:
Add `Cont<A, R>` and `ContEffect` into Arrow Core.

Ideally in the future the `cont<R, A>` DSL can completely replace the current Arrow Computations machinery.

I'm afraid to fully switch to the `Cont` runtime for the existing computation blocks is not possible in a binary compatible way since the receiver of the DSL needs to change. `either` would simply work in terms of `ContEffect<E>`, and no longer needs a specific interface `EitherEffect<E, A>`. This will also allow us to drop the `A` constraint, and avoid the `*` projection and casting many cases (and downstream implementations).

Switching to this new `Cont` system for the computation blocks has serveral benefits.
 - It uses `kotlin.coroutines.CancellationException` to properly signal cancellation to the coroutine system
 - Much simpler runtime, so easier to understand and maintain.
 - Implements a generic runtime for computation that result in `E`, `A` or `Throwable` so we can generically implement monadic operations, and `fold` these `Continuation` based programs over the different cases.
 
 
TODO:
 - [ ] Write docs
 - [x] Add test suite for Cont

 As a deprecation path we could take following approach:
   - [x] 1.0.x we deprecate all the related effect types (done in this PR)
   - [ ] 1.1.x (or 1.2.x) we set the deprecated stuff to `level = DeprecationLevel.HIDDEN`, and that allows us to declare a new `either { }` block by lowering the receiver from `EitherEffect<E, A>` to `ContEffect<E>`. (See example below)
    => This doesn't break the binary, but hides the code from the Kotlin API. 
    => Only removing code from Arrow Computations but fully preserving Arrow Core's API.
   - [ ] 2.x.x (or later) we remove the HIDDEN deprecated code (breaking the binary)
   
   
```kotlin
object either {
  @Deprecated("Gone from Kotlin API", level = DeprecationLevel.HIDDEN)
  public suspend inline operator fun <E, A> invoke(crossinline c: suspend EitherEffect<E, *>.() -> A): Either<E, A> =
    Effect.suspended(eff = { EitherEffect { it } }, f = c, just = { it.right() })
  
  public suspend operator fun <E, A> invoke(c: suspend ContEffect<E>.() -> A): Either<E, A> =
    cont(c).toEither()
}  
```

See reference implementation, also includes `restrictedCont` mentioned in code comments.
https://github.com/nomisRev/Continuation

### DIFF
--- a/arrow-libs/core/arrow-continuations/src/commonMain/kotlin/arrow/continuations/Effect.kt
+++ b/arrow-libs/core/arrow-continuations/src/commonMain/kotlin/arrow/continuations/Effect.kt
@@ -2,6 +2,7 @@ package arrow.continuations
 
 import arrow.continuations.generic.DelimitedScope
 
+@Deprecated("Prefer using Cont<R, A>")
 public fun interface Effect<F> {
   public fun control(): DelimitedScope<F>
 
--- a/arrow-libs/core/arrow-continuations/src/commonMain/kotlin/arrow/continuations/Effect.kt
+++ b/arrow-libs/core/arrow-continuations/src/commonMain/kotlin/arrow/continuations/Effect.kt
@@ -4,9 +4,11 @@ import arrow.continuations.generic.DelimitedScope
 
 @Deprecated("Prefer using Cont<R, A>")
 public fun interface Effect<F> {
+  @Deprecated("Use ContEffect<E>.shift instead")
   public fun control(): DelimitedScope<F>
 
   public companion object {
+    // Replace by restrictedCont
     public suspend inline fun <Eff : Effect<*>, F, A> suspended(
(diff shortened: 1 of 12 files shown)


nomisRev:
This will effectively deprecate Arrow-Continuations, and replace it with `Cont<R, A>`.

i-walker (approved):
Awesome, and very concise

nomisRev:
I'm going to close this PR. Some other work has been done in, some review/discussion/changes happened there, and I'm currently documenting everything there.

After it's done, it will probably go to a new package `arrow.continuation` and might replace/deprecate `arrow.computations`.

Thanks for the review and feedback @i-walker!

PR #2575, opened 2021-11-09
Add head/tail destructuring declarations support for NonEmptyList

psteiger:
Allows for destructuring `NonEmptyList` into head and tail in an idiomatic fashion:

```
val (head, tail) = nonEmptyListOf(1, 2, 3, 5, 6)
```


serras:
@psteiger thanks for your collaboration, and sorry for taking so long to have a look at it! I've updated the API dump, which should make the test go green. Feel free to turn the PR into a "non-draft" state if you think it's ready.

psteiger:
Hi @serras , thanks for reviewing. 

In fact I tend to drop this PR, unfortunately, because of the problems discussed in the issue linked here. Mainly, the fact that Kotlin already provides `component1` up to `component5` operators as extensions on `List`, which are un-overridable (because they're extensions). 

So even though I strongly believe the best mental model for destructuring a `Nel` is by two, and only two, components (head and tail), I don't see how to conciliate this with Kotlin's stdlib.

serras:
@psteiger understood! feel free to close the PR or leave it open.

psteiger:
@serras I'll close it. 

By the way, that was a good talk on Arrow @ Kotlin by Jetbrains YouTube channel. 👍

PR #2584, opened 2021-11-25
Simplify Gradle layout, + KNIT, - ANK and fix site code fences

raulraja:
This PR is a joint effort with @JavierSegoviaCordoba and @nomisRev.

1. Simplifies the gradle setup into a single project with all the modules, still preserves the folder layout.
2. Moves a lot of the gradle config into https://github.com/arrow-kt/arrow-gradle-config 
3. Replaces Ank in favor of Knit and fixes all the docs broken snippets.
4. Works around many dokka bugs where Dokka is incorrectly rendering markdown and code fences. For example, everything processed by DokkaGFM was returned as HTML escaped strings.
5. Removes the need to have a `build/_site` which is where Ank was postprocessing the `docs` folder. @franciscodr We need you to please review the publishing workflow as the nesting in the docs does not exist anymore and most likely all steps related to removing the index file and copying the different versions need paths adjustments. I did my best but I don't understand how that works and I have no way to test it. Before this gets merged we need this final commit by @franciscodr ensuring the publishing of the site is going to work.


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Ior.kt
```diff
@@ -35,10 +35,12 @@ public sealed class Ior<out A, out B> {
    *
    * Example:
    * ```
+ * <!--- KNIT example-ior-01.kt -->
```
i-walker, line 38: question: 
This is generating empty kotlin files without the example
Was this intended or rather something like?
```markdown

* ```kotlin
* Left("tulip").isRight           // Result: false
* Right("venus fly-trap").isRight // Result: true
* Both("venus", "fly-trap").isRight // Result: false
* ```
<!--- KNIT example-ior-01.kt -->
```

nomisRev: Ye, that doesn't look right.

raulraja: Seems like Knit adds a bunch of empty file headers as this and we committed it since they compile. I'll do another pass today over empty files.

raulraja: Just pushed a bunch of knit fixes for the Ior file and others.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -748,12 +781,14 @@ public sealed class Either<out A, out B> {
    *
    * Example:
    * ```
+ * <!--- KNIT example-either-34.kt -->
```
i-walker, line 784: same here

i-walker: will take care of these small changes

franciscodr:
@raulraja @nomisRev You can check how the site will look in this [link](https://dlgjlm1httido.cloudfront.net/)

nomisRev (approved):
LGTM

i-walker (approved):
👍🏾 Nice

franciscodr (approved):
🚢

PR #2588, opened 2021-12-05
Implement filter function for Optional

lenguyenthanh:
The same functionality `filer` in Monocle [Optional.filter](https://www.optics.dev/Monocle/api/monocle/Optional$.html#filter[A](predicate:A=%3EBoolean):monocle.Optional[A,A])
- [x] Tests
- [x] More document
- [x] Update Api description
- [ ] Laws


serras:
Note that in optics libraries in other languages, filtering is usually not part of Optional, but of Fold. [See the Haskell version](https://hackage.haskell.org/package/optics-core-0.4/docs/Optics-AffineFold.html#v:filtered) for example. The reason is that the `set` operation is not so well-defined: can I set a value which does _not_ satisfy the predicate?

serras:
@lenguyenthanh I've turned this PR into a new one introducing `OptionalGetter` in #2611. Feel free to add your comments there.

lenguyenthanh:
@serras cool, I'll take a look. Should we close this pr now?

nomisRev:
@lenguyenthanh yes, I'm going to close this PR now. Feel free to re-open if necessary.
Thanks for the work, and the improvement that came from it 🙌

PR #2594, opened 2021-12-14
Fixing documentation workflow

franciscodr:
This pull request adapts the workflow to generate and publish the Arrow documentation to the new versioning plugin.


.github/workflows/publish.yml
```diff
@@ -58,6 +58,12 @@ jobs:
       - name: Build
         run: ./gradlew build
 
+      - name: Upload Arrow version file
+        uses: actions/upload-artifact@v2
+        with:
+          name: arrow-version
+          path: build/semver/version.txt
```
JavierSegoviaCordoba, line 65: This step can read the file and add the version to the next job:

https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idoutputs

This suggestion is based on the premise that uploading/downloading the file should be slower than just reading the file and setting the version to the output. If the time is the same then it doesn't matter.

.github/workflows/publish.yml
```diff
@@ -114,85 +121,97 @@ jobs:
...
+      - name: Set up Java
+        uses: actions/setup-java@v2.4.0
+        with:
+          distribution: 'zulu'
+          java-version: 11
+
+      - name: Download Arrow version file
+        uses: actions/download-artifact@v2
+        with:
+          name: arrow-version
```
JavierSegoviaCordoba, line 138: This is not required with the latest change, no?

franciscodr: True! Good spot!

.github/workflows/publish.yml
franciscodr: I think it's a good suggestion. I'll fix it

Issue #2535, opened 2021-09-23
Add Traverse and Sequence operations for kotlin.Sequences


pablisco:
We have traverse and Sequence operations for Iterable instances. We could probably add these for `kotlin.Sequence` instances too. This is specially nice for \`traverseResult\` and `sequenceResult`as we could short circuit the sequence when we get a failure :slightly\_smiling\_face: 

Additionally, we may want to check these for Flows too

pakoito:
Additionally indeed hahaha

pablisco:
@pakoito Too many things at the same time :sweat\_smile:

abendt:
i have created a PR that adds _sequenceResult_ and _traverseResult_

i-walker:
Can we close this @pablisco as the reason, why this promotes Anti-patterns is described here https://github.com/arrow-kt/arrow/pull/2602#issuecomment-1000769802

i-walker:
Closing this as Pablisco agreed on slack

PR #2601, opened 2021-12-19
add traverseResult and sequenceResult for kotlin.Sequence

abendt:
https://github.com/arrow-kt/arrow/issues/2535

this adds _sequenceResult_ and _traverseResult_ for _kotlin.Sequence_

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
@@ -695,7 +695,7 @@ public fun <E, A, B> Sequence<A>.traverseValidated(
   }
 }.map { it.asSequence() }
 
-public fun <A, B> Sequence<A>.traverseResult(f: (A) -> Result<B>): Result<Sequence<B>> {
+public inline fun <A, B> Sequence<A>.traverseResult(f: (A) -> Result<B>): Result<Sequence<B>> {
   val acc = mutableListOf<B>()
 
   forEach { a ->
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
@@ -621,6 +621,9 @@ public fun <A> Sequence<Option<A>>.sequenceOption(): Option<Sequence<A>> =
 public fun <E, A> Sequence<Validated<E, A>>.sequenceValidated(semigroup: Semigroup<E>): Validated<E, Sequence<A>> =
   traverseValidated(semigroup, ::identity)
 
+public fun <A> Sequence<Result<A>>.sequenceResult(): Result<Sequence<A>> =
+  traverseResult(::identity)
+
 public fun <A> Sequence<A>.some(): Sequence<Sequence<A>> =
   if (none()) emptySequence()
   else map { generateSequence { it } }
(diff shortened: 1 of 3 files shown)


nomisRev (approved):
Thanks @abendt 👍

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
```diff
@@ -692,6 +695,20 @@ public fun <E, A, B> Sequence<A>.traverseValidated(
   }
 }.map { it.asSequence() }
 
+public fun <A, B> Sequence<A>.traverseResult(f: (A) -> Result<B>): Result<Sequence<B>> {
```
nomisRev, line 698:
```suggestion
public inline fun <A, B> Sequence<A>.traverseResult(f: (A) -> Result<B>): Result<Sequence<B>> {
```

nomisRev (changes requested):
Actually, sorry @abendt. That function should be `inline` to allow for the suspension to go through.

abendt:
Another question @nomisRev. i also wanted to have a look at the _traverseX/sequenceX_ operators for Flows.
Is this a correct signature for e.g. the Result variant?

`public suspend fun <A, B> Flow<A>.traverseResult(f: suspend (A) -> Result<B>): Result<Flow<B>>`

nomisRev:
Well the `Sequence` signature is actually already a bit missleading since it's actually processing **all** the values of the `Sequence`/`Flow` before returning. So it's actually not a sequencing/streaming operator, but a terminal operator.

```
public suspend fun <A, B> Sequence<A>.traverseResult(f: suspend (A) -> Result<B>): Result<List<B>>
public suspend fun <A, B> Flow<A>.traverseResult(f: suspend (A) -> Result<B>): Result<List<B>>
```

abendt:
> Well the `Sequence` signature is actually already a bit missleading since it's actually processing **all** the values of the `Sequence`/`Flow` before returning. So it's actually not a sequencing/streaming operator, but a terminal operator.

true, that makes sense. Should i change the signature of the `Sequence.traverseResult` then too? Downside i see: there are the existing _traverseOption/Validated/..._  functions. These should ideally be changed too for consistency. Ofc this would be a breaking API change.

wdyt?

nomisRev (changes requested):
Hey @abendt,

> rue, that makes sense. Should i change the signature of the Sequence.traverseResult then too?
> Ofc this would be a breaking API change.

I've been thinking about it, and I think we should actually deprecate the current `traverse` functions on `Sequence` since they're missleading. They're the equivalent of `sequence.toList().traverseXXX`.

So these functions actually promote an anti-pattern, so I would actually be in favor of deprecating them all together instead of updating them since an alias for `toList().traverse` seems redundant.

Since 1.0.0 we have to be very careful as to what functionality/methods we add since we can not easily remove/change the API. Having bloated, very large APIs surface is a downside not rather than an upside.

What do you think?

nomisRev:
I looked into if we can make a sequence safe traverse function but since `SequenceScope` is designed as an `abstract class` with an `internal constructor` we cannot mix the effect as we could do if `SequenceScope` was an interface.
Otherwise we could create `interface EitherSequence<E, A> : EitherEffect<E>, SequenceScope<A>`.

raulraja:
Thanks, @abendt. I propose we put this on hold, including adding additional functions to Arrow until the maintainers regroup in January. There is a plan to meet to discuss the Arrow Roadmap for 2022, including a review of needed methods and others that will get deprecated before 2.0.

PR #2602, opened 2021-12-20
add Flow traverseX/sequenceX operators 

abendt:
https://github.com/arrow-kt/arrow/issues/2535


nomisRev:
I think with the same arguments of my comments on the Sequence PR in mind I think it's counterproductive to add this to Arrow since it promotes an anti-pattern. It's the equivallent of `toList().traverseXXX`.

https://github.com/arrow-kt/arrow/pull/2601#pullrequestreview-839854083

What do you think?

i-walker:
#2535 has been closed can we close this PR @abendt ? Sorry for the late response

PR #2606, opened 2021-12-22
Upgrade to kotlinx-coroutines 1.6.0

serras:


nomisRev:
😕

nomisRev:
@serras I'm actually not sure if the broken test was regression, or faulty on our end..
I rewrote the test, and this makes a lot more sense to me and the behavior is correct. What do you think?


cc\\ @raulraja

nomisRev:
I can try rewriting this tomorrow to a simple exampel share it on Kotlin Slack with the KotinX maintainers.

serras:
> Red 😕

Yep, my intention was to make it easy to collaborate in this. I managed to fix the compilation problems, but some of the problems with tests are a bit out of my league.

nomisRev (approved):

raulraja (approved):
thanks @serras @nomisRev !

PR #2611, opened 2021-12-24
[Optics] Optional getter

serras:
PR #2588 has shown that there's a missing element in the optics hierarchy: folds focus on 0..N, getters focus on exactly 1, so we are missing something which focuses on 0..1. This is `OptionalGetter`, the "get" part of `Optional` and `Prism`. In other libraries this is called an `AffineFold`. However, I find that name very obscure. Feel free to suggest changes.

In particular, the operation of filtering is one such `OptionalGetter`. The implementation is a copy of the one provided by @lenguyenthanh in #2588 (thank you for that!).

A technical note: you may notice that it is **not** the case that `Getter<S, A>` extends `OptionalGetter<S, A>`. This is because the right way to describe the relation is that `Getter<S, A>` is a `POptionalGetter<S, T, A>` **for any choice of `T`** (since it never returns it). Alas, Kotlin does not allow us to declare that relationship. One way to handle this would be to use something like `OptionalGetter<S, Nothing, A>`, but unfortunately the `T` must be invariant :(


arrow-site/docs/docs/optional_getter/README.md
```diff
@@ -0,0 +1,33 @@
+---
+layout: docs-optics
+title: Fold
```
nomisRev, line 3:
```suggestion
title: OptionalGetter
```

nomisRev (approved):
Looks good to me, a small mistake in the `title` of the markdown doc.

raulraja (approved):

Issue #2418, opened 2021-06-09
["Request"] Optic composition with standard Kotlin `KProperty`s


magneticflux-:
What version are you currently using?
0.13.2

What would you like to see?

Having extension functions similar to the existing ones like `Getter<A,B> + Getter<B,C> -> Getter<A,C>` would be super handy.

I found out that this code works (because `Getter` is a functional interface?):
```kotlin
fun example(getter: arrow.optics.Getter<T, V>) { }

fun main() {
    val prop: KProperty1<T, V> = TODO("External")
    example(prop)
}
```

So naturally now I want to be able to use typical optic compositions on the results.

This won't be able to replace the `@optics` annotation since it can't derive the equivalent `Setter`-type lenses, but it would make it easier to create optics on non-constructor parameters of data classes.

I am currently using this snippet:
```kotlin
operator fun <T, V, U> KProperty1<T, V>.plus(other: Getter<V, U>): Getter<T, U> {
    return Getter {
        other.get(this(it))
    }
}
```

nomisRev:
Hey @magneticflux-,

That's a great idea, and a great addition to Optics! If you're up for adding your snippet in a small PR that'd be awesome.

> (because Getter is a functional interface?):

Since `Getter<S, A>` is a functional interface it matches anything of the shape `(source: S) -> A`. 
`fun interface` works like `@FunctionalInterface` in Java where you can pass anything of the same shape of the single abstract method.

serras:
@magneticflux- since some time we have an `arrow-optics-reflect` which allows turning any `KProperty` into a lens. If you have time to have a look and give feedback, that would be awesome!

magneticflux-:
Looks good, thank you!

Something like Scala's implicit conversions would be awesome, but unfortunately I don't think that's on the roadmap for Kotlin since it's a bit niche.

PR #2612, opened 2021-12-24
Optics based on kotlin.reflect

serras:
The issue #2418 (by @magneticflux-) has given me an idea for an API for optics based on `KProperty`. It's a bit rough on the edges (for example, there's no way to check at compile-time that you are referring to a field in a `data class`, so the access may fail at runtime). On the other hand, it allows you to create optics easily for types you do not control:

```kotlin
data class Person(val name: String, val age: String, val friends: List<Person>)

val friendNames = Person::friends.every compose Person::name.lens
friendNames.modify(someFriends) { it.capitalize() }
```

The prism part is done using a combinator `instance<Parent, Children>`. Maybe a better name is needed, though.


arrow-libs/optics/arrow-optics/build.gradle.kts
```diff
@@ -26,12 +26,14 @@ kotlin {
     jvmMain {
       dependencies {
         implementation(libs.kotlin.stdlibJDK8)
+        api(libs.kotlin.reflect)
```
nomisRev, line 29: 😱

nomisRev:
As a rule Arrow never exports unsafe functions from the main APIs.
I was hoping to solve this issue with a compiler plugin, but they've been on hold for a long time since Optics was released.
Soon they might become stable though, and creating optics for classes you don't own will be automatically solved as a part of that.

With that in mind, that'd also mean we'd have to deprecate this.
We could try looking into doing this with [Reflekt](https://github.com/JetBrains-Research/reflekt) though.

nomisRev:
Or perhaps in a second module `arrow-optics-reflect` so it can be opted-in.

serras:
> Or perhaps in a second module `arrow-optics-reflect` so it can be opted-in.

I can do that, sure. The goal here was to provide a simple bridge which works for now.

nomisRev:
Yes, that makes perfect sense. We've historically done that in other projects too, there is a small reflection based module in Arrow Endpoints for `Schema` too but hopefully we can replace all those with nicer plugins next year 🥳

Thanks for adding this! When moved to `arrow-optics-reflect`, and perhaps a small addition in the docs we can merge 👍

serras:
@nomisRev I've just pushed a commit which moves this to a new `arrow-optics-reflect` package.

arrow-libs/optics/arrow-optics-reflect/build.gradle.kts
```diff
@@ -0,0 +1,32 @@
...
+apply(from = property("ANIMALSNIFFER_MPP"))
+
+kotlin {
+  sourceSets {
+    jvmMain {
+      dependencies {
+        api(projects.arrowCore)
+        api(projects.arrowOptics)
+        api(libs.kotlin.stdlibCommon)
+        implementation(libs.kotlin.stdlibJDK8)
```
nomisRev, line 19: So this exposes the API of Kotlin Std, and uses the JDK8 implementation?

I'm still not a 100% sure how this works, and what is the correct config :/ 
I don't we're doing different things in different modules.. but off-topic for this PR.

nomisRev (commented):
Yes, that looks perfect @serras! Thanks for moving it 👍
Some docs would be great too ❤️

Let me know when you think this is ready, and I'll review and ✔️

arrow-libs/optics/arrow-optics-reflect/build.gradle.kts
serras: I'm not sure 100% either. In any case, I've removed the `api` dependency on `stdlibCommon` and just left the `implementation` for the JDK version.

nomisRev:
@serras should become green after merging `main`.

nomisRev (approved):

raulraja (approved):

PR #2613, opened 2021-12-26
Introduce `combineAllOrNull` extension for `Iterable`

Quantum64:
We have found it useful to have an equivalent of the `combineAll` iterable extension for semigroups. The naming choice follows the direction of the standard library, where many iterable extensions use the `orNull` name suffix to denote an optional result.

Let me know what you think!


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -710,6 +710,12 @@ public fun <A> Iterable<A>.combineAll(MA: Monoid<A>): A = MA.run {
   }
 }
 
+public fun <A> Iterable<A>.combineOrNull(SG: Semigroup<A>): A? = SG.run {
```
nomisRev, line 713: Thanks @Quantum64 for your first contribution!

Awesome that you've followed naming of the Kotlin Std! 
Should we name it `combineAllOrNull` to be in line with the `combineAll` methods that relies on `Monoid`?

(CI is red due to some issues on Kotlin MPP & Github Actions cross-builds. We're working on it)

Quantum64: I agree that `combineAllOrNull` is a more appropriate name

nomisRev:
Looks great @Quantum64, but we should add tests for this before we can merge.
If you merge `main` CI should turn green again, it has finally been stabilised (🤞).

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -710,6 +710,12 @@ public fun <A> Iterable<A>.combineAll(MA: Monoid<A>): A = MA.run {
   }
 }
 
+public fun <A> Iterable<A>.combineAllOrNull(SG: Semigroup<A>): A? = SG.run {
```
raulraja, line 713: seeing this makes me think if the issue here is that what we call `combine` the std lib calls `reduce` in collections. If our semigroup op was called `reduce`, this and other functions that end up implemented in terms of reduce would be redundant. 

Given `Semigroup<A>` is a fun interface observing shape `(A, A) -> A` isn't `combineAllOrNull` the same as `reduceOrNull` ?

nomisRev: > Given Semigroup<A> is a fun interface observing shape (A, A) -> A isn't combineAllOrNull the same as reduceOrNull ?

Yes, I think you're correct!

i-walker:
Since this is a duplicate of `reduceOrNull` from the standard lib, can we deprecate this @Quantum64 ?

PR #2621, opened 2021-12-29
Non throwing Retrofit Either adapter

lukaszkalnik:
Currently implemented Retrofit adapters (`ArrowEitherCallAdapter` and `ArrowResponseECallAdapter`) only wrap HTTP errors (i.e. valid server responses with codes 4xx and 5xx) in `Left`.
They still throw exceptions in the following cases:
1. Loss of connection to network/server
2. Timeout
3. Malformed JSON
4. Response body is null (which is a valid use case when you get the response 204 No Content)

Especially for mobile apps this is problematic, as losing a connection or experiencing a timeout is quite a normal case for them. The application has to deal with it gracefully.

Therefore I decided to implement a Retrofit adapter which converts all possible exceptions to domain errors (i.e. wrapping them with `Left`).

This adapter requires the developer to define his Retrofit API passing the provided `CallError` type as the `Left` type argument for returned `Either`:

 ```
 import arrow.core.Either
 import arrow.retrofit.adapter.either.networkhandling.CallError
 import retrofit2.http.GET
 
 data class User(val name: String)

 interface MyService {

   @GET("/user/me")
   suspend fun user(): Either<CallError, User>
 }
 ```

The adapter then automatically wraps ocurring exceptions in appropriate `CallError` subclasses.

This adapter also allows for `null` response body in case the defined Retrofit API expects a `Unit` return type. This is useful e.g. when expecting a `204 No Content` response:

```
  // This allows a null response body
  @POST("/")
  suspend fun postSomething(@Body something: String): Either<CallError, Unit>
```

~~*Note that both `ArrowEitherCallAdapter` and `ArrowResponseECallAdapter` always throw when response body is null, effectively not allowing processing of responses like 204 No Content, which might be worth fixing.*~~ This will be fixed by #2625


hoc081098:
Very nice improvement

lukaszkalnik:
The CI build fails with the following message:
```
* What went wrong:
A problem occurred configuring project ':arrow-optics-test'.
> java.util.concurrent.TimeoutException
```

nomisRev:
@lukaszkalnik is this ready for review? I saw you were still making some improvements yesterday.

lukaszkalnik:
Yes @nomisRev , it is now finalized and ready for review. Looking forward to your review 👍

arrow-libs/core/arrow-core-retrofit/src/main/kotlin/arrow/retrofit/adapter/either/networkhandling/NetworkEitherCallAdapter.kt
```diff
@@ -0,0 +1,87 @@
...
+          Unit.right() as Either<CallError, R>
+        } else {
+          UnexpectedCallError(
+            IllegalStateException(
+              "Response code is ${code()} but body is null.\n" +
+                "If you expect response body to be null then define your API method as returning Unit:\n" +
+                "@POST fun postSomething(): Either<CallError, Unit>"
+            )
+          ).left()
+        }
```
nomisRev, line 59: I'm curious. What happens if I'd have `Person` as a `json` in `body` but it's optional.
So I'd like to have `suspend fun fetchOptionalPerson(id: UUID): Either<CallError, Person?>`, would that be possible in this case?

nomisRev (approved):
Looks great to me @lukaszkalnik!

I am wondering what you think about `ResponseE` because it seems redundant to me in comparison to both `Either` adapters. One could also do `Either<CallError, Response>`, no?

arrow-libs/core/arrow-core-retrofit/build.gradle.kts
```diff
@@ -2,6 +2,7 @@ plugins {
   id(libs.plugins.kotlin.jvm.get().pluginId)
   alias(libs.plugins.arrowGradleConfig.kotlin)
   alias(libs.plugins.arrowGradleConfig.publish)
+  alias(libs.plugins.kotlinx.serialization) // Needed for testing only
```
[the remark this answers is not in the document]
lukaszkalnik: Kotlinx Serialization consists of two parts: [a compiler plugin and a library](https://github.com/Kotlin/kotlinx.serialization#setup).
This is only the compiler plugin, it is needed for the code generation for the `@Serializable` annotated `ResponseMock` class used in tests.
If you scroll down, `dependencies` you will see the JSON serialization library as a `testImplementation` dependency.

lukaszkalnik:
> I am wondering what you think about `ResponseE` because it seems redundant to me in comparison to both `Either` adapters. One could also do `Either<CallError, Response>`, no?

Yes, but to have information about the type of the parsed body you would need to nested parameterize `Response`:
```
import retrofit2.Response

@GET("/")
suspend fun getResponseE(): Either<ErrorMock, Response<ParsedType>>
```
I guess it's there for convenience.

arrow-libs/core/arrow-core-retrofit/src/main/kotlin/arrow/retrofit/adapter/either/networkhandling/NetworkEitherCallAdapter.kt
lukaszkalnik: No, unless your json parsing Converter (which you set when configuring your Retrofit instance) allows `null` bodies.
All 3 converters I tested throw an exception in this case. Maybe you can configure it for some converters, I didn't check.

Also it doesn't seem like a valid REST idiom to me. There are no "optional" resources. If a `Person` has been found, server will return a `Person`. If it hasn't been found, server should return "404 Not Found" or a similar error response.

nomisRev: Right, that makes perfect sense! Thanks for the clear reply!

> Also it doesn't seem like a valid REST idiom to me.

Anything out of the 2xx range is not considered an `Either.Right` 👍

nomisRev:
This looks good to me @lukaszkalnik! Thanks for the great work, and contribution.
Let's try to get a second review on this.

Perhaps @pakoito, @pablisco, @LordRaydenMK, or @JorgeCastilloPrz can give a good review here? 🙏 I think those are the contributors that are most familiar with Retrofit.

arrow-libs/core/arrow-core-retrofit/src/main/kotlin/arrow/retrofit/adapter/either/networkhandling/NetworkEitherCallAdapter.kt
```diff
@@ -0,0 +1,87 @@
...
+
+      override fun onResponse(call: Call<R>, response: Response<R>) {
+        callback.onResponse(this@EitherCall, Response.success(response.toEither()))
+      }
+
+      private fun Response<R>.toEither(): Either<CallError, R> {
+        // Http error response (4xx - 5xx)
+        if (!isSuccessful) {
+          val errorBody = errorBody()?.string() ?: ""
+          return HttpError(code(), errorBody).left()
```
LordRaydenMK, line 40: this should probably use `requireNotNull` or `!!`.

AFIK it's not possible to construct an error response without a body: https://github.com/square/retrofit/blob/master/retrofit/src/main/java/retrofit2/Response.java

LordRaydenMK (commented):
Looks good to me, left a few questions/suggestions

arrow-libs/core/arrow-core-retrofit/src/main/kotlin/arrow/retrofit/adapter/either/networkhandling/CallError.kt
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Http request returned an error response.
+ */
+public data class HttpError(val code: Int, val body: String) : CallError()
+
+/**
+ * IO error: no network, socket timeout, malformed JSON etc. Check the [cause] for details.
+ */
+public data class IOError(val cause: IOException) : CallError()
```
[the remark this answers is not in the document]
lukaszkalnik: True, this will depend on the deserializing adapter the programmer is using.
* GSON:
```
    Either.Left(IOError(cause=com.google.gson.stream.MalformedJsonException: Use JsonReader.setLenient(true) to accept malformed JSON at line 1 column 1 path $))
```
* Moshi:
```
    Either.Left(IOError(cause=com.squareup.moshi.JsonEncodingException: Use JsonReader.setLenient(true) to accept malformed JSON at path $))
```
* Kotlinx serialization:
```
Either.Left(UnexpectedCallError(cause=kotlinx.serialization.json.internal.JsonDecodingException: Expected start of the object '{', but had 'EOF' instead
    JSON input: not a valid JSON))
```
So it's impossible to write some generic documentation. I can simply remove this statement.

arrow-libs/core/arrow-core-retrofit/src/main/kotlin/arrow/retrofit/adapter/either/networkhandling/NetworkEitherCallAdapter.kt
lukaszkalnik: Ok, I see that the ` error(ResponseBody body, okhttp3.Response rawResponse)` factory method indeed uses `requireNonNull(body)`.
What wonders me is why then the Retrofit public API marks the `errorBody()` as nullable.

arrow-libs/core/arrow-core-retrofit/src/main/kotlin/arrow/retrofit/adapter/either/networkhandling/CallError.kt
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Error hierarchy when calling remote server.
+ */
+public sealed class CallError
+
+/**
+ * Http request returned an error response.
+ */
+public data class HttpError(val code: Int, val body: String) : CallError()
```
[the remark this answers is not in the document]
lukaszkalnik: As this adapter is really meant just for translating the HTTP responses into the domain layer, I don't think we need to pass full responses.
Usually apps have OkHttp logging activated anyway so full responses are visible in logs.
I think including the `message` is a good idea though.

arrow-libs/core/arrow-core-retrofit/src/main/kotlin/arrow/retrofit/adapter/either/networkhandling/NetworkEitherCallAdapter.kt
```diff
@@ -0,0 +1,91 @@
...
+    object : Callback<R> {
+
+      override fun onResponse(call: Call<R>, response: Response<R>) {
+        callback.onResponse(this@EitherCall, Response.success(response.toEither()))
+      }
+
+      private fun Response<R>.toEither(): Either<CallError, R> {
+        // Http error response (4xx - 5xx)
+        if (!isSuccessful) {
+          val errorBody = errorBody()!!.string()
```
hoc081098, line 39: IMO
```kotlin
val errorBody = checkNotNull(errorBody()) { "Response.errorBody() must be not null!" }.string()
```

lukaszkalnik: Yes, I was also thinking about this. But it's impossible for `errorBody` to be null, unless a bug is introduced in Retrofit (in which case not only this adapter will have problems, but the whole Retrofit based world will break in some way).

`Response<T>` has fields `@Nullable T body` and `@Nullable ResponseBody errorBody`. The reason they are both nullable is that always only one of them is not null. In case of an error response (which in my adapter is checked by `if (!isSuccessful)`, `errorBody` is guaranteed by Retrofit implementation logic to be non-null. Look at the [implementation](https://github.com/square/retrofit/blob/9b32c965ef31275627e35a862f584bcc7b9fa4e5/retrofit/src/main/java/retrofit2/Response.java#L90) of both `Response.error()` methods - they both already perform the check.

LordRaydenMK (approved):

arrow-libs/core/arrow-core-retrofit/src/main/kotlin/arrow/retrofit/adapter/either/networkhandling/NetworkEitherCallAdapter.kt
hoc081098: That makes sense

arrow-libs/core/arrow-core-retrofit/src/main/kotlin/arrow/retrofit/adapter/either/networkhandling/NetworkEitherCallAdapter.kt
hoc081098: That makes sense

nomisRev:
@arrow-kt/maintainers should add @lukaszkalnik and @LordRaydenMK to the maintainers team?

This PR looks good to me, and ready to be merged. Thanks for the review @LordRaydenMK & @hoc081098! 🙏

andre-krueger:
What about having a route that can return a 500 with no body, but for other failure status codes the body is of a custom type, that you want to match in the error handling?
When using CallError, that information is lost and you would have to deserialize the body from the HttpError class.

The only solution I can think of is to use an OkHttp Interceptor, that throws a custom exception when the response code is 500 and then handle that with Either.catch in the api methods. That way one can keep the custom body.

Any other solution?

lukaszkalnik:
> What about having a route that can return a 500 with no body, but for other failure status codes the body is of a custom type, that you want to match in the error handling? 
> When using CallError, that information is lost and you would have to deserialize the body from the HttpError class.

It's not lost, error code and body are still in the `HttpError` class, only you have to parse them manually.

> 
> The only other solution I can think of is to use an OkHttp Interceptor, that throws a custom exception when the response code is 500 and then handle that with Either.catch in the api methods. That way one can keep the custom error type.
> 
> Any other solution?

You never have guarantees about what really comes inside the error body. As Jake Wharton writes [here](https://github.com/square/retrofit/issues/2118#issuecomment-265800627), errors may come from proxies, wifi portals, applications, you may suddenly have a human-readable HTML page inside the error response, and now you're dealing with two errors.

I'm not sure if making this adapter more complex by allowing custom response body type is the way to go. I would rather have a choice - for general usage just take this adapter. When you need project-specific response body parsing then I would suggest creating your own adapter which supports that.

The main premise of this adapter is to avoid throwing exceptions in case of network errors and to handle them as domain errors.

> 
> [This](https://github.com/arrow-kt/arrow/blob/main/arrow-libs/core/arrow-core-retrofit/src/main/kotlin/arrow/retrofit/adapter/either/eitherAdapter.kt#L25-L27) doesn't seem to check if the errorBody actually contains any data (only that errorBody is not null), which is not the case when there is no body. With that check in place and declaring the Left type (the custom error type) of the api method as optional, would that be a solution?

This is not my code. I'm not sure what you mean here.

lukaszkalnik:
@andre-krueger Ok, I think I understood what you mean.
You want to keep the error type free to define by the user of this adapter, so that in case of a strictly defined HTTP error body (for 4xx/5xx responses) it can be automatically parsed.

This unfortunately won't work for the following reasons:
1. The error type has to be opinionated if this adapter should convert network exceptions (like connection lost, timeout, malformed JSON etc.) to domain errors. This means that the adapter can only return an error as an instance of the `sealed class` hierarchy which contains the `IOError` and `UnexpectedCallError` types (which wrap the network exceptions). If we let the user define the error type, then we have no way to enforce having these types in the hierarchy which we could then use to wrap the network errors.
2. Currently we have 3 adapters in this module, but we only have one factory for them. Which adapter is instantiated is dynamically checked based on the declared wrapper type (`Either` or `ResponseE`) and error type. This adapter is automatically created by the factory when the user declares `CallError` as error type. I admit that this can be trivially solved with just creating a custom factory for this adapter, which always returns exactly this adapter.

lukaszkalnik:
@nomisRev will there be some further review on this or could this be merged?

andre-krueger:
@lukasz-kalnik-gcx 
Ok.
When playing around with the EitherCallAdapterFactory code, it wasn't possible to return an empty ResponseBody [here](https://github.com/arrow-kt/arrow/blob/main/arrow-libs/core/arrow-core-retrofit/src/main/kotlin/arrow/retrofit/adapter/either/eitherAdapter.kt#L34) anyway (common for 500 status codes, which normally don't have any response body).
When that PR is merged, I think it's the best solution to use the CallError class and just do the deserialization for routes that both can return 500 and a custom error body.

andre-krueger:
I was using an express server to test the 500 response and was sending this for the route:
`res.status(500).send()`

That's what curl replied with, so it seems like an empty body.

```
< HTTP/1.1 500 Internal Server Error
< X-Powered-By: Express
< Date: Fri, 11 Feb 2022 12:49:37 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
< Content-Length: 0
```

So it failed [here](https://github.com/arrow-kt/arrow/blob/main/arrow-libs/core/arrow-core-retrofit/src/main/kotlin/arrow/retrofit/adapter/either/eitherAdapter.kt#L30) with an exception (it was EOFException, when I remember correctly).

lukasz-kalnik-gcx:
> I was using an express server to test the 500 response and was sending this for the route: `res.status(500).send()`
> 
> That's what curl replied with, so it seems like an empty body.
> 
> ```
> < HTTP/1.1 500 Internal Server Error
> < X-Powered-By: Express
> < Date: Fri, 11 Feb 2022 12:49:37 GMT
> < Connection: keep-alive
> < Keep-Alive: timeout=5
> < Content-Length: 0
> ```
> 
> So it failed [here](https://github.com/arrow-kt/arrow/blob/main/arrow-libs/core/arrow-core-retrofit/src/main/kotlin/arrow/retrofit/adapter/either/eitherAdapter.kt#L30) with an exception (it was EOFException, when I remember correctly).

Yes, so the body was indeed empty and whatever JSON converter you have registered failed parsing it and threw the EOFException.
You can mitigate it with a custom converter like described in this [Stackoverflow question](https://stackoverflow.com/questions/56135835/how-can-i-convert-empty-objects-as-null-with-moshi-and-retrofit).

andre-krueger:
Unfortunately, no luck with that.
Would have been good if it was as easy as to make the Either.Left error type nullable in the Retrofit api definitions, in order to allow an empty error body.

I will wait until these changes are merged - until now I just return the custom error in an interceptor in case of the 500 status response.

raulraja (approved):
great addition, thanks @lukaszkalnik  !

i-walker (approved):
Thanks @lukaszkalnik 👍🏾

nomisRev:
Thank you so much @lukasz-kalnik-gcx for you patience, and the great addition 🙏 Finally merged!

PR #2624, opened 2022-01-04
Fix Retrofit adapter test configuration

lukaszkalnik:
Current test configuration reuses one `MockWebServer` instance across all tests. This is incorrect, as any changes to the mock server configuration influence subsequent tests.
To experience the problem on the current `main` branch, change the test order in that you move any test asserting a successful response, e.g. `should return ResponseMock for 200 with valid JSON`, after the last test `should throw when server disconnects`. The test for a successful response will now start failing.

This PR fixes that in that the `server` and `service` variables have been made `lateinit var`s and are initialized before any test.
To test if this fixes the problem, just change the test order in the way described above on this PR's branch and see that the tests all still pass regardless of their order.

I have also renamed the confusingly named `SuspedApiClientTest` to `SuspendApiTestClient`. The old name suggested that it was a test (because of the `Test` suffix) when in reality it is a test client.


nomisRev (approved):

lukaszkalnik:
@nomisRev the build is failing with the following error:
```
* What went wrong:
A problem occurred configuring project ':arrow-optics-test'.
> java.util.concurrent.TimeoutException
```

nomisRev:
This should be solved now @lukaszkalnik if you merge `main`.

PR #2625, opened 2022-01-04
Retrofit adapter: allow null response body for methods returning Unit

lukaszkalnik:
If the developer defined some service methods as returning `Unit` it means he's not interested in the response body. It will be mostly the case for responses like `204 No Content` for which the server may return a null response body.

The current implementation will always throw an exception when null response body has been received, regardless of the expected return type defined by the developer.

This PR fixes that. It adds support for receiving `null` response body for methods defined like in the example below (note that the right type is defined as `Unit`):
```
@POST("/")
suspend fun postSomething(@Body something: String): Either<Error, Unit>
```


nomisRev (approved):
Thanks for the great addition @lukaszkalnik!

raulraja (approved):
thanks @lukaszkalnik !

i-walker (approved):

PR #2631, opened 2022-01-10
Update README.md

i-walker:
- remove Snapshot publication in docs

### DIFF
--- a/arrow-site/docs/docs/core/README.md
+++ b/arrow-site/docs/docs/core/README.md
@@ -268,7 +268,7 @@ configuration:
 ```xml
 
 <repository>
-    <snapshotss>
+    <snapshots>
         <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
         <updatePolicy>always</updatePolicy>
     </snapshots>
--- a/arrow-site/docs/docs/core/README.md
+++ b/arrow-site/docs/docs/core/README.md
@@ -138,20 +138,7 @@ dependencies {
 
 #### Next development version
 
-If you want to try the latest features, replace `1.0.1` with `1.0.2-SNAPSHOT` and add this
-configuration:
-
-```
-allprojects {
-    repositories {
-        ...
-        maven(url = "https://oss.sonatype.org/content/repositories/snapshots/")
-    }
-
-    // To use latest artifacts
-    configurations.all { resolutionStrategy.cacheChangingModulesFor(0, "seconds") }
-}
-```
+If you want to try the latest features, replace `1.0.1` with on of the latest `alpha`, `beta` or `rc` publications.
 
 </div>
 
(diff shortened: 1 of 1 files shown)


franciscodr:
Task linked: [CU-1zv9arp test Maven fix](https://app.clickup.com/t/1zv9arp)

arrow-site/docs/docs/core/README.md
```diff
@@ -234,7 +234,7 @@ Add the dependencies that you want to use:
 
 <dependency>
     <groupId>io.arrow-kt</groupId>
-    <artifactId>arrow-core</artifactId>
+    <artifactId>arrow-core-jvm</artifactId>
```
nomisRev, line 237: The fix is that we fix the bug in publishing, so this `-jvm` isn't needed anymore.
This should be fixed in the latest alpha release, can you check in the newlly created Maven template?

i-walker: the alpha release isn't published, the CI commit is broken, as the tag has to be updated

i-walker: Just pushed the latest one. I am trying it out

i-walker: It works in the latest Snapshot 1.0.2 only

i-walker: The only problem is that it doesn't load the jvm binaries, so for it to work we still need the `-jvm`. But from i see in sonatype the last snapshot was released in November https://oss.sonatype.org/content/repositories/snapshots/io/arrow-kt/arrow-core/ for `arrow-core`

nomisRev: We're not releasing SNAPSHOT anymore, we're instead releasing fixed alpha version to mavenCentral.

nomisRev: Strange, afaik for KotlinX Coroutines it's possible to pull the JVM artifact from Maven without `-jvm`.
Users have also reported that this worked for 1.0.0, but broke in 1.0.1

i-walker: Just posting this for everyone, 
the maven support may work with `1.0.2-alpha.43`

arrow-site/docs/docs/core/README.md
```diff
@@ -268,8 +268,8 @@ configuration:
 ```xml
 
 <repository>
-    <snapshotss>
-        <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
+    <url>https://oss.sonatype.org/content/repositories/snapshots/</url>
+    <snapshots>
         <updatePolicy>always</updatePolicy>
     </snapshots>
```
raulraja, line 274:
```suggestion
```
We are no longer producing snapshots so this piece of config should not be required.

raulraja (approved):

nomisRev (approved):

PR #2633, opened 2022-01-11
Propagate Arrow version to publish_doc job

franciscodr:
The `publish_doc` can't access the Arrow version defined as an output for the `build` job because it doesn't depend on that job directly. So we need to propagate that value from the `publish` job to be able to read the version of the library.


i-walker (approved):
Thanks Paco 🙌🏾

raulraja (approved):

PR #2636, opened 2022-01-12
[Optics] Show error for generic classes in KSP plug-in

serras:

PR #2640, opened 2022-01-14
revert ksp plugin to jvm only

i-walker:
Since #2636 alpha releases are broken, as there are no mpp dependencies for ios or other native targets
e.g.: :arrow-optics-ksp-plugin:generateMetadataFileForIosArm32Publication  fails


nomisRev:
Doesn't this break KSP MPP usage?

nomisRev (approved):

PR #2660, opened 2022-02-01
Porting Control

i-walker:
all of this was possible by https://github.com/nomisRev/Continuation proposal

This is solely porting it into arrow. In the following PR's I'll add depreciations and rewrites


nomisRev:
We should also move the tests, there is a couple of suites to verify it correctly works with Arrow Fx, KotlinX Coroutines, and its base behavior.

i-walker:
Going to start a new branch and PR where the git history of   https://github.com/nomisRev/Continuation  is included :D

Issue #2547, opened 2021-10-04
["Request"] ResultEffect leaks to the global namespace


pakoito:
What version are you currently using?
1.0.1

What would you like to see?
`arrow.core.computations.ResultEffect.bind` is available globally, so you can do disallowed things such as

```
Either.catch {
    either.bind()
}
```

nomisRev:
Ye, it's indeed publically available as through the `object`. Which exposes the behavior in an unsafe way.
We should fix this without making a binary breaking change 😅

Think we could either make it `internal` after deprecation of the `public` object, this would result in least changes and would keep the JVM binary the same.

Or we should deprecate it completely making it private, but without any breakage in the Kotlin API whilst keeping the old JVM API around as `DeprecationLevel.HIDDEN`.

nomisRev:
It's only been released in `1.0.0` should it shouldn't have much usage out there.

pakoito:
Ye go for it IMO

nomisRev:
@i-walker the `result { }` builder is still missing from the new `arrow.core.continuations` package. Correct?
We should add it there, and deprecate whatever we have in the current `result { }` builder.
Then we can take this issue into account, and get rid of it.

I've heard more reports of it being problematic. You can also do a global import for `Either<Throwable, A>.bind()` through this code that throws 😱

i-walker:
I am on it @nomisRev, it will be available here https://github.com/arrow-kt/arrow/pull/2669

i-walker:
Can we close this PR @pakoito since it has been deprecated with a Warning in https://github.com/arrow-kt/arrow/pull/2669?

pakoito:
Yay! Go for it!

PR #2661, opened 2022-02-03
Transfer Cont from @nomisrev to arrow

i-walker:
Thanks to @nomisRev proposal 🙌🏾
I've slightly changed the names from `Cont` to `Effect` and `ContEffect` to `EffectScope`, e.g.:

```kotlin
class DomainError(msg: String) 

context(CoroutineScope, EffectScope<DomainError>)
suspend fun program(): Int {
    val fa = async<Int> { shift(DomainError("error")) }
    val fb = async { delay(2000); 1 }
    return fa.await() + fb.await()
}
```

Feedback is welcome ☺️


raulraja (approved):
thanks @nomisRev @i-walker !

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
```diff
@@ -0,0 +1,242 @@
...
+  public suspend fun toOption(orElse: suspend (R) -> Option<A>): Option<A> = fold(orElse, ::Some)
+
+  /**
+   * [fold] the [Effect] into an [A?]. Where the shifted value [R] is mapped to
+   * [null], and result value [A].
+   */
+  public suspend fun orNull(): A? = fold({ null }, ::identity)
+
+  /** Runs the [Effect] and captures any [NonFatal] exception into [Result]. */
+  public suspend fun attempt(): Effect<R, Result<A>> = effect {
```
nomisRev, line 104:
```suggestion
  public fun attempt(): Effect<R, Result<A>> = effect {
```

All functions that return `Effect` should be non-suspending.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
```diff
@@ -0,0 +1,242 @@
...
+    try {
+      Result.success(bind())
+    } catch (e: Throwable) {
+      Result.failure(e.nonFatalOrThrow())
+    }
+  }
+
+  public suspend fun handleError(recover: suspend (R) -> A): Effect<Nothing, A> = effect {
```
nomisRev, line 112:
```suggestion
  public fun handleError(recover: suspend (R) -> A): Effect<Nothing, A> = effect {
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
```diff
@@ -0,0 +1,242 @@
...
+    fold(recover, ::identity)
+  }
+
+  public suspend fun <R2> handleErrorWith(recover: suspend (R) -> Effect<R2, A>): Effect<R2, A> = effect {
```
nomisRev, line 116:
```suggestion
  public fun <R2> handleErrorWith(recover: suspend (R) -> Effect<R2, A>): Effect<R2, A> = effect {
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
```diff
@@ -0,0 +1,242 @@
...
+    fold({ recover(it).bind() }, ::identity)
+  }
+
+  public suspend fun <B> redeem(recover: suspend (R) -> B, transform: suspend (A) -> B): Effect<Nothing, B> =
```
nomisRev, line 120:
```suggestion
  public fun <B> redeem(recover: suspend (R) -> B, transform: suspend (A) -> B): Effect<Nothing, B> =
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
```diff
@@ -0,0 +1,242 @@
...
+    effect {
+      fold(recover, transform)
+    }
+
+  public suspend fun <R2, B> redeemWith(
```
nomisRev, line 125:
```suggestion
  public fun <R2, B> redeemWith(
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
```diff
@@ -0,0 +1,242 @@
...
+// `recover`.
+@PublishedApi
+internal class Internal(val token: Token, val shifted: Any?, val recover: suspend (Any?) -> Any?) :
+  ShiftCancellationException() {
+  override fun toString(): String = "ShiftCancellationException($message)"
+}
+
+// Class that represents a unique token by hash comparison
+@PublishedApi
+internal class Token {
```
nomisRev, line 148: Can we re-use the existing `Token` that is already in Arrow?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
```diff
@@ -0,0 +1,242 @@
...
+ *     val x = Either.Right(1).bind()
+ *     val y = Validated.Valid(2).bind()
+ *     val z: Int = None.bind { "Option was empty" }
+ *     x + y + z
+ *   }.fold({ it shouldBe "Option was empty" }, { fail("Int can never be the result") })
+ * }
+ * ```
+ * <!--- KNIT example-effect-02.kt -->
+ */
+public suspend inline fun <R, A> effect(crossinline f: suspend EffectScope<R>.() -> A): Effect<R, A> =
```
nomisRev, line 208:
```suggestion
public inline fun <R, A> effect(crossinline f: suspend EffectScope<R>.() -> A): Effect<R, A> =
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/option.kt
```diff
@@ -0,0 +1,30 @@
...
+
+import arrow.core.None
+import arrow.core.Option
+import arrow.core.Some
+import arrow.core.identity
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.contract
+import kotlin.jvm.JvmInline
+
+public suspend inline fun <A> option(crossinline f: suspend OptionEffectScope.() -> A): Option<A> =
```
nomisRev, line 11: I think we need`option { }` and `option.eager { }` to replace the existing behavior in `arrow.core.computations`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/ior.kt
```diff
@@ -0,0 +1,44 @@
+package arrow.core.continuations
+
+import arrow.continuations.generic.AtomicRef
+import arrow.continuations.generic.updateAndGet
+import arrow.core.EmptyValue
+import arrow.core.Ior
+import arrow.core.identity
+import arrow.typeclasses.Semigroup
+
+public suspend inline fun <E, A> ior(
```
nomisRev, line 10: We should also expose `object ior` with `suspend operator fun invoke` and `fun eager` like the other builders

EffectAndStructuredConcurrency.md
```diff
@@ -0,0 +1,202 @@
+# Semantics of Structured Concurrency + Effect<R, A>
```
nomisRev, line 1: I don't think this doc will show up on the Arrow Website this way.
Do we need to move it to the KDoc of the `Effect` interface? Or do we need to put it in a specific place and include it into the sidebar of the Arrow Core website?

nomisRev (changes requested):
Thank you @i-walker for moving this code here 🙏
Couple of things still look incorrect. Great rename to `Effect` and `EffectScope` I like it!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
i-walker: the only difference is that the one in arrow.generics is not transforming the hashCode with an unsigned Integer. If that's fine, I'll use it. Otherwise I could also change the one in arrow.generics?

nomisRev: Ah, it's in `arrow-continuations` 👍 
That module will be deprecated, and dissapear over time so lets keep this one instead.

It's not necessary to call `toUInt()` first and so we can improve this code with `override fun toString(): String = "Token(${hashCode().toString(16)})"`. It was needed in an older Kotlin version IIRC.

EffectAndStructuredConcurrency.md
i-walker: It's now in Arrow Core and Fx

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/option.kt
i-walker: I've exposed an eager version, but that would only point to eagerEffect with no dsl changes on top as in `option { }`
We could do it if the eagerEffect `shift` is not suspend.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffectScope.kt
```diff
@@ -45,6 +46,9 @@ public interface EagerEffectScope<R> {
       is Some -> value
     }
 
+  public suspend fun <B> Eval<B>.bind(): B =
+    value()
```
nomisRev, line 50: This is redundant, since you can just use `value()` instead of `bind`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EffectScope.kt
```diff
@@ -174,10 +175,34 @@ public interface EffectScope<R> {
...
+   *   effect<String, Int> {
+   *     val x: Int = eval.bind()
+   *     x
+   *   }.fold({ default }, ::identity) shouldBe eval.value()
+   * }
+   * ```
+   * <!--- KNIT example-effect-scope-08.kt -->
+   */
+  public suspend fun <B> Eval<B>.bind(): B =
+    value()
```
nomisRev, line 205: idem

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/option.kt
```diff
@@ -30,10 +30,30 @@ public value class OptionEffectScope(private val cont: EffectScope<None>) : Effe
...
+  public suspend fun <B> ensureNotNull(value: B?): B =
+    ensureNotNull(value) { None }
+}
+
 @Suppress("ClassName")
 public object option {
-  public inline fun <A> eager(crossinline f: suspend EagerEffectScope<None>.() -> A): Option<A> =
-    eagerEffect(f).toOption()
+  public inline fun <A> eager(crossinline f: suspend OptionEagerEffectScope.() -> A): Option<A> =
+    @Suppress("ILLEGAL_RESTRICTED_SUSPENDING_FUNCTION_CALL")
```
nomisRev, line 55: Why is this needed here?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/result.kt
```diff
@@ -0,0 +1,51 @@
+package arrow.core.continuations
+
+import arrow.core.Either
+import arrow.core.Validated
+import arrow.core.identity
+
+public object ResultEffectScope {
```
nomisRev, line 7: We cannot expose this as a `public object`, see issue: https://github.com/arrow-kt/arrow/issues/2547

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/result.kt
```diff
@@ -0,0 +1,51 @@
...
+
+  public fun <A> Either<Throwable, A>.bind(): A =
+    fold({ throw it }, ::identity)
+
+  public fun <A> Validated<Throwable, A>.bind(): A =
+    fold({ throw it }, ::identity)
+}
+
+@Suppress("ClassName")
+public object result {
```
nomisRev, line 20: If we only have the `inline `version, we probably don't need the `object` indirection here.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/continuations/EagerEffectSpec.kt
```diff
@@ -0,0 +1,164 @@
...
+    val e = RuntimeException("test")
+    Either.catch { eagerEffect<Nothing, Nothing> { throw e }.runCont() } shouldBe Either.Left(e)
+  }
+
+  "Rethrows suspended exceptions" {
+    val e = RuntimeException("test")
+    Either.catch { eagerEffect<Nothing, Nothing> { throw e }.runCont() } shouldBe Either.Left(e)
+  }
+
+  // Fails https://github.com/Kotlin/kotlinx.coroutines/issues/3005
```
nomisRev, line 139: Can we fix this by not using the method reference for `inline suspend` somewhere?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/option.kt
i-walker: removed it 👍🏾

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/result.kt
i-walker: removed it 👍🏾

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/continuations/EagerEffectSpec.kt
i-walker: yeah this was not failing

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/option.kt
i-walker: Updated it with an dsl

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/ior.kt
i-walker: Done 👍🏾

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
```diff
@@ -0,0 +1,114 @@
...
+    f: (R) -> EagerEffect<R2, B>,
+    g: (A) -> EagerEffect<R2, B>
+  ): EagerEffect<R2, B> = eagerEffect { fold(f, g).bind() }
+}
+
+@PublishedApi
+internal class Eager(val token: Token, val shifted: Any?, val recover: (Any?) -> Any?) :
+  ShiftCancellationException() {
+  override fun toString(): String = "ShiftCancellationException($message)"
+}
```
nomisRev, line 73: I think we should use the same trick here as we do for `Effect`, so users have an escape hatch to differentiate between `JobCanncellationException` and `ShiftCanncelationException` when they have to in low-level niche use-cases.

arrow-site/docs/docs/arrow/core/continuations/README.md
```diff
@@ -0,0 +1,202 @@
+# Semantics of Structured Concurrency + Effect<R, A>
```
nomisRev, line 1: This is not the documentation we want, this was part of the design process/documentation.
What we want instead is the [README.MD](https://github.com/nomisRev/Continuation/blob/main/README.md)

I'm thinking we should make it part of the KDoc on top of `Effect`, and link the page to `Effect` as well.

Currently, a lot of users have a hard time understanding the difference between `Effect` and `EagerEffect` and when or why they should be using `either { }` vs `either.eager { }`. This is currently completely undocumented inside `arrow-continuations`/`arrow.core.computations`.

I think we should add proper documentation in this PR before we merge, otherwise, it's going to be dropped to the bottom of the backlog again.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
i-walker: added tests in EagerEffect and Effect for low-level use-cases

arrow-site/docs/docs/arrow/core/continuations/README.md
i-walker: Updated most places I was aware of, can I get another review

i-walker:
I removed `EagerEffect<R, A>.toEffect(): Effect<R, A>`, because `effect { }` and `eagerEffect { }` can fold or `bind()` an `EagerEffect` value and it would introduce multiple paths of how to operate with this value doing something like `toEffect().bind()`, even though `bind()` is sufficient.

nomisRev (approved):
Great work @i-walker! Thanks for all the great effort moving this into Arrow 🥳 👏 👏 👏

arrow-site/docs/_data/sidebar-fx.yml
```diff
@@ -20,6 +20,23 @@ options:
...
+        url: /apidocs/arrow-core/arrow.core.continuations/-effect-scope/
+
+      - title: arrow.core.continuations.EagerEffect
+        url: /apidocs/arrow-core/arrow.core.continuations/-eager-effect/
+
+      - title: arrow.core.continuations.EagerEffectScope
+        url: /apidocs/arrow-core/arrow.core.continuations/-eager-effect-scope/
+
+      - title: Semantics of Structured Concurrency and Effect
+        url: /arrow/core/continuations/
```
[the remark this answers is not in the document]
i-walker: Makes sense 👍🏾

arrow-site/docs/_data/sidebar-core.yml
```diff
@@ -36,6 +36,25 @@ options:
       - title: Extensions
         url: /apidocs/arrow-core/arrow.core/index.html#functions
 
+  - title: Continuations and programs
+
+    nested_options:
+
+      - title: arrow.core.continutions.Effect
+        url: /apidocs/arrow-core/arrow.core.continuations/-effect/
```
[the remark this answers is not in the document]
i-walker: Yes sounds great too, it would be great if we could link EffectScope with EagerEffectScope and showing they cover similar signatures. 
And rearranging it sounds super. 

On that note now that we moved on from Data type + extension (aka implementation of a Type class) we might want to consider just having Data types and computations. 

The few typeclasses we have left can also have a dedicated place, if necessary 
Just as a note, though.

arrow-site/docs/docs/fx/README.md
```diff
@@ -63,6 +63,7 @@ s [Coroutines Guide](https://kotlinlang.org/docs/coroutines-guide.html) on Kotli
   - [Pure & Referentially Transparent Functions]({{ '/fx/purity-and-referentially-transparent-functions/' | relative_url }})
   - [Kotlin's Std Coroutines package]({{ '/fx/coroutines/' | relative_url }})
   - [Why suspend over IO monad]({{ '/effects/io/' | relative_url }})
+  - [Semantics of Structured Concurrency and Effect]({{ '/arrow/core/continuations/' | relative_url }})
```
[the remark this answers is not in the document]
i-walker: We have [this](https://arrow-kt.io/docs/fx/coroutines/), but it doesn't go as deep as `Semantics of Structured Concurrency and Effect` maybe we can create a ticket and I'd be happy to contribute it. Or help in whatever way

i-walker:
Thanks for extensive review, learnt a lot 🤗

arrow-site/docs/docs/core/README.md
```diff
@@ -68,6 +68,7 @@ boilerplate and enable direct syntax including [monad comprehensions and computa
   - [Kotlin's Std Lib Guide](https://kotlinlang.org/api/latest/jvm/stdlib/)
   - [Pure & Referentially Transparent Functions]({{ '/fx/purity-and-referentially-transparent-functions/' | relative_url }})
   - [Why suspend over IO monad]({{ '/effects/io/' | relative_url }})
+  - [Semantics of Structured Concurrency and Effect]({{ '/arrow/core/continuations/' | relative_url }})
```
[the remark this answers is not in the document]
i-walker: yes we can leave it as a task in a ticket regarding how/ or what we should show on the site

PR #2669, opened 2022-02-22
deprecate signatures in computations  and add Result for continuations

i-walker:
I am deprecating the computations module in a seperate PR


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/result.kt
```diff
@@ -0,0 +1,68 @@
...
+    cont.shift(r)
+
+  public suspend fun <B> Option<B>.bind(): B =
+    bind { NullPointerException("Option is None") }
+
+  public suspend fun <B> Result<B>.bind(): B =
+    fold(::identity) { shift(it) }
+
+  public suspend fun ensure(value: Boolean): Unit =
+    ensure(value) { IllegalStateException("expected true, but was false") }
```
i-walker, line 43: @nomisRev is this Exception fine or should we a custom one, which doesn't include the stacktrace?
e.g.: `public class EnsureContinuationException(msg: String) : IllegalStateException(msg, null)` or something along the lines

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/computations/either.kt
```diff
@@ -93,14 +99,24 @@ public suspend fun <E, B : Any> EitherEffect<E, *>.ensureNotNull(value: B?, orLe
...
 public fun interface RestrictedEitherEffect<E, A> : EitherEffect<E, A>
 
+@Deprecated(deprecateInFavorOfEffectOrEagerEffect, ReplaceWith("either", "arrow.core.continuations.either"))
 @Suppress("ClassName")
 public object either {
+  @Deprecated(deprecateInFavorOfEagerEffect, ReplaceWith("either.eager(c)", "arrow.core.continuations.either"))
   public inline fun <E, A> eager(crossinline c: suspend RestrictedEitherEffect<E, *>.() -> A): Either<E, A> =
     Effect.restricted(eff = { RestrictedEitherEffect { it } }, f = c, just = { it.right() })
 
+  @Deprecated(deprecateInFavorOfEffect, ReplaceWith("either(c)", "arrow.core.continuations.either"))
```
nomisRev, line 113: 🥳

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/computations/nullable.kt
```diff
@@ -76,14 +81,24 @@ public suspend fun <B : Any> NullableEffect<*>.ensureNotNull(value: B?): B {
   return value ?: control().shift(null)
 }
 
+@Deprecated(deprecateInFavorOfOptionEagerEffectScope)
 @RestrictsSuspension
 public fun interface RestrictedNullableEffect<A> : NullableEffect<A>
 
+@Deprecated(deprecateInFavorOfEffectOrEagerEffect, ReplaceWith("option", "arrow.core.continuations.option"))
```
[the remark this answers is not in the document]
i-walker: Sure I'll add a delegated one to option.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/nullable.kt
```diff
@@ -0,0 +1,10 @@
+package arrow.core.continuations
+
+@Suppress("ClassName")
+public object nullable {
+  public inline fun <A> eager(crossinline f: suspend OptionEagerEffectScope.() -> A): A? =
+    option.eager(f).orNull()
```
i-walker, line 6: there is one issue I found is problematic in the signature of `A?.bind():A` in OptionEffectScope or the eager version. It is allows this:
```kotlin
// context(EffectScope<DomainError>)
val a: Int? = null
a.bind().bind().bind()
```
without the compiler or Intellij able to grey out that the other two `bind()` are not necessary. I've added a contract for that case here, but still won't highlight it. Am I missing something?

raulraja: These kinds of accesses are not detected by the compiler but by the IDEA plugin. The plugin will tell you `"".toString().toString()"` is redundant but it does not know `.bind().bind()` is redundant because it has no knowledge of bind(). It only knows these https://github.com/JetBrains/intellij-kotlin/blob/e2c105911e1fca5b7fe6b9dc10cf9f75463a3c13/idea/src/org/jetbrains/kotlin/idea/intentions/RemoveRedundantCallsOfConversionMethodsIntention.kt#L36-L51 

Making it universal it's not possible, some operators with other semantics may perform side effects and it's legit to chain their calls:
```kotlin
acquire().acquire()
release(2)
```

i-walker: Thanks Raul, makes sense.

nomisRev: Right, that's why we had a separate `nullable` DSL that worked *only* for `nullable` types.
Otherwise `A?.bind()` took precedence over `Option<A>.bind()` 🤔

I still think the `nullable` DSL is useful enough to keep around though, but then we should keep it separate again.
We made this mistake in the past and had to fix it through deprecations and removal. This should be traceable by checking the git history of the Option and Nullable computation DSL.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/computations/nullable.kt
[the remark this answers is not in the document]
nomisRev: See other comments. We need to check git history of `nullable` and `option`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/computations/nullable.kt
```diff
@@ -7,6 +7,7 @@ import kotlin.contracts.ExperimentalContracts
 import kotlin.contracts.contract
 import kotlin.coroutines.RestrictsSuspension
 
+@Deprecated("NullableEffect is being replaced by arrow.core.continuations.OptionEffectScope")
```
nomisRev, line 10:
```suggestion
```

This is for a subsequent PR.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/computations/nullable.kt
```diff
@@ -67,6 +68,7 @@ public fun interface NullableEffect<A> : Effect<A?> {
  * ```
  * <!--- KNIT example-nullable-computations-02.kt -->
  */
+@Deprecated(deprecateInFavorOfOptionEffectScope)
```
nomisRev, line 71:
```suggestion
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/computations/nullable.kt
```diff
@@ -76,14 +78,24 @@ public suspend fun <B : Any> NullableEffect<*>.ensureNotNull(value: B?): B {
   return value ?: control().shift(null)
 }
 
+@Deprecated(deprecateInFavorOfOptionEagerEffectScope)
```
nomisRev, line 81:
```suggestion
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/computations/nullable.kt
```diff
@@ -76,14 +78,24 @@ public suspend fun <B : Any> NullableEffect<*>.ensureNotNull(value: B?): B {
...
 @RestrictsSuspension
 public fun interface RestrictedNullableEffect<A> : NullableEffect<A>
 
+@Deprecated(deprecateInFavorOfEffectOrEagerEffect)
```
nomisRev, line 85:
```suggestion
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/computations/nullable.kt
```diff
@@ -76,14 +78,24 @@ public suspend fun <B : Any> NullableEffect<*>.ensureNotNull(value: B?): B {
...
 @Suppress("ClassName")
 public object nullable {
+  @Deprecated(deprecateInFavorOfEagerEffect)
```
nomisRev, line 88:
```suggestion
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/computations/nullable.kt
```diff
@@ -76,14 +78,24 @@ public suspend fun <B : Any> NullableEffect<*>.ensureNotNull(value: B?): B {
...
   public inline fun <A> eager(crossinline func: suspend RestrictedNullableEffect<A>.() -> A?): A? =
     Effect.restricted(eff = { RestrictedNullableEffect { it } }, f = func, just = { it })
 
+  @Deprecated(deprecateInFavorOfEffect)
```
nomisRev, line 92:
```suggestion
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/computations/nullable.kt
```diff
@@ -76,14 +78,24 @@ public suspend fun <B : Any> NullableEffect<*>.ensureNotNull(value: B?): B {
...
   public suspend inline operator fun <A> invoke(crossinline func: suspend NullableEffect<*>.() -> A?): A? =
     Effect.suspended(eff = { NullableEffect { it } }, f = func, just = { it })
 }
+
+internal const val deprecateInFavorOfOptionEffectScope: String =
+  "Is being replaced by arrow.core.continuations.OptionEffectScope"
+
+internal const val deprecateInFavorOfOptionEagerEffectScope: String =
+  "Is being replaced by arrow.core.continuations.OptionEagerEffectScope"
```
nomisRev, line 101:
```suggestion

```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/option.kt
```diff
@@ -22,8 +22,11 @@ public value class OptionEffectScope(private val cont: EffectScope<None>) : Effe
   public suspend fun <B> Option<B>.bind(): B =
     bind { None }
 
-  public suspend fun <B> B?.bind(): B =
-    this ?: shift(None)
+  @OptIn(ExperimentalContracts::class)
+  public suspend fun <B> B?.bind(): B {
+    contract { returns() implies (this@bind != null) }
```
nomisRev, line 27: This doesn't conflict with `Either<None, A>.bind()` inherited from `EffectScope<None>?

nomisRev: In essence, this is the equivalent as `ensureNotNull` so it's kinda redundant.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/option.kt
```diff
@@ -50,8 +53,11 @@ public value class OptionEagerEffectScope(private val cont: EagerEffectScope<Non
...
+    return this ?: shift(None)
+  }
```
nomisRev, line 60: Idem, other comment.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/result.kt
nomisRev: We could also omit these overrides and then we don't make a decision for the users, since they can use `ensure(...) { }` from `EffectScope<Throwable>`.

This is maybe a better initial approach until users request such behavior to be added.
Same for `Option<B>.bind()` and `ensureNotNull`.

cc\\ @pakoito

i-walker: Will add that instead

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/option.kt
i-walker: Agree

nomisRev (approved):

PR #2672, opened 2022-02-23
deprecate arrow.continuation

i-walker:


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/generic/AtomicRef.kt
```diff
@@ -0,0 +1,52 @@
+package arrow.core.generic
```
nomisRev, line 1: Should this be under `generic` 🤔  Perhaps we should put it under `continuation`?
It's not really an `arrow.core` type, but it's required for the `continuation` runtime.

I think it's useful enough to expose since I've been using it already in some downstream MPP projects.
It's something I really miss in MPP projects and is not easily available in Kotlin since `atomic-fu` is not stable yet and still has some issues to be used in libraries.

nomisRev (commented):
Looks great @i-walker! 🙏 
Just one thought on the `arrow.core.generic` package for `AtomicRef`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/generic/AtomicRef.kt
i-walker: Yes, agree we can put this under arrow.generic. Maybe even create a small micro lib just with this binary,  bc it can live without the overhead of arrow.core. What do you think? I am good with both options

i-walker: I just remembered that `arrow.generic` is a dedicated package for generic programming in the incubator, so should we keep it under arrow.core.continuation @nomisRev ?

nomisRev: Yes, let's just keep it in `arrow.core.continuation`. We can always push it down into another dependency if we really need to later without making a binary breaking change.

i-walker: done :D

nomisRev (approved):

PR #2675, opened 2022-02-25
add Nullable proposal

i-walker:


arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/continuations/NullableSpec.kt
```diff
@@ -0,0 +1,114 @@
...
+  "if expression short circuit" {
+    nullable {
+      val number = "s".length.bind()
+      val string = if (number != 1) {
+        number.toString()
+      } else {
+        null
+      }.bind()
+      string
+    } shouldBe null
```
nomisRev, line 112: Can we also add a test checking if we can bind `Either<Nothing, A>`?

nomisRev (approved):
Looks good to me!

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/continuations/NullableSpec.kt
i-walker: Added :)

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/continuations/NullableSpec.kt
```diff
@@ -111,4 +112,11 @@ class NullableSpec : StringSpec({
       string
     } shouldBe null
   }
+
+  "Either<Nothing, A> short circuit" {
```
nomisRev, line 116:
```suggestion
"Either<Nothing, A> can be bind"
```

`Either<Nothing, A>` can never result in `Left` and thus also never short-circuit.

nomisRev (approved):

PR #2677, opened 2022-03-02
Resource.asFlow

### LINKED ISSUE #2632: ["Request"] Convenience conversion from Resource<R> to Flow<R>
What version are you currently using? 1.0.1

What would you like to see? `Resource` to `Flow` converter, just like [fs2.Stream[F[_], O].resource(r: Resource[F, O])](https://oss.sonatype.org/service/local/repositories/releases/archive/co/fs2/fs2-core_2.13/3.2.0/fs2-core_2.13-3.2.0-javadoc.jar/!/fs2/Stream$.html#resource[F[_],O](r:cats.effect.kernel.Resource[F,O])(implicitF:cats.effect.kernel.MonadCancel[F,_]):fs2.Stream[F,O])

Example:
```
// Presumably, something along the lines of:
fun <R> Resource<R>.asFlow(): Flow<R> = flow {
  use { acquired ->
    emit(acquired)
  }
}
```

```
// So, given
val data: Resource<InputStream> = TODO()
//    and
fun InputStream.chunked(size: Int): Flow<ByteArray> = TODO()


// This can be done:
val chunks: Flow<ByteAray> = data.asFlow().flatMapConcat { it.chunked(64) }
```

i-walker:
resolves #2632

### DIFF
--- a/arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
+++ b/arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
@@ -13,6 +13,8 @@ import kotlin.coroutines.CoroutineContext
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.NonCancellable
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.withContext
 
 /**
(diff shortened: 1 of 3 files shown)


arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -761,3 +763,13 @@ private suspend fun List<suspend (ExitCase) -> Unit>.cancelAll(
...
+
+/**
+ * binds and emits [A] of the resource
+ */
+public fun <A> Resource<A>.asFlow(): Flow<A> =
+  flow {
+    arrow.fx.coroutines.continuations.resource {
+      emit(bind())
+    }
+  }
```
nomisRev, line 775: This implementation is not correct, since now you return `Resource` to `flow { }` which is then never invoked.

```suggestion
  flow {
    use {
      emit(it)
    }
  }
```

nomisRev (approved):

PR #2678, opened 2022-03-02
fix imports for ksp plugin

i-walker:
resolves #2618 
Status:
**In progres**

PR #2681, opened 2022-03-08
File name generation fix

i-walker:
a mistake from  https://github.com/arrow-kt/arrow/pull/2678 
files before #2678 MyGenerated.kt after org.example.MyGenerated.kt


nomisRev:
Actually, I think in the `kapt` version we used `org.example.MyGenerated.kt` for the file names but not 100% sure.
Otherwise, you can have conflicts between two models with the same name.
For example, `io.github.persistence.dto.User` and `io.github.model.User`.

Changing the name of the file like this doesn't impact/break anything, right? @i-walker
If not I think we should leave your change, using FQN is safer than simpleName.

nomisRev:
Discussed in person. Misunderstood the fix, and this should be merged 👍

nomisRev (approved):

PR #2685, opened 2022-03-10
Update Set up Documentation

i-walker:
remove kapt docs 
set up for ksp plugin with Maven is still under investigation https://github.com/google/ksp/issues/47


raulraja (approved):
Thanks @i-walker !

nomisRev (approved):

PR #2686, opened 2022-03-10
rename Sequence sequenceX, traverseX and deprecate Sequence.some

i-walker:
Rename to either `sequence` or `traverse`: 
- sequenceEither
- sequenceValidated
- sequenceOption
- traverseEither
- traverseOption
- traverseValidated

Deprecate:
- some

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
@@ -641,7 +641,7 @@ public fun <E, A> Sequence<Validated<E, A>>.sequence(semigroup: Semigroup<E>): V
 public fun <E, A> Sequence<Validated<E, A>>.sequenceValidated(semigroup: Semigroup<E>): Validated<E, Sequence<A>> =
   sequence(semigroup).map { it.asSequence() }
 
-@Deprecated("Deprecated legacy Api", ReplaceWith("map { generateSequence { this } }"))
+@Deprecated("some is being deprecated in favor of map", ReplaceWith("map { generateSequence { this } }"))
 public fun <A> Sequence<A>.some(): Sequence<Sequence<A>> =
   if (none()) emptySequence()
   else map { generateSequence { it } }
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
@@ -610,15 +610,20 @@ public fun <A, B> Sequence<Validated<A, B>>.separateValidated(): Pair<Sequence<A
   return asep to bsep
 }
 
-@Deprecated("Terminal operators on Sequence are being deprecated. Please use either one of the supported terminal operators of Sequence and opt for supported [traverse] functions")
+public fun <E, A> Sequence<Either<E, A>>.sequence(): Either<E, List<A>> =
+  traverse(::identity)
+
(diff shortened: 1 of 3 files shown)


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
```diff
@@ -612,15 +612,19 @@ public fun <A, B> Sequence<Validated<A, B>>.separateValidated(): Pair<Sequence<A
   return asep to bsep
 }
 
+@Deprecated("Terminal operators on Sequence are being deprecated. Please use either one of the supported terminal operators of Sequence and opt for supported [traverse] functions")
```
raulraja, line 615: What is the replacement that a user would perform in this case? The current message is a bit cryptic. If we know the expression that would replace `sequenceeEther` then we can add it to a `ReplaceWith`.

i-walker: here a user could opt to use `toList().sequenceEither()` instead, if that sounds good to all, I can add it?

nomisRev: There are several options here:

- Deprecate and change the signature to `Sequence<Either<E, A>>.sequenceEither(): Either<E, List<A>>`, which doesn't lie about the fact that it's terminal. This would be more optimized than `toList().sequenceEither()`.

- Deprecate and `ReplaceWith` `toList().sequenceEither()`.

Now that I think about it, the latter might be problematic since that cannot short-circuit an infinite `Sequence` on `Either.Left` but deprecating towards a `List` in the return type is going to be very difficult without breaking the Kotlin API (the binary can be kept in tact by using HIDDEN).

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
```diff
@@ -610,15 +611,37 @@ public fun <A, B> Sequence<Validated<A, B>>.separateValidated(): Pair<Sequence<A
...
 
+@Deprecated(
+  "sequenceValidated is being renamed to sequence to simplify the Arrow API",
+  ReplaceWith("sequence(semigroup).map { it.asSequence() }", "arrow.core.sequence")
+)
 public fun <E, A> Sequence<Validated<E, A>>.sequenceValidated(semigroup: Semigroup<E>): Validated<E, Sequence<A>> =
-  traverseValidated(semigroup, ::identity)
+  sequence(semigroup).map { it.asSequence() }
 
+@Deprecated("Deprecated legacy Api", ReplaceWith("map { generateSequence { this } }"))
```
i-walker, line 644:
```suggestion
@Deprecated("some is being deprecated in favor of map", ReplaceWith("map { generateSequence { this } }"))
```
Is this message better @nomisRev ?

nomisRev: Yes, that is great 👍  I was more concerned about the APIs we're renaming but not removing.

nomisRev (approved):

raulraja (approved):
It looks excellent @i-walker, thanks. Since we are renaming `sequenceX` to `sequence`, is there a better name for this operation?. I feel in Kotlin this operation may be confusing by name, and you'd think you'd get a `Sequence` back instead of traversing over identity.

i-walker:
Since sequence == traverse(::identity) we could introduce `f` to have a default value ::identity for 2.0. Then we could drop sequence since the Traverse type class doesn't exist anymore
wdyt @raulraja ?

Issue #2682, opened 2022-03-09
["Request"] `Iterable::fold` and `Iterable::combineAll` are duplcates


brezinajn:
What version are you currently using? 1.0.1

What would you like to see?

`Iterable::fold`
https://github.com/arrow-kt/arrow/blob/main/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt#L983

and
`Iterable::combineAll`
https://github.com/arrow-kt/arrow/blob/main/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt#L707

are doing the same and one of them probably should be deprecated.

Original debate on arrow slack: https://kotlinlang.slack.com/archives/C5UPMM0A0/p1646820327599249

nomisRev:
I agree with @brezinajn, I prefer `fold`. `fold` is also the name used in the Kotlin Std so everyone in Kotlin should already be familiar with it.

i-walker:
Will do :D

PR #2687, opened 2022-03-10
deprecate combineAll in favour of fold

i-walker:
resolves https://github.com/arrow-kt/arrow/issues/2682

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
@@ -1495,7 +1495,7 @@ public fun <A, B> Either<A, B>.combine(SGA: Semigroup<A>, SGB: Semigroup<B>, b: 
       is Right -> Either.Right(SGB.run { this@combine.value.combine(b.value) })
     }
   }
-@Deprecated("use fold instead", ReplaceWith("fold(MA, MB)"))
+@Deprecated("use fold instead", ReplaceWith("fold(MA, MB)", "arrow.core.fold"))
 public fun <A, B> Iterable<Either<A, B>>.combineAll(MA: Monoid<A>, MB: Monoid<B>): Either<A, B> =
   fold(MA, MB)
 
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
@@ -1496,11 +1496,8 @@ public fun <A, B> Either<A, B>.combine(SGA: Semigroup<A>, SGB: Semigroup<B>, b: 
     }
   }
 
-@Deprecated("use fold instead", ReplaceWith("fold(MA, MB)", "arrow.core.fold"))
+@Deprecated("use fold instead", ReplaceWith("fold(Monoid.either(MA, MB))", "arrow.core.fold", "arrow.typeclasses.Monoid"))
 public fun <A, B> Iterable<Either<A, B>>.combineAll(MA: Monoid<A>, MB: Monoid<B>): Either<A, B> =
-  fold(MA, MB)
-
-public fun <A, B> Iterable<Either<A, B>>.fold(MA: Monoid<A>, MB: Monoid<B>): Either<A, B> =
   fold(Right(MB.empty()) as Either<A, B>) { acc, e ->
     acc.combine(MA, MB, e)
   }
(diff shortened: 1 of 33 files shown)


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -704,6 +704,7 @@ public fun <A, B> Iterable<Ior<A, B>>.unalign(): Pair<List<A>, List<B>> =
 public inline fun <A, B, C> Iterable<C>.unalign(fa: (C) -> Ior<A, B>): Pair<List<A>, List<B>> =
   map(fa).unalign()
 
+@Deprecated("use fold instead", ReplaceWith("fold(MA)"))
```
nomisRev, line 707:
```suggestion
@Deprecated("use fold instead", ReplaceWith("fold(MA)", "arrow.core.fold"))
```

nomisRev (changes requested):
Thank you for taking care of this @i-walker. We should do the same for `Sequence`.

i-walker:
Agree, I‘ll take of it in this PR

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -904,12 +904,17 @@ public fun <A> A.some(): Option<A> = Some(this)
 
 public fun <A> none(): Option<A> = None
 
+@Deprecated("use fold instead", ReplaceWith("fold(MA)", "arrow.core.fold"))
```
nomisRev, line 907: This fold doesn't exist, should we just deprecate to `fold(Monoid.option(MA))` ?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -1496,7 +1496,11 @@ public fun <A, B> Either<A, B>.combine(SGA: Semigroup<A>, SGB: Semigroup<B>, b:
     }
   }
 
+@Deprecated("use fold instead", ReplaceWith("fold(MA, MB)", "arrow.core.fold"))
```
nomisRev, line 1499: Should instead of adding this new fold method deprecate towards `fold(Monoid.either(MA, MB))`?
These overloads seem a bit redundant.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -1496,7 +1496,11 @@ public fun <A, B> Either<A, B>.combine(SGA: Semigroup<A>, SGB: Semigroup<B>, b:
...
 public fun <A, B> Iterable<Either<A, B>>.combineAll(MA: Monoid<A>, MB: Monoid<B>): Either<A, B> =
+  fold(MA, MB)
+
+public fun <A, B> Iterable<Either<A, B>>.fold(MA: Monoid<A>, MB: Monoid<B>): Either<A, B> =
```
nomisRev, line 1503: See another comment. This seems a bit redundant.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/map.kt
```diff
@@ -436,7 +436,11 @@ public fun <K, A> Map<K, A>.combine(SG: Semigroup<A>, b: Map<K, A>): Map<K, A> =
   else b.foldLeft(this@combine) { my, (k, a) -> my + Pair(k, a.maybeCombine(my[k])) }
 }
 
+@Deprecated("use fold instead", ReplaceWith("fold(SG)", "arrow.core.fold"))
```
nomisRev, line 439: Idem as others.

```suggestion
@Deprecated("use fold instead", ReplaceWith("fold(Monoid.map(SG))", "arrow.core.fold"))
```

nomisRev (commented):
I think we can reduce the API a bit, by promoting better usage of `Monoid`.
What do you think @i-walker. cc\\ @raulraja

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
raulraja: I also think these overloads are redundant if we already have one instance for Either that we could take as a single argument. 

I never use `Monoid` if I can avoid it and prefer a method that takes an `empty` value and an `assoc` function. That method already exists for Iterable, and it's just `fold`. In actual use cases in Kotlin, people are not going to provide their own `Monoid` or `Semigroup` instances to pass them around. These instances we provide are mainly for primitive types, but most use cases have more complex models inside the Iterable. Since users would not bother to construct complicated instances for `Monoid` over their model, these functions quickly become useless and need to be rewritten to basic `fold` or similar.

nomisRev (approved):
LGTM! Thanks @i-walker 🙌

PR #2694, opened 2022-03-25
Fix Either#isNotEmpty()'s KDoc sample

ChristianGarcia:
# Issue

The KDoc for `Either#isNotEmpty()` is listing a sample for `isEmpty()` instead, with the results for `isNotEmpty()` as comments.

# Solution

1. Show the correct sample for `Either#isNotEmpty()`:
```
  //sampleStart
  Either.Left("foo").isNotEmpty()  // Result: false
  Either.Right("foo").isNotEmpty() // Result: true
  //sampleEnd
}
```
2. Launch `gradle knit` to update the generated sample file.


raulraja (approved):
thanks @ChristianGarcia !

i-walker (approved):
Thanks @ChristianGarcia, 
to resolve the CI, you can run `gradlew :arrow-core:knit` 👍🏾

ChristianGarcia:
> Thanks @ChristianGarcia, to resolve the CI, you can run `gradlew :arrow-core:knit` 👍🏾

Ah! I was wondering how those samples would be updated. Now I know :). Thanks for the tip @i-walker!

CI is passing now.

PR #2715, opened 2022-05-03
rename traverseX and sequenceX functions in core

i-walker:
- Ior
- Either
- Option
- Map
- NonEmptyList
- Validated

Note:
traverseNullable, which has an issue in type resolution for a lambda of `(A) -> C?` where C is nullable, other traverse function resolve fine


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Ior.kt
```diff
@@ -543,41 +544,61 @@ public sealed class Ior<out A, out B> {
...
+  @OptIn(ExperimentalTypeInference::class)
+  @OverloadResolutionByLambdaReturnType
+  public inline fun <C> traverse(fa: (B) -> Option<C>): Option<Ior<A, C>> =
     fold(
       { a -> Some(Left(a)) },
       { b -> fa(b).map { Right(it) } },
       { a, b -> fa(b).map { Both(a, it) } }
     )
 
+  @Deprecated("traverseOption is being renamed to traverse to simplify the Arrow API", ReplaceWith("traverse(f)"))
```
raulraja, line 578: Is it ok that the argument is called `fa` but the replacement just uses `f`?

raulraja (approved):
thanks @i-walker !

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Ior.kt
i-walker: good catch didn't see that, will change it :)

nomisRev (approved):
> traverseNullable, which has an issue in type resolution for a lambda of (A) -> C? where C is nullable, other traverse function resolve fine

Pretty nice we don't have this issue with `sequence`!

At some point in the future we might consider removing all these functions in favor of `map` + `bind`, especially given the power that context receivers are offering us but that will be for 2.x.x or later.

Thanks for taking charge of this @i-walker ❤️ 🙌

PR #2716, opened 2022-05-06
Add new workflow for publishing the landing page

franciscodr:
The landing page for the Arrow site should only be updated when there are changes in the `arrow-site` folder and shouldn't be related to any other GitHub Action workflow for generating and publishing artifacts.

Having a dedicated workflow for the landing page will help us to simplify the rest of the workflows by removing unrelated steps.


nomisRev (approved):

raulraja (approved):
Thanks @franciscodr !

PR #2717, opened 2022-05-08
Fix publish sync

JavierSegoviaCordoba:
To avoid picking incorrect tags, we need to do only a checkout merging all jobs into one in the 	 `publish` workflow.


i-walker (approved):

franciscodr (approved):
Looks good. Thanks @JavierSegoviaCordoba!

PR #2722, opened 2022-05-16
feat: contravariance for R, A in Effect, EagerEffect and `in` for *EffectScope

i-walker:
Currently, this doesn't work:

```kotlin
sealed interface DomainError
data class SomeDomainError(val s: String) : DomainError
object UnknownError: DomainError

private fun serverRestarts(): Effect<DomainError, Int> = TODO()
private fun fetchServerLocation(): Effect<SomeDomainError, String> = TODO()

fun serverLocation(): Effect<DomainError, String> =
  effect {
    val restarts = serverRestarts().bind()

    if (restarts > 4) {
      shift(UnknownError)
    } else {
      fetchServerLocation().bind()
    }
  }
```

since `R` in `Effect` or `EagerEffect` isn't contravariant


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
```diff
@@ -568,7 +568,7 @@ import kotlin.coroutines.resume
  * ```
  * <!--- KNIT example-effect-guide-13.kt -->
  */
-public interface Effect<R, A> {
+public interface Effect<out R, A> {
```
nomisRev, line 571: We should probably make `A` also `out`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
```diff
@@ -96,7 +96,7 @@ public interface EagerEffect<R, A> {
 
   public fun <B> map(f: (A) -> B): EagerEffect<R, B> = flatMap { a -> eagerEffect { f(a) } }
 
-  public fun <B> flatMap(f: (A) -> EagerEffect<R, B>): EagerEffect<R, B> = eagerEffect {
+  public fun <B> flatMap(f: (A) -> EagerEffect<@UnsafeVariance R, B>): EagerEffect<R, B> = eagerEffect {
```
nomisRev, line 99: Damn, these were methods we removed for `Effect` but seem left over for `Effect`.

nomisRev (commented):
We should also make `EffectScope<in R>` such that the following code compiles:

```kotlin
sealed interface MyError
sealed interface SubError : MyError
sealed interface OtherError: MyError {
    object Actual : OtherError
}

context(EffectScope<SubError>)
suspend fun subprogram(): Unit =
    println("Hello SubProgram!")

context(EffectScope<OtherError>)
suspend fun otherprogram(): Unit =
    println("Hello OtherProgram!")

context(EffectScope<OtherError>)
suspend fun fail(): Unit =
    shift(OtherError.Actual)

suspend fun main() {
    effect<MyError, Unit> {
        subprogram()
        otherprogram()
        fail()
    }.fold(::println, {  })
    // Hello SubProgram!
    // Hello OtherProgram!
    // OtherError$Actual@7cd62f43
}
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
nomisRev: That way it will behave the same as `Either` does today:

```kotlin
val either: Either<String, Int> = 1.right()
val widen: Either<CharSequence, Number> = either
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
i-walker: should we remove flatMap?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
i-walker: Will do ☺️👍🏾

nomisRev:
Do you mind adding tests for the scenarios we discussed in this PR?

i-walker:
Sure 😄

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
nomisRev: Maybe we should deprecate it... it probably has not been in use a lot anyway. cc\\ @raulraja @franciscodr

nomisRev (approved):

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
raulraja: I think there is no need for flatMap or any lambda taking combinators related to the functor, applicative, and monad hierarchies whenever there is the ability to `bind`. The smaller the API the better. For the scala 3 version of some of these abstractions, I'm avoiding implementing any of those until we need them.

raulraja (approved):
thanks @i-walker !

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
nomisRev: Agreed, so let's deprecate these methods so we avoid people using them.

i-walker: I'll add the deprecations it in this PR right away

i-walker: I am pushing this into a different branch for a cleaner git history

Issue #2727, opened 2022-05-23
["Request"] ease conversion from Kotlin Iterable to NonEmptyList


christophejan:
### What version are you currently using?

1.1.2

### What would you like to see?

```fun <T> Iterable<T>.toNonEmptyList(): Option<NonEmptyList<T>>```
or
```fun <T> fromIterable(i: Iterable<T>): Option<NonEmptyList<T>>```

### Why?

I think that currently, we have to convert ```Iterable``` first to ```List``` then to ```NonEmptyList``` : ```NonEmptyList.fromList(iterable.toList())```

nomisRev:
Hey @christophejan,

Thank you for the ticket. The current way of transforming is indeed a bit Kotlin unidiomatic.
I would like to take the opportunity to also remove `Option` from the signature, in this case, there is no technical reason to use `Option` so we can safely use `?` instead.

With that in mind WDYT about `fun <T> Iterable<T>.toNelOrNull()` and `fun <T> Iterable<T>.toNonEmptyListOrNull()`?

christophejan:
Hi @nomisRev 

Thank you very much. That would be perfect for me.

I'm a rather new Arrow user. I've suggested methods using the same pattern than ```fromList``` but I've no idea why ```Option```  rather than ```?``` was used on it.

nomisRev:
> I've no idea why Option rather than ? was used on it.

There are purely historic, so you could say it's a bit of legacy API that has not yet been revamped :)

PR #2728, opened 2022-05-24
Improve Nel constructor

nomisRev:
Closes https://github.com/arrow-kt/arrow/issues/2727


franciscodr (approved):

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -427,7 +425,8 @@ public inline fun <E, A, B> NonEmptyList<A>.traverse(f: (A) -> Either<E, B>): Ei
     }
   }
   // Safe due to traverse laws
-  return NonEmptyList.fromListUnsafe(acc).right()
+  return (acc.toNonEmptyListOrNull()
+    ?: throw IndexOutOfBoundsException("Empty list doesn't contain element at index 0.")).right()
```
i-walker, line 429: we should mention in the Docs that this throws an IndexOutOfBounds

nomisRev: It was already doing that before with the unsafe access, I could've put `toNonEmptyListOrNull()!!` in the `ReplaceWith` but that would've broken the original signature in case some1 is relying on `try/catch` with `IndexOutOfBounds`.

This construction is actually safe here, so we might as well just use `!!`. We the resulting list from `traverse` will be `Nel`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -464,7 +462,7 @@ public inline fun <E, A, B> NonEmptyList<A>.traverse(
         is Invalid -> semigroup.run { Invalid(acc.value.combine(res.value)) }
       }
     }
-  }.map { NonEmptyList.fromListUnsafe(it) }
+  }.map { requireNotNull(it.toNonEmptyListOrNull()) }
```
i-walker, line 465: 🙌🏾

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/typeclasses/Index.kt
```diff
@@ -197,9 +198,9 @@ public fun interface Index<S, I, A> {
         POptional(
           getOrModify = { l -> l.all.getOrNull(i)?.right() ?: l.left() },
           set = { l, a ->
-            NonEmptyList.fromListUnsafe(
               l.all.mapIndexed { index: Int, aa: A -> if (index == i) a else aa }
-            )
+                      .toNonEmptyListOrNull()
+                      ?: throw IndexOutOfBoundsException("Empty list doesn't contain element at index 0.")
```
i-walker, line 203: nit: can we use here also a requireNotNull unless that was intended

i-walker (approved):
Thanks @nomisRev !

nomisRev:
Thanks for turning this green @i-walker ❤️

PR #2729, opened 2022-05-26
Add command to install bundler

franciscodr:
The workflow for publishing the landing page runs on Ubuntu which doesn't have Bundler installed by default, so we need to add the install command to access the tool.


i-walker (approved):
Thanks @franciscodr :D I created https://github.com/arrow-kt/arrow/pull/2730, so that the build becomes green

nomisRev (approved):

PR #2734, opened 2022-05-30
Convert Arrow core effect into a non-inline function

franciscodr:


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/either.kt
```diff
@@ -7,6 +7,6 @@ public object either {
   public inline fun <E, A> eager(crossinline f: suspend EagerEffectScope<E>.() -> A): Either<E, A> =
     eagerEffect(f).toEither()
 
-  public suspend inline operator fun <E, A> invoke(crossinline f: suspend EffectScope<E>.() -> A): Either<E, A> =
+  public suspend inline operator fun <E, A> invoke(noinline f: suspend EffectScope<E>.() -> A): Either<E, A> =
```
i-walker, line 10: Do we have to change this in the other variants like `nullable` too @nomisRev?

raulraja: I believe here `noinline` over `f` will result the body to be inlined like:
```
effect(f).toEither()
```
in place of the call to `invoke(f)`. If `effect` is also inlined it will then further expand `effect` preserving `f` as no inlined

If the function is not `inline` then it will just be:
```kotlin
invoke(f)
```

If the issue with IOS is related to inline is best to make it not inline at all IMO

nomisRev (approved):

raulraja (approved):

PR #2738, opened 2022-06-03
revert arrow-gradle-config to 0.10.1

i-walker:
there was an error introduced in 0.10.2, which is still under development, to maintain compatibility to Arrow 1.1.2 this PR reverts it back to 0.10.1, 

For context, Kotlin slack discussion:
https://kotlinlang.slack.com/archives/C8UK6RTHU/p1653734285366849

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,6 +1,6 @@
 [versions]
 animalSniffer = "1.5.4"
-arrowGradleConfig = "0.10.2"
+arrowGradleConfig = "0.10.1"
 assertj = "3.22.0"
 coroutines = "1.6.1"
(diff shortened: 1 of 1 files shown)


raulraja (approved):
cc @JavierSegoviaCordoba

JavierSegoviaCordoba (approved):

JavierSegoviaCordoba:
we should ensure new alpha version of arrow config is working correctly so we can do a normal release of it. In integrations it looks like it is working as the folk that caught the issue hasn't reported it anymore

i-walker:
Yes, definitely moving forward we could draft a test suite with an environment to check that everything stays consistent

PR #2739, opened 2022-06-03
add merge to validated

agusotto96:
Adds a merge function to Validated, like the one found in Either

### DIFF
--- a/arrow-libs/core/arrow-core/api/arrow-core.api
+++ b/arrow-libs/core/arrow-core/api/arrow-core.api
@@ -2319,6 +2319,7 @@ public abstract class arrow/core/Validated {
 	public static final fun lift (Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;)Lkotlin/jvm/functions/Function1;
 	public final fun map (Lkotlin/jvm/functions/Function1;)Larrow/core/Validated;
 	public final fun mapLeft (Lkotlin/jvm/functions/Function1;)Larrow/core/Validated;
+	public final fun merge (Larrow/core/Validated;)Ljava/lang/Object;
 	public final fun swap ()Larrow/core/Validated;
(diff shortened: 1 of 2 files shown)


nomisRev (approved):

agusotto96:
Hi @nomisRev! Thanks a lot for answering, glad to help in any way, I'm really enjoying this library and all the resources provided by JetBrains and 47 Degrees.
By the way, I see that one of the checks failed, is this normal? Also, I have a few other small operators that I find interesting, should I keep adding them in PRs or should I open an issue?

i-walker:
To keep consistency we may want to add this as an extension function like we did for Either https://github.com/arrow-kt/arrow/blob/c28e7da09ef07921efbc2419796c0edb45bb25ee/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt#L1383

nomisRev (changes requested):
@i-walker, you're correct.
It needs to be placed outside of `Validated`. I didn't catch that on the diff with the indentation 😱 

This way it will not be conveniently accessible.

nomisRev:
@agusotto96 I think CI failed with a false negative due to CI crashing.

> org.gradle.launcher.daemon.client.DaemonDisappearedException: Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)

agusotto96:
Perfect, will check those out! I moved the function outside of Validated, hope its ok this time :sweat_smile:

agusotto96 (commented):
Perfect, will check those out! I moved the function outside of Validated, hope its ok this time :sweat_smile:

i-walker (approved):
Thanks @agusotto96 👍🏾

nomisRev (approved):

nomisRev:
Thank you for the contribution @agusotto96! And congrats on your first contribution to Arrow 🥳

agusotto96:
Thanks! @i-walker @nomisRev

PR #2746, opened 2022-06-23
catch for EffectScope

serras:
This PR adds a small utility method for "simulating" `catch` as we have for exceptions. If this is useful, a bunch of other utility methods can be implemented on top of it, but first I wanted to know whether you like this `catch`.


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EffectScope.kt
```diff
@@ -240,3 +240,17 @@ public suspend fun <R, B : Any> EffectScope<R>.ensureNotNull(value: B?, shift: (
...
 }
+
+/**
+ * Catch any possible `shift`ed outcome from [action], and transform it with [handler].
+ * This reads like a `catch` block, but for `EffectScope`.
+ *
+ * The [handler] may `shift` into a different `ErrorScope`, which is useful to
+ * simulate re-throwing of exceptions.
+ */
+public suspend fun<E, R, A> EffectScope<E>.catch(
```
nomisRev, line 251: Needs to be an extension due to variance?

This duplicates behaviour with [`handleErrorXXX`](https://github.com/arrow-kt/arrow/blob/051847de3b1d3186fc9843864aaf642a2f4a854f/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt#L661) but it exposes a nicer lambda for `handler` where you can leverage the DSL again by exposing `EffectScope<E>` on the receiver of the lambda.

I am absolutely up for improving these APIs, and renaming them but we should probably avoid duplicated APIs as much as possible to keep the API surface reasonable.

That being said this signatures reminds me a lot of [`Flow#catch`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/catch.html), and it uses the same naming so I am in favour of `catch` over the current `handleErrorXXX` & `redeemXXX` methods. This seems to cover all 4 APIs with a single nice API.

nomisRev:
If we speak in terms of `catch` then I think we also revisit renaming `shift` to `raiseError` or similar. WDYT?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EffectScope.kt
serras: About not being an extension, I just wrote it directly like that, although it could also be part of `EffectScope` itself. In fact, this opens the decades-old debate: it `catch` part of the "error effect", or is it a "effect transformer" (my definition of `catch` says that the latter).

serras:
> If we speak in terms of `catch` then I think we also revisit renaming `shift` to `raiseError` or similar. WDYT?

Or we can go with `reset` instead of `catch` if you prefer the delimited continuations terminology. In my perfect world we would keep both: `EffectScope`/`shift`/`reset` speak of the general concept, wheread `ErrorScope`/`throw`/`catch` speak of the most important use case.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EffectScope.kt
i-walker: This also an alternative, using the `UnsafeVariance`, which is fine here.

// in EffectScope interface
```kotlin
public suspend fun <B> Effect<R, B>.catch(recover: suspend EffectScope<R>.(@UnsafeVariance R) -> B): B =
    fold({ recover(it) }, ::identity)
```

nomisRev:
> Or we can go with reset instead of catch if you prefer the delimited continuations terminology. In my perfect world we would keep both: EffectScope/shift/reset speak of the general concept

@serras I thought `reset` had a different meaning in delimited continuations 🤔 Since from a `reset` block you can typically `shift` multiple values, no? I am not sure we have to stick to the delimited continuations terminology perse, from, from a Kotlin user point of view I don't think it brings any benefit.

> Personally, I'd prefer to stay within the realm of EffectScope as much as I could, and only refer to Effect to "run" the action at the very top level.

While I personally agree with this, I don't expect this to become the most popular usage in Kotlin/Arrow. That heavily depends though on how we want to model Arrow towards 2.0. Which I think with all these new changes is on the horizon. For example, a really cool usage which I think would be popular is the following.

```kotlin
suspend fun fetchUser(id: UUID): Either<DbError, User> = ...

// This pattern covers handleErrorWith, handleError, redeem & redeemWith
either<ServiceError, ServiceUser> {
  catch({ fetchUser(uuid) }) { dbError ->
     shift(dbError.toServiceError())
  }
}
```

This is also another example for me that people will not feel that `shift` is an idiomatic name in Kotlin.
And that the signature that your propose absolutely makes sense!

It's also 100% in line with what KotlinX Coroutines already exposes in its datatypes (Flow).
It would **drastically** reduce API surface, which decreases learning curve etc.
1 pattern/api for all effect-style code / data types.

serras:
One more design question, after looking at [`Flow`'s `catch`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/catch.html): to completely mirror that API we should provide a `catch` that can be used as:

```kotlin
{ fetchUser(uuid) }.catch { dbError -> shift(dbError.toServiceError()) }
```

However, I don't know how to write such function (since there's already a receiver), nor whether that would be good for our use case. Another possibility is to mark the function infix, so we can write:

```kotlin
{ fetchUser(uuid) } catch { dbError -> shift(dbError.toServiceError()) }
```

but maybe then we get a conflict due to a keyword being used.

serras:
> @serras I thought `reset` had a different meaning in delimited continuations 🤔 Since from a `reset` block you can typically `shift` multiple values, no? I am not sure we have to stick to the delimited continuations terminology perse, from, from a Kotlin user point of view I don't think it brings any benefit.

My understanding was that `reset` was a way to mark the "scope" of a `shift`, in the sense that a call to `shift` would bubble up to the nearest `reset`. In that sense, this works a bit like this, since we force the action to run and capture everything bubbling up. But I agree that it gives no benefit for Kotlin users.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EffectScope.kt
```diff
@@ -240,3 +240,17 @@ public suspend fun <R, B : Any> EffectScope<R>.ensureNotNull(value: B?, shift: (
...
+  action: suspend EffectScope<R>.() -> A,
+  handler: suspend EffectScope<E>.(R) -> A
```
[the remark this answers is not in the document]
serras: Done in the latest commit!

nomisRev:
> One more design question, after looking at [Flow's catch](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/catch.html): to completely mirror that API we should provide a catch that can be used as:
>```kotlin
> { fetchUser(uuid) }.catch { dbError -> shift(dbError.toServiceError()) }
> ```

I'm personally not a big fun of these APIs, since they don't play well with inference and they most of the times don't show up in auto-completion. Maybe `@BuilderInference` solves the inference issue, but not sure.
In that regard I would say that `Effect<E, A>.catch(...)` would be a better signature, but I prefer your original signature and I think it makes perfect sense in our DSL style API.

KotlinX Coroutines in also takes similar approach of DSL-like blocks for `suspend () -> A` signatures. Even more so when they can be written using `inline fun` since then the DSL works for both non-suspend and suspending code. Such as [`withContext`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/with-context.html), [`timeoutOrNull`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/with-timeout-or-null.html).

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffectScope.kt
```diff
@@ -214,3 +214,17 @@ public suspend fun <R, B : Any> EagerEffectScope<R>.ensureNotNull(value: B?, shi
...
+ * Catch any possible `shift`ed outcome from [action], and transform it with [handler].
+ * This reads like a `catch` block, but for `EagerEffectScope`.
+ *
+ * The [handler] may `shift` into a different `EagerErrorScope`, which is useful to
+ * simulate re-throwing of exceptions.
+ */
+public fun<E, R, A> EagerEffectScope<E>.catch(
+  f: suspend EagerEffectScope<R>.() -> A,
+  recover: EagerEffectScope<E>.(R) -> A
+): A = eagerEffect(f).fold(
```
pakoito, line 228: NIT: Recover first, operation last.

nomisRev: Also for a DSL like this? I know it's the convention for `bimap`, `fold` etc but to me it feels counter-intuitive here since this models an operation like `try { } catch { }` but for typed errors.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EffectScope.kt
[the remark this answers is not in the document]
i-walker: I have to correct this review @serras, the Effect API already follows the parameter convention we want to apply in later iterations of Arrow so, please change that `f` to `transform`. I had to revisit this chapter, sorry

i-walker (approved):
Thanks @serras, going through our discussion I prefer the original encoding too. And it gives us room in case we come up with improvements. 

I corrected an earlier review above

nomisRev (commented):
> About not being an extension, I just wrote it directly like that, although it could also be part of EffectScope itself. In fact, this opens the decades-old debate: it catch part of the "error effect", or is it a "effect transformer" (my definition of catch says that the latter).

Sorry @serras I don't want to drag this out 😅 These APIs are paving the way to new great stuff in Arrow so I think they deserver a little bit of extra attention.

Currently as it stands it requires `import arrow.core.continuations.catch`, and it'll be listed last when suggesting APIs in IntelliJ IIRC since members get preference and don't require an import in Kotlin.

So IMO the ergonomics are better as a member of `EffectScope`, but since `R` is defined as `out` it would require `@UnsafeVariance` as @i-walker suggested. And I think that hinders other ergonomics. I.e. that you cannot "upcast"? (right? 🤔)

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffectScope.kt
serras: Looking at things like [Flow's `catch`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines.flow/catch.html), I think that the convention is action, then recovery.

serras:
@nomisRev and I have been working a bit on the ergonomics, and came to a two-step blocks,

```kotlin
attempt { /* something */ } catch { e -> /* handle e */ }
```

We found no other (easy) way for the compiler to infer the types of every block, a problem that surfaced once I started adding tests.

i-walker:
Using `attempt` might introduce confusion since, we're using Effect#attempt [for instance here](https://github.com/arrow-kt/arrow/blob/31884b3f34a38246d9d902c306baca8256f39064/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt#L649) and EagerEffect#attempt to run continuations. Is there another descriptor name we can use

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffectScope.kt
```diff
@@ -186,6 +188,38 @@ public interface EagerEffectScope<in R> {
...
+   * ```
+   * attempt { ... } catch { ... }
+   * ```
+   *
+   * The [effect] may `shift` into a different `ErrorScope`, giving
+   * the change to a later `recover` to change the shifted value.
+   * This is useful to simulate re-throwing of exceptions.
+   */
+  @OptIn(ExperimentalTypeInference::class)
+  public suspend fun <E, A> attempt(
```
i-walker, line 206: In the Kotlin Stdlib functions build using the BuilderInference would start with `build*` e.g.: `buildList`, `buildString`, etc.

So here we could name this:
`buildEffect` and `buildEagerEffect`
wdyt cc @nomisRev, @serras

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffectScope.kt
```diff
@@ -186,6 +188,38 @@ public interface EagerEffectScope<in R> {
...
+    @BuilderInference
+    f: suspend EagerEffectScope<E>.() -> A,
+  ): suspend EagerEffectScope<E>.() -> A = f
+
+  /**
+   * Finishes an "attempt block" by providing the way in which to [recover]
+   * from a `shift`ed outcome. This function should be used in combination
+   * with `attempt`.
```
i-walker, line 214: Maybe this Doc suggestion  is a bit more clear - I kept the attempt terminology. 
```suggestion
   * Runs the receiver within an [eagerEffect] and running [fold] function over the computation.
   *
   * When the [EagerEffect] has shifted with [R] it will [recover] the shifted value to [A], and when it
   * ran the computation to completion it will return the value [A].
   * [catch] is used in combination with [attempt].
```

I would probably prefer a real example below like: 

```
* ```kotlin
 * import arrow.core.Either
 * import arrow.core.None
 * import arrow.core.Option
 * import arrow.core.Validated
 * import arrow.core.continuations.eagerEffect
 * import io.kotest.assertions.fail
 * import io.kotest.matchers.shouldBe
 *
 * fun main() {
 *   eagerEffect<String, Int> {
 *     val x = Either.Right(1).bind()
 *     val y = Validated.Valid(2).bind()
 *     val z = 
 *      attempt { Option(3).bind { "Option was empty" } } catch { 0 }
 *     x + y + z
 *   }.fold({ fail("Shift can never be the result") }, { it shouldBe 6 })
* ```
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffectScope.kt
nomisRev: I think that's inaccurate, it's only true for some APIs. Makes sense that `buildXXX` DSLs benefit from `@BuilderInference`, but it's also heavily used in KotlinX Coroutines wherever generic DSLs are being used.

https://github.com/Kotlin/kotlinx.coroutines/search?q=BuilderInference

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffectScope.kt
nomisRev: Shouldn't this snippet show the behaviour of `catch` in this case?

```kotlin
attempt { None.bind { "Option was empty" } } catch { 0 }
```

serras: I changed your comments by your suggestions, but avoided the reference to `fold`, since I feel this is not a part of the API the user needs to understand in order to know what `catch` is doing.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/continuations/EffectSpec.kt
```diff
@@ -8,10 +8,7 @@ import io.kotest.assertions.fail
 import io.kotest.core.spec.style.StringSpec
 import io.kotest.matchers.shouldBe
 import io.kotest.property.Arb
-import io.kotest.property.arbitrary.boolean
-import io.kotest.property.arbitrary.int
-import io.kotest.property.arbitrary.orNull
-import io.kotest.property.arbitrary.string
+import io.kotest.property.arbitrary.*
```
i-walker, line 11: nit: can you make those imports explicit

serras: Done!

nomisRev (approved):
Thanks for this great new API @serras 👏 🥳
This is another **huge** improvement towards Arrow 2.0 🎉

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EffectScope.kt
```diff
@@ -212,6 +214,55 @@ public interface EffectScope<in R> {
...
+   *     val z =
+   *      attempt { None.bind { "Option was empty" } } catch { 0 }
+   *     x + y + z
+   *   }.fold({ fail("Shift can never be the result") }, { it shouldBe 3 })
+   * }
+   * ```
+   * <!--- KNIT example-effect-scope-09.kt -->
+   */
+  public suspend infix fun <E, A> (suspend EffectScope<E>.() -> A).catch(
+    recover: EffectScope<R>.(E) -> A,
```
nomisRev, line 264: Should this not be `recover: suspend EffectScope<R>.(E) -> A`? My eye only fell on this just now.

serras: No, the handler executes in the same context available to the outer thing (which is called R here). The idea is that you “close over” errors of type E, and may turn them into errors of type R, if that makes sense.

nomisRev: Yes, makes sense but not sure I understand why suspend is not allowed in recover 🤔

nomisRev: Doesn’t this prevent you from calling shift or bind?

serras: Oh, I forgot it when I moved it from being an extension function! I’ll push an update tonight.

nomisRev: Awesome, feel free to merge after! Great contribution, excited to see this getting merged 🥳

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffectScope.kt
```diff
@@ -186,6 +188,56 @@ public interface EagerEffectScope<in R> {
...
+   *     val z =
+   *      attempt { None.bind { "Option was empty" } } catch { 0 }
+   *     x + y + z
+   *   }.fold({ fail("Shift can never be the result") }, { it shouldBe 3 })
+   * }
+   * ```
+   * <!--- KNIT example-eager-effect-scope-08.kt -->
+   */
+  public infix fun <E, A> (suspend EagerEffectScope<E>.() -> A).catch(
+    recover: EagerEffectScope<R>.(E) -> A,
```
i-walker, line 239:
```suggestion
    recover: suspend EagerEffectScope<R>.(E) -> A,
```

shift in EagerEffect here is also suspend

serras: Thanks for catching this!

serras: It seems that it doesn't work with `suspend` here. I've tried adding `suspend` to the top-level too, but I cannot make it work, so I've reverted it. I think this is fine, even though we might find a more general `suspend`-based version in the future.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/continuations/EffectSpec.kt
```diff
@@ -109,6 +110,32 @@ class EffectSpec :
       }
     }
 
+    "attempt - catch" {
+      checkAll(Arb.int(), Arb.long()) { i, l ->
```
nomisRev, line 114: Can we add these tests for EagerEffect too? I think we’ll run into the issue that shift cannot be called from the EagerEffect catch method, but I’m not 100% sure.

i-walker (approved):

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/continuations/EffectSpec.kt
nomisRev: Thank you @serras! I was wrong 😁

serras:
@i-walker after our meeting today I'm not sure whether I should merge this or not. If you say yes, I'll do it :)

i-walker:
No, go ahead this is fine to be merged :) @serras

PR #2747, opened 2022-06-23
Add `kotlin.mpp.enableCompatibilityMetadataVariant=true`

JavierSegoviaCordoba:


nomisRev (approved):
Nice work @JavierSegoviaCordoba 👏 👏 
I think this is the best middle ground until all of this stabilises in the Kotlin eco-system.

As soon as Kotest adds the `kotlin.mpp.enableCompatibilityMetadataVariant` flag, or it becomes enabled by default we can remove this workaround. Yes?

JavierSegoviaCordoba:
> Nice work @JavierSegoviaCordoba 👏 👏 I think this is the best middle ground until all of this stabilises in the Kotlin eco-system.
> 
> As soon as Kotest adds the `kotlin.mpp.enableCompatibilityMetadataVariant` flag, or it becomes enabled by default we can remove this workaround. Yes?

Yeah, we have to create a PR in Kotest if this works.

About the default, it will never be true because it is a flag to add support to how KMP was working in old Kotlin versions when Kotlin Multiplatform becomes stable all old libraries should be already using the new way (if they aren't abandoned).

raulraja (approved):

i-walker (approved):
thanks @JavierSegoviaCordoba !!

JavierSegoviaCordoba:
Looks like disabling tests is not enough in order to avoid them being compiled when you run test tasks, so we can't run test tasks in GitHub Actions in order to be able to merge the PR (and later be able to create an alpha).

nomisRev:
@JavierSegoviaCordoba can we remove the `sourceSets.test` completely 😅
That way it should not be included in the build, right?

JavierSegoviaCordoba:
@nomisRev we are running windows specifying test source sets which probably fail. We have to change the workflow to avoid that too.

.github/workflows/pull_request.yml
```diff
@@ -29,13 +29,13 @@ jobs:
         uses: gradle/gradle-build-action@v2.2.0
         if: matrix.os != 'windows-latest'
         with:
-          arguments: --full-stacktrace build
+          arguments: --full-stacktrace assemble
```
i-walker, line 32: are we still able to run tests in CI with this configuration?

JavierSegoviaCordoba: No, the idea is publishing an alpha to try this one, and later when the alpha is available, revert the commit as we talked a few days ago

cc @nomisRev @raulraja

i-walker: Perfect, thanks for clarifying.

nomisRev:
We're still using `build` in some places @i-walker so this failed on `main`.

i-walker:
I have the fix here https://github.com/arrow-kt/arrow/pull/2749

PR #2750, opened 2022-06-27
Resource deprecate traverseResource, fixes arity-4

nomisRev:
This fix for arity-4 is non-binary breaking but it removes a redundant type argument from `Resource#zip`.
It deprecates `traverseResource` and makes some improvements to the docs to better highlight the resource DSL.


arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -507,6 +497,7 @@ public sealed class Resource<out A> {
   public companion object {
 
     @PublishedApi
+    @Deprecated("This will be removed from the binary in Arrow 2.0", level = DeprecationLevel.ERROR)
     internal val unit: Resource<Unit> = just(Unit)
```
i-walker, line 501: Should we also remove those in Schedule, arrow.fx.coroutines.builder.kt?

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -647,36 +644,35 @@ public infix fun <A> Resource<A>.releaseCase(release: suspend (A, ExitCase) -> U
...
- *   res.forEach(::println)
+ *   }.forEach(::println)
  * }
  * ```
  * <!--- KNIT example-resource-10.kt -->
  */
-public inline fun <A, B> Iterable<A>.traverseResource(crossinline f: (A) -> Resource<B>): Resource<List<B>> =
+@OptIn(ExperimentalTypeInference::class)
+@OverloadResolutionByLambdaReturnType
+public inline fun <A, B> Iterable<A>.traverse(crossinline transform: (A) -> Resource<B>): Resource<List<B>> =
```
i-walker, line 666:
```suggestion
public inline fun <A, B> Iterable<A>.traverse(crossinline f: (A) -> Resource<B>): Resource<List<B>> =
```
to keep the parameter name the same across the other `traverse` functions

nomisRev: We should've not named the new APIs with `f` ><

Very annoying that this is a API breaking change in Kotlin, and there are not strategies for refactoring/deprecating this. Arrow should try to stick to Kotlin Std naming for parameters as well, but given that there is no way to migrate this not sure how we could improve it for Arrow 2.0.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
nomisRev: This was not even being used, but was also not showing up as _unused_ :/ I think any _unused_ internal stuff should be deprecated, and removed.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
nomisRev: I had to make a manual commit @i-walker since the change from `transform` to `f` was not represented for the method body in the suggestion.

nomisRev:
Need to fix docs @i-walker. Will do asap.

serras (approved):
I don't fully understand why `//sampleStart` and `//sampleEnd` are no longer required, but other than that, this LGTM.

PR #2751, opened 2022-06-27
Fix build, inject kotlin.mpp.enableCompatibilityMetadataVariant in CI

nomisRev:
The [1.1.3-alpha.23](https://github.com/arrow-kt/arrow/tree/1.1.3-alpha.23) release on the Arrow Fx Coroutines Utils project, and confirmed that the `kotlin.mpp.enableCompatibilityMetadataVariant` flag fixes the issue.

To still run tests on CI, and to still publish Arrow with the `kotlin.mpp.enableCompatibilityMetadataVariant` I made it `false` by default and inject it during the `publish.yml`.
This way CI can still run with the same guarantees as before.

### DIFF
--- a/.github/workflows/generate-alpha-tag.yaml
+++ b/.github/workflows/generate-alpha-tag.yaml
@@ -27,7 +27,7 @@ jobs:
       - name: Gradle build
         uses: gradle/gradle-build-action@v2.2.0
         with:
-          arguments: --full-stacktrace assemble "-Psemver.stage=alpha"
+          arguments: --full-stacktrace build "-Psemver.stage=alpha"
 
       - name: Stop Gradle daemons
         run: ./gradlew --stop
--- a/.github/workflows/generate-tag.yaml
+++ b/.github/workflows/generate-tag.yaml
@@ -47,7 +47,7 @@ jobs:
       - name: Gradle build
         uses: gradle/gradle-build-action@v2.2.0
         with:
-          arguments: --full-stacktrace assemble "-Psemver.scope=${{ github.event.inputs.scope }}" "-Psemver.stage=${{ github.event.inputs.stage }}"
+          arguments: --full-stacktrace build "-Psemver.scope=${{ github.event.inputs.scope }}" "-Psemver.stage=${{ github.event.inputs.stage }}"
 
       - name: Stop Gradle daemons
         run: ./gradlew --stop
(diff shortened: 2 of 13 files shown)


i-walker (approved):
👍🏾 ❤️

JavierSegoviaCordoba (approved):

raulraja (approved):

.github/workflows/generate-alpha-tag.yaml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Gradle build
         uses: gradle/gradle-build-action@v2.2.0
         with:
-          arguments: --full-stacktrace assemble "-Psemver.stage=alpha"
+          arguments: --full-stacktrace build "-Psemver.stage=alpha"
```
JavierSegoviaCordoba, line 30: Here is it not necessary to inject it?

.github/workflows/generate-tag.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
       - name: Gradle build
         uses: gradle/gradle-build-action@v2.2.0
         with:
-          arguments: --full-stacktrace assemble "-Psemver.scope=${{ github.event.inputs.scope }}" "-Psemver.stage=${{ github.event.inputs.stage }}"
+          arguments: --full-stacktrace build "-Psemver.scope=${{ github.event.inputs.scope }}" "-Psemver.stage=${{ github.event.inputs.stage }}"
```
i-walker, line 50:
```suggestion
          arguments: --full-stacktrace build -Pkotlin.mpp.enableCompatibilityMetadataVariant=true "-Psemver.scope=${{ github.event.inputs.scope }}" "-Psemver.stage=${{ github.event.inputs.stage }}"
```

here too then

.github/workflows/generate-alpha-tag.yaml
nomisRev: We want to run the full build, with tests before we create the tag/release, right?
The publish workflow is completely separate from this, so they shouldn't interfere with each other.

.github/workflows/generate-tag.yaml
nomisRev: `build` +  `-Pkotlin.mpp.enableCompatibilityMetadataVariant=true` does not work.

Since Kotest is released without `Pkotlin.mpp.enableCompatibilityMetadataVariant` we cannot resolve it's dependencies for `Pkotlin.mpp.enableCompatibilityMetadataVariant` :/

> We want to run the full build, with tests before we create the tag/release, right?
> The publish workflow is completely separate from this, so they shouldn't interfere with each other.

PR #2756, opened 2022-07-07
Update kotlin-js-store/yarn.lock

serras:
This should fix the [annoying Dependabot alerts](https://github.com/arrow-kt/arrow/security/dependabot). I know that @JavierSegoviaCordoba has [opened an issue for Renovate](https://github.com/renovatebot/renovate/issues/13573) but it seems that this process won't be automated soon.

### DIFF
--- a/kotlin-js-store/yarn.lock
+++ b/kotlin-js-store/yarn.lock
@@ -7,6 +7,46 @@
   resolved "https://registry.yarnpkg.com/@discoveryjs/json-ext/-/json-ext-0.5.7.tgz#1d572bfbbe14b7704e0ba0f39b74815b84870d70"
   integrity sha512-dBVuXR082gk3jsFp7Rd/JI4kytwGHecnCoTtXFb7DB6CNHp4rg5k1bhg0nWdLGLnOV71lmDzGQaLMy8iPLY0pw==
 
+"@jridgewell/gen-mapping@^0.3.0":
+  version "0.3.2"
+  resolved "https://registry.yarnpkg.com/@jridgewell/gen-mapping/-/gen-mapping-0.3.2.tgz#c1aedc61e853f2bb9f5dfe6d4442d3b565b253b9"
+  integrity sha512-mh65xKQAzI6iBcFzwv28KVWSmCkdRBWoOh+bYQGW3+6OZvbbN3TqMGo5hqYxQniRcH9F2VZIoJCm4pa3BPDK/A==
+  dependencies:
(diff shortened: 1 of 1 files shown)


JavierSegoviaCordoba:
Problem you can't change that lock file directly, it will come back to the old state.

The dependencies has to be fixed via Gradle DSL but I don't do it before.

serras:
I think that if you have it and run `gradle build`, then the dependencies are pinned to whatever is there. So if we just update them once, everybody else will have the same dependencies.

nomisRev (approved):

JavierSegoviaCordoba:
Are you sure it is not regenerates with each assemble? I have seen changes in it without calling any special command, just build

serras:
> Are you sure it is not regenerates with each assemble? I have seen changes on it without calling any special command, just build

What I've noticed is that Gradle/Kotlin never changes it, unless I manually remove the `kotlin-js-store` folder, in which case the new `yarn.lock` is generated.

nomisRev:
> What I've noticed is that Gradle/Kotlin never changes it, unless I manually remove the kotlin-js-store folder, in which case the new yarn.lock is generated.

I thought this was indeed the case. A quick search on the KotlinLang Slack has pointed me in a slightly different direction. Not sure if this is the correct approach, or if both work. I could use this fix in some other projects though so would be great to fix.

https://github.com/rjaros/kvision/blob/14d9e3637dabd492743a827d5ff195abdad18594/build.gradle.kts#L28

nomisRev:
Some other sources seem to state that originally it was placed in `build/js`, and that it is now part of the project structure so you can change and fix versions like @serras did in this PR.
So I think we can continue with @serras his fix, and only look for a more elaborate solution if it turns out that this doesn't work. At least I am okay with moving forward like that for this issue.

nomisRev (approved):

JavierSegoviaCordoba:
@nomisRev I think it changes when you update the Kotlin version, until then Kotlin is not being able to know there are new versions (probably they are hardcoded in KGP) if you don't provide the versions to KGP with its Gradle extension.

raulraja (approved):
thanks @serras !

JavierSegoviaCordoba:
I have checked this and I am 100% sure this archive will be modified automatically again, and probably we can have problems like shipping something which we think is correct but it isn't, and probably even my semver plugin can fail to generate the correct version because with a modified file the repo is not clean so it will not be able to create a non-dirty version.

PR #2763, opened 2022-07-14
[2761] Add explicit visibility modifier for all the generated optics

### LINKED ISSUE #2761: ["Request"] Specify visibility modifier explicitly in the generated code
**1.1.2**
## What would you like to see?
In my kotlin project I am using Arrow Optics with `@optics` code generation.
I am trying to use `explicitApi()` to publish it as a library.
But build fails with `Visibility must be specified in explicit API mode` in the generated code (I am using ksp).
I would suggest adding explicit visibility modifier in the generated code, e.g.
```
// now
inline val ... = ...

// after fix
public inline val ... = ...
// or
internal inline val ... = ...
```
Visibility of generated extensions can be the same as companion object visibility.

### Some thoughts
Because generated code has no documentation, it makes sense to make it `internal` and manually provide `public` wrappers for some of the generated optics. But in this case, we still need to have explicit `internal` visibility modifier.

aangairbender:
Fixes #2761

- Determines the visibility modifier based on Companion object visibility modifier (internal or public otherwise)
- Prefixed all the generated optics with determined visibility modifier

This fix prefixes optics with `internal` modifier if companion object is `internal`. Otherwise optics are prefixed with `public` modifier.

### DIFF
--- a/arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/internals/domain.kt
+++ b/arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/internals/domain.kt
@@ -12,7 +12,7 @@ data class ADT(val pckg: KSName, val declaration: KSClassDeclaration, val target
   val sourceName = declaration.simpleName.asString().replaceFirstChar { it.lowercase(Locale.getDefault()) }
   val simpleName = declaration.simpleName.asString()
   val packageName = pckg.asString()
-  val visibilityQualifier = when (declaration.companionObject?.getVisibility()) {
+  val visibilityModifierName = when (declaration.companionObject?.getVisibility()) {
     Visibility.INTERNAL -> "internal"
     else -> "public"
   }
--- a/arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/internals/domain.kt
+++ b/arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/internals/domain.kt
@@ -1,7 +1,10 @@
 package arrow.optics.plugin.internals
 
+import arrow.optics.plugin.companionObject
+import com.google.devtools.ksp.getVisibility
 import com.google.devtools.ksp.symbol.KSClassDeclaration
 import com.google.devtools.ksp.symbol.KSName
+import com.google.devtools.ksp.symbol.Visibility
 import java.util.Locale
 
 data class ADT(val pckg: KSName, val declaration: KSClassDeclaration, val targets: List<Target>) {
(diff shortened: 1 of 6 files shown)


aangairbender:
Thank you. I am excited to make my first contribution 😄 
I have a small concern though. All the existing tests are passing, but the logic clearly changed a bit. I didn't find a nice way to add unit test covering the change. What should I do?

arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/internals/domain.kt
```diff
@@ -1,14 +1,21 @@
...
 import com.google.devtools.ksp.symbol.KSName
+import com.google.devtools.ksp.symbol.Visibility
 import java.util.Locale
 
 data class ADT(val pckg: KSName, val declaration: KSClassDeclaration, val targets: List<Target>) {
   val sourceClassName = declaration.qualifiedNameOrSimpleName
   val sourceName = declaration.simpleName.asString().replaceFirstChar { it.lowercase(Locale.getDefault()) }
   val simpleName = declaration.simpleName.asString()
   val packageName = pckg.asString()
+  val visibilityModifierName = when (declaration.companionObject?.getVisibility()) {
```
serras, line 15: Why handle only `INTERNAL` and not the rest of visibility modifiers?

nomisRev: - `private` we can not handle here, because we cannot generate a `private` function for a `private` value of another file.

- `protected` doesn't make sense either because it's only allowed in combination with inheritance, so it's also not accessible. (Right?)

nomisRev:
Testing 🤔 I am not super familiar with testing for KSP, or compiler related code maybe @serras can give you better advice.
I think the simplest would be to add an `internal data class` similar to the other tests, https://github.com/arrow-kt/arrow/blob/main/arrow-libs/optics/arrow-optics-ksp-plugin/src/test/kotlin/arrow/optics/plugin/PrismTests.kt

And then something like this maybe?

```kotlin
internal data class Test(internal val str: String) {
  internal companion object
}

Test::class.visibility == KVisibility.INTERNAL
Test::str.visibility == KVisibility.INTERNAL
```

serras:
> Testing 🤔 I am not super familiar with testing for KSP, or compiler related code maybe @serras can give you better advice. 

I think the easier way is to create sources which showcase that compilation fails if you try to use the optic from a different module. The current testing utilities are geared towards one single file, but it should be easy to create a [variant of `compile`](https://github.com/arrow-kt/arrow/blob/main/arrow-libs/optics/arrow-optics-ksp-plugin/src/test/kotlin/arrow/optics/plugin/Compilation.kt#L40-L64) which uses two files instead.

serras:
Even without further tests, I think this PR is ready for merge, and solves a very real problem when using the optics plug-in.

serras (approved):

PR #2765, opened 2022-07-14
Fix #2764 & #2760

nomisRev:
This PR fixes issues reported in #2764 & #2760.

After quite some experimenting, and poking around I found that the issue lies within this line of code:

```kotlin
public suspend fun <B> Effect<R, B>.bind(): B = fold(this@EffectScope::shift, ::identity)
```

Originally I thought the issue was `fold(this@EffectScope::shift` resulting in incorrect code, but that was not the case.

This change result in **les** `Continuation` instantiation, since it now re-uses the outer `EffectScope` when possible.
By doing so we avoid additional suspend points, and less `Continuation` being created.

I've not been able to pinpoint where the exact issue lies, but I was able to write a failing test based on the code shared in both tickets and then fix it by making these code changes.

I locally also tested the scenarios shared in #2764 & #2760, and it fixes both.
Hopefully @jrgonzalezg & @Vaysman can verify once the `alpha` is released.

Issue #2760, opened 2022-07-12
Crashes on effect shift when using Arrow with ktor client


jrgonzalezg:
Hi!. I'm trying to migrate an Android app's network layer to the new Effect / effect APIs in Arrow but there seems to be some issue or incompatibility with ktor's HttpClient. The code works fine on the happy path, but as soon as it gets to a `shift` the app crashes and it does not even stop on catch blocks or fold error handling so it seems to be bypassing the error handling mechanism.

The function I'm trying to implement looks like this:
```
suspend inline fun <reified R, reified A> get(
    url: Url,
    acceptedErrorCodes: List<Int> = emptyList()
  ): Effect<BackendFailure<R>, A> =
    effect {
      try {
        val httpResponse: HttpResponse = getHttpResponse(url)

        val httpStatusCode: HttpStatusCode = httpResponse.status

        if (httpStatusCode.isSuccess() || httpStatusCode.value in acceptedErrorCodes) {
          when (val backendResponse: BackendResponse<R, A> = httpResponse.body()) {
            is BackendResponse.Failure -> shift(BackendFailure.Known(backendResponse.data))

            is BackendResponse.Success -> backendResponse.data
          }
        } else {
          shift(BackendFailure.Unknown(null))
        }
      } catch (throwable: Throwable) {
        shift(BackendFailure.Unknown(throwable.nonFatalOrThrow().message))
      }
    }
```

but same issue also happened with more minimal code without the try-catch and simplified error type. This happens both with arrow-stack 1.1.2 and 1.1.3-alpha.29. The stacktraces of the crashes look like this:
```
kotlinx.coroutines.CoroutinesInternalError: Fatal exception in coroutines machinery for AwaitContinuation(DispatchedContinuation[Dispatchers.Main.immediate, Continuation at io.ktor.client.engine.HttpClientEngine$DefaultImpls.executeWithinCallContext(HttpClientEngine.kt:100)@96aed1b]){Completed}@4a75eb8. Please read KDoc to 'handleFatalException' method and report this incident to maintainers
        at kotlinx.coroutines.DispatchedTask.handleFatalException(DispatchedTask.kt:144)
        at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:115)
        at android.os.Handler.handleCallback(Handler.java:938)
        at android.os.Handler.dispatchMessage(Handler.java:99)
        at android.os.Looper.loopOnce(Looper.java:201)
[...]

PR #2769, opened 2022-07-19
Add mapLeft to Effect and EagerEffect

jrgonzalezg:
On the new Effect and EagerEffect types there was no easy way to map an `Effect<R, A>` to an `Effect<R2, A>`. This is useful for instance when error types need to be refined from lower level errors to higher level errors. This PR simplifies that by adding a `mapLeft` function to both `Effect` and `EagerEffect` implemented in terms of `handleErrorWith`.

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
@@ -117,7 +117,7 @@ public interface EagerEffect<out R, out A> {
 
   public fun <R2> mapLeft(f: (R) -> R2): EagerEffect<R2, A> =
     eagerEffect {
-      toEither().fold({ shift(f(it)) }, { it })
+      toEither().fold({ shift(f(it)) }, ::identity)
     }
 
   public fun <B> redeem(f: (R) -> B, g: (A) -> B): EagerEffect<Nothing, B> = eagerEffect {
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
@@ -116,8 +116,8 @@ public interface EagerEffect<out R, out A> {
     }
 
   public fun <R2> mapLeft(f: (R) -> R2): EagerEffect<R2, A> =
-    handleErrorWith {
-      eagerEffect { shift(f(it)) }
+    eagerEffect {
+      toEither().fold({ shift(f(it)) }, { it })
     }
 
   public fun <B> redeem(f: (R) -> B, g: (A) -> B): EagerEffect<Nothing, B> = eagerEffect {
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
@@ -115,6 +115,11 @@ public interface EagerEffect<out R, out A> {
       toEither().fold({ r -> f(r).bind() }, ::identity)
     }
 
+  public fun <R2> mapLeft(f: (R) -> R2): EagerEffect<R2, A> =
+    handleErrorWith {
+      eagerEffect { shift(f(it)) }
+    }
+
   public fun <B> redeem(f: (R) -> B, g: (A) -> B): EagerEffect<Nothing, B> = eagerEffect {
     fold(f, g)
   }
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
@@ -663,8 +663,8 @@ public sealed interface Effect<out R, out A> {
   }
 
   public fun <R2> mapLeft(f: (R) -> R2): Effect<R2, A> =
-    handleErrorWith {
-      effect { shift(f(it)) }
+    effect {
+      fold({ shift(f(it)) }, { it })
     }
 
   public fun <B> redeem(recover: suspend (R) -> B, transform: suspend (A) -> B): Effect<Nothing, B> =
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
@@ -662,6 +662,11 @@ public sealed interface Effect<out R, out A> {
     fold({ recover(it).bind() }, ::identity)
   }
 
+  public fun <R2> mapLeft(f: (R) -> R2): Effect<R2, A> =
+    handleErrorWith {
+      effect { shift(f(it)) }
+    }
+
   public fun <B> redeem(recover: suspend (R) -> B, transform: suspend (A) -> B): Effect<Nothing, B> =
     effect {
       fold(recover, transform)
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
@@ -664,7 +664,7 @@ public sealed interface Effect<out R, out A> {
 
   public fun <R2> mapLeft(f: (R) -> R2): Effect<R2, A> =
     effect {
-      fold({ shift(f(it)) }, { it })
+      fold({ shift(f(it)) }, ::identity)
     }
 
   public fun <B> redeem(recover: suspend (R) -> B, transform: suspend (A) -> B): Effect<Nothing, B> =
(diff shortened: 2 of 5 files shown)


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
```diff
@@ -115,6 +115,11 @@ public interface EagerEffect<out R, out A> {
       toEither().fold({ r -> f(r).bind() }, ::identity)
     }
 
+  public fun <R2> mapLeft(f: (R) -> R2): EagerEffect<R2, A> =
+    handleErrorWith {
+      eagerEffect { shift(f(it)) }
+    }
```
i-walker, line 121: a more efficient implementation is
```suggestion
  public fun <R2> mapLeft(f: (R) -> R2): EagerEffect<R2, A> =
    eagerEffect {
      toEither().fold({ shift(f(it)) }, ::identity)
    }
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
```diff
@@ -662,6 +662,11 @@ public sealed interface Effect<out R, out A> {
     fold({ recover(it).bind() }, ::identity)
   }
 
+  public fun <R2> mapLeft(f: (R) -> R2): Effect<R2, A> =
+    handleErrorWith {
+      effect { shift(f(it)) }
+    }
```
i-walker, line 668: here too
```suggestion
  public fun <R2> mapLeft(f: (R) -> R2): Effect<R2, A> =
    effect {
      fold({ shift(f(it)) }, ::identity)
    }
```

i-walker:
Thanks @jrgonzalezg for raising this issue 👍🏾

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
i-walker: I meant efficient in terms of the encoding, and specifically for mapLeft not creating too many continuations. 

> And speaking about efficiency, is ::identity less or more efficient than a simple { it } or why is it preferred?
same answer as above and ::identity is rather a convention both suffice 

> Also why does EagerEffect need the toEither() while Effect doesn't?
That has to do with `shift` being suspend and fold in EagerEffect is not inline

regarding https://github.com/arrow-kt/arrow/pull/2765 can you reproduce what you mean/or show what crashes in an example repo?

jrgonzalezg: ok, I will change, although I don't see how it could create more continuations. It could at most add (if not optimized by the compiler) an extra function call due to implementing it in terms of the other operator, right? Also `handleErrorWith` is not even `suspend` so I don't see how it could involve any differences in terms of continuations.

I am not sure about the `::identity` part if you mean I should use it cause it is an Arrow convention (I guess it is not a Kotlin one) or if  `{ it }` is also fine. I will go with `{ it }` on first update, change to `::identity` afterwards if you meant that was better / needed too.

And about #2765 I just mentioned it here cause it makes me feel uneasy about the `::identity`. But yes, as I said I'm trying to come up with some example for the #2765 follow up (#2765 did improve things, but I still have some other crashes which look very similar in the stacktraces). It is just not trivial to figure out the exact minimal triggers since I get it only on some rather complex part of my code mixed with things that are not really good to have on an issue report so I'm trying to get rid of that "noise" while keeping the issue happening and I have not been yet able to...

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
```diff
@@ -115,6 +115,11 @@ public interface EagerEffect<out R, out A> {
       toEither().fold({ r -> f(r).bind() }, ::identity)
     }
 
+  public fun <R2> mapLeft(f: (R) -> R2): EagerEffect<R2, A> =
+    eagerEffect {
+      toEither().fold({ shift(f(it)) }, { it })
```
i-walker, line 120: can we use here ::identity

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
```diff
@@ -662,6 +662,11 @@ public sealed interface Effect<out R, out A> {
     fold({ recover(it).bind() }, ::identity)
   }
 
+  public fun <R2> mapLeft(f: (R) -> R2): Effect<R2, A> =
+    effect {
+      fold({ shift(f(it)) }, { it })
```
i-walker, line 667: same here

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EagerEffect.kt
jrgonzalezg: ok, I don't like that cause what does a reference to an inlined function really even "references" / "points to"? It is just much harder to grasp the meaning of the code / what the compiler will do with those. And I still suspect the `::identity` that existed on `bind()` may had been one of the triggers for #2760 but ok, it is changed now.

I will also keep trying to get an example for the coroutines internal errors that are still happening on shift (even after #2765) but I'm not having much luck with it. It seems to be related to using the fold variant with `error` handler at some depth of effect nesting, but I can't figure out a minimal example. Anyway, that would be part of a different issue / PR if I manage to get a simple reproducible scenario for it.

jrgonzalezg:
I don't understand why CI fails after merging main into the branch, seems flaky or broken on main 🤔

i-walker:
we're investigating that gradle error, since it's impacting all new PR's

i-walker (approved):
Lets get this merged with the release candidate
Thanks @jrgonzalezg

nomisRev:
Hey @jrgonzalezg,

Sorry for keeping this PR open for so long, but we've been discussing some other options for error handlers. More in line with what Kotlin already has, in for example `Flow#catch` and I think we've come up with something nice.

The piece of documentation here shows it: https://github.com/arrow-kt/arrow/pull/2782/files#diff-0f50c505222285587fdac4f1597856cab0570b6f5ce48954af9dbc61e0888c7aL204

But it should also cover your use-case, as you can use `catch` to `mapLeft`. Where `mapLeft` becomes: `effect.catch { r -> shift(f(r)) }`.

How do you feel about such API?
I think the reduction in API surface by replacing all errors handers with `catch` outweighs the fact that you are writing `shift` every time. The snippet below also shows that the resulting code is almost exactly as long.

```kotlin
val effect: Effect<String, Unit> = effect { shift("error") }

val x: Effect<Int, Unit> = effect.mapLeft { str -> str.length }
val y: Effect<Int, Unit> = effect.catch { str -> shift(str.length) }
```

The reason I am asking your feedback is because I discussed with @i-walker the possibility of back-porting `catch` to Arrow 1.1.x as soon as possible. That way people can already use to this API before 2.0, but if we'd introduce `catch` then merging this PR might not make sense.

What do you think @jrgonzalezg ? Feel free to also comment your thoughts on the PR I introduced this feature in, some more context can be found there. https://github.com/arrow-kt/arrow/pull/2782

nomisRev:
> In short I think that the final approach should ideally make the effect system "fully functional", allowing to operate on handled / (yet) unhandled errors, map on them / tap on them / recover from them to success and so on.

I 100% agree, and currently in the linked PR there is an overload to also handle `Throwable` which defaults to `{ throw it }`.
This allows you to take control over `R`, `Throwable` as you say which are the 3 channels within `Effect`.

Currently we have:

```kotlin
@BuilderInference
 public fun <E, E2, A> Effect<E, A>.catch(
   recover: suspend Shift<E2>.(Throwable) -> A,
   resolve: suspend Shift<E2>.(E) -> A,
 ): Effect<E2, A> = effect {
   fold(
     { recover(it) },
     { resolve(it) },
     { it }
   )
 }
```

jrgonzalezg:
ok, I haven't seen the variant with `recover` that may work, although you still need to do the `resolve` part even if you are not interested on handling like that. I guess library users can always add extra methods on top of what Arrow may provide if they want and it may then be ok to include only the core operators in the library itself. So Arrow could provide that new catch and users may add a mapLeft or a `catch` variant (maybe with other name) with only recover and so on 👍. I also don't see in #2782 any defaulting to `throw it`, right? I am not sure if it will be good or not to have default implementations on those any way. I think I would prefer a derived operator that hides the defaulting and only exposes what you need to fill than an operator that has a lot of default implementations and makes it more difficult to discern what is required for what you want to do.

Should we close this PR then? or are you still deciding?

nomisRev:
Hey @jrgonzalezg,

After our conversation yesterday I continued some work on the PR, and I turned it into a draft since there were still some unanswered questions like this one. I have added a couple of overloads, that I think satisfy the requirements here.
So now in addition to `catch` there is also `attempt` which only works over `Throwable`, and requires less type arguments.

It also has a new `reified` variant that allows you to refine errors, so you can do stuff like this:

```kotlin
object User
object Error

effect<Error, User> {
  throw IllegalArgumentException("BOOM")
}.attempt { e: IllegalArgumentException -> shift(Error) }
```

> I also don't see in https://github.com/arrow-kt/arrow/pull/2782 any defaulting to throw it, right?

It was a manual/custom overload, not a default param in the `catch` function. Sorry, I miss-phrased that.

> So Arrow could provide that new catch and users may add a mapLeft or a catch variant (maybe with other name) with only recover and so on 👍.

Yes, a lot of people might add custom variants. I expect them to be something like this though, something I do myself from time-to-time.

```kotlin
sealed interface DbError {
  data class ForeignKey(val underlying: SQLException) : DbError
}

fun <A> catchForeignKeyViolation(action: suspend () -> A): Effect<DbError.ForeignKey, A> =
  effect { /** sql delight code */ }
    .attempt { ex: SqlException -> if(ex.isForeignKeyViolation()) shift(ForeignKey(ex)) else throw ex }

fun <E, A> Effect<E, A>.catchUncaughtException(): Effect<DomainError, A> =
  attempt { t: Throwable -> shift(UncaughtException(t)) }
```

jrgonzalezg:
Thanks @nomisRev!. It looks better now! :). I'll close this PR since it is not direction we are going for. I have some doubts about "discoverability" of catch when the user is thinking on say a `mapLeft` but I can imagine this may change with user background (like being familiar or not with `Flow.catch`) and can be considered part of the learning curve to use Arrow.

PR #2770, opened 2022-07-21
Refine the Arrow version selector

franciscodr:
The current Arrow version selector is not fully up-to-date. It's showing old versions for which there are no longer documentation and pointing out version `1.0.1` as the latest stable version.

This pull request proposes a few changes:
- Remove the old versions
- Update the latest stable version to `1.1.x`
- Remove the previous version item, which is a bit confusing

This is a screenshot of how the Arrow version selector would look like after applying these changes:

Screen Shot 2022-07-21 at 12 40 22

### DIFF
--- a/arrow-site/docs/_data/doc-versions.yml
+++ b/arrow-site/docs/_data/doc-versions.yml
@@ -1,34 +1,25 @@
 docVersions:
 
-  - title: "v0.10.5"
-    url: https://arrow-kt.io/docs/0.10/
-    previous: https://arrow-kt.io/docs/0.9/
-
-  - title: "v0.11.0"
-    url: https://arrow-kt.io/docs/0.11/
-    previous: https://arrow-kt.io/docs/0.10/
-
-  - title: "v0.12.1"
+  - title: "Version 0.12.x"
     url: https://arrow-kt.io/docs/0.12/
     previous: https://arrow-kt.io/docs/0.11/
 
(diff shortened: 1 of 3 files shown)


arrow-site/docs/_data/doc-versions.yml
```diff
@@ -1,34 +1,25 @@
...
-    previous: https://arrow-kt.io/docs/0.9/
-
-  - title: "v0.11.0"
-    url: https://arrow-kt.io/docs/0.11/
-    previous: https://arrow-kt.io/docs/0.10/
-
-  - title: "v0.12.1"
+  - title: "Version 0.12.x"
     url: https://arrow-kt.io/docs/0.12/
     previous: https://arrow-kt.io/docs/0.11/
```
serras, line 5:
```suggestion

```

It feels a bit strange to be able to see the docs of a version which is not in the list.

serras (approved):
Just a small nitpick, but otherwise I say yes to the (new) dress.

arrow-site/docs/_data/doc-versions.yml
franciscodr: Yes. you're right. And probably, we don't need that information anymore, as we have removed the `Previous` link`

raulraja (approved):

i-walker (approved):

Issue #2718, opened 2022-05-09
["Request"] [Arrow Optics] Allow duplicate data class name


risalfajar:
What version are you currently using?
1.1.2

What would you like to see?
Let's imagine a structure like this:
```kotlin
@optics
data class LocalPerson(
    //....
) {
    @optics
    data class PersonalData(
        //...some fields
    )
}

@optics
data class ForeignPerson(
    //....
) {
    @optics
    data class PersonalData(
        //...some fields
    )
}
```
Notice that there are two `PersonalData` classes, even if they're inside different classes, this exception will throw:
```
[ksp] kotlin.io.FileAlreadyExistsException: MyApp/app/build/generated/ksp/debug/kotlin/com/example/myapp/service/data/model/PersonalData.kt
	at com.google.devtools.ksp.processing.impl.CodeGeneratorImpl.createNewFile(CodeGeneratorImpl.kt:75)
	at com.google.devtools.ksp.processing.CodeGenerator$DefaultImpls.createNewFile$default(CodeGenerator.kt:59)
	at arrow.optics.plugin.OpticsProcessor.processClass(OpticsProcessor.kt:58)
	at arrow.optics.plugin.OpticsProcessor.process(OpticsProcessor.kt:26)
	at com.google.devtools.ksp.AbstractKotlinSymbolProcessingExtension$doAnalysis$4$1.invoke(KotlinSymbolProcessingExtension.kt:223)
	at com.google.devtools.ksp.AbstractKotlinSymbolProcessingExtension$doAnalysis$4$1.invoke(KotlinSymbolProcessingExtension.kt:221)
	at com.google.devtools.ksp.AbstractKotlinSymbolProcessingExtension.handleException(KotlinSymbolProcessingExtension.kt:316)
	at com.google.devtools.ksp.AbstractKotlinSymbolProcessingExtension.doAnalysis(KotlinSymbolProcessingExtension.kt:221)
	at org.jetbrains.kotlin.cli.jvm.compiler.TopDownAnalyzerFacadeForJVM.analyzeFilesWithJavaIntegration(TopDownAnalyzerFacadeForJVM.kt:120)
	at org.jetbrains.kotlin.cli.jvm.compiler.TopDownAnalyzerFacadeForJVM.analyzeFilesWithJavaIntegration$default(TopDownAnalyzerFacadeForJVM.kt:96)
[...]

Issue #2552, opened 2021-10-14
["Request"] Generic Type parameter support for optics


Sintrastes:
What version are you currently using?

1.0.0

What would you like to see?

The generated output of arrow-optics for an `@optics` annotated data class with generic parameter(s) fails to compile. I'd like to see this use-case supported.

This issue has already been reported on the [old](https://github.com/arrow-kt/arrow-optics/issues/68) repo for arrow optics -- but I did not see any activity either there or here. I originally saw this issue with older versions of arrow, but just confirmed that that same issue occurs with 1.0.0.

Fully supporting this would complicate the codegen a little bit, as making the generated lenses would require them to be `fun`s instead of `val`s -- but beyond that, I think this should be pretty straightforward to support -- as in the generated optics, we should be able to just carry-over the generic parameter context to the definition of the generic lenses.

I wouldn't mind taking a look at this myself as a first issue for contributing to arrow, honestly -- if someone could maybe point me to the appropriate place to make changes.

raulraja:
@Sintrastes I just got news that the new FIR API is available in master and we are gonna have to adapt meta to it https://github.com/JetBrains/kotlin/blob/master/plugins/fir/fir-plugin-prototype/src/org/jetbrains/kotlin/fir/plugin/FirAllOpenComponentRegistrar.kt . I think it's worth exploring together with this new compiler plugin API what a pure synthetic optics plugin looks like that does not rely on code generation and provides automatic field access to all declarations in Kotlin as you type. If you are not in a rush and would like to explore this with @nomisRev @serras and I, we can coordinate together. Otherwise, if you just want to fix the current codegen candidate for publishing is currently located in the arrow-meta repository. https://github.com/arrow-kt/arrow-meta/blob/main/plugins/optics/optics-plugin/src/main/kotlin/arrow/meta/plugins/optics/OpticsPlugin.kt Feel free to ping us in slack with any questions.

raulraja:
> Let's say I have an interface I don't control (or maybe a data class if that's problematic at all), I, with a property val I.someProperty: String, if I start typing I., the IDE will show me I.someProperty: Lens<I, String> as a completion candidate. If I enter that in and use it in my code, the appropriate lens will be generated "behind the scenes" at compile-time of the module where it is used, but (for instance), no IR or code needs to be generated for any properties of I that I did not use as a lens.

[...]

PR #2776, opened 2022-07-28
Support generics and inner classes in `@optics`

serras:
Fices #2718 and #2552

The optics generated for classes with generics are (inline) functions instead of `val`s, due to the way companion objects work in Kotlin (only one instance per class, not per generic instantiation).

Most of the code is straightforward, except for prisms. The hard part there is that if you have:

```kotlin
@optics
sealed class PrismSealed<A,B>(val field: A, val nullable: B?) {
 data class PrismSealed1(private val a: String?) : PrismSealed<String, String>("", a)
 data class PrismSealed2<C>(private val b: C?) : PrismSealed<String, C>("", b)
 companion object
}
```

you need to take the type parameters **of the subclass** into account.

### DIFF
--- a/arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/OpticsProcessor.kt
+++ b/arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/OpticsProcessor.kt
@@ -38,12 +38,6 @@ class OpticsProcessor(private val codegen: CodeGenerator, private val logger: KS
       return
     }
 
-    // check that it does not have type arguments
-    /* if (klass.typeParameters.isNotEmpty()) {
-      logger.error(klass.qualifiedNameOrSimpleName.typeParametersErrorMessage, klass)
-      return
-    } */
-
     // check that the companion object exists
     if (klass.companionObject == null) {
       logger.error(klass.qualifiedNameOrSimpleName.noCompanion, klass)
--- a/arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/OpticsProcessor.kt
+++ b/arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/OpticsProcessor.kt
@@ -39,10 +39,10 @@ class OpticsProcessor(private val codegen: CodeGenerator, private val logger: KS
     }
 
     // check that it does not have type arguments
-    if (klass.typeParameters.isNotEmpty()) {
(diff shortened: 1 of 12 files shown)


arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/OpticsProcessor.kt
```diff
@@ -39,10 +39,10 @@ class OpticsProcessor(private val codegen: CodeGenerator, private val logger: KS
     }
 
     // check that it does not have type arguments
-    if (klass.typeParameters.isNotEmpty()) {
+    /* if (klass.typeParameters.isNotEmpty()) {
       logger.error(klass.qualifiedNameOrSimpleName.typeParametersErrorMessage, klass)
       return
-    }
+    } */
```
nomisRev, line 45: Should we remove this?

arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/internals/processor.kt
```diff
@@ -69,12 +66,13 @@ internal fun evalAnnotatedPrismElement(
...
+      element.getSealedSubclasses().map {
         Focus(
-          it,
-          it.substringAfterLast(".").replaceFirstChar { c -> c.lowercase(Locale.getDefault()) }
+          it.primaryConstructor?.returnType?.resolve()?.qualifiedString() ?: it.qualifiedNameOrSimpleName,
+          it.simpleName.asString().replaceFirstChar { c -> c.lowercase(Locale.getDefault()) },
+          it.superTypes.first().resolve()
         )
-      }
+      }.toList()
```
nomisRev, line 75: Does `element.getSealedSubclasses()` return a `Sequence`?

nomisRev (approved):
Great addition @serras! Thank you for picking this up 🙏
Been outstanding for a very long time.

arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/internals/processor.kt
serras: It does. Since this is a small sequence anyway, I went the easy route and just turned it into a `List` at the end.

nomisRev (approved):
Great work @serras 🙏

i-walker (approved):
Thanks @serras 🙌🏾

lenguyenthanh:
this is awesome!

PR #2777, opened 2022-07-29
Generalized copy for optics

serras:
This brings a small DSL which looks pretty much like Kotlin's built-in `copy`, but uses optics under the hood.

```kotlin
val p = Person("Alejandro", Address("Ergens", "Utrecht", "NL"))
val p2 = p.copy {
  Person.name transform { it.uppercase() }
  Person.address.street set "Nergens"
  inside(Person.address) {
    Address.city set "Madrid"
    Address.country set "ES"
  }
}
```

### DIFF
--- a/arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Copy.kt
+++ b/arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Copy.kt
@@ -68,4 +68,4 @@ private class CopyImpl<A>(var current: A): Copy<A> {
  * ```
  */
 public fun <A> A.copy(f: Copy<A>.() -> Unit): A =
-  CopyImpl(this).also { f(it) }.run { current }
+  CopyImpl(this).also(f).current
--- a/arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Copy.kt
+++ b/arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Copy.kt
@@ -1,0 +1,71 @@
+package arrow.optics
+
+public interface Copy<A> {
+  /**
+   * Changes the value of the element(s) pointed by the [Setter].
+   */
+  public infix fun <B> Setter<A, B>.set(b: B)
+
+  /**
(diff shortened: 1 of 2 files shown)


arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Copy.kt
```diff
@@ -0,0 +1,71 @@
...
+  /**
+   * Transforms the value of the element(s) pointed by the [Traversal].
+   */
+  public infix fun <B> Traversal<A, B>.transform(f: (B) -> B)
+
+  /**
+   * Declares a block in which all optics are nested within
+   * the given [field]. Instead of:
+   *
+   * ```
```
nomisRev, line 18: Should we turn these into Knit examples?
Maybe we should discuss a documentation strategy for Arrow 2.0 in the weekly meeting? 🤔

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Copy.kt
```diff
@@ -0,0 +1,71 @@
...
+ * you can write:
+ *
+ * ```
+ * person.copy {
+ *   Person.address.city set "Madrid"
+ * }
+ * ```
+ */
+public fun <A> A.copy(f: Copy<A>.() -> Unit): A =
+  CopyImpl(this).also { f(it) }.run { current }
```
nomisRev, line 71:
```suggestion
  CopyImpl(this).also(f).current
```

nomisRev (approved):
Small suggestion, and question.
Besides that it looks great @serras!! I think is an API many people have been looking for! 👏 👏 

There have been many questions in the passed to update multiple fields using Optics.

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Copy.kt
serras: Yep, it would be nice to have Knit examples working instead of simple blocks.

i-walker (approved):
🤩 Wow that is really nice

PR #2780, opened 2022-07-30
value class impl of Option (called Maybe) prototype

kyay10:
This PR adds a boxing-free version of the Option data type that works performantly for nested Options (technically up to UInt.MAX_VALUE nestings :grin:)

Maybe sadly still boxes if used inside a generic context (i.e. put in a `List` or an `Either` or printed out etc), but it won't box for placing it inside itself.
~Maybe has its type parameter not-nullable on purpose because if `A` could be nullable, `Maybe<A>?` would box. In fact, it doesn't really make sense to have a nullable type inside `Maybe`, since one could just use `Maybe.fromNullable` instead.~ UPDATE: I've implemented support for nullability inside of `Maybe` now. It works in a similar fashion to `NestedNothing` so that we can indicate where exactly the null appeared (i.e. at which nesting level).
The biggest 2 things that we lose with `Maybe` instead of `Option is:

1. We're forced to use `Maybe.fold` instead of pattern-matching over a `Maybe`: This should be fixed when [KT-27576 Support-inline-sealed-classes](https://youtrack.jetbrains.com/issue/KT-27576/Support-inline-sealed-classes) is implemented. This will allow for `is Just` to be a type cast, and therefore granting access to a `.value` property. (Pattern matching in that way will actually result in boxing for nested Maybe due to the peculiar way that allowed me to optimize fold being inherently about inline functions. More on that later!)

2. Any function that wants to `fold` over a Maybe has to do so using a `reified A`. This might seem very weird at first, but the implementation details should make it clear. Basically, whenever `Just(aValueOfTypeMaybe)` is called, `Just` accesses the `underlying` value of `aValueOfTypeMaybe` and wraps that inside Maybe instead. What this means is that, at runtime, we can't distinguish between a `Just(Just(42))` vs a `Just(42)`. Therefore, we rely on compile-time information by using the reified type to check if the function that is passed to `fold` expects a `Maybe` or a different type. This reified requirement propagates through all functions that do an operation on a generic `Maybe`, but seemingly this shouldn't affect normal consumers of Arrow. This reified requirement cannot be remedied for now unless Kotlin eventually provides some feature to declare that a `value class` should *never* box (something like a `@NeverBoxes` annotation).

The way that nested Maybes are handled for `Just` is that the value is represented simply as whatever its true value is, no matter the level of nesting. For `Nothing`, however, it's quite interesting. Basically, and this is the whole point of using `Maybe` and not just nullable types, we want to be able to know at which level of nesting did a `Nothing` occur. Therefore, it uses internally a `NestedNothing` which has its level of nesting in a property (the first 22 nestings of that class is cached as an optimization). Normal `Nothing` is just the value `NestedNothing(0)` inside a maybe. Whenever `Just` is called on a Maybe whose underlying is `NestedNothing(n)`, it returns a Maybe whose underlying is `NestedNothing(n + 1)`. Conversely, whenever `fold(ifEmpty, ifJust)` is called on a Maybe whose underlying is `NestedNothing(n)`, it first checks if the Maybe is equal to `NestedNothing(0)` and if so it calls `ifEmpty`; otherwise, it calls `ifJust` on `NestedNothing(n-1)` inside a maybe.

Some weird implementation details:
At first I thought that, whenever a `Maybe` is used in a generic context (i.e. as a value of a type parameter `A` or as an `Any`) that it will *always* box. You can even see my initial commit has 2x the final code since I duplicated all the functions that access Maybe just to prevent boxing. Turns out, that is not true for `inline` functions. The compiler is surprisingly much smarter than I initially figured, and in fact it will not box if it can determine that inside of the function the `Maybe` is never used generically. In fact, the compiler even evaluates `aValueOfTypeMaybe is Maybe<*>` at *compile-time*, and it performs dead-code elimination which then ensures that the value isn't used in a boxing way. This is how `Just` is implemented:
```
public inline fun <T : Any> Just(value: T): Maybe<T> {
  // This `is` check gets inlined by the compiler if `value` is statically known to be Maybe
  return _Just(if (value is Maybe<*>) value.underlying else value)
}
```
If `Just` is called on a value that is *known* to be a maybe, the code gets inlined into `_Just(value.underlying)`, which is a non-boxing usage of that maybe value.

Additionally, if a lambda parameter of an `inline` function has a generic parameter `T`, and the compiler knows that `T` is provided completely by the caller, it won't box that value of `T` if it turns out to be a `Maybe`. This means that `aValueOfTypeMaybe.let { /* some code */ }` does not box, which is very neat. 

The really odd-yet-powerful thing is that this works even if the value is not provided by the caller, but is instead provided to the lambda directly. So, what I did is that fold checks if the type parameter `A` is `Maybe` (by checking if a normal Maybe is an instance of it (which is true when `A` is `Maybe` or `Any`) and checking that `Unit` is *not* an instance of it (which rules out `Any`). I figured this might be more performant than a class equality check, and such an `instanceOf` check should be easy to optimize by JIT or a minifier) and if `A` is indeed `Maybe`, it calls the provided lambda with the value wrapped inside a `Maybe`. 

Miraculously, the compiler understands that the lambda is called with a `Maybe` directly, and so it doesn't box it! However, the flip side of this is that every call to `fold` will duplicate its `ifJust` lambda. This means that nesting `fold`s inside of each other generates exponentially more bytecode (interestingly as a result I couldn't implement `zip` like that, and so `zip` actually boxes all of its arguments then unboxes them immediately). This code duplication does result in bigger file sizes, so be warned.

When it comes to the `MaybeInternals` opt-in annotation, I added it just to prevent any internal `Maybe` apis from being used literally anywhere in Arrow unless the dev knows exactly what they're doing. For those internal functions, you can see an interesting symmetry:
`fold` is dual to `Just` (since the first ensures a value is a `Maybe` and the second checks that a value is a `Maybe`)
`value` is dual to `_Just` (since the first un-nests a `NestedNohting` and the second nests a `NestedNothing`
`Maybe.Companion.construct` is dual to `Maybe.underlying` (they give access to `constructor-impl` and `box-impl` in the first case and the raw value of a Maybe and `unbox-impl` in the second case)

You'll find that there's very few cases where a function is duplicated with slightly varying type-arguments, a prime example of which is `Maybe<A>.orNull(): A?` and `Maybe<Maybe<A>>.orNull(): Maybe<A>?`. This is because, sadly, functions that return a generic type `A` based on a value inside of a `Maybe` will box whenever that type is substitued for a `Maybe`. This happens unless the return value is uniquely determined by lambdas that that function takes (`fold` is the perfect example of this where it returns some type `R` that is returned by `ifEmpty` and `ifJust`, and as a result `fold` doesn't need a duplicate `foldMaybe`). This is why `zip`, as I mentioned earlier, is not optimized properly: internally it relies on `Maybe<A>.orNull()`, which will box if `A` is a `Maybe`. Writing `zip` differently requires either maintaining 2^10 = 1024 versions of it for all combinations of its ten type parameters being a `Maybe`. Writing `zip` to use fold results in a method that is literally too large it crashes the compiler (I mean you can imagine how big 1024 copies of a method must be). Also there is some weirdness with `fold` for instance if `R` is determined to be `Maybe<A>?` since then it boxes and unboxes for some reason. To combat that, I just `return null` inside of `Maybe.orNull` in the `ifEmpty` case, so that `fold` returns `Maybe<A>` in this case without nullability.

I've checked that all of these functions (except `zip`) does not box unless it absolutely needs to (e.g. if a `Maybe` is going inside an `Either` or a `Pair`).  Even functions for `EffectScope` have been optimized so that it doesn't box even between shifts (only case I haven't checked yet is if there's a suspension inside a shift since that might keep the `Maybe` across suspension points and therefore box it). I did this mainly by decompiling using cfr-0.152 and looking at the decompiled Java code. If you wish to embark upon that same journey, make sure to have some aspirin nearby, oh and keep in mind that sometimes certain inline functions with lambdas *look* like they're boxing, except that whenever they're used with those lambdas actually specified, the boxing goes away (prime example is `flatMap` which looks like it boxes in bytecode but whenever it is used (e.g. inside `map`) it doesn't box at all.

Also, I couldn't add `Maybe<B>.bind(shift: () -> R)` inside `EffectScope` because `EffectScope` is an interface and interfaces aren't allowed to have `inline` functions. This can be solved by using `context` receivers, or maybe by providing an unsafe variant of `Maybe<A>.bind(shift)` that considers the value to be an `Any` and unsafe-casts it alongside a `Maybe<Maybe<A>>.bind(shift)` that considers the value to be a `Maybe`. such usage of `EffectScope` with `bind(shift)` doesn't seem that common though since most would just rely on `maybe { }` and `maybe.eager { }` instead, which do work using the expected `Maybe.bind()` syntax.

Finally, this is a draft PR because it is missing documentation ~and tests~ UPDATE: I've added tests that are based on the `Option` tests already included. I also haven't done benchmarks yet to verify that this is much faster than `Option` or not, and whether this will have performance on par with normal nullable types. I'll port the documentation over from `Option` ASAP. I also didn't dare touch any usage of `Option` in `optics` and just left them as-is since I don't know if they'd benefit from equivalent `Maybe` functions.

### DIFF
--- a/arrow-libs/core/arrow-core-test/src/commonMain/kotlin/arrow/core/test/generators/Generators.kt
+++ b/arrow-libs/core/arrow-core-test/src/commonMain/kotlin/arrow/core/test/generators/Generators.kt
@@ -1,23 +1,7 @@
 package arrow.core.test.generators
 
-import arrow.core.Const
-import arrow.core.Either
-import arrow.core.Endo
-import arrow.core.Eval
-import arrow.core.Ior
-import arrow.core.Option
-import arrow.core.Tuple10
-import arrow.core.Tuple4
-import arrow.core.Tuple5
-import arrow.core.Tuple6
-import arrow.core.Tuple7
-import arrow.core.Tuple8
-import arrow.core.Tuple9
-import arrow.core.Validated
-import arrow.core.left
-import arrow.core.right
+import arrow.core.*
 import arrow.core.test.concurrency.deprecateArrowTestModules
-import arrow.core.toOption
 import io.kotest.property.Arb
 import io.kotest.property.arbitrary.bind
 import io.kotest.property.arbitrary.boolean
@@ -209,6 +193,10 @@ public fun <A> Arb.Companion.endo(arb: Arb<A>): Arb<Endo<A>> = arb.map { a: A ->
 @Deprecated(deprecateArrowTestModules)
 public fun <B> Arb.Companion.option(arb: Arb<B>): Arb<Option<B>> =
   arb.orNull().map { it.toOption() }
+
+@Deprecated(deprecateArrowTestModules)
+public fun <B : Any> Arb.Companion.maybe(arb: Arb<B>): Arb<Maybe<B>> =
+  arb.orNull().map { it.toMaybe() }
 
 @Deprecated(deprecateArrowTestModules)
 public fun <E, A> Arb.Companion.either(arbE: Arb<E>, arbA: Arb<A>): Arb<Either<E, A>> {
--- a/arrow-libs/core/arrow-core/api/arrow-core.api
+++ b/arrow-libs/core/arrow-core/api/arrow-core.api
@@ -692,6 +692,136 @@ public final class arrow/core/MapKt {
 	public static final fun zip (Ljava/util/Map;Ljava/util/Map;Lkotlin/jvm/functions/Function3;)Ljava/util/Map;
 }
 
+public final class arrow/core/Maybe {
+	public static final field Companion Larrow/core/Maybe$Companion;
+	public static final synthetic fun box-impl (Ljava/lang/Object;)Larrow/core/Maybe;
+	public fun equals (Ljava/lang/Object;)Z
+	public static fun equals-impl (Ljava/lang/Object;Ljava/lang/Object;)Z
+	public static final fun equals-impl0 (Ljava/lang/Object;Ljava/lang/Object;)Z
+	public static final fun fromNullable-a5TdXEY (Ljava/lang/Object;)Ljava/lang/Object;
+	public static final fun fromNullable-rCUBF3A (Ljava/lang/Object;)Ljava/lang/Object;
+	public fun hashCode ()I
+	public static fun hashCode-impl (Ljava/lang/Object;)I
+	public static final fun invoke-rCUBF3A (Ljava/lang/Object;)Ljava/lang/Object;
+	public static final fun isDefined-impl (Ljava/lang/Object;)Z
+	public static final fun isEmpty-impl (Ljava/lang/Object;)Z
+	public static final fun isJust-impl (Ljava/lang/Object;)Z
+	public static final fun isNotEmpty-impl (Ljava/lang/Object;)Z
+	public static final fun isNothing-impl (Ljava/lang/Object;)Z
+	public static final fun nonEmpty-impl (Ljava/lang/Object;)Z
+	public static final fun tapNothing-rCUBF3A (Ljava/lang/Object;Lkotlin/jvm/functions/Function0;)Ljava/lang/Object;
+	public fun toString ()Ljava/lang/String;
+	public static fun toString-impl (Ljava/lang/Object;)Ljava/lang/String;
+	public static final fun tryCatch (Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function0;)Ljava/lang/Object;
+	public static final fun tryCatchMaybe (Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function0;)Ljava/lang/Object;
+	public static final fun tryCatchOrNothing (Lkotlin/jvm/functions/Function0;)Ljava/lang/Object;
+	public static final fun tryCatchOrNothingMaybe (Lkotlin/jvm/functions/Function0;)Ljava/lang/Object;
+	public final synthetic fun unbox-impl ()Ljava/lang/Object;
+}
+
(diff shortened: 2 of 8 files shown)


kyay10:
I just realized that we can support nullable types for `Maybe` and we'd be able to differentiate between a `Maybe<String?>?` being null or having a null inside of it (by doing something very analogous to `NestedNothing`. Perhaps I can even implement it through negative nestings of `NestedNothing`) and we can still have `Maybe.underlying: Any` to prevent boxing for `orNull()`

raulraja:
@kyay10 Thanks for the great and detailed PR. After reading your code and looking at your original description, I think this could be an improvement for Arrow 2.0. and in that case, maybe it makes sense to think about porting this to `Option`.

> At first I thought that, whenever a Maybe is used in a generic context (i.e. as a value of a type parameter A or as an Any) that it will always box. You can even see my initial commit has 2x the final code since I duplicated all the functions that access Maybe just to prevent boxing. Turns out, that is not true for inline functions. The compiler is surprisingly much smarter than I initially figured, and in fact it will not box if it can determine that inside of the function the Maybe is never used generically

Could you provide decompiled examples of the current `Option` vs `Maybe` example operations so we can understand how this behaves and they are different in terms of generated bytecode?

Thanks!

nomisRev:
Hey @kyay10,

I'm going to close this PR for now. It's been stale for quite some time, and we've been unable to confirm if it'll result in actual performance benefits while introducing a lot of complexity.

If you have any more time, or interest to pursue this work feel free to re-open the PR with new findings. You can also find all of us on #arrow-contributors on KotlinLang to discuss further.

Thanks for your interest, hard work, and investigation!

PR #2784, opened 2022-08-02
Remove Tuple10 to Tuple22, Const, Eval, computation blocks, and arrow-continuations

serras:


nomisRev:
🔥

raulraja (approved):
thanks @serras !

i-walker (approved):
Thanks @serras ! 🙌🏾

raulraja (approved):

serras:
The JS and Native builds are failing because they expect `arrow-continuations`, but this is in fact one of the things removed by this PR.

```
e: Could not find "io.arrow-kt:arrow-continuations" in [/Users/serras/arrow/arrow, /Users/serras/.konan/klib, /Users/serras/.konan/kotlin-native-prebuilt-macos-x86_64-1.7.22/klib/common, /Users/serras/.konan/kotlin-native-prebuilt-macos-x86_64-1.7.22/klib/platform/ios_simulator_arm64]
```

I expected the problem to go away by excluding `arrow-core` from the dependencies of `kotest-extensions-arrow`, thinking that it would also exclude anything based on it. But I guess my assumption was wrong...

nomisRev:
I am unsure we can rely on the kites extension for Arrow in the 2.x.x since they're build using 1.x.x, and both versions are not compatible.. 🤔

serras:
I guess I’ll have to revert the removal of the `test` modules, then.

nomisRev:
@serras if you want I can decouple them in 1.x.x, and we can merge that to 2.x.x

serras:
Wouldn’t we have the same problem? We cannot depend on the test module until there’s a 2.x-compatible version, right?

serras:
Another option would be to stop depending on things like `ArrowFxSpec`, but keep the few utilities and generators in `commonTest` for each project.

serras:
@nomisRev @i-walker @raulraja this is now really ready to be merged, I've managed to remove the dependency on the tests on `main`, and take the whole thing up to here

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/EitherZipTest.kt
```diff
@@ -18,6 +18,25 @@ import io.kotest.property.arbitrary.short
 import io.kotest.property.arbitrary.string
 import io.kotest.property.checkAll
 
+data class Tuple10<out A, out B, out C, out D, out E, out F, out G, out H, out I, out J>(
```
nomisRev, line 21: Is this still being used somewhere?

gradle.properties
```diff
@@ -29,6 +29,7 @@ kotlin.native.ignoreDisabledTargets=true
 kotlin.mpp.enableCompatibilityMetadataVariant=false
 # https://youtrack.jetbrains.com/issue/KT-45545#focus=Comments-27-4773544.0-0
 kapt.use.worker.api=false
+kotest.proptest.default.iteration.count=10
```
nomisRev, line 32: Oof, this is a very low number. Any reason why? Any reason why this was introduced here? 🤔

nomisRev (commented):
Looking great overall. Wondering why the iteration count was dropped from 1000 to 10 in Arrow Optics.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/EitherZipTest.kt
serras: Oh, this should have been removed

gradle.properties
serras: In the older tests this was set using a subclass of `StringSpec`, so I changed to the recommended variant. However, I forgot to change the number 🙈

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/EitherZipTest.kt
serras: I replaced it with `Tuple9` in the tests

gradle.properties
serras: Done! Now it's using Kotest's default

nomisRev (approved):

nomisRev:
Thanks @serras 🙌 That was a big chunk.

PR #2791, opened 2022-08-06
Backport flattening token/shift into Cont

nomisRev:
In exploration of Effect for Arrow 2.0 I discovered that we can flatten the `Token`, `Continuation` and `EffectScope` types into a single type, and thus reduce `fold` from 3 allocations to 1.

This improvement is back-portable to `1.1.x` in a binary compatible way.
It increases the public API binary of `internal Fold` a bit but entirety of `Effect` will be binary breaking in 2.0 anyway.


raulraja (approved):
Nice! glad to find out that it's possible 🙌 .

i-walker (approved):

PR #2792, opened 2022-08-07
Enable builder inference for optics `copy`

serras:
This should help type inference and show nicer in the IDE

### DIFF
--- a/arrow-libs/optics/arrow-optics/build.gradle.kts
+++ b/arrow-libs/optics/arrow-optics/build.gradle.kts
@@ -20,6 +20,10 @@ if (enableCompatibilityMetadataVariant) {
   }
 }
 
+tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
+  kotlinOptions.freeCompilerArgs += listOf("-Xenable-builder-inference", "-opt-in=kotlin.RequiresOptIn")
+}
+
 kotlin {
   sourceSets {
     commonMain {
--- a/arrow-libs/optics/arrow-optics/build.gradle.kts
+++ b/arrow-libs/optics/arrow-optics/build.gradle.kts
@@ -21,7 +21,7 @@ if (enableCompatibilityMetadataVariant) {
 }
 
 tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
-  kotlinOptions.freeCompilerArgs += listOf("-Xenable-builder-inference", "-opt-in=kotlin.RequiresOptIn")
+  kotlinOptions.freeCompilerArgs += listOf("-Xenable-builder-inference")
 }
 
 kotlin {
--- a/arrow-libs/optics/arrow-optics/build.gradle.kts
+++ b/arrow-libs/optics/arrow-optics/build.gradle.kts
@@ -20,10 +20,6 @@ if (enableCompatibilityMetadataVariant) {
   }
 }
 
-tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
-  kotlinOptions.freeCompilerArgs += listOf("-Xenable-builder-inference")
-}
-
 kotlin {
   sourceSets {
     commonMain {
--- a/arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Copy.kt
+++ b/arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Copy.kt
@@ -1,5 +1,11 @@
 package arrow.optics
 
+import kotlin.experimental.ExperimentalTypeInference
+
+@DslMarker
+public annotation class OpticsCopyMarker
+
+@OpticsCopyMarker
 public interface Copy<A> {
   /**
    * Changes the value of the element(s) pointed by the [Setter].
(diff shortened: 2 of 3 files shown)


arrow-libs/optics/arrow-optics/build.gradle.kts
```diff
@@ -20,6 +20,10 @@ if (enableCompatibilityMetadataVariant) {
   }
 }
 
+tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
+  kotlinOptions.freeCompilerArgs += listOf("-Xenable-builder-inference", "-opt-in=kotlin.RequiresOptIn")
```
nomisRev, line 24: Are both these flags needed?

First one is to allow builder inference inside Arrow, or to be able to expose annotation?
Second one is to allow OptIn without a compiler warning?

I’m always confused about the OptIn, because it works fine without this flag.

nomisRev (approved):

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Copy.kt
```diff
@@ -1,5 +1,11 @@
 package arrow.optics
 
+import kotlin.experimental.ExperimentalTypeInference
+
+@DslMarker
+public annotation class OpticsCopyMarker
+
+@OpticsCopyMarker
```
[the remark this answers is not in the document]
serras: As I understand from https://kotlinlang.org/docs/type-safe-builders.html#scope-control-dslmarker, you put the annotation in the interface which will be in scope in the builder, which in this case is `Copy`.

arrow-libs/optics/arrow-optics/build.gradle.kts
serras: I am not sure, the docs at https://kotlinlang.org/docs/using-builders-with-builder-inference.html#writing-your-own-builders are by no means clear. But I'll delete the opt-in one.

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Copy.kt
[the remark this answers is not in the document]
serras: As I understand the `@DslMarker` is the thing that marks the member of that interface as "keywords", whereas `@BuilderInference` changes how type checking works to better suit this kind of pattern.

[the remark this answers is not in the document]
nomisRev: Okay, I asked because I used `@DslMarker` differently in the Arrow 2.0 branch and it worked for me in IDEA 🤔

arrow-libs/optics/arrow-optics/build.gradle.kts
nomisRev: > But I'll delete the opt-in one.

I'm not sure if we need it yes, or no 😅 Just wanted to clarify, it can be that we're currently missing it.

nomisRev: As you said "are by no means clear" :/ I feel the same for the `opt-in` one.

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Copy.kt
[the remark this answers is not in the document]
i-walker: Do we need the DslMarker for this feature? - Just asking :D

arrow-libs/optics/arrow-optics/build.gradle.kts
```diff
@@ -20,6 +20,10 @@ if (enableCompatibilityMetadataVariant) {
   }
 }
 
+tasks.withType<org.jetbrains.kotlin.gradle.tasks.KotlinCompile>().configureEach {
+  kotlinOptions.freeCompilerArgs += listOf("-Xenable-builder-inference")
+}
+
```
i-walker, line 26: AFAIK this is not needed. Since we have BuilderInference already working with *EffectScope in core and dont have to add it

serras: Just pushed a commit removing it

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Copy.kt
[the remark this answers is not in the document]
serras: According to the [docs](https://kotlinlang.org/docs/type-safe-builders.html#scope-control-dslmarker) (emphasis mine):

> To make the compiler start controlling scopes you only have to **annotate the types of all receivers used in the DSL with the same marker annotation**. For instance, for HTML Builders you declare an annotation `@HTMLTagMarker`:
>
> ```
> @DslMarker
> annotation class HtmlTagMarker
> ```
>
> An annotation class is called a DSL marker if it is annotated with the `@DslMarker` annotation.
>
> [...]
>
> After you've added this annotation, the Kotlin compiler knows **which implicit receivers are part of the same DSL** and allows to call members of the nearest receivers only.

i-walker (approved):
Thanks @serras 👍🏾

PR #2793, opened 2022-08-07
Setup Kover as test coverage tool

### LINKED ISSUE #2557: ["Request"] Setup Kover as test coverage tool
What would you like to see?

I would like to see Kover as a test coverage tool in Arrow: https://github.com/Kotlin/kotlinx-kover

- [ ] We should create tickets for low-hanging fruit of test coverage

MarkMarkyMarkus:
Closes #2557.

Run `./gradlew koverMergedHtmlReport` to build merged html report for all projects.

Example:
arrow-kover-example


serras:
@MarkMarkyMarkus is it possible to obtain these reports as part of the GitHub Action we are currently running?

MarkMarkyMarkus:
> @MarkMarkyMarkus is it possible to obtain these reports as part of the GitHub Action we are currently running?

Yes, it can be found in artifacts. But downloading _reports-*-latest.zip_ from artifacts is not very convenient.
Maybe it would be better to upload the report to something like [Codecov](https://about.codecov.io/)?

nomisRev:
Tagging in @JavierSegoviaCordoba, because I know he has some experience with this. I remember you telling me about some issues with codecov?

IIRC ./gradlew build already generates the reports.

JavierSegoviaCordoba:
I didn't get it working with codecov and I stopped to use codecov too.

JavierSegoviaCordoba:
Additionally, I got it working with sonarcloud, not sure if you are interested in using that platform tho

MarkMarkyMarkus:
> IIRC ./gradlew build already generates the reports.

No, Kover reports are not generated by the `build` task.

I tested the integration with CodeCov:

arrow-codecov-example

Looks strange, I think it's related to [this issue](https://github.com/Kotlin/kotlinx-kover/issues/16) ([comment](https://github.com/Kotlin/kotlinx-kover/issues/189#issuecomment-1167436909)).

serras:
After trying a lot, it turns out that the problem with Kover here is related to [GitHub Actions permissions](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/). In particular PRs from forks are **not** allowed to add comments to the PR, and thus the action here fails.

serras:
In particular, you can go to the Action logs, and see the `GITHUB_TOKEN` permissions. For this PR, for example, you have `pull_requests: read`, but for branches within the repo you have `pull_requests: write`.

serras:
@i-walker @nomisRev @raulraja does it make sense to have this if GitHub won't allow reporting it as part of the PR?

MarkMarkyMarkus:
I think the previous steps failed due to Gradle crashing while running the tests. It looks like the tests require a lot of resources that the Github runner can't provide (couldn't complete `gradlew build` step).

JavierSegoviaCordoba:
@serras we can at least track the coverage between alpha releases.

nomisRev:
>  does it make sense to have this if GitHub won't allow reporting it as part of the PR?

I see tickets related to this on the Kover repo, so it seems they're solving this issue. IMO merging this work in preparation of future improvements to Kover makes sense. I expect the official tool to have a proper solution for this use-case.

nomisRev:
> I think the previous steps failed due to Gradle crashing while running the tests. It looks like the tests require a lot of resources that the Github runner can't provide (couldn't complete gradlew build step).

@MarkMarkyMarkus yes.. for some time sometimes Github Actions fails when building the projects. We have a hard time reproducing this, and I've never been able to reproduce it locally. It seems to have something to do with the `Gradle configuration steps` where it times out on a hardcoded timeout. https://github.com/gradle/gradle/issues/19446 There seems to be a PR with a fix open 🥳

JavierSegoviaCordoba:
With the new APIs you have been working and simplifying everything, the build time will be lower too right?

Maybe that plus separating a few artifacts into multiple repositories can solve the problem.

nomisRev:
> With the new APIs you have been working and simplifying everything, the build time will be lower too right?

Yes, build time will go down and complexity of the project in general too but that might take until next year until the code is fully out of the repo. We might also still split into separate repos soon-ish too, like we did for `suspendapp`. I hope to do the same for Arrow (Fx) Resilience.

serras (approved):
Nice!

nomisRev:
Hey @MarkMarkyMarkus,
Thank you so much for the contribution! And congrats on your first contribution being merged 🙌 🥳
Sorry it took so very long to get this picked back up and merged 🙈 

@serras in a comment above you mentioned it doesn't work for external forks, due to the `GITHUB_TOKEN`. For now I just disable the steps for forks, so we at least get it on PRs of maintainers. If we enable a minimum coverage limit, we can enable the kover step to fail the build if coverage drops below it.

PR #2795, opened 2022-08-09
Feature/remove validated

Atternatt:
- Removed `Validated` From the project
- Had several knit reports (want to discuss and fix)


Atternatt:
Hey @nomisRev In this Pr I removed only the `Validated` and I also removed the example files that contained `Validated` examples. (also renamed the examples to be consecutive).
At this point I tried to run a grade build and I've got a lot of knit warnings and lately an error:
```Execution failed for task ':knitCheck'.
> knitCheck task failed, see log for details (use '--info' for detailed log).
  Run 'knit' task to write 23 missing/outdated files.```

I just want to double check how to proceed in this case and how to fix it (as well for the checks that the actions are not passing that looks like a lot of them)

**Comment by serras** [comment]
@Atternatt The easiest fix is to remove all the Validated examples from Knit and run `gradle knit` again

**Comment by serras** [comment]
@Atternatt maybe this is a bit too much asking, but could this be based on #2784? I think we are working on removing many similar things (like computation blocks), and duplicating work would be a waste of resources

**Comment by Atternatt** [comment]
> @Atternatt maybe this is a bit too much asking, but could this be based on #2784? I think we are working on removing many similar things (like computation blocks), and duplicating work would be a waste of resources

I was going to pick that feature but @nomisRev told me that you already picked it so we were discussing what I could do. Then we came to this that consisted of removing `Validated` from Arrow. 
So far, I removed all the references in the project and the example files. Perhaps we could remove a common file with coss dependencies between Validated and Tuple?

**Comment by serras** [comment]
@Atternatt on a second thought, just go on. My removal touches some parts, but merging should take care of it.

**Comment by Atternatt** [comment]
on the other hand, would you mind to help me understand what the build errors for the github actions are in order to fix them?

**Comment by serras** [comment]
Sure thing! The first step would be to run `gradle apiDump` to update the API files and then commit and push. Since you are removing (lots) of functions the API Compatibility Validator screams at you… but this is the point of starting a new point release, right?

**Comment by serras** [comment]
To fix the Knit problems, remove all the files in `arrow-libs/core/arrow-core/src/jvmTest/kotlin/examples` and run `gradle knit` to recreate the examples. Knit expects to find a file for every example, but since you are removing a few, it can no longer backreference them and fails.

**Comment by Atternatt** [comment]
while removing documentation refering to `Validated` I found something I'm not sure how to fix in `arrow-site/docs/docs/patterns/errorhandling/README.md:229`. 
How should we update this example?

**Comment by nomisRev** [comment]
@Atternatt the error handling docs are IMO up for revamping after these changes.

We’ve also changed how to do error handling in Effect, and I hope to change all APIs accordingly. So that would mean this doc would be almost completely outdated. So feel free to remove, or ignore it for now.

**Comment by Atternatt** [comment]
got it @nomisRev If there are more fails on the build through GitHub actions actions I'll start removing these files

File: arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt [hunk 1]
```diff
@@ -330,57 +330,6 @@ public fun <A> Iterable<Result<A>>.sequenceResult(): Result<List<A>> =
...
-@Deprecated("traverseValidated is being renamed to traverse to simplify the Arrow API", ReplaceWith("traverse(semigroup, f)", "arrow.core.traverse"))
-public inline fun <E, A, B> Iterable<A>.traverseValidated(
-  semigroup: Semigroup<E>,
-  f: (A) -> Validated<E, B>
-): Validated<E, List<B>> =
-  traverse(semigroup, f)
-
-@OptIn(ExperimentalTypeInference::class)
-@OverloadResolutionByLambdaReturnType
-public inline fun <E, A, B> Iterable<A>.traverse(
```
nomisRev, line 342: Replacing with `mapAccumulating` is for a follow-up PR?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
```diff
@@ -709,36 +699,57 @@ internal class Token {
...
 )
-internal class FoldContinuation<B>(
-  private val token: Token,
+internal class FoldContinuation<R, B>(
   override val context: CoroutineContext,
   private val error: suspend (Throwable) -> B,
-  private val parent: Continuation<B>
-) : Continuation<B> {
+  private val parent: Continuation<B>,
+) : Continuation<B>, Token(), EffectScope<R> {
```
nomisRev, line 706: I think `main` was merged a couple times instead of `arrow2` branch. Not sure how do-able rebasing on `arrow2` would be, or cherry picking last couple commits.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
```diff
@@ -179,36 +179,6 @@ class IterableTest : UnitSpec() {
       }
     }
 
-    "traverse Validated stack-safe" {
```
nomisRev, line 182: We'd need to bring these tests back for `mapAccumulating` for `Iterable`, `Map`, `NonEmptyList`, `Sequence`. Probably not anymore for single value containers.

nomisRev (approved):
Great work @Atternatt !! 👏 👏 🥳 🥳 

The next step would be to remove all redundant `traverse` operations, and to cover the desired functionality with `mapAccumulating` but perhaps cleaner, and easier to manage in a separate PR. WDYT?

serras:
@Atternatt this is great work! As @nomisRev mentions, there's still the hard part of merging it with the current `arrow2` branch (is it up to date with the latest `main`?); I could help with that from Monday on if you prefer to focus on more interesting stuff.

Atternatt:
hello @nomisRev and @serras I've been out for a couple of days but I'm back.
> The next step would be to remove all redundant traverse operations, and to cover the desired functionality with mapAccumulating but perhaps cleaner, and easier to manage in a separate PR. WDYT?

I was thinking exactly the same, creating separated PR's in order to have a clear work. The main p[point will be creating the mapAccumulating tests without forgetting the ones I removed for validated now

Atternatt:
> @Atternatt this is great work! As @nomisRev mentions, there's still the hard part of merging it with the current `arrow2` branch (is it up to date with the latest `main`?); I could help with that from Monday on if you prefer to focus on more interesting stuff.

Is up to date with `arrow2` branch but not with `main` @serras

Atternatt:
@nomisRev @serras What do you think about creating a typealias for `Either<List<L>, R>` that could be named `Validated<L,R>`? does it make sense?

serras:
> Is up to date with `arrow2` branch but not with `main` @serras

Great! Then we can merge for the Arrow 2.c series

serras:
@Atternatt I think the perfect alias would be `Either<NonEmptyList<E>, A>`, with some extension methods to create a `Left` with a singleton list.

Atternatt:
sounds good!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
Atternatt: yes

Atternatt: what do you mean with the issue with the git history? Am I doing something wrong? how should I do it?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
Atternatt: ohh, damn! yes! I made a mistake! merged with main instead of `arrow2` 🤦🏻

Atternatt: let me try to do a rebase with `arrow2` let's see if I don' die

Atternatt:
> > Is up to date with `arrow2` branch but not with `main` @serras
> 
> Great! Then we can merge for the Arrow 2.c series

sorry @serras I was wring, I made a mistake and merged main to this branch instead of `arrow2` 🤦🏻

nomisRev:
> I think the perfect alias would be Either<NonEmptyList<E>, A>, with some extension methods to create a Left with a singleton list.

Is the idea to make everything work over `NonEmptyList` since we can apply `Semigroup<Nel<A>>` and have some methods to flatten `E` rather than have it work over `Semigroup<E>`?

Atternatt:
Something happened after merging `arrow2` into the branch

serras:
@Atternatt what is the status of this?

nomisRev:
@Atternatt, we have a large number of conflicts 😱 Should we fix these first, and run the tests before reviewing?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -1551,105 +1490,228 @@ public fun <A, B> Iterable<Either<A, B>>.combineAll(MA: Monoid<A>, MB: Monoid<B>
...
+  ff: Either<E, F>,
+  g: Either<E, G>,
+  h: Either<E, H>,
+  i: Either<E, I>,
+  f: (A, B, C, D, EE, F, G, H, I) -> Z,
+): Either<E, Z> = zip(SE, b, c, d, e, ff, g, h, i, unit) { a, b, c, d, e, ff, g, h, i, _ ->
+  f(a, b, c, d, e, ff, g, h, i)
 }
 
+public inline fun <reified E, A, B, C, D, EE, F, G, H, I, J, Z> Either<E, A>.zip(
```
nomisRev, line 1603: `E` should not be reified in these `zip` implementations

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -283,153 +283,36 @@ public inline fun <B, C, D, E, F, G, H, I, J, K, L> Iterable<B>.zip(
...
-}
-
-@Deprecated("sequenceNullable is being renamed to sequence to simplify the Arrow API", ReplaceWith("sequence()", "arrow.core.sequence"))
-public fun <A> Iterable<A?>.sequenceNullable(): List<A>? =
-  sequence()
-
-public fun <A> Iterable<A?>.sequence(): List<A>? =
-  traverse(::identity)
+public inline fun <Error, A, B> Iterable<A>.mapAccumulating(
+  transform: (A) -> Either<NonEmptyList<Error>, B>,
```
nomisRev, line 314: Hmm, thinking if this should be `EitherNel<Error, B>` or just `Either<Error, B>` 🤔
I think the latter occurs more often, but perhaps we should do both. Or do both signatures conflict with each other?

arrow-libs/fx/arrow-fx-coroutines/src/commonTest/kotlin/arrow/fx/coroutines/ParTraverseEitherTest.kt
```diff
@@ -3,25 +3,26 @@ package arrow.fx.coroutines
...
-class ParTraverseEitherTest : ArrowFxSpec(
+//todo(#2728): @marc check if this test is still valid after removing traverse
+/*
+class parMapEitherTest : ArrowFxSpec(
   spec = {
-    "parTraverseEither can traverse effect full computations" {
+    "parMapEither can traverse effect full computations" {
       val ref = Atomic(0)
-      (0 until 100).parTraverseEither {
+      (0 until 100).parMapEither {
```
nomisRev, line 19: Where is this `parMapEither` coming from? Is this `parMapAccumulating`?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -283,153 +283,36 @@ public inline fun <B, C, D, E, F, G, H, I, J, K, L> Iterable<B>.zip(
...
-}
+public inline fun <Error, A, B> Iterable<A>.mapAccumulating(
+  crossinline combine: (first: Error, second: Error) -> Error,
+  transform: (A) -> Either<Error, B>,
+): Either<Error, List<B>> =
+  fold<A, Either<Error, ArrayList<B>>>(Right(ArrayList(collectionSizeOrDefault(10)))) { acc, a ->
+    when (val res = transform(a)) {
+      is Right -> when (acc) {
+        is Right -> acc.also { acc.value.add(res.value) }
+        is Left -> acc
```
nomisRev, line 294: @Atternatt looking at this implementation I think this is incorrect 🤔
The `combine` parameter is also unused, this looks like `traverseEither`.

Did we put all the relevant tests back?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
nomisRev: As we discussed in the bi-weekly Arrow meeting, I think something like this makes most sense.

```kotlin
public inline fun <Error, A, B> Iterable<A>.mapAccumulating(
  transform: (A) -> Either<Error, B>
): Either<NonEmptyList<Error>, List<B>> {
  val right = ArrayList<B>(collectionSizeOrDefault(10))
  val left = ArrayList<Error>()
  forEach { a ->
    when(val res = transform(a)) {
      is Either.Left -> left.add(res.value)
      is Either.Right -> right.add(res.value)
    }
  }
  return left.toNonEmptyListOrNull()?.left() ?: right.right()
}
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
nomisRev: I think we're going to need something nasty like `EMPTY_VALUE` and a specialised `combine` for it to do this efficiently :/

nomisRev:
```kotlin
public inline fun <Error, A, B> Iterable<A>.mapAccumulating(
  combine: (first: Error, second: Error) -> Error,
  transform: (A) -> Either<Error, B>
): Either<Error, List<B>> {
  val right = ArrayList<B>(10)
  var left: Any? = EMPTY_VALUE
  forEach { a ->
    when(val res = transform(a)) {
      is Either.Left -> left = if (left == EMPTY_VALUE) res.value else combine(EMPTY_VALUE.unbox(left), res.value) 
      is Either.Right -> right.add(res.value)
    }
  }
  return if (left == EMPTY_VALUE) right.right() else EMPTY_VALUE.unbox(left)
}
```

I think this would be the most efficient implementation.

nomisRev:
This PR proposes following changes:

# Integrate `Validated` into `Either`

Based on some discussion on KotlinLang, and quite some confusion why there is both `Validated` and `Either` especially amongst beginners we wanted to explore merging the two. So a small recap between the difference.

`Either<E, A>` and `Validated<E, A>` model the same ADT, which is the disjunction between `E` and `A`. `Either` allows you to write _happy-path_ code, and allows sequencing `Either<E, A>` values that depend on each-other using `bind` (`Monad`). Whilst `Validated` only allows combining values using `zip`, where you combine both `(E, E) -> E` and `(A, B) -> C`. (`Applicative`). The functionality of `Validated` can also be exposed over `Either`, and this PR proposes exactly that.

The reason for doing this is that we want to heavily decrease the API surface, and the resulting learning curve. After the 1.x.x series we realised that combing with the Arrow DSL having both types is not useful, and this is a good opportunity to heavily decrease the API surface. The [1.x.x Either API deprecation PR](https://github.com/arrow-kt/arrow/pull/2830) explains this a little bit as well.

## Integrating `Validated` into `Either`

In 1.x.x we deprecated `Either#zip` in favour of `either { f(bind(), b.bind(), ...) }` which also solves the arity-n issue of `zip`. This PR replaces the `Either#zip` behaviour with the _accumulating errors_ powers of `Validated`, **but we can also consider a different name than `zip`** such as `fa.validate(fb, ::Tuple2) ` or `fa.accumulate(fb, ::Tuple2)`. This PR proposes 3 different signatures to work with `Either` for validation purposes (only arity-2 shown below).

```kotlin
/** The most generic signature */
inline fun <E, A, B, C> Either<E, A>.zip(
  combine: (E, E) -> E,
  other: Either<E, B>,
  transform: (A, B) -> C
): Either<E, C>

/**  Automatically accumulate inside NonEmptyList, without requiring the original `E` to be wrapped */
inline fun <E, A, B, C> Either<E, A>.zip(
  other: Either<E, B>,
  transform: (A, B) -> C
): Either<NonEmptyList<E>, C>

/**
 * A function to combine results of the above function, or validations that would previously return `ValidatedNel`.
 * Regular `Either` values can be lifted into `NonEmptyList<E>` by using `toEitherNel()`.
 */
inline fun <E, A, B, C> Either<NonEmptyList<E>, A>.zip(
  other: Either< NonEmptyList<E>, B>,
  transform: (A, B) -> C
): Either<NonEmptyList<E>, C>
```

### traverse

Besides combining independent values of `Validated` a second big use-case for using it is `traverse` which is completely being deprecated for the Arrow DSL, and therefore we also need to find a solution for where you'd previously use `traverse + Validated`. To replace `traverse` one can simply use the DSL + `Iterable#map`, and therefore we thought `mapAccumulating` offers us an idiomatic name for _accumulating typed errors_ which is what `traverse + Validated` did.

```kotlin
either<String, List<Int>> {
   listOf(1, 2, 3).map { raise("fail") }
   //listOf(1, 2, 3).map { Either.Left("fail").bind() }
} // Either.Left("fail")

either<String, List<Int>> {
  listOf(1, 2, 3).mapAccumulating(String::plus) { raise("fail") }
  //listOf(1, 2, 3).mapAccumulating(String::plus) { Either.Left("fail").bind() }
} //  Either.Left("failfailfail")
```

In the same fashion as `zip` `mapAccumulating` also provides 2 different methods.

```kotlin
public inline fun <Error, A, B> Iterable<A>.mapAccumulating(
  semigroup: Semigroup<Error>,
  transform: Raise<Error>.(A) -> B,
): Either<Error, List<B>>

public inline fun <Error, A, B> Iterable<A>.mapAccumulating(
  transform: Raise<Error>.(A) -> B,
): Either<NonEmptyList<Error>, List<B>>
```

It works over `Raise<E>`, but returns a concrete `Either`.  Snippet below compares `traverse` to `mapAccumulating` usages:

```kotlin
fun process(i: Int): Either<String, Int> = i.right()

listOf(1, 2, 3).traverse { process(it) }
listOf(1, 2, 3).mapAccumulating { process(it).bind() }
```

This API is optimised for context receivers, or working with `Raise<E>` extension functions. Otherwise it requires the additional call to `bind()`.

#### Additional changes
##### Semigroup

To make this PR compile, and these functions to accept `(E, E) -> E` where `fun interface Semigroup` is defined the abstract function of `Semigroup` had to change shape from `E.(E) -> E` to `(E, E) -> E`. This is done in a non-binary breaking way, but we can still consider removing `Semigroup` in a different PR or changing its APIs.

raulraja (approved):
Thanks @Atternatt @nomisRev !

pakoito:
`mapAccumulating` -> `mapAccumulatingLeft` and I'd still keep the equivalent of `sequence` to flatten lists of `Either`

nomisRev:
Hey @pakoito,
Thanks for sharing, great to hear that you've also experienced this and looked for a middle ground in 1.x.x 😁

> mapAccumulating -> mapAccumulatingLeft

I felt `mapAccumulating` was already too long 😂 The name is definitely open for suggestions 🙏 I personally feel `mapAccumulatingLeft` is a bit long, but `mapAccumLeft` is perhaps a bit foreign. Should it reflect `Raise`, or did you choose `Left` in the name because of the `Either` return type? Perhaps `collectRaise` 🤔 `collect` conflicts a bit with `Flow#collect` though..

Do you like the new signatures?

> I'd still keep the equivalent of sequence to flatten lists of Either

We can definitely do that, but `sequence` is perhaps not the best name for this 🤔 Any suggestions?

pakoito:
The convention in the stdlib is "XOrY", so this could follow:

For mapping: `mapOrAccumulate`, `mapOrRaiseFirst`

For sequencing, "merge" isn't taken:  `mergeOrRaiseFirst` + `mergeOrAggregate`.

nomisRev:
> For mapping: mapOrAccumulate, mapOrRaiseFirst

`mapOrRaiseFirst` do you mean short-circuit behaviour? That is just `map`.

```kotlin
object Error

fun process(i: Int): Either<Error, Int> = i.right()
fun Raise<Error>.process2(i: Int): Int = i

either<Error, List<Int>> {
  listOf(1, 2, 3).map { process(it).bind() }
  listOf(1, 2, 3).map { process2(it) }
}
```

I do like `mapOrAccumulate`! It's more descriptive than `mapAccumulate` since with `mapAccumulate` the `Accumulate` seems to refer to _happy-path_ whilst `Or` is a Kotlin Std established pattern that it refers to _unhappy-path_. Thanks @pakoito ❤️

> For sequencing, "merge" isn't taken: mergeOrRaiseFirst + mergeOrAccumulate.

I was originally thinking of `flatten` but didn't propose it since there is no actually `flattening` (of `List` or otherwise) happening. Arrow already has `Either<A, A>.merge`, do you think it would be confusing to use `merge` here as well?

> Nested zips will be a bit meh, but I'll probably reach for mergeOrAccumulate + deconstruction instead.

Do you mean nested zips to achieve arities higher than 9? `mergeOrAccumulate + deconstruction` is untyped 😭

pakoito:
> Arrow already has Either<A, A>.merge, do you think it would be confusing to use merge here as well?

I'd rename that merge to `converge` or similar, if that frees the rerm. It's an extfun over a different type so it'll light up during migration.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/typeclasses/Semigroup.kt
```diff
@@ -6,23 +6,27 @@ import arrow.core.Endo
...
-import arrow.core.Validated
 import arrow.core.combine
 import arrow.core.compose
 import kotlin.jvm.JvmName
 import kotlin.jvm.JvmStatic
 
 public fun interface Semigroup<A> {
+  
+  // TODO: think of better name
+  public fun append(a: A, b: A): A
```
i-walker, line 17: is this still part of the Validated removal?
And is `append` going to replace `combine`

i-walker (approved):
Looks good, I would opt to separate the changes in Semigroup in a separate PR. 
Is there going to be another PR on writing the stack-safety tests for NonEmptyList, Iterable, etc, or can I pick that up?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/typeclasses/Semigroup.kt
nomisRev: I addressed it briefly in my comment, the problem is that the current `combine` signature doesn't match the signature we want for `zip`. So these changes belong together. There has been some talk about removing `Semigroup`, which would remove this changes (and horrible API name). And would replace the `Semigroup<E>` argument with a simple lambda `(E, E) -> E`. cc\\ @raulraja 

> Additional changes
>
> Semigroup
>
> To make this PR compile, and these functions to accept (E, E) -> E where fun interface Semigroup is defined the abstract function of Semigroup had to change shape from E.(E) -> E to (E, E) -> E. This is done in a non-binary breaking way, but we can still consider removing Semigroup in a different PR or changing its APIs.

nomisRev:
> I'd rename that merge to converge or similar, if that frees the term. It's an extfun over a different type so it'll light up during migration.

@pakoito I'm hesitant to rename it since it's been quite well established by now. It's a quite popular API afaik, and there will already be so many changes. Also `(List<Either<E, A>>) -> Either<E, List<A>>` doesn't really merge. When I explain it to someone unfamiliar with the pattern I typically say it _flips_ the container `List` and `Either`.  But I don't like _flip_ for an API name.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/typeclasses/Semigroup.kt
i-walker: Sorry I missed that :)

serras:
About the `sequence` function, I feel like the following expresses the intent quite well.

```kotlin
either { thing.map { it.bind() } } 
```

I could see that this is a bit complicated to write the first time, though. However, we should remember that the intention is to move people towards the `Raise` DSL, where you don't need explicit `bind` at all.

Going into the `bind` direction, may I suggest instead to have the following function?

```kotlin
fun <E, A> Iterable<Either<E, A>>.bindAll(): List<A> = map { it.bind() }
```

This would be just a short-hand for "bind everything in this `Iterable`", in other words, "check that everything in this `Iterable` is correct". This is slightly more generic than `sequence`, because we can use it within an `either` block, but still write `either { listOfEithers.bindAll() }` if needed.

serras:
I'm merging this PR so we can continue work in other branches. I've moved the discussion about the "missing function" for iterables of `Either`s to https://github.com/arrow-kt/arrow/discussions/2861.

PR #2797, opened 2022-08-17
[PROPOSAL] Effect without suspending shift

nomisRev:
This is proposal is second proposal alongside #2782 after some discussions with @raulraja based on some other experimental / explatory implementations.

# Design choices
## Removal of suspend from shift

Both `EagerEffectScope` and `EffectScope` implement `shift` using a subclass of `CancellationException`. This is done to make `shift` behave as a _proper short-circuit_ when mixed with KotlinX Coroutines Structured Concurrency. Such that following bracket laws are respected.

```kotlin
effect<String, Int> {
  bracketCase(
    acquire = { 1 }
    use = { shift<Int>("error") }
    releaseCase = { _, exitCase -> println(exitCase) } // prints ExitCase.Cancelled
   )
}.toEither() // Either.Left("error")
```

Due to both implementations of `EffectScope` and `EagerEffectScope` simply throw `CancellationException` the `shift` function does not require `suspend`. So we can unify it under a single `Shift`, and we do not longer require `@RestrictSuspension`.

This changes the `typealias` of `Effect`, and `EagerEffect` to only differ in `suspend`, and means that we can now write truly polymorphic code when we're inside the context of `Shift`, since we can require `suspend` only when we rely on its _capabilities_. (I'm using extension functions below to replace/emulate the not yet released `context` receivers.)

```kotlin
public typealias Effect<R, A> = suspend Shift<R>.() -> A
public typealias EagerEffect<R, A> = Shift<R>.() -> A

fun Shift<String>.error(): Int = shift("error")

suspend fun Shift<String>.parallel(): Int = parZip({ error() }, { error() }) { x, y -> x + y }
```

## Inline fold

Since we now have a single `Shift`, we can now also implement a single `fold` which is `inline` so it allows for `suspend` to pass through! Below is the definition of our _base_ implementation, and we can derive all implementations for `Effect`, `EagerEffect`, `Either`, etc from it.

```kotlin
public inline fun <R, A, B> fold(
  @BuilderInference program: Shift<R>.() -> A,
  error: (error: Throwable) -> B,
  recover: (shifted: R) -> B,
  transform: (value: A) -> B,
): B
```

What is interesting here is that `fold` is `inline`, so depending on the _caller_ it will allow `suspend` or not.
This however means we can now also implement our `either` DSL as `inline`, which means we no longer need to differentiate between `suspend` and `eager`.

```kotlin
public inline fun <E, A> either(@BuilderInference block: Shift<E>.() -> A): Either<E, A>
```

This allows us to remove an additional complexity pain-point from Arrow. Since there is quite some confusion as to why we need `eager` **and** `suspend operator fun invoke`. This new `fold` implementation removes that limitation completely, and automatically works correctly depending on the callers context.

## Catch / Attempt

As an added benefit we can now implement `catch` for `inline` for `Either`, alongside `catch` for `Effect` and `EagerEffect`. Because the `catch` implementation for `Either` is `inline` it still allows you to use `suspend` if desired.

```kotlin
public inline fun <E2, E, A> Either<E, A>.catch(block: Shift<E2>.(E) -> A): Either<E2, A> = TODO()

val error: Either<String, Int> = Either.Left("error")

val example2: Either<Nothing, Int> = error.catch { it.length }

suspend fun example(): Either<List<Char>, Int> =
  error.catch { str ->
    delay(100)
    shift(str.toList())
  }
```

## Conclusion

Because we removed `suspend` from `shift` we are able to:
 1. Unify `EagerShift` and `Shift` into a single type
 2. `inline` fold to unify internal implementation, and remove all internal complexity of low-level coroutines
 3. `inline` monadic builders, supporting both regular and suspending code seamlessly
 4. Implement `catch` for suspending, and eager data types.


raulraja:
Thanks, @nomisRev I think this is a significant improvement in both runtime and declarations. 👍 ❤️

JavierSegoviaCordoba:
Really great job 👏

nomisRev:
This PR takes precedence over #2782 due to some new findings. I converted it back into a `DRAFT` because it needs some documentation before it can be merged.

All code is complete, and all tests are present. So the code can be reviewed.


I want to challenge the `EffectScope<E>` to `Shift<E>` renaming. It was something that was discussed offline, but I want to open it up here. `Shift` is not very descriptive about what happens. Ideally some1 should be able to grasp what is happening when looking at this code:

```kotlin
context(Shift<Error>)
suspend fun example(): Int = shift(Error)
```

Naming that is more explicit about _errors_ can be more descriptive:

```kotlin
context(Raise<Error>)
suspend fun example(): Int = raise(Error)
```

or

```kotlin
context(CanFail<Error>)
suspend fun example(): Int = fail(Error)
```

It's a bit imprecise because `RaiseError`/`CanFail` can be used also for short-circuiting but not _error_ paths perse.
99% of the use-cases will use it for errors... Similar to how `MonadError` might be used in Scala/Haskell for it's short-circuit behaviour in non-failure settings.

Open to all suggestions, or alternative names. I didn't suggest `ShortCircuit` because I felt it was non-descriptive like `Shift`.

sksamuel:
`RaiseError` is more clear to the non-FP user than Shift.

i-walker:
Since this was raised a couple of times, we could either favour failure-settings and keep only `Raise` or have `Raise` as an type-alias since `Shift` does encompass both scenarios failure-settings or just utilising short-circuit behaviour with non-failure settings. 
If the type-alias isn't an option, maybe too verbose or documentation wise.
I'd prefer `Shift` over `Raise` for the best case. 
We can also hold a pool in Github just like we did for `bind`?
wdyt cc @raulraja , @nomisRev @serras @arrow-kt/maintainers

nomisRev:
@i-walker we've started a small poll on KotlinLang, but we can easily hold other polls and accumulate the data.

I think we've been able to gather some good feedback already, https://kotlinlang.slack.com/archives/C5UPMM0A0/p1661170633105679

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Builders.kt
```diff
@@ -0,0 +1,104 @@
...
+  public fun <B> Option<B>.bind(): B = bind { shift(None) }
+  public fun ensure(value: Boolean): Unit = ensure(value) { None }
+  
+  public fun <B> ensureNotNull(value: B?): B {
+    contract { returns() implies (value != null) }
+    return ensureNotNull(value) { None }
+  }
+}
+
+public class IorShift<E> @PublishedApi internal constructor(semigroup: Semigroup<E>, private val effect: Shift<E>) :
```
raulraja, line 73: These may be an exception to what I said above, but it's also not a value class since it takes more than one arg.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/ErrorHandlers.kt
```diff
@@ -0,0 +1,111 @@
...
+ * If you don't need an `error` value when wrapping your foreign code you can use `Nothing` to fill the type parameter.
+ *
+ * ```kotlin
+ * val y = effect<Nothing, User> {
+ *   throw IllegalArgumentException("builder missed args")
+ * }.attempt<IllegalArgumentException, Error, User> { shift(Error) }
+ * ```
+ * <!--- KNIT example-effect-error-03.kt -->
+ */
+@JvmName("attemptOrThrow")
```
raulraja, line 85: nit: it does not really `throw` since it captures what it throws back into an effect when it does not match `T`.
`orThrow` implies when I read it that it is unsafe, and that a call when it fails blows up with an exception. Perhaps we can use something like `attemptWithThrowableType` or a similar JVM name indicating we are just matching on type args and letting others unmatched go through but this method is actually safe.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Fold.kt
```diff
@@ -0,0 +1,86 @@
...
+  } catch (e: Throwable) {
+    error(e.nonFatalOrThrow())
+  }
+}
+
+/**
+ * **AVOID USING THIS TYPE, it's meant for low-level cancellation code** When in need in low-level code,
+ * you can use this type to differentiate between a foreign [CancellationException] such as JobCancellationException, and the one from [Effect].
+ */
+public class ShiftCancellationException(
```
raulraja, line 70: can this be internal or private?

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/continuations/EagerEffectSpec.kt
```diff
@@ -28,13 +30,13 @@ class EagerEffectSpec : StringSpec({
...
-
+  
   "try/catch - finally works" {
     checkAll(Arb.string(), Arb.int()) { s, i ->
       val promise = CompletableDeferred<Int>()
-      eagerEffect<String, Int> {
+      eagerEffect {
         try {
-          shift(s)
+          shift<Int>(s)
```
raulraja, line 39: Perhaps we should explore after this PR if `shift` can be implemented by returning `Nothing` and if it brings better or worse ergonomics / type inference

raulraja (commented):
Looking great and getting even simpler 👏 @nomisRev!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Builders.kt
```diff
@@ -0,0 +1,104 @@
...
+  }
+  
+  public fun <B> ensureNotNull(value: B?): B {
+    contract { returns() implies (value != null) }
+    return ensureNotNull(value) { null }
+  }
+}
+
+@JvmInline
+public value class ResultShift(private val cont: Shift<Throwable>) : Shift<Throwable> {
```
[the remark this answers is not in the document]
nomisRev: The reason for this type is to achieve the following _context receiver_ method. So sadly I think we need this type until we can add this top-level defintion.

```kotlin
context(Shift<Throwable>)
suspend fun <A> Result<A>.bind(): A = ...
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Builders.kt
nomisRev: Sam issue, it can be implemented for `StateShift<State, R>` if we have such a type. (Local prototype, shared with you on Slack). Can share gist here for everyone that is interested, otherwise I'll raise it as a follow-up PR.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/ErrorHandlers.kt
nomisRev: Right... I think I originally had this on the DSL method, where it made more sense since it threw immediately into the `suspend` context. It's only needed when you call this API from Java, so it's not super critical.

`attemptRefined`, `attempt`, or maybe just `attemptReified` since it also would have a `Class<T>` param in the Java API?

For reference from Scala APIs:
 -   Cats-effect: `def rethrow[F : MonadError[F, E], A, EE <: E](fa: F[Either[EE, A]]): F[A]`
 - ZIO: `def refineOrDie[E1](pf: PartialFunction[E, E1])(implicit ev1: <:<[E, Throwable], ev2: CanFail[E]): ZIO[R, E1, A]`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Fold.kt
nomisRev: This could be `@PublishedApi internal`, meaning that it would still be part of the public API.

The reason I kept it public is that people could use it to distinct `CancellationException` between `JobCancellationException` and `ShiftCancellationException`, but `JobCancellationException` is internal in KotlinX...

TBH I haven't found an actual use-case for this, but it's needs to be public from the Java binary anyway since it appears inside `inline`.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/continuations/EagerEffectSpec.kt
nomisRev: Agreed, that would be a good thing to explore. I found the argument that `fun error` returns `Nothing` very compelling. Comparing it to `emptyList` was probably incorrect.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Builders.kt
i-walker: Can you send in a gist link, please? 😄

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/ErrorHandlers.kt
raulraja: I just realized that this might not matter since you can't call inline reified methods from Java unless you provide an additional wrapper function that takes a java `Class` as an argument. In any case, a better name would be, as you said, `attemptReified` or similar.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Fold.kt
raulraja: I'd make it `@PublishedApi internal` so it's not advertised until we have a use case in Kotlin or need outside the library. By the shape of the type, it seems this is not something people will ever construct manually. The other use case is handling it as an exception which we do ourselves internally.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Mappers.kt
```diff
@@ -0,0 +1,45 @@
...
+import arrow.core.Ior
+import arrow.core.None
+import arrow.core.Option
+import arrow.core.Some
+import arrow.core.Validated
+import kotlin.jvm.JvmMultifileClass
+import kotlin.jvm.JvmName
+
+/** Run the [Effect] by returning [Either.Right] of [A], or [Either.Left] of [E]. */
+public suspend fun <E, A> Effect<E, A>.toEither(): Either<E, A> = either { invoke() }
```
raulraja, line 16: Should these functions be `inline`?

raulraja (approved):
great work @nomisRev !

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Fold.kt
nomisRev: I went a step further, and changed the internals a bit so I could make this type private! 🥳

I agree, since there is not strong use-case to have it public it should not be visible. If such a use-case arises in the future we can ease this restriction.

nomisRev:
Oops. There was only a single approval. @serras do we need to protect the `arrow-2` branch with same rules as `main`?

If there are any additional comments, feel free to post them here or in a new issue and we'll tackle them in as subsequent PR. :pray:

PR #2798, opened 2022-08-17
Remove traverse and replace zip

Atternatt:


Atternatt:
@nomisRev So far I've created this zip function taking as reference the one you shared with me some days ago: [link](https://github.com/nomisRev/kotlin-gists/blob/d5943ef9dd09f765d32506508c481d3b80ac447d/src/main/kotlin/arrow2/Either.kt#L285)

do you think the implementation make sense?

nomisRev:
Hey @Atternatt,

Thanks for looking into this. The implementation is super ugly I know it 😭
We have it like that in Arrow 1.x.x as well. It's the most efficient, and simplest implementation we've found but definitely open to suggestions that read/look better.

https://github.com/arrow-kt/arrow/blob/5d0bc9e214f9b92f8492a8202a7c165d5e81fa9e/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Validated.kt#L526


I saw you're using `NonEmptyList` which creates an underlying `List` and then `foldRight` will also introduce a `while` with a new `List`. So this change might include a bunch of overhead that is not really improving readability too much. WDYT?

Atternatt:
I see the point. You are right I does not make sense in this case to use fold if we are adding 2 extra lists. Perhaps the easiest and most effective way is just implement it as it was originally done. will change it.

Atternatt:
Now that we are replacing the zip on `Either` does it make sense this change I did in `Ior` ?

```
public inline fun <AA, C, D> bitraverseEither(
    fa: (A) -> Either<AA, C>,
    fb: (B) -> Either<AA, D>
  ): Either<AA, Ior<C, D>> =
    fold(
      { a -> fa(a).map { Left(it) } },
      { b -> fb(b).map { Right(it) } },
      { a, b ->
        fa(a).flatMap { aa ->
          fb(b).map { c -> Both(aa,c) }
        }
      }
    )

  public inline fun <reified AA, C, D> bitraverseAccumulated(
    SA: Semigroup<AA>,
    fa: (A) -> Either<AA, C>,
    fb: (B) -> Either<AA, D>
  ): Either<AA, Ior<C, D>> =
    fold(
      { a -> fa(a).map { Left(it) } },
      { b -> fb(b).map { Right(it) } },
      { a, b -> fa(a).zip(SA, fb(b)) { aa, c -> Both(aa, c) } }
    )
```

nomisRev:
Hey @Atternatt, we can talk about this tomorrow in the bi-weekly. TBH I think we can get rid of a lot of these methods since they're pretty complex, and therefore not used since using Arrow DSLs is much simpler.

PR #2804, opened 2022-08-22
[BUGFIX] Result.zip

nomisRev:
Fixes Result.zip which results in NullPointerException instead of the correct `null` success value.


frko:
Great how you pick this up quickly and provided a fix @nomisRev. Happy to observe the sense of ownership here.

i-walker (approved):
Thanks @nomisRev 🙌🏾

franciscodr (approved):

Issue #2812, opened 2022-09-01
["Request"] Improve Iterable extension impl


raulraja:
Some of our Iterable extensions, such as `partitionEither` have non-optimal implementations that can be easily fixed to produce fewer allocations and iterations.

https://github.com/arrow-kt/arrow/blob/6aa1875a687fb2c1c0e639bdd73be4a05bdfbf7a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt#L1002

Discussion with Mitchell Cohen from slack regarding `partitionEither` examples: https://kotlinlang.slack.com/archives/C5UPMM0A0/p1662064718102909?thread_ts=1559561395.141900&cid=C5UPMM0A0

Khepu:
The exact same pattern is followed here:
https://github.com/arrow-kt/arrow/blob/6aa1875a687fb2c1c0e639bdd73be4a05bdfbf7a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt#L1014

This also applies to `Sequence` and `Option`, would changes for all of those be included in the same PR?

Tests for these methods are lacking from what I could tell, would it make sense to create those?

raulraja:
> This also applies to Sequence and Option, would changes for all of those be included in the same PR?
that would be great

> Tests for these methods are lacking from what I could tell, would it make sense to create those?
yes, that'd be great, thanks!

nomisRev:
Hey @Khepu,

> Tests for these methods are lacking from what I could tell, would it make sense to create those?
Yes, that would be greatly appreciated 🙏 There is some tests missing here, and there but we're trying to fill them in over time whenever we have to touch related code.

> This also applies to Sequence and Option, would changes for all of those be included in the same PR?

Yes, you can include `separateValidated` and both methods for `Sequence` as well. I'm not sure this applies to `Option`, since `Option` only holds a single value, but I'm not sure which methods you're referring to.

Also feel free to put it in a separate PR if that is easier for you.

PR #2814, opened 2022-09-02
add separateEither and separateValidated test

hoc081098:
related to #2812

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
+++ b/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
@@ -479,5 +479,25 @@ class IterableTest : UnitSpec() {
         listOfOption.flattenOption() shouldBe listOfOption.mapNotNull { it.orNull() }
       }
     }
+
+    "separateEither" {
+      checkAll(Arb.list(Arb.int())) { ints ->
+        val list = ints.map {
+          if (it % 2 == 0) it.left()
+          else it.right()
+        }
+        list.separateEither() shouldBe ints.partition { it % 2 == 0 }
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
```diff
@@ -479,5 +479,25 @@ class IterableTest : UnitSpec() {
...
       }
     }
+
+    "separateEither" {
+      checkAll(Arb.list(Arb.int())) { ints ->
+        val list = ints.map {
+          if (it % 2 == 0) it.left()
+          else it.right()
+        }
+        list.separateEither() shouldBe ints.partition { it % 2 == 0 }
```
raulraja, line 489: thanks @hoc081098 !. 

Maybe I'm missing something obvious or it's too late here but I believe this test is not correct because the assertion compares results of different types.

`Pair<List<Left<L>>, List<Right<R>>>` vs `Pair<List<Int>, List<Int>`. Just calling `partition` won't result in a new list of either but instead in a list of ints. 

What puzzles me is why tests are passing.

hoc081098: `separateEither` returns `Pair<List<L>>, List<R>>` (values inside Lefts/Rights)

raulraja: My bad @hoc081098, for a second there I thought `separateEither` returned the results still in the Either structure.

raulraja (approved):
thanks @hoc081098 !

nomisRev (approved):
LGTM @hoc081098

PR #2815, opened 2022-09-02
Refactor - #2812 iterable separate performance

Khepu:
related to #2812

A single pass is now made over the original `Iterable` instead of 2. I am not sure if using mutable lists and a single pair would offer any significant advantages but I am open to changing it if that is considered better.

I ran the tests introduced in #2814 at my local setup and they seem to pass.

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
@@ -999,11 +999,13 @@ public fun <A, B> Iterable<Validated<A, B>>.uniteValidated(): List<B> =
  * @receiver Iterable of Validated
  * @return a tuple containing List with [Either.Left] and another List with its [Either.Right] values.
  */
-public fun <A, B> Iterable<Either<A, B>>.separateEither(): Pair<List<A>, List<B>> {
-  val asep = flatMap { gab -> gab.fold({ listOf(it) }, { emptyList() }) }
-  val bsep = flatMap { gab -> gab.fold({ emptyList() }, { listOf(it) }) }
-  return asep to bsep
-}
+public fun <A, B> Iterable<Either<A, B>>.separateEither(): Pair<List<A>, List<B>> =
+  fold(listOf<A>() to listOf<B>()) { (lefts, rights), either ->
+    when (either) {
+      is Left -> lefts + either.value to rights
+      is Right -> lefts to rights + either.value
+    }
+  }
 
 /**
  * Separate the inner [Validated] values into the [Validated.Invalid] and [Validated.Valid].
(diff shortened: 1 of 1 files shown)


raulraja (approved):
thanks @Khepu !

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -997,23 +997,27 @@ public fun <A, B> Iterable<Validated<A, B>>.uniteValidated(): List<B> =
  * @receiver Iterable of Validated
  * @return a tuple containing List with [Either.Left] and another List with its [Either.Right] values.
  */
-public fun <A, B> Iterable<Either<A, B>>.separateEither(): Pair<List<A>, List<B>> {
-  val asep = flatMap { gab -> gab.fold({ listOf(it) }, { emptyList() }) }
-  val bsep = flatMap { gab -> gab.fold({ emptyList() }, { listOf(it) }) }
-  return asep to bsep
-}
+public fun <A, B> Iterable<Either<A, B>>.separateEither(): Pair<List<A>, List<B>> =
+  fold(listOf<A>() to listOf<B>()) { (lefts, rights), either ->
```
nomisRev, line 1001: Should we use `mutableListOf` here and use `add` inside of the `fold` body? That would avoid having to recreate `Pair` on every element of the `List`. The returned type can still be _read-only_ `List`.

nomisRev: It can also be rewritten to `while` if we'd pattern similar to Kotlin Std implementations.

```kotlin
val left = ArrayList<A>(collectionSizeOrDefault(10))
val right = ArrayList<B>(collectionSizeOrDefault(10))
forEach { either ->
  when(either) {
     is Left -> left.add(either.value)
     is Right -> right.add(either.value)
   }
}
Pair(left, right)
```

nomisRev: This eliminates List.size allocations.

Khepu: I think mimicking the Kotlin stdlib is a good way to go. Looking at the implementation it's gonna be a lot like this (based on `map`):

```kotlin
public fun <A, B> Iterable<Either<A, B>>.separateEither(): Pair<List<A>, List<B>> {
  val left = ArrayList<A>(collectionSizeOrDefault(10))
  val right = ArrayList<B>(collectionSizeOrDefault(10))

  for (either in this)
    when(either) {
      is Left -> left.add(either.value)
      is Right -> right.add(either.value)
  }

  return Pair(left, right)
}
```

If that is something we are happy with, let me know!

nomisRev: Yes, that looks perfect to me @Khepu!

nomisRev (approved):
Looks perfect! Thank you @Khepu 🙌

PR #2816, opened 2022-09-03
Tracing POC

nomisRev:
Discussing some ideas for new features with @raulraja he mentioned tracing for `Effect`, so I was playing around with some ideas and come up with a small POC.

Since `raise` already relies on `CancellationException` to co-operate with the coroutine system we currently pay a performance penalty for `raise` and the stack trace creation that comes with `CancellationException` by default. This penalty is not needed, and is stack trace creation is [by default also disabled in KotlinX Coroutines](https://github.com/Kotlin/kotlinx.coroutines/blob/d88a8c227a295f91251960c8fe6b21a02ddf704b/kotlinx-coroutines-core/jvm/src/Exceptions.kt#L40). 

This PR solves the issue of _debugging_/_tracing_, and the performance penalty we currently pay for _raise_. 
We can now disable the stack trace creation, and we can ad-hoc enable it for _any_ `Effect` you compose _traced_ upon or any `Raise` program you run within a `traced` block.

```kotlin
public fun main() {
  val error = effect<String, Int> { raise("error") }
  error.traced { traced -> traced.printStackTrace() }
    .fold({ require(it == "error") }, { error("impossible") })
}
```
```text
arrow.core.continuations.RaiseCancellationException: Raised Continuation
  at arrow.core.continuations.DefaultRaise.raise(Fold.kt:77)
  at MainKtKt$main$error$1.invoke(MainKt.kt:6)
  at MainKtKt$main$error$1.invoke(MainKt.kt:6)
  at arrow.core.continuations.Raise$DefaultImpls.bind(Shift.kt:22)
  at arrow.core.continuations.DefaultRaise.bind(Fold.kt:74)
  at arrow.core.continuations.Effect__TracingKt$traced$2.invoke(Tracing.kt:46)
  at arrow.core.continuations.Effect__TracingKt$traced$2.invoke(Tracing.kt:46)
  at arrow.core.continuations.Effect__FoldKt.fold(Fold.kt:92)
  at arrow.core.continuations.Effect.fold(Unknown Source)
  at MainKtKt.main(MainKt.kt:8)
  at MainKtKt.main(MainKt.kt)
```

# Notes

- Since the `Effect` runtime is shared amongst DSL such as `either`, `result`, `option`, `ior`, etc the `traced` operator can be added, or used for all these DSLs, and types. Including any of the DSL methods that result in shifting such as `ensure`, `ensureNotNull`, `bind`, etc.

- Currently only JVM is disabling stack-trace generation. Not sure if we also need to disable it for JS, or Native. (KotlinX Coroutines only disables it for JVM)

- What API would we want on `Traced<R>`? Currently we include mentions to `ShiftCancellationException` and the dynamic call to `DefaultShift#shift` which is not part of the user stack trace. Should we slightly change the API such that this is improved?


arrow-libs/core/arrow-core/src/jsMain/kotlin/arrow/core/continuations/ShiftCancellationExceptionNoTrace.kt
```diff
@@ -0,0 +1,7 @@
+package arrow.core.continuations
+
+
```
i-walker, line 3: nit: spacing

i-walker (approved):
Looks great

nomisRev:
Thank you @franciscodr for applying the fixes ❤️

nomisRev:
Closing in favor of #2946

PR #2817, opened 2022-09-03
Add Atomic module, and StateShift

nomisRev:
This PR splits `Atomic` to a base module. This is something that is often needed in many projects, and it would be useful to have it as a smaller base binary.

Additionally it also solves the use-case we have for `Ior` but it offers a lower level builder for it.

Currently `StateShift<S, A>` is just a boilerplate type for `context(Atomic<S>, Shift<A>)`.


i-walker (approved):
Looks great! 
One question are there going to be some general purpose tests for StateShift, Atomic or are we adding them in another PR?

nomisRev:
@i-walker I wanted to gather some feedback on this API. I’ll add tests and more docs before making it ready for review.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Builders.kt
```diff
@@ -87,18 +95,9 @@ public class IorShift<E> @PublishedApi internal constructor(semigroup: Semigroup
...
-    val leftState = leftState.get()
-    return if (leftState === EmptyValue) Ior.Right(res)
-    else Ior.Both(EmptyValue.unbox(leftState), res)
-  }
-  
-  @Suppress("UNCHECKED_CAST")
+
   private fun combine(other: E): E =
-    leftState.updateAndGet { state ->
-      if (state === EmptyValue) other else EmptyValue.unbox<E>(state).combine(other)
```
raulraja, line 102: do we still need `EmptyValue` after these changes?

nomisRev: I'd need to check, maybe it's not used anywhere anymore after this change. I choose `Option` here, but the `Option` allocations could be removed using `EmptyValue` again.

raulraja (approved):
thanks @nomisRev

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/StateRaise.kt
```diff
@@ -0,0 +1,39 @@
...
+
+public typealias StateEffect<State, R, A> = suspend StateRaise<State, R>.() -> A
+
+public typealias EagerStateEffect<State, R, A> = StateRaise<State, R>.() -> A
+
+public inline fun <State, R, A, B> fold(
+  initial: State,
+  @BuilderInference program: StateRaise<State, R>.() -> A,
+  error: (state: State, error: Throwable) -> B,
+  recover: (state: State, shifted: R) -> B,
```
JavierSegoviaCordoba, line 22: `shifted` should be `raised`?

PR #2818, opened 2022-09-03
Refactor - #2812 sequence separate performance

Khepu:
related to #2812 similar to #2815 and #2814

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
@@ -593,11 +593,13 @@ public fun <A> Sequence<A>.salign(
  * @receiver Iterable of Validated
  * @return a tuple containing Sequence with [Either.Left] and another Sequence with its [Either.Right] values.
  */
-public fun <A, B> Sequence<Either<A, B>>.separateEither(): Pair<Sequence<A>, Sequence<B>> {
-  val asep = flatMap { gab -> gab.fold({ sequenceOf(it) }, { emptySequence() }) }
-  val bsep = flatMap { gab -> gab.fold({ emptySequence() }, { sequenceOf(it) }) }
-  return asep to bsep
-}
+public fun <A, B> Sequence<Either<A, B>>.separateEither(): Pair<Sequence<A>, Sequence<B>> =
+  fold(sequenceOf<A>() to sequenceOf<B>()) { (lefts, rights), either ->
+    when (either) {
+      is Left -> lefts + either.value to rights
+      is Right -> lefts to rights + either.value
+    }
+  }
 
 /**
  * Separate the inner [Validated] values into the [Validated.Invalid] and [Validated.Valid].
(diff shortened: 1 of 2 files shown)


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
```diff
@@ -593,23 +593,27 @@ public fun <A> Sequence<A>.salign(
  * @receiver Iterable of Validated
  * @return a tuple containing Sequence with [Either.Left] and another Sequence with its [Either.Right] values.
  */
-public fun <A, B> Sequence<Either<A, B>>.separateEither(): Pair<Sequence<A>, Sequence<B>> {
-  val asep = flatMap { gab -> gab.fold({ sequenceOf(it) }, { emptySequence() }) }
-  val bsep = flatMap { gab -> gab.fold({ emptySequence() }, { sequenceOf(it) }) }
-  return asep to bsep
-}
+public fun <A, B> Sequence<Either<A, B>>.separateEither(): Pair<Sequence<A>, Sequence<B>> =
```
hoc081098, line 596: IMO, the return type should be `Pair<List<A>, List<B>>`. Because we evaluate this sequence and return sequence is not lazy

Khepu: I do agree with that, is it something that I can change in this PR or would it be done along with other breaking changes?

nomisRev: @hoc081098 that is a great point! We should probably post-pone that breaking change for Arrow 2.0. Shall we create an Arrow 2.0 ticket for the breaking change, and go ahead with these changes?

@Khepu the same improvement I commented on the `Iterable.kt` PR can probably be applied here too.

Khepu: Just looked into how Sequence methods are implemented in the STD. It will be a bit different so as to avoid realizing the entire sequence just to separate it. Most of the implementations included an internal class that implements a custom iterator object. I should be able to have a PoC tomorrow though!

nomisRev:
Coming back to this PR after some time. Sorry for the inactivity.

I think @hoc081098 is correct, and it should actually be deprecated. See my message below from Slack.

> This signature cannot exist whilst preserving laziness by definition.
If you lazily compute Either, then there is no way to return `Pair<Sequence<Either<_, >, Sequence<Either<_,_>>>` without computing all the results first.

This is also confirmed by looking at the documentation of `Sequence#fold`, https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.sequences/fold.html

> The operation is terminal.

All that being said, I still think we should merge this PR. Anyone currently using this operator will benefit from the performance fix! And that way @Khepu efforts don't go to waste 👍
We should however deprecate it in favor of `Pair<List<Either<_, >, List<Either<_,_>>>` for 2.x.x.

nomisRev:
cc\\ @hoc081098 WDYT?

hoc081098:
@nomisRev I absolutely agree with you 💯

i-walker (approved):

PR #2820, opened 2022-09-07
Add Resource.allocated() to decompose Resource into it's allocate and…

custommonkey:
… release

Decomposes a [Resource]<A> into a Pair<suspend () -> A, suspend (A, ExitCase) -> Unit>, containing the allocation and release functions which make up the [Resource].

This can be used to integrate Resources with code which cannot be run within the [use] function.

### DIFF
--- a/arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
+++ b/arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
@@ -13,6 +13,11 @@ public sealed class ExitCase {
 
   public data class Cancelled(val exception: CancellationException) : ExitCase()
   public data class Failure(val failure: Throwable) : ExitCase()
+
+  public companion object {
+    public fun ExitCase(error: Throwable): ExitCase =
+      if (error is CancellationException) Cancelled(error) else Failure(error)
+  }
 }
 
 /**
--- a/arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
+++ b/arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
@@ -4,8 +4,8 @@ import arrow.core.continuations.AtomicRef
 import arrow.core.continuations.update
 import arrow.core.identity
 import arrow.core.prependTo
+import arrow.fx.coroutines.ExitCase.Companion.ExitCase
 import arrow.fx.coroutines.continuations.ResourceScope
-import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.DelicateCoroutinesApi
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.NonCancellable
@@ -181,9 +181,8 @@ public sealed class Resource<out A> {
           val a = dsl(effect)
           f(a)
         } catch (e: Throwable) {
-          val ex = if (e is CancellationException) ExitCase.Cancelled(e) else ExitCase.Failure(e)
           val ee = withContext(NonCancellable) {
-            effect.finalizers.get().cancelAll(ex, e) ?: e
+            effect.finalizers.get().cancelAll(ExitCase(e), e) ?: e
           }
           throw ee
         }
@@ -497,9 +496,8 @@ public sealed class Resource<out A> {
           }
           allocate to release
         } catch (e: Throwable) {
-          val ex = if (e is CancellationException) ExitCase.Cancelled(e) else ExitCase.Failure(e)
           val ee = withContext(NonCancellable) {
-            effect.finalizers.get().cancelAll(ex, e) ?: e
+            effect.finalizers.get().cancelAll(ExitCase(e), e) ?: e
           }
           throw ee
         }
--- a/arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
+++ b/arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
@@ -6,6 +6,7 @@ import arrow.core.identity
 import arrow.core.prependTo
 import arrow.fx.coroutines.continuations.ResourceScope
 import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.DelicateCoroutinesApi
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.flow.Flow
@@ -476,6 +477,7 @@ public sealed class Resource<out A> {
    *
    * This can be used to integrate Resources with code which cannot be run within the [use] function.
    */
+  @DelicateCoroutinesApi
   public suspend fun allocated(): Pair<suspend () -> A, suspend (@UnsafeVariance A, ExitCase) -> Unit> =
     when (this) {
       is Bind<*, A> ->
--- a/arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
+++ b/arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
@@ -5,13 +5,13 @@ import arrow.core.continuations.update
 import arrow.core.identity
 import arrow.core.prependTo
 import arrow.fx.coroutines.continuations.ResourceScope
-import kotlin.coroutines.CoroutineContext
 import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.withContext
+import kotlin.coroutines.CoroutineContext
 import kotlin.experimental.ExperimentalTypeInference
 
 /**
@@ -318,7 +318,7 @@ public sealed class Resource<out A> {
   public inline fun <B, C, D> zip(
     b: Resource<B>,
     c: Resource<C>,
-    crossinline map: (A, B, C) -> D,
+    crossinline map: (A, B, C) -> D
   ): Resource<D> =
     arrow.fx.coroutines.continuations.resource {
       map(bind(), b.bind(), c.bind())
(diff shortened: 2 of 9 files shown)


arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -464,12 +464,48 @@ public sealed class Resource<out A> {
...
+        }.allocated()
+      is Allocate -> acquire to release
+      is Defer -> resource().allocated()
+      is Dsl -> {
+        val effect = ResourceScopeImpl()
+        val allocated = try {
+          val allocate: suspend () -> A = suspend { dsl(effect) }
+          val release: suspend (A, ExitCase) -> Unit = { _, e ->
+            effect.finalizers.get().cancelAll(e)
+            Unit
```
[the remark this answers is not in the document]
nomisRev:
```suggestion
            effect.finalizers.get().cancelAll(e)?.let { throw it }
```

arrow-libs/fx/arrow-fx-coroutines/src/commonTest/kotlin/arrow/fx/coroutines/ResourceTest.kt
```diff
@@ -443,6 +453,44 @@ class ResourceTest : ArrowFxSpec(
...
+    suspend fun checkAllocated(mkResource: (() -> Int, (Int, ExitCase) -> Unit) -> Resource<Int>) {
+      val released = CompletableDeferred<Int>()
+      val releaseValue = Random.nextInt()
+      val seed = Random.nextInt()
+
+      val (allocate, release) = mkResource({ seed }) { _, _ -> released.complete(releaseValue) }.allocated()
+
+      allocate() shouldBe seed
+
+      release(1, ExitCase.Completed)
```
nomisRev, line 466: We're only testing the happy path with `ExitCase.Completed`, we should also test `ExitCase.Failed` and `ExitCase.Cancelled` code branches.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -464,12 +464,48 @@ public sealed class Resource<out A> {
...
     arrow.fx.coroutines.continuations.resource {
       parZip(ctx, { this@Resource.bind() }, { fb.bind() }) { a, b -> f(a, b) }
     }
 
+  /**
+   * Decomposes a [Resource]<A> into a Pair<suspend () -> A, suspend (A, ExitCase) -> Unit>, containing
+   * the allocation and release functions which make up the [Resource].
+   *
+   * This can be used to integrate Resources with code which cannot be run within the [use] function.
+   */
```
nomisRev, line 478: I think we need to be more elaborate in this piece of documentation. It's a very delicate API, and should also carefully explain its correct usage.

```suggestion
  /**
   * Deconstruct [Resource] into an `acquire` and `release` handlers.
   * The `release` action **must** always be called with resource [A] returned from `acquire`, if the `release` step is never called, then the resource [A] will leak. The `acquire` and `release` steps are already made `NonCancellable` to guarantee correct invocation like `Resource` or `bracketCase`.
   *
   * ```kotlin
   * val resource = Resource({ println("Acquire") }) { _, exitCase ->
   *  println("Release $exitCase")
   * }
   * 
   * suspend fun main(): Unit {
   *   val (acquire, release) = resource.allocated()
   *   val a = acquire()
   *   try {
   *     /** Do something with A */
   *     release(a, ExitCase.Completed)
   *   } catch(e: Throwable) {
   *      val e2 = runCatching { release(a, ExitCase(e)) }.exceptionOrNull()
   *      throw Platform.composeErrors(e, e2)
   *   }
   *}
   * ```
   * 
   * This is a **delicate** API. It is easy to accidentally create resource or memory leaks `allocated` is used.
   * A `Resource` allocated by `allocated` is not subject to the guarantees that [Resource] makes,
   * instead the caller is responsible of correctly invoking the `release` handler at the appropriate time.
   * This API is useful for building inter-op APIs between [Resource] and non-suspending code, such as Java libraries.
   */
```

While write this doc, we should probably expose an smart-constructor for `val exit = if(e is CancellationException) ExitCase.Cancelled(e) else ExitCase.Failure(e)`. Do you mind adding it in this PR? We're also using it several times inside this PR. Something like the following function heavily cleans up this user-example above, and it would also clean-up several instances in our internal Resource code I think.

```kotlin
/** Construct an ExitCase from a [Throwable] */
public fun ExitCase(error: Throwable): ExitCase =
  if (error is CancellationException) ExitCase.Cancelled(error) else ExitCase.Failure(error)
```

nomisRev (commented):
Couple of nits, but in general looks great! Very nice work @custommonkey for your first-contribution 🙌 🥳

nomisRev:
I see CI failed over the Binary API Validator. For some reason the CI task to automatically update this doesn't work for forks :/
To fix this you should run `./gradlew apiDump` and push the changes ;)

custommonkey:
Thanks for the great feedback. 

The use case I have in mind is kotest integration, we've been using Resource for testing in scala for a couple of years, so very nice to see you've already been looking at this space.

nomisRev:
Btw @custommonkey, if you have experience with `Resource` in Scala you might find this PR interesting. https://github.com/arrow-kt/arrow/pull/2786

If you have any feedback I'd love to hear it 🙏 Feel free to comment on the closed PR, here or create a ticket requesting functionality for Arrow 2.0. A lot of Arrow users/maintainers also hang around in the KotlinLang #arrow channel, if you're looking for a place to chat about Arrow ;)

arrow-libs/fx/arrow-fx-coroutines/src/commonTest/kotlin/arrow/fx/coroutines/ResourceTest.kt
custommonkey: I've updated the tests to call release with each of the three exist cases,  although I'm not sure that I'm really testing the error path effectively.

nomisRev (approved):
Looks great to me!
This could even be used to write a `@JUnitRule` for Arrow's Resource 🥳

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
```diff
@@ -13,6 +13,11 @@ public sealed class ExitCase {
 
   public data class Cancelled(val exception: CancellationException) : ExitCase()
   public data class Failure(val failure: Throwable) : ExitCase()
+
+  public companion object {
+    public fun ExitCase(error: Throwable): ExitCase =
```
i-walker, line 18: I was wondering if we could name this `fromThrowable`, so it is aligned with the naming conventions in other parts of Arrow?

i-walker (commented):
A small nit other than that it looks really great thanks @custommonkey !

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
nomisRev: I was wondering about this too. Do we currently have `fromThrowable` anywhere? Or is the in analog to `Option.fromNullable`?

I'm not sure if that's a great name either, but I cannot think of a similar construct in the Kotlin Std.

i-walker: > I'm not sure if that's a great name either, but I cannot think of a similar construct in the Kotlin Std.
I meant to edit the Kotlin Std lib part it away 😅

Yes it would be analog to `Option.fromNullable`, but maybe the top-level `exitCase` would suffice wdyt @nomisRev ?

nomisRev: I.e.
 - [`Result`](https://github.com/JetBrains/kotlin/blob/ea836fd46a1fef07d77c96f9d7e8d7807f793453/libraries/stdlib/src/kotlin/util/Result.kt#L92) is `success` and `failure` on the `Companion`.
 - [Duration.kt](https://github.com/JetBrains/kotlin/blob/ea836fd46a1fef07d77c96f9d7e8d7807f793453/libraries/stdlib/src/kotlin/time/Duration.kt#L113) has `milliseconds`, `seconds` as extensions on `Int` inside the `Companion` 🤯 
 - [Random](https://github.com/JetBrains/kotlin/blob/ea836fd46a1fef07d77c96f9d7e8d7807f793453/libraries/stdlib/src/kotlin/random/Random.kt#L318) top-level constructors
 
 Don't think we have access to `@InlineOnly` though.
 
 Some alternatives:
  - (Companion, Result) `fun throwable(error: Throwable): ExitCase`
  - (Companion, Duration) `fun Throwable.toExitCase(): ExitCase`
  - (top-level, Random) `fun ExitCase(error: Throwable): ExitCase`

PS: @i-walker I think we use `fromNullable` because we need to explicitly distinct between the constructor that allows `Option<A?>` and `Option<A>`, otherwise you can never have `Some(null)` which you want in generic cases.

i-walker: >  I think we use fromNullable because we need to explicitly distinct between ...

Gotcha !

And the alternatives look good to me :D

custommonkey: Not quite sure where we are with this. I'm happy to make changes

nomisRev: Sorry, @custommonkey.

We were wondering if there is a convention for such APIs in Kotlin, since we've not established a clear convention within Arrow. Looking at the Kotlin Std APIs it doesn't _seem_ to be consistent within the Std either..

TBH I like your proposal because it works well in this case.

`Option` can follow the same guidelines where we can place a `fun Option(...)` constructor on the companion to replace `fromNullable`, and we can keep `null.toOption()` as well.
This would then also belong on the `Companion` similar to how it's encoded for `Duration.kt`.

WDYT @i-walker?
If Imran agrees, there is no action required from you @custommonkey and we'll go ahead with merging the PR as is.

i-walker: Ok, let's keep it how it is. Maybe down the line we could draft a standard for 2.0, but this has low priority in comparison to other efforts in Arrow

i-walker (approved):
Thanks @custommonkey for your contribution !! 🙌🏾

PR #2822, opened 2022-09-14
Make the server disconnect test more general

lukasz-kalnik-gcx:
In the `NetworkEitherCallAdapterTest`, it turns out on some host OS/JVM combinations a more general exception (SocketException) is thrown when the server disconnects.
Original problem report: https://kotlinlang.slack.com/archives/C5UPMM0A0/p1663048397387609

As the SocketException is parent of ConnectException, it should be enough to make the test a bit more general. Anyway the main point of the test is that the error is wrapped into a `Left<IoError>`.


nomisRev (approved):

serras (approved):
LGTM

nomisRev:
Thanks @lukasz-kalnik-gcx 🙏

Issue #2807, opened 2022-08-23
NonEmptyList breaks List equality contract


wfhartford:
Implementations of List generally have a equals function which allows for different list implementations to be considered equal if their contents are equal. This does not appear to be the case with the `NonEmptyList.equals()` function. Is this intentional or a bug?

Of the following two test cases, the first fails while the second succeeds:
```kotlin
  test("nel equality 1") {
    ("thing".nel() == listOf("thing")) shouldBe true
  }
  test("nel equality 2") {
    (listOf("thing") == "thing".nel()) shouldBe true
  }
```

The current implementation of `NonEmptyList.equals()` clearly requires that the other object have the same runtime class:
```kotlin
  override fun equals(other: Any?): Boolean {
    if (this === other) return true
    other?.let {
      if (it::class != this::class) return false
      (other as NonEmptyList<*>)
      if (all != other.all) return false
      return true
    } ?: return false
  }
```

PR #2825, opened 2022-09-23
Use `super` equals and hashCode overrides for NonEmptyList

RusticFlare:
Fixes https://github.com/arrow-kt/arrow/issues/2807 by removing the overrides of `equals` and `hashCode` in `NonEmptyList`, so that we use the implementations in `AbstractList`.

One thing to note is that Kotest's `shouldBe` doesn't use the `equals` function for `Iterable`, so I've tested the result of `==` explicitly (and left a comment to that effect).

(This is my first PR here 👋 so let me know if I've missed anything)

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
@@ -205,18 +205,11 @@ public class NonEmptyList<out A>(
   public fun extract(): A =
     this.head
 
-  override fun equals(other: Any?): Boolean {
-    if (this === other) return true
-    other?.let {
-      if (it::class != this::class) return false
-      (other as NonEmptyList<*>)
-      if (all != other.all) return false
-      return true
-    } ?: return false
-  }
+  override fun equals(other: Any?): Boolean =
+    super.equals(other)
 
   override fun hashCode(): Int =
-    all.hashCode()
+    super.hashCode()
 
   override fun toString(): String =
     "NonEmptyList(${all.joinToString()})"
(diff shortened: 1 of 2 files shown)


nomisRev:
Hey @RusticFlare,

This change looks a bit dangerous to me, since it breaks `shouldBe` from Kotest. I know many people rely on `shouldBe` on Kotest.

I am wondering if the fix for that ticket is rather not simply removing `if (it::class != this::class)` from the `equals` implementation. That way it would solve the issue from the ticket, but wouldn't break `shouldBe` if not mistaken.

nomisRev:
Or is `shouldBe` incorrect for this kind of comparisons 🤔

RusticFlare:
Hi, it doesn't break `shouldBe`. I just couldn't use `shouldBe` to write a failing test because `shouldBe` doesn't use `equals` on `Iterable`. The current behaviour is an obvious bug and I don't think there is any value in overriding the `equals` and `hashCode` on `AbstractList`.

nomisRev:
Okay, thanks for clarifying @RusticFlare! I miss-understood.

We cannot break the binary, so instead of removing I think we should manually delegate to the super class for now. We can break the binary in `2.x.x`. Then CI should turn green, it's now turning red on a binary breaking change.

> API check failed for project arrow-core.

RusticFlare:
@nomisRev I've made the suggested change and hopefully the comment about `shouldBe` is clearer now

RusticFlare:
Looks like CI is green 🎉 I can make another PR to remove the methods for `2.x.x` once this is merged.

nomisRev (approved):
Thank you for your first contribution @RusticFlare 🙏:clap::tada: Sorry for the hold-up.

i-walker (approved):

PR #2833, opened 2022-10-04
Improve Either.getOrHandle() docs

lukasz-kalnik-gcx:
This fixes the `getOrHandle()` doc - there was incorrect information that the function somehow returns `Right`, whereas it always unwraps the value.


nomisRev:
Hey @lukasz-kalnik-gcx,
Thanks for the fix 🙏

I think you're missing `./gradlew knit` to [generate and validate the documentation](https://github.com/arrow-kt/arrow/blob/main/CONTRIBUTING.md#how-to-generate-and-validate-the-documentation).

lukasz-kalnik-gcx:
Thanks @nomisRev , I just restored the original code example instead (should build now).

raulraja (approved):
thanks @lukasz-kalnik-gcx

nomisRev (approved):

Issue #2810, opened 2022-08-31
Returning `Nothing` from shift instead of being polymorphic


nomisRev:
https://github.com/arrow-kt/arrow/pull/2797#discussion_r958247534

Also discussed on Slack: https://kotlinlang.slack.com/archives/C5UPMM0A0/p1661181942207909

serras:
To me `Nothing` seems like the right type, to make them more like [exceptions](https://kotlinlang.org/docs/exceptions.html#the-nothing-type). In particular, the docs make this point, which I think is very relevant:

> When you call this function, the compiler will know that the execution doesn't continue beyond the call.

serras:
Closed by #2839

Issue #2805, opened 2022-08-23
Proposal to change return type of `shift` to `Nothing`


roomscape:
The `shift` function in `EffectScope` and `EagerEffectScope` is currently declared as 
```kotlin
suspend fun <B> shift(r: R): B
```

Instead it could be declared as
```kotlin
suspend fun shift(r: R): Nothing
```

[Initially proposed](https://kotlinlang.slack.com/archives/C5UPMM0A0/p1661181942207909) in the #kotlinlang Slack.

# Benefits

Using `Nothing` would:
- Take advantage of the type system to convey information about the behaviour of the function
- Allow the IDE to warn about unreachable code after a call to `shift`
- Allow smart-casting after a call to `shift`
- Remove the need to specify a type argument in some cases where omitting it would currently be a compile error

Minimal example that does not compile before the change, but would compile after the change:

```kotlin
val result: Either<String, Int> = either.eager {
    shift("Not an int") // Not enough information to infer type variable B
    1
}
```

Example of a missed opportunity for a smart cast:

```kotlin
val result: Either<String, Int> = either.eager {
    val x: String? = null
    if (x == null) shift<Any>("Not an int")
    x.toInt() // Only safe (?.) or non-null asserted (!!.) calls are allowed on a nullable receiver of type String?
}
```

# Comparison with standard library

`Nothing` is used in a few places in standard Kotlin functions, notably:
- [error](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/error.html): `fun error(message: Any): Nothing`
- [TODO](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-t-o-d-o.html): `fun TODO(): Nothing`

There are also examples of bottom/empty types that don't use `Nothing`, e.g:
- [emptyList](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/empty-list.html) `fun <T> emptyList(): List<T>`

The semantics of `shift` make it more similar to functions like `error`, so it would be idiomatic to use `Nothing` here.

# Compatibility

- ✅ `Nothing` is assignable to any type, so any code that uses the return value from `shift` would continue to compile.
- ℹ️ A new warning would be introduced because code after a call to `shift` would be identified as unreachable. This could be a breaking change for anyone using the `-Werror` compiler flag.
- ❌ **Breaking change**: calling `shift` with an explicit generic type, e.g. `shift<Nothing>()`, would become a compile error when it wasn't before.

[...]

PR #2839, opened 2022-10-10
Change return type of raise to Nothing

roomscape:
Implements #2805 / #2810


arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/continuations/EffectSpec.kt
```diff
@@ -302,7 +302,7 @@ class EffectSpec :
     "catch - error path and recover" {
       checkAll(Arb.int().suspend(), Arb.string().suspend()) { int, fallback ->
         effect<Int, String> {
-          raise<String>(int())
+          raise(int())
           fail("It should never reach this point")
```
roomscape, line 306: This line now warns about unreachable code. Same in the other similar tests. Is this line still useful, or should it be removed?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
```diff
@@ -355,7 +355,7 @@ import kotlin.jvm.JvmMultifileClass
  *    val error = "Error"
  *    val exit = CompletableDeferred<ExitCase>()
  *   effect<String, Int> {
- *     parZip({ awaitExitCase<Int>(exit) }, { raise<Int>(error) }) { a, b -> a + b }
+ *     parZip({ awaitExitCase<Int>(exit) }, { raise(error) }) { a: Int, b: Int -> a + b }
```
roomscape, line 358: The explicit lambda signature is needed here because otherwise the type of `b` is unconstrained and the `+` operator function can't be resolved.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/continuations/EffectSpec.kt
nomisRev: Yes, that test is still useful. It verifies that passed `raise` no code will be executed, which is also confirmed by the IDEA inspection now 🥳

So, I think we should suppress this warning if possible. (Not sure what `@Supress` is needed for this).

nomisRev (approved):
Looks great to me! Thank you @roomscape for this PR 🙏
And congrats on really awesome first contribution 🥳 🎉 🙌

Making the suggestion, and making the fix in a PR ❤️

nomisRev:
@roomscape this changed the binary a bit, so you'll need to run `./gradlew apiDump`

raulraja (approved):
thanks @roomscape !

i-walker (approved):

PR #2847, opened 2022-11-16
Resource API deprecation, and preparation for 2.x.x. & back port improvements

nomisRev:
Propose the deprecate almost the entirety of the Resource API, as discussed in the [2.x.x PR](https://github.com/arrow-kt/arrow/pull/2786).

# Deprecated APIs

- zip: composing resource using resource { f(bind(), bind()) } supports arity-n
- parZip: can be inherited from Arrow Fx Coroutines methods, and naturally compose inside the DSL.

# flatMap & co

For `Either` we've decided to keep these methods, should we keep them for `Resource` as well?

- map/tap/flatMap: this methods can be replaced entirely by the DSL, should we keep

```kotlin
resA.flatMap {
  resB
}

resource {
  resA.bind()
  resB.bind()
}

resA.map(f)
resource { f(resA.bind()) }

resA.tap(f)
resource { resA.bind().also(f) }
```

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -25,6 +25,9 @@ jobs:
           distribution: 'zulu'
           java-version: 11
 
+      - name: Install Watch
+        run: xcrun simctl create "Apple Watch Series 5 - 44mm" "Apple Watch Series 5 (44mm)"
+
       - name: build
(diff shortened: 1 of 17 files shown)


i-walker (approved):
Thanks 🙌🏾 @nomisRev

serras (approved):

Issue #2799, opened 2022-08-17
Can not compose Traversal with Getter


feser:
Documentation for Traversal mentions that a `Traversal` can be composed with `Getter` which results in `Fold`.
https://arrow-kt.io/docs/optics/traversal/#composition

However, only following compose functions are available for `Traversal` which don't include `Getter`.
https://github.com/arrow-kt/arrow/blob/5d0bc9e214f9b92f8492a8202a7c165d5e81fa9e/arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Traversal.kt#L47
https://github.com/arrow-kt/arrow/blob/5d0bc9e214f9b92f8492a8202a7c165d5e81fa9e/arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Setter.kt#L56

This is required to get nested attribute without modifying it.
For example, getting `employeeStreetName` in Traversal usage example.
https://arrow-kt.io/docs/optics/traversal/#usage

Thanks in advance!

serras:
It seems that the documentation is the culprit here. A `Traversal` in Arrow Optics is defined as something which modifies 0 to N elements, so it makes no sense to compose it with a `Getter`.

Just to double check, @raulraja @nomisRev, is this right? If so I'll open a PR to update the docs

nomisRev:
Yes, I think you're right @serras.
`Getter` cannot compose with `Traversal` because it cannot satisfy the `modify` behavior of `Traversal.

serras:
The new hierarchy in Arrow 2.0 (merged in #2873) is greatly simplified. This problem won't happen then. Closing this issue as a result.

PR #2851, opened 2022-11-18
Use the more common optics hierarchy also in Arrow

serras:
The issue #2799 revealed that what we've called `Traversal` in Arrow Optics doesn't match what other communities call `Traversal`. In particular, here `Traversal` was nothing else than a `Setter`, whereas commonly a `Traversal` is the combination of `Fold` + `Setter` (so you can focus and modify 0 .. N elements). The documentation still shows this was the intention, as mentioned in #2799.

This PR changes the hierarchy to be similar to Haskell's or Monocle's. For comparison, here's the one used in Haskell's `optics` and `microlens` (note that in Haskell `AffineFold` is what we call an `OptionalGetter`, and `AffineTraversal` what we call an `Optional`; We also have no `Review`, `Prism` directly introduces `reverseGet`.)
hierarchy optics - page 1
One nice result of this change is that now every optics type is the child of at most two parents, as opposed to the previous implementation where sometimes we had 5 superclasses in the declaration.

This is **not** backwards-compatible, for that reason I'm targeting Arrow 2.0. Regardless of that, I think that this is the right thing to do to align ourselves with the rest of FP communities.

This PR also includes a few changes coming from the new hierarchy:
- Variance annotations have been included. For example, `Fold<in S, out A>`, since you "take `S`" and "get back `A`s".
- A new `PGetter` has been introduced, to align with the rest of the hierarchy.
- Before `Every` was a combination of `Fold` and `Traversal`. This optic doesn't make sense anymore, but the companion object does. So I've changed it to be simply an `object`.


serras:
Closed in favor of #2873

PR #2852, opened 2022-11-18
Update various versions

serras:
It seems that Renovate is not working so well, so I've updated a few versions.

~~Importantly, the Kotlin version is still at 1.6.21, which means that KSP and kotlin-compile-testing also stay in the corresponding versions.~~ This PR also updates Kotlin, `kotlin-compile-testing`, Dokka, and KSP to the 1.7.21 series.


build.gradle.kts
```diff
@@ -86,3 +86,13 @@ apiValidation {
 
   ignoredProjects.addAll(ignoreApiValidation)
 }
+
+rootProject.plugins.withType<org.jetbrains.kotlin.gradle.targets.js.yarn.YarnPlugin> {
```
nomisRev, line 90: 🙏

nomisRev:
Yes, not sure what is up with Renovate. I would like to make a release this week with 1.7.21 since soon 1.8.0 will also be released. I think it would be best if we can still release 1.7.x before we move to 1.8.x. Right?

We've also made some fixes for Optics KSP that people rely on through stable alpha versions.

serras:
If you want, I can update Kotlin (and associated libraries) to 1.7.21 as part of this PR.

In fact, the Windows build is failing, and the only solution I've come around in Internet are "first update to newer Kotlin".

serras:
@nomisRev I've just pushed an update to 1.7.21.

I seems that [freezing](https://kotlinlang.org/docs/native-migration-guide.html#update-your-code) is deprecated, and I think it's used in the `atomic` module.

JavierSegoviaCordoba:
build is passing locally?

serras:
Nope.

I think I need to revert the changes for the yarn part, because we’re getting problems there. I’ll do so in a couple of hours.

build.gradle.kts
serras: Unfortunately it seems that blindly updating the versions breaks the build 😞 So I reverted it, I guess we'll have to live with Dependabot alerts...

serras:
I don't know why the Action is getting againa lot of "the daemon was unexpectedly killed" errors :/

serras:
The other problem is that the `arrow-fx-stm` tests fail in native targets, they just keep running indefinitely.

arrow-libs/core/arrow-continuations/src/commonTest/kotlin/generic/TestSuite.kt
```diff
@@ -4,97 +4,21 @@ import arrow.continuations.Reset
...
     "short circuit" {
-      runScope<Either<String, Int>> {
+      Reset.restricted<Either<String, Int>> {
         val no: Int = shift { Left("No thank you") }
         throw IllegalStateException("This should not be executed")
       } shouldBe Left("No thank you")
     }
-
-    // some examples from http://homes.sice.indiana.edu/ccshan/recur/recur.pdf
-    if (capabilities().contains(ScopeCapabilities.MultiShot)) {
```
nomisRev, line 31: The compile error in `arrow-continuations` was coming from here @serras.

Not entirely sure why, but perhaps a regression bug in the compiler due to `abstract class ContTestSuite` being combined with `abstract class UnitSpec` in combination with the `abstract fun capabilities()`. Might be interesting to share with Kotlin team(?).

As you can see `SingleShotContTestSuite` was the only implementer of `ContTestSuite`, and it was returning `emptySet` for `capabilities()`. So these tests were not being run, and they never ran. We left the code for reasons that don't matter anymore, so removing them is non-impactful and safe.

nomisRev: I found this by opening the `js` file that was linked in the stack trace at runtime, `/var/folders/qs/9sb020z91_z1_50v18bnmf0000gn/T/_karma_webpack_466164/common.js:5773`. There I saw it was calling `ContTestSuite` and that led me to the `capabilities` function.

serras: I'm amazed you found it. Thanks very much! We're just missing the STM problem then...

build.gradle.kts
nomisRev: I guess it was not related to this. Resolving the comment

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/OS.kt
```diff
@@ -0,0 +1,8 @@
+package arrow.fx.coroutines
+
+public expect object OS {
```
hoc081098, line 3: How about putting this class in test sourceSet

serras: I tried that, but `nativeTest` does not work as `sourceSet` currently. So I would need to create 20x `test` source sets, one per platform

arrow-libs/fx/arrow-fx-coroutines/src/commonTest/kotlin/arrow/fx/coroutines/ResourceTest.kt
```diff
@@ -177,23 +183,26 @@ class ResourceTest : ArrowFxSpec(
       return Pair(promises.map { it.second }, res)
     }
 
-    "parZip - deep finalizers are called when final one blows" {
+    "parZip - deep finalizers are called when final one blows".config(enabled = !OS.isApple) {
```
nomisRev, line 186: Are these still failing? They should pass for `iOS` :(

.github/workflows/pull_request.yml
```diff
@@ -25,23 +26,55 @@ jobs:
...
+      - name: JS tests (Ubuntu)
+        uses: gradle/gradle-build-action@v2
+        if: matrix.os == 'ubuntu-latest'
+        with:
+          arguments: --full-stacktrace jsTest
+      - name: Linux Native tests (Ubuntu)
+        uses: gradle/gradle-build-action@v2
+        if: matrix.os == 'ubuntu-latest'
+        with:
+          arguments: --full-stacktrace linuxX64Test
```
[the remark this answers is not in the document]
serras: That makes sense, let me change those

arrow-libs/core/arrow-continuations/build.gradle.kts
```diff
@@ -50,3 +50,12 @@ kotlin {
...
 }
+
+tasks["jsBrowserTest"].enabled = false
+tasks["jsNodeTest"].enabled = false
+tasks["mingwX64Test"].enabled = false
+tasks["linuxX64Test"].enabled = false
+tasks["iosX64Test"].enabled = false
+tasks["tvosX64Test"].enabled = false
+tasks["watchosX64Test"].enabled = false
+tasks["macosX64Test"].enabled = false
```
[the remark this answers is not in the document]
serras: I can try to remove those, but they kept failing in the latest commits

arrow-libs/fx/arrow-fx-coroutines/src/commonTest/kotlin/arrow/fx/coroutines/ResourceTest.kt
serras: Not in my machine, not in GitHub's :(

arrow-libs/core/arrow-continuations/build.gradle.kts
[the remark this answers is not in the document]
serras: It works now :)

arrow-libs/fx/arrow-fx-coroutines/src/commonTest/kotlin/arrow/fx/coroutines/ResourceTest.kt
serras: Unfortunately I think this would need to stay. Anyway we're dropping them for Arrow 2.0...

.github/workflows/pull_request.yml
```diff
@@ -25,23 +26,55 @@ jobs:
...
+          arguments: --full-stacktrace compileJava compileKotlin
 
-      - name: build
+      - name: Compile test files
         uses: gradle/gradle-build-action@v2
-        if: matrix.os != 'windows-latest'
+        if: matrix.os == 'ubuntu-latest'
         with:
-          arguments: --full-stacktrace build
+          arguments: --full-stacktrace compileTestJava compileTestKotlin
```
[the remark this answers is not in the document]
serras: The idea is to somehow work around the problems with memory that Gradle has by making the build in several steps.

PR #2853, opened 2022-11-25
Property delegation for bind

serras:
This is a proposal for an alternative syntax for `bind` using [delegated properties](https://kotlinlang.org/docs/delegated-properties.html#property-delegate-requirements). The idea is that instead of writing:

```kotlin
either {
  val x = oneThing().bind()
  val y = anotherThing().bind()
  x + y
}
```

you can use `by` and remove the `bind`

```kotlin
either {
  val x by oneThing()
  val y by anotherThing()
  x + y
}
```

The goal here is to provide a more intuitive syntax for beginners, since `bind` is not a common word outside of Scala/Haskell circles. The idea would be to say that if you want to "execute" an `Either`, `Result`, ... inside the block you can "bridge" it with `by`.

### Some comments about the implementation

The way Kotlin translates these delegated properties is in two steps. Assuming `val x by t`
1. It calls `provideDelegate` over the type of `t`, which returns something of a new type `D`. In our case, `provideDelegate` is exactly `bind`, so the type `D` is the type `A` "contained" in the `Either`, `Result`, ...
2. It calls `getValue` over the type `D`. In our case this is a no-op, since `provideDelegate` has done all the work.

It's also allowed to *not* give a `provideDelegate`. But one should be aware that `getValue` is called every time we access the variable. So if we made `getValue` the one calling `bind`, code like

```kotlin
val x by thing()
x + x
```

would actually be equivalent to

```kotlin
val t = thing()
t.bind() + t.bind()
```

whereas we only want one `bind`!


serras:
Makes sense. I'm closing this since the decision was already made.

PR #2854, opened 2022-11-27
Add CountDownLatch

nomisRev:
There was a discussion about adding some new types to Arrow Fx, for its evolution towards 2.0.
This PR adds a `CountDownLatch` type, a KMP & suspend version of `java.util.concurent.CountDownLatch`.

Originally written [here](https://github.com/nomisRev/arrow-fx-coroutines-utils/blob/main/src/commonMain/kotlin/io/github/nomisrev/CountDownLatch.kt), but I removed the `interface` to simplify the code a bit.

PR #2857, opened 2022-12-04
Add CyclicBarrier

nomisRev:
Similar to CountDownLatch added in #2854 this PR introduces [CyclicBarrier ](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/CyclicBarrier.html).

Originally written [here](https://github.com/nomisRev/arrow-fx-coroutines-utils/blob/main/src/commonMain/kotlin/io/github/nomisrev/CyclicBarrier.kt), but I removed the interface to simplify the code a bit.

The goal for 2.x.x of Arrow Fx Coroutines is to offer a _core_ module of concurrent data types, and operations in Kotlin.


serras (approved):
LGTM. Nice to see that these "bare Kotlin" implementations are not that complicated :)

i-walker (approved):
🙌 Thanks !

PR #2858, opened 2022-12-04
Temp rollback error handling deprecations until 1.2.x

nomisRev:
In #2830 I proposed a bunch of deprecations, and back-ported the error handling APIs designed in the 2.x.x branch. To back -port the PR introduced some machinery that was _source compatible_ with 2.x.x.

Since we're planning a _patch_ release this week, and a _minor_ released after with the 2.x.x `Raise` DSL back-ported to 1.2.x. This will be done so users will have a long grace period to slowly migrate to _Raise_.

Therefore I propose to put off deprecation the error handlers until we have the `Raise` DSL available in 1.2.x, this way users will be able to migrate to the long-term solution straight away. Without introducing temporary machinery, and unnecessary breakage.


i-walker (approved):

serras:
How would this PR be merged into `arrow-2`?

nomisRev:
@serras These changes were introduced after `1.1.3`, and are unrelated to `2.x.x`. It was an attempt to back port `recover` & `catch` signature in a _source-compatible_ way.

The back-port of the `Raise` DSL will re-introduce these signatures for `1.2.x`, and they will supersede everything that is currently in the 2.x.x branch. To back-port for `1.2.x` I propose, as discussed, to introduce the `Raise` DSL into `arrow.core.raise.*` and then when merging this into `2.x.x` we simply remove the `arrow.core.computations.*` package and replace it by `arrow.core.raise`.

The `recover` & `catch` signatures for `Either` can then directly rely on `arrow.core.raise.Raise`.

PR #2872, opened 2022-12-13
'accumulate' for Raise

serras:
This PR proposed a one-for-all accumulating map for `Raise`. The signature is not the greatest at this moment, because `Raise` needs to appear in receiver position, so the `list` goes as first argument.

```kotlin
public fun <R, A, B> Raise<NonEmptyList<R>>.accumulate(
  list: Iterable<A>,
  block: Raise<R>.(A) -> B
): List<B>
```

Interestingly, this function is actually generic over the way we run the `Raise`.

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Raise.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Raise.kt
@@ -3,6 +3,7 @@
 package arrow.core.continuations
 
 import arrow.core.Either
+import arrow.core.NonEmptyList
 import arrow.core.None
 import arrow.core.Option
 import arrow.core.Some
@@ -246,4 +247,24 @@ public inline fun <R> Raise<R>.ensure(condition: Boolean, raise: () -> R): Unit 
 public inline fun <R, B : Any> Raise<R>.ensureNotNull(value: B?, raise: () -> R): B {
   contract { returns() implies (value != null) }
   return value ?: raise(raise())
+}
+
+/**
+ * Accumulate the errors obtained by executing the [block]
+ * over every element of [list].
+ */
+@EffectDSL
+public fun <R, A, B> Raise<NonEmptyList<R>>.accumulate(
+  list: Iterable<A>,
+  block: Raise<R>.(A) -> B
+): List<B> {
+  val errors = mutableListOf<R>()
+  val results = mutableListOf<B>()
+  list.forEach {
+    fold<R, B, Unit>({ block(it) }, { errors.add(it) }, { results.add(it) })
+  }
+  if (errors.isNotEmpty())
+    raise(NonEmptyList.fromListUnsafe(errors))
+  else
+    return results
 }
--- a/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/continuations/EffectSpec.kt
+++ b/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/continuations/EffectSpec.kt
@@ -1,6 +1,7 @@
 package arrow.core.continuations
 
 import arrow.core.Either
+import arrow.core.NonEmptyList
 import arrow.core.identity
 import arrow.core.left
 import arrow.core.right
@@ -14,6 +15,7 @@ import io.kotest.property.arbitrary.arbitrary
 import io.kotest.property.arbitrary.boolean
 import io.kotest.property.arbitrary.flatMap
 import io.kotest.property.arbitrary.int
+import io.kotest.property.arbitrary.list
 import io.kotest.property.arbitrary.long
 import io.kotest.property.arbitrary.orNull
 import io.kotest.property.arbitrary.string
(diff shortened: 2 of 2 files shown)


raulraja (approved):
Thanks @serras, looks great !

nomisRev:
That signature makes me sad 😞 I'm personally not really in favour of adding this method until we have context receivers. I think the pattern of working over `Raise<E>` will not really catch on they become stable.

You can of course call `bind` here inside of `block`, but then this seems like a micro-optimisation compared to the current signature.

Working `E` doesn't solve any problems, since you typically want `Nel<E>`. The signature below suffers from a second problem which is that it's not `inline`, which is not possible inside interfaces. So it doesn't allow foreign _suspend_ to pass through, which is a must in this model of "transformers". To solve this it would also need to be put outside of the `interface` which makes this suffer from the same issue.

```kotlin
interface Raise<E> {
  fun <A, B> Iterable<A>.mapOrAccumulate(combine: (E,E) -> E, transform: Raise<E>.(A) -> B): List<B>
}
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Raise.kt
```diff
@@ -247,3 +248,23 @@ public inline fun <R, B : Any> Raise<R>.ensureNotNull(value: B?, raise: () -> R)
   contract { returns() implies (value != null) }
   return value ?: raise(raise())
 }
+
+/**
+ * Accumulate the errors obtained by executing the [block]
+ * over every element of [list].
+ */
+@EffectDSL
+public fun <R, A, B> Raise<NonEmptyList<R>>.accumulate(
```
nomisRev, line 257:
```suggestion
public inline fun <R, A, B> Raise<NonEmptyList<R>>.accumulate(
```

serras:
@nomisRev I agree that context receivers will make this much nicer. However, I don't think that the following looks so terrible,

```kotlin
accumulate(elements) {
  checkValidity(it)
}
```

when compared to

```kotlin
elements.accumulate {
  checkValidity(it)
}
```

nomisRev:
Alright, that's fair. Is `accumulate` the correct name for this operator? It still has the shape of `map`, right?

I think we should add the other signature inside `Raise` already since that fits.

```kotlin
interface Raise<in E> {
  fun <A, B> Iterable<A>.accumulate(
    semigroup: Semigroup<E>,
    transform: Raise<E>.(A) -> B
  ): List<B>
}
```

serras:
@nomisRev I updated the code to use the more general version you mention; in fact I've added an even more general one because to implement the version with `NonEmptyList<E>` you may want a way to "embed" a single `E` into the non-empty list.

i-walker:
Is it possible to instead use a combination of mapLeft and mapAccumulate without the embed parameter. This way the method isnt too convoluted @serras? Unless that is what we were going for

serras:
@i-walker indeed, I think you're right. I've pushed a new commit in which `mapOrAccumulate` within `Raise` is simpler, and I use a combination with `mapLeft` to implement the `NonEmptyList` version

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Raise.kt
```diff
@@ -62,6 +64,47 @@ public interface Raise<in R> {
...
+   * over every element of [this] using the given [semigroup].
+   * The error type [E] of the [block] may be different from
+   * the error type [R] of the surrounding computation.
+   */
+  public fun <E, A, B> Iterable<A>.mapOrAccumulate(
+    embed: (E) -> R,
+    semigroup: Semigroup<@UnsafeVariance R>,
+    block: Raise<E>.(A) -> B
+  ): List<B> {
+    var error: R? = null
```
nomisRev, line 76: Using `null` here is unsafe due to the nested null problem.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Raise.kt
```diff
@@ -211,6 +242,16 @@ public interface Raise<in R> {
   ): A = fold({ catch(it) }, { raise(it) }, { it })
 }
 
+/**
+ * Execute an [action], and map every error of type [E]
+ * into one of type [R].
+ */
+@EffectDSL
+public inline fun <R, E, A> Raise<R>.mapLeft(
```
i-walker, line 250: since `mapLeft` is just recover can we remove this? and inline it into `mapOrAccumulate`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Raise.kt
serras: I didn't know about `EmptyValue`, this has been fixed now.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Raise.kt
serras: Done! I've kept `mapLeft` in the code, because it might be useful elsewhere.

i-walker: The method is still there can we add this in a separate PR :)

serras: Indeed, I've just removed it to make this PR self-contained

nomisRev (approved):

PR #2873, opened 2022-12-13
Simplify optics to Traversal/Optional/Lens/Prism

serras:
**Note**: this PR superseedes #2851 

Following the discussion at https://github.com/arrow-kt/arrow/discussions/2859, here's an even more simplified version of the optics hierarchy, consisting only of `Traversal`, `Optional`, `Lens`, and `Prism`.

In addition, nullable field no longer generate `Optional<A, S>`, but rather `Lens<A, S?>` in KSP. However, the developer can add `.notNull` to get the old behavior back (`notNull: Lens<A, S?> -> Optional<A, S>`); this matches the explicitness of `every`, and doesn't confuse users by losing the ability to set that field to `null`.

There's also an effort of trimming down the amount of methods exposed in the DSL. For example, before we had:

- `Prism<A, S>.field: Optional<S, B>`
- `Optional<A, S>.field: Optional<S, B>`

but those can be trimmed down to a single one, because `Prism` is a subclass of `Optional`. This is a huge gain for the KSP generation, because it goes from 8 additional methods being generated to only 3!


serras:
Note that this is **not** backwards-compatible, since users who relied on KSP generating an `Optional` for nullable fields need to update their code.

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/dsl/index.kt
```diff
@@ -10,38 +10,28 @@ import arrow.optics.typeclasses.Index
...
-/**
- * DSL to compose [Index] with a [Prism] for a structure [S] to focus in on [A] at given index [I]
- *
- * @receiver [Prism] with a focus in [S]
- * @param ID [Index] instance to provide a [Optional] to focus into [S] at [I]
- * @param i index [I] to focus into [S] and find focus [A]
- * @return [Optional] with a focus in [A] at given index [I]
- */
-public fun <T, S, I, A> Prism<T, S>.index(ID: Index<S, I, A>, i: I): Optional<T, A> = this.compose(ID.index(i))
+public fun <T, S, I, A> Lens<T, S>.index(idx: Index<S, I, A>, i: I): Optional<T, A> = this.compose(idx.index(i))
```
nomisRev, line 17: A `Lens` will be an `Optional`, right? So can we omit this syntax like we do for `nullable`?

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/dsl/at.kt
```diff
@@ -10,38 +10,28 @@ import arrow.optics.typeclasses.At
...
-/**
- * DSL to compose [At] with a [Prism] for a structure [S] to focus in on [A] at given index [I].
- *
- * @receiver [Prism] with a focus in [S]
- * @param AT [At] instance to provide a [Lens] to zoom into [S] at [I]
- * @param i index [I] to zoom into [S] and find focus [A]
- * @return [Optional] with a focus in [A] at given index [I].
- */
-public fun <T, S, I, A> Prism<T, S>.at(AT: At<S, I, A>, i: I): Optional<T, A> = this.compose(AT.at(i))
+public fun <T, S, I, A> Lens<T, S>.at(at: At<S, I, A>, i: I): Lens<T, A> = this.compose(at.at(i))
```
nomisRev, line 17: A `Lens` will be an `Optional`, right? So can we omit this syntax like we do for `nullable`?

nomisRev (approved):
Looks great @serras !! 👏 👏 Awesome improvements!

nomisRev:
Unrelated: should we start adding issues on Github for the low-coverage files? Those seem like good first issues, now that we have the coverage reporting enabled it seems like a good step to make.

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/dsl/index.kt
serras: Done in the new commit

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/dsl/at.kt
serras: Not really, because the result for `at.at(i)` is a `Lens`. So if we remove this overload we would be losing a bit of information.

raulraja (approved):

PR #2877, opened 2022-12-29
Prisms for Either

### LINKED ISSUE #2865: ["Request"] Add extension to look into Either instances in Optics

1.1.3

What would you like to see?

There is no out of the box way to focus into an Either in Optics. Given:
```
@optics
data class Foo(
    val bar: Either<Throwable, Bar>,
) { companion object }
```
There should be a way to modify `bar` if it's a `Right`/`Left` directly.
```
Foo.bar.right.modify(...) // Unresolved reference: right
```

serras:
Fixes #2865

### DIFF
--- a/arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Prism.kt
+++ b/arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Prism.kt
@@ -179,20 +179,42 @@ public interface PPrism<S, T, A, B> : POptional<S, T, A, B>, PSetter<S, T, A, B>
     /**
      * [Prism] to focus into an [arrow.core.Either.Left]
      */
-    @JvmStatic @JvmName("eitherLeft")
-    public fun <L, R> left(): Prism<Either<L, R>, L> =
+    @JvmStatic
+    public fun <L, R> left(): Prism<Either<L, R>, L> = pLeft()
+      
+    /**
+     * [Prism] to focus into an [arrow.core.Either.Right]
+     */
+    @JvmStatic
+    public fun <L, R> right(): Prism<Either<L, R>, R> = pRight()
+
+    /**
+     * [Prism] to focus into an [arrow.core.Either.Left]
+     */
+    @JvmStatic
+    public fun <L, R, E> pLeft(): PPrism<Either<L, R>, Either<E, R>, L, E> =
       Prism(
-        getOrModify = { e -> e.fold({ it.right() }, { e.left() }) },
+        getOrModify = { e ->
+          when (e) {
+            is Either.Left -> e.value.right()
+            is Either.Right -> e.left()
+          }
+        },
         reverseGet = { it.left() }
       )
 
     /**
      * [Prism] to focus into an [arrow.core.Either.Right]
      */
-    @JvmStatic @JvmName("eitherRight")
-    public fun <L, R> right(): Prism<Either<L, R>, R> =
-      Prism(
-        getOrModify = { e -> e.fold({ e.left() }, { it.right() }) },
+    @JvmStatic
+    public fun <L, R, B> pRight(): PPrism<Either<L, R>, Either<L, B>, R, B> =
+      PPrism(
+        getOrModify = { e ->
+          when (e) {
+            is Either.Left -> e.left()
+            is Either.Right -> e.value.right()
+          }
+        },
         reverseGet = { it.right() }
       )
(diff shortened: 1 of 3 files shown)


arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Prism.kt
```diff
@@ -172,6 +174,26 @@ public interface PPrism<S, T, A, B> : POptional<S, T, A, B>, PSetter<S, T, A, B>
         getOrModify = { option -> option.fold({ Either.Right(Unit) }, { Either.Left(option) }) },
         reverseGet = { _ -> None }
       )
+
+    /**
+     * [Prism] to focus into an [arrow.core.Either.Left]
+     */
+    @JvmStatic
+    public fun <L, R> pLeft(): Prism<Either<L, R>, L> =
```
nomisRev, line 182: It's prefixed `p` but implement non-polymorphic Optic. The pattern is prefix with `p` for `PPrism` and expose non-prefixed variant for regular `Prism`.

nomisRev (changes requested):
`p` prefix for non-polymorphic optic.

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Prism.kt
serras: Oh, I misunderstoof the convention, I thought `p` stood for "prism". The problem here is that `right` and `left` are already taken by other operations in `PPrism`; any ideas of better naming?

serras: Fixed in the latest commit

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Prism.kt
```diff
@@ -172,6 +175,26 @@ public interface PPrism<S, T, A, B> : POptional<S, T, A, B>, PSetter<S, T, A, B>
...
+
+    /**
+     * [Prism] to focus into an [arrow.core.Either.Right]
+     */
+    @JvmStatic @JvmName("eitherRight")
+    public fun <L, R> right(): Prism<Either<L, R>, R> =
+      Prism(
+        getOrModify = { e -> e.fold({ e.left() }, { it.right() }) },
+        reverseGet = { it.right() }
+      )
```
nomisRev, line 197: This is what I meant @serras . The `@JvmName` is not needed here.

```suggestion

    /**
     * [Prism] to focus into an [arrow.core.Either.Left]
     */
    @JvmStatic
    public fun <L, R> left(): Prism<Either<L, R>, L> = pLeft()
      
    /**
     * [Prism] to focus into an [arrow.core.Either.Right]
     */
    @JvmStatic
    public fun <L, R> right(): Prism<Either<L, R>, R> = pRight()

    /**
     * [Prism] to focus into an [arrow.core.Either.Left]
     */
    @JvmStatic
    public fun <L, R, E> pLeft(): PPrism<Either<L, R>, Either<E, R>, L, E> =
      Prism(
        getOrModify = { e ->
          when (e) {
            is Either.Left -> e.value.right()
            is Either.Right -> e.left()
          }
        },
        reverseGet = { it.left() }
      )

    /**
     * [Prism] to focus into an [arrow.core.Either.Right]
     */
    @JvmStatic
    public fun <L, R, B> pRight(): PPrism<Either<L, R>, Either<L, B>, R, B> =
      PPrism(
        getOrModify = { e ->
          when (e) {
            is Either.Left -> e.left()
            is Either.Right -> e.value.right()
          }
        },
        reverseGet = { it.right() }
      )
```

serras: Without `JvmName` I get a conflict with the same (instance) method in `Prism`

nomisRev:
> Without JvmName I get a conflict with the same (instance) method in Prism

🤦 I know why I reasoned about this incorrectly. Normally this doesn't conflict, but due to `@JvmStatic` it does. `@JvmStatic` puts it as a `static` method not the class, instead of a method on the `Companion` nested class.

Too bad, but I guess perf improvement > slightly worse Java interop.

nomisRev (approved):

raulraja (approved):

PR #2880, opened 2022-12-30
`zipOrAccumulate` for `Raise` (waiting for #2912 to be merged)

serras:
This generalizes #2872 by providing a `zipOrAccumulate` operation similar to `mapOrAccumulate`


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Raise.kt
```diff
@@ -286,6 +345,56 @@ public inline fun <R, B : Any> Raise<R>.ensureNotNull(value: B?, raise: () -> R)
...
-  recover({ block(it) }, { raise(it.nel()) })
-}
+): List<B> =
+  list.mapOrAccumulate(Semigroup.nonEmptyList()) { elt -> mapErrorNel { block(elt) } }
+
+/**
+ * Accumulate the errors from running both [action1] and [action2].
+ */
+@EffectDSL
+public inline fun <R, A, B, C> Raise<NonEmptyList<R>>.zipOrAccumulate(
```
nomisRev, line 355: For `Either` these are just called `zip`, should we also rename them to `zipOrAccumulate`? We removed regular `zip` since it's equivalent to `bind` for short-circuit behavior and we re-used the `zip` name for _accumulating_ behavior.

So we should be consistent between `Raise` and other data types.

Additionally, in `Either` is has signature `action1.zip(action2, block)` but it was requested on KotlinLang why it's not `Either.zip(action1, action2, block)` that would also align both these signatures further. It might even help back-porting them to 1.x.x easier, WDYT?

serras: I prefer too that version of `zip`, so I'm happy to make the change into a different PR, if you want.

nomisRev:
> I prefer too that version of zip, so I'm happy to make the change into a different PR, if you want.

@serras not sure I understood what you prefer 😅

serras:
I prefer writing `Either.zip(action1, action2) { transform }`.

@nomisRev what should we do about the `zip` naming? I feel like changing the one in `Either` would be too breaking, so maybe it's better to change the one in `Raise` to `zip`?

nomisRev:
>  I feel like changing the one in Either would be too breaking

It's a new api that is not yet in `1.1.x`, so it's non breaking. The current `zip` we have in `1.1.x` is the traditional Apply _short-circuiting_ behavior, whilst this zip and the one added in `arrow-2` is accumulating. So we're free to name it however we think is best for this API.

nomisRev:
Also, here I propose to deprecate the current `zip` for `1.2.0`.

https://github.com/arrow-kt/arrow/pull/2912/files#diff-44a762c203185e0889fe0fc1fbaf6a2ee7ed2960cac415256742a6fbbc4e23d3R1762

nomisRev:
Just for keeping all history included in the PR history. In the Arrow meeting of 2 February 2023 the decision was made to use `Either.zipOrAccumulate` for the concrete version of this API. The relevant PR for this work is in #2916.

nomisRev:
@serras #2916 is merged, so this can be updated and merged 🙌

serras:
Closed in favor of #2916

PR #2884, opened 2023-01-05
Improve debugging experience of leaked shift calls

nomisRev:
Currently it's possible to leak `shift`, and it results in `CancellationException`.
This PR proposes to turn that into `IllegalStateException` such that it's not dealt with like a special case by Structured Concurrency and we can also provide a more descriptive error message.


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
```diff
@@ -500,14 +502,30 @@ import kotlin.coroutines.resumeWithException
...
+ *     )
  *   }
  * }
  * ```
  * <!--- KNIT example-effect-guide-11.kt -->
+ * ```text
+ * ErrorA
+ * ```
+ *
+ * The example here will always print `ErrorA`, but never `ErrorB`. This is because `fa` is awaited first, and when it's `shifts` it will cancel `fb`.
```
franciscodr, line 520:
```suggestion
 * The example here will always print `ErrorA`, but never `ErrorB`. This is because `fa` is awaited first, and when it `shifts` it will cancel `fb`.
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
```diff
@@ -716,17 +777,20 @@ internal class FoldContinuation<R, B>(
...
     error: suspend (Throwable) -> B,
     parent: Continuation<B>,
   ) : this(context, error, parent)
-  
+
   lateinit var recover: suspend (R) -> Any?
-  
+
+  // Add AtomicBoolean to arrow-atomic
+  val isActive: AtomicRef<Boolean> = AtomicRef(true)
```
serras, line 792: I'm a bit worried about exposing the `AtomicRef` directly. Is it possible to make this `internal`?

nomisRev: This is inside an `internal class` so it's `internal` by inheriting the visibility modifier.

serras: But it shows in the `arrow-core.api` file... 🤔

nomisRev: That's because `internal` is public in the (Java) binary 😭

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/Effect.kt
```diff
@@ -500,14 +502,30 @@ import kotlin.coroutines.resumeWithException
...
+ * ```
+ *
+ * The example here will always print `ErrorA`, but never `ErrorB`. This is because `fa` is awaited first, and when it `shifts` it will cancel `fb`.
+ * If instead we used `awaitAll`, then it would print `ErrorA` or `ErrorB` due to both `fa` and `fb` being awaited in parallel.
  *
  * #### launch
  *
+ * It's **not allowed** to call `shift` from within `launch`, this is because `launch` creates a separate unrelated child Job/Continuation.
+ * Any calls to `shift` inside of `launch` will be ignored by `effect`, and result in an exception being thrown inside `launch`.
+ * Because KotlinX Coroutines ignores `CancellationException`, and thus swallows the `shift` call.
```
franciscodr, line 527:
```suggestion
 * It's **not allowed** to call `shift` from within `launch`. This is because `launch` creates a separate unrelated child Job/Continuation.
 * Any calls to `shift` inside of `launch` will be ignored by `effect` and result in an exception being thrown inside `launch`.
 * Because KotlinX Coroutines ignores `CancellationException`, and thus swallows the `shift` call.
```

franciscodr (approved):

serras (approved):

i-walker (approved):

nomisRev:
@serras I was able to get rid of `isActive` after all. I'm really confused why `complete` doesn't show up in the binary 🤔

PR #2886, opened 2023-01-09
Implement 'align' using 'buildList'

serras:
It seems that [`zipAll`](https://github.com/JetBrains/kotlin/pull/4568) is not making to the standard library anytime soon. However, by looking at it I thought of a more performant way to implement `align` in Arrow Core, using `buildList` and manual iteration with `hasNext`.


nomisRev (approved):
Thanks for the great improvement 🙏 LGTM!

Should we rename `align` to `zipAll` towards 2.0? 🤔

franciscodr (approved):

franciscodr:
> > Should we rename `align` to `zipAll` towards 2.0? 🤔
> 
> I like the current names more than `zipAll` (which to me sounds like something which takes a `Iterable<Iterable<A>>` and zips them somehow), but that might be only my inner Haskeller speaking...

If we visualize how a zip fastener works, I see it more clearly if we have two separate lists as the input of the function rather than a list containing other lists.

PR #2887, opened 2023-01-16
Fix knitCheck & re-add check

nomisRev:
@serras in [2874](https://github.com/arrow-kt/arrow/pull/2874/files#diff-a0fe23534b616d51ce686d2a1bcd1a78bc75074aef1a2f6ee96c9469991e1a4c) you removed `check` and split into `jvmTest` and `jsTest`.

Was this because you were experiencing flakiness on CI again? This has resulted in a couple of false negatives on main since only `check` guarantees that all tasks to verify the project are ran.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -87,7 +87,7 @@ jobs:
           name: 'reports-windows'
           path: '**/build/reports/**'
 
-  jvm:
+  check:
     runs-on: ubuntu-latest
     timeout-minutes: 60
 
@@ -115,7 +115,7 @@ jobs:
       - name: check
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: check
+          arguments: check -x linuxX64Test -x jsTest
 
       - name: Create code coverage report
         if: "! github.event.pull_request.head.repo.fork "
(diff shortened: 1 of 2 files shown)


serras:
I added it because of flakiness, and then kept it that way because I noticed that by having different actions GitHub parallelizes those, so the feedback came earlier. But I wasn't aware at all that some tasks were not run (except for the `checkApi`, which is implicitly done in the `update_api` step).

serras (approved):

nomisRev:
@serras if needed for flakiness we can run `linuxX64Test ` and `js jsTest` in a different job by excluding them from `check` by using `-x jsTest -x linuxX64Test`.

franciscodr (approved):

serras:
> @serras if needed for flakiness we can run `linuxX64Test ` and `js jsTest` in a different job by excluding them from `check` by using `-x jsTest -x linuxX64Test`.

I guess that would be the perfect solution: in one run `jvmTest` + all other checks, and in others run the native and JS tests.

nomisRev:
I made that change @serras 👍 &  re-requested your review.

serras (approved):

PR #2895, opened 2023-01-19
Use Duration in Schedule

### LINKED ISSUE #2744: ["Improvement"] Use Duration directly inside Schedule instead of Double

serras:
Fixes #2744 

I've only made the changes in the `arrow-fx-resilience` module, but they could be copied to `arrow-fx-coroutines` if desired.

### DIFF
--- a/arrow-libs/fx/arrow-fx-resilience/src/commonMain/kotlin/arrow/fx/resilience/Schedule.kt
+++ b/arrow-libs/fx/arrow-fx-resilience/src/commonMain/kotlin/arrow/fx/resilience/Schedule.kt
@@ -278,6 +278,9 @@ public sealed class Schedule<Input, Output> {
    * Combines with another schedule by combining the result and the delay of the [Decision] with the functions [zipContinue], [zipDuration] and a [zip] function
    */
   @ExperimentalTime
+  @Deprecated(NanosDeprecation,
+    ReplaceWith("combine(other, zipContinue, { a, b -> zipDuration(a.toDouble(DurationUnit.NANOSECONDS), b.toDouble(DurationUnit.NANOSECONDS)).nanoseconds }, zip)",
+      "kotlin.time.DurationUnit", "kotlin.time.Duration.Companion.nanoseconds"))
   public fun <A : Input, B, C> combineNanos(
     other: Schedule<A, B>,
     zipContinue: (cont: Boolean, otherCont: Boolean) -> Boolean,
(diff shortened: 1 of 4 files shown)


arrow-libs/fx/arrow-fx-resilience/src/commonMain/kotlin/arrow/fx/resilience/Schedule.kt
```diff
@@ -268,24 +267,24 @@ public sealed class Schedule<Input, Output> {
...
-  public abstract fun <A : Input, B, C> combineNanos(
+  @ExperimentalTime
+  public fun <A : Input, B, C> combineNanos(
     other: Schedule<A, B>,
     zipContinue: (cont: Boolean, otherCont: Boolean) -> Boolean,
     zipDuration: (duration: Double, otherDuration: Double) -> Double,
     zip: (Output, B) -> C
-  ): Schedule<A, C>
+  ): Schedule<A, C> =
+    combine(other, zipContinue, { a, b -> zipDuration(a.toDouble(NANOSECONDS), b.toDouble(NANOSECONDS)).nanoseconds }, zip)
```
nomisRev, line 287: I think we can get rid of this function now.

arrow-libs/fx/arrow-fx-resilience/src/commonMain/kotlin/arrow/fx/resilience/Schedule.kt
```diff
@@ -301,11 +300,11 @@ public sealed class Schedule<Input, Output> {
...
    */
-  @ExperimentalTime
-  public fun modify(f: suspend (Output, Duration) -> Duration): Schedule<Input, Output> =
-    modifyNanos { output, d -> f(output, d.nanoseconds).toDouble(NANOSECONDS) }
+  public abstract fun modify(f: suspend (Output, Duration) -> Duration): Schedule<Input, Output>
 
-  public abstract fun modifyNanos(f: suspend (Output, Double) -> Double): Schedule<Input, Output>
+  @ExperimentalTime
+  public fun modifyNanos(f: suspend (Output, Double) -> Double): Schedule<Input, Output> =
+    modify { output, d -> f(output, d.toDouble(NANOSECONDS)).nanoseconds }
```
nomisRev, line 307: We don't need any of the `Nanos functions anymore now that `Duration` is finally stable.

nomisRev (approved):
Great, LGTM!

We can get rid of all the `xxxNano` code since that was a work-around to expose `Duration` in a binary compat way without losing precision. We can leave this for another PR though ☺️

arrow-libs/fx/arrow-fx-resilience/src/commonMain/kotlin/arrow/fx/resilience/Schedule.kt
serras: I've marked them for deprecation, since they were part of the exposed API

nomisRev:
@serras deprecation makes a lot sense indeed 👍 This gives users smooth transition to new module, and breaking change only in 2.0.0 🙌

i-walker (approved):

PR #2896, opened 2023-01-19
Fix semantics of `untilOutput` and `untilInput`

### LINKED ISSUE #2846: ["BUG"][arrow-fx-coroutines] Scheduled.retryOrElseEither never stops if first `untilOutput` is true and following untilxx are false
Hi,

when creating a Schedule with multiple until[input,output]/while[input,output] conditions is run using `retryOrElseEither` it will retry forever even if a preceding predicate is met if the following predicates still result in a continue boolean.

Reproducible by adding the following test into ScheduleTest.kt:
```kotlin
    "Schedule stops retrying if first of more predicates is met" {
      val ex = Throwable("Hello")

      val schedule = Schedule.exponential<Throwable>(1.0.milliseconds)
        .untilOutput { it > 50.0.milliseconds }
        .untilInput<Throwable> { it is IllegalStateException }

      val result: Either<Throwable, Unit> = withTimeout(10.seconds) {
        schedule.retryOrElseEither({
          throw ex
        }, { t, _ -> t })
      }

      result.fold({ it shouldBe ex }, { fail("The impossible happened") })
    }
```

See also:
https://gist.github.com/kartoffelsup/896edcb154a13f2b5dc9cf639ac08311

I tried to fix this myself but frankly the code of Scheduled confuses me a bit 🤯 🙈 

I also reported this in slack: https://kotlinlang.slack.com/archives/C5UPMM0A0/p1667847505701159

p.s.
Thanks for creating/maintaining arrow. I have a lot of fun using it.

serras:
Fixes #2846

It seems that this was a semantic problem. Here's the diff between the previous and the proposed version:

```diff kotlin
public fun untilOutput(f: suspend (Output) -> Boolean): Schedule<Input, Output> =
-    !whileOutput(f)
+    whileOutput { !f(it) }
```

The former version applies `whileOutput(f)` and then negates, which applies to the *entire* `Schedule`. However, this is not right if you have several conditions, the right thing to do is to negate *only* the given `f`.

### DIFF
--- a/arrow-libs/fx/arrow-fx-resilience/src/commonMain/kotlin/arrow/fx/resilience/Schedule.kt
+++ b/arrow-libs/fx/arrow-fx-resilience/src/commonMain/kotlin/arrow/fx/resilience/Schedule.kt
@@ -370,16 +370,16 @@ public sealed class Schedule<Input, Output> {
     check { input, _ -> f(input) }
 
   /**
-   * `untilOutput(f) = whileOutput(f).not()`
+   * `untilOutput(f) = whileOutput { !f(it) }`
    */
   public fun untilOutput(f: suspend (Output) -> Boolean): Schedule<Input, Output> =
-    !whileOutput(f)
+    whileOutput { !f(it) }
 
   /**
-   * `untilInput(f) = whileInput(f).not()`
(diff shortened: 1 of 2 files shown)


nomisRev (approved):
Awesome! Thank you @serras

Issue #2565, opened 2021-11-01
["Request"]  NonEmptySet


hoc081098:
What version are you currently using?
1.0.0

What would you like to see?
NonEmptySet, ref:
- https://hackage.haskell.org/package/non-empty-0.3/docs/Data-NonEmpty-Set.html
- https://typelevel.org/cats/api/cats/data/index.html#NonEmptySet[A]=cats.data.NonEmptySetImpl.Type[A]

- https://github.com/quickbirdstudios/NonEmptyCollections

ibado:
Hi there :wave: I'd like to take this one! Is that okan? @serras

serras:
Sure, I'm really happy to see somebody taking over this. Since this doesn't conflict with anything in Arrow 1.x, please target `main` when doing the corresponding PR. Thanks for your help!

nomisRev:
This has been fixed by @ibado in #2899

PR #2899, opened 2023-01-22
Implement basic NonEmptySet

ibado:
This is a super basic implementation for #2565. After review, if this is fine, I can add any other functionality may be required.

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
@@ -29,9 +29,8 @@ public value class NonEmptySet<out T> private constructor(
   }
 
   @Suppress("RESERVED_MEMBER_INSIDE_VALUE_CLASS")
-  override fun hashCode(): Int {
-    return elements.hashCode()
-  }
+  override fun hashCode(): Int =
+    elements.hashCode()
 }
 
 public fun <T> nonEmptySetOf(first: T, vararg rest: T): NonEmptySet<T> =
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
@@ -2,23 +2,19 @@ package arrow.core
 
 public class NonEmptySet<out T>(
   private val first: T,
-  rest: Set<T>,
+  private val rest: Set<T>,
 ) : AbstractSet<T>() {
 
-  public val elements: Set<T>
-
-  init {
-      elements = setOf(first) + rest
-  }
+  public val elements: Set<T> by lazy { setOf(first) + rest }
 
   public operator fun plus(other: NonEmptySet<@UnsafeVariance T>): NonEmptySet<T> =
-    NonEmptySet(this.first, elements + other.elements)
+    NonEmptySet(this.first, rest + other.elements)
 
   public operator fun plus(s: Set<@UnsafeVariance T>): NonEmptySet<T> =
-    NonEmptySet(first, this.elements + s)
+    NonEmptySet(first, rest + s)
 
   public operator fun plus(s: @UnsafeVariance T): NonEmptySet<T> =
-    NonEmptySet(first, this.elements.plus(s))
+    NonEmptySet(first, rest + s)
 
   override val size: Int
     get() = elements.size
@@ -26,22 +22,13 @@ public class NonEmptySet<out T>(
   override fun iterator(): Iterator<T> = elements.iterator()
 
   override fun toString(): String = "NonEmptySet(${elements.joinToString()})"
-
-  override fun equals(other: Any?): Boolean = when {
-    other === this -> true
-    other !is Set<*> -> false
-    other is NonEmptySet<*> -> elements == other.elements
-    else -> elements == other
-  }
-
-  override fun hashCode(): Int = elements.hashCode()
 }
 
 public fun <T> nonEmptySetOf(fist: T, vararg rest: T): NonEmptySet<Any?> =
   NonEmptySet(fist, rest.toSet())
 
 public fun <T> Collection<T>.toNonEmptySetOrNull(): NonEmptySet<T>? =
-  firstOrNull()?.let { NonEmptySet(it, this.toSet()) }
+  firstOrNull()?.let { NonEmptySet(it, this.drop(1).toSet()) }
 
 public fun <T> Collection<T>.toNonEmptySetOrNone(): Option<NonEmptySet<T>> =
   toNonEmptySetOrNull().toOption()
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
@@ -1,6 +1,9 @@
 package arrow.core
 
-public class NonEmptySet<out T> private constructor(
+import kotlin.jvm.JvmInline
+
+@JvmInline
+public value class NonEmptySet<out T> private constructor(
   private val elements: Set<T>
 ) : Set<T> by elements {
 
@@ -19,11 +22,13 @@ public class NonEmptySet<out T> private constructor(
 
   override fun toString(): String = "NonEmptySet(${this.joinToString()})"
 
+  @Suppress("RESERVED_MEMBER_INSIDE_VALUE_CLASS")
   override fun equals(other: Any?): Boolean = when (other) {
-    is Set<*> -> elements == other
-    else -> false
+    is NonEmptySet<*> -> elements == other.elements
+    else -> elements == other
   }
 
+  @Suppress("RESERVED_MEMBER_INSIDE_VALUE_CLASS")
   override fun hashCode(): Int {
     return elements.hashCode()
   }
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
@@ -4,16 +4,24 @@ public class NonEmptySet<out T> private constructor(
   private val elements: Set<T>
 ) : Set<T> by elements {
 
-  public constructor(first: T, rest: Set<T>): this(setOf(first) + rest)
+  public constructor(first: T, rest: Set<T>) : this(setOf(first) + rest)
 
-  public operator fun plus(set: Set<@UnsafeVariance T>): NonEmptySet<T> = NonEmptySet(elements + set)
+  public operator fun plus(set: Set<@UnsafeVariance T>): NonEmptySet<T> =
+    NonEmptySet(elements + set)
 
-  public operator fun plus(element: @UnsafeVariance T): NonEmptySet<T> = NonEmptySet(elements + element)
+  public operator fun plus(element: @UnsafeVariance T): NonEmptySet<T> =
+    NonEmptySet(elements + element)
+
+  public fun <R> map(transform: (@UnsafeVariance T) -> R): NonEmptySet<R> =
+    NonEmptySet(elements.map { transform(it) }.toSet())
+
+  override fun isEmpty(): Boolean = false
 
   override fun toString(): String = "NonEmptySet(${this.joinToString()})"
 
-  override fun equals(other: Any?): Boolean {
-    return elements == other
+  override fun equals(other: Any?): Boolean = when (other) {
+    is Set<*> -> elements == other
+    else -> false
   }
 
   override fun hashCode(): Int {
@@ -21,8 +29,8 @@ public class NonEmptySet<out T> private constructor(
   }
 }
 
-public fun <T> nonEmptySetOf(fist: T, vararg rest: T): NonEmptySet<T> =
-  NonEmptySet(fist, rest.toSet())
+public fun <T> nonEmptySetOf(first: T, vararg rest: T): NonEmptySet<T> =
+  NonEmptySet(first, rest.toSet())
 
 public fun <T> Collection<T>.toNonEmptySetOrNull(): NonEmptySet<T>? =
   firstOrNull()?.let { NonEmptySet(it, minus(it).toSet()) }
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
@@ -1,0 +1,47 @@
+package arrow.core
+
+public class NonEmptySet<out T>(
+  private val first: T,
+  rest: Set<T>,
+) : AbstractSet<T>() {
+
+  public val elements: Set<T>
+
+  init {
+      elements = setOf(first) + rest
+  }
+
+  public operator fun plus(other: NonEmptySet<@UnsafeVariance T>): NonEmptySet<T> =
+    NonEmptySet(this.first, elements + other.elements)
+
+  public operator fun plus(s: Set<@UnsafeVariance T>): NonEmptySet<T> =
+    NonEmptySet(first, this.elements + s)
+
+  public operator fun plus(s: @UnsafeVariance T): NonEmptySet<T> =
+    NonEmptySet(first, this.elements.plus(s))
+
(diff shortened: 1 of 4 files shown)


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
```diff
@@ -0,0 +1,47 @@
...
+  override fun iterator(): Iterator<T> = elements.iterator()
+
+  override fun toString(): String = "NonEmptySet(${elements.joinToString()})"
+
+  override fun equals(other: Any?): Boolean = when {
+    other === this -> true
+    other !is Set<*> -> false
+    other is NonEmptySet<*> -> elements == other.elements
+    else -> elements == other
+  }
```
hoc081098, line 35: I think it should be `= super.equals(other)`, likes NonEmptyList

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
```diff
@@ -0,0 +1,47 @@
...
+public class NonEmptySet<out T>(
+  private val first: T,
+  rest: Set<T>,
+) : AbstractSet<T>() {
+
+  public val elements: Set<T>
+
+  init {
+      elements = setOf(first) + rest
+  }
```
serras, line 12:
```suggestion
 public val elements: Set<T> = setOf(first) + rest
```

or even better, if we don't want to compute it until we need it

```suggestion
 public val elements: Set<T> by lazy { setOf(first) + rest }
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
```diff
@@ -0,0 +1,47 @@
...
+
+  override fun hashCode(): Int = elements.hashCode()
+}
+
+public fun <T> nonEmptySetOf(fist: T, vararg rest: T): NonEmptySet<Any?> =
+  NonEmptySet(fist, rest.toSet())
+
+public fun <T> Collection<T>.toNonEmptySetOrNull(): NonEmptySet<T>? =
+  firstOrNull()?.let { NonEmptySet(it, this.toSet()) }
```
serras, line 44: This will include `first` twice... in theory it's fine since it represents a set, but I'm not sure if we can maybe guarantee that `first` is not present in `rest`

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/NonEmptySetTest.kt
```diff
@@ -0,0 +1,25 @@
...
+      nonEmptySet.elements.toNonEmptySetOrNull().shouldNotBeNull() shouldBe nonEmptySet
+    }
+  }
+
+  "iterable.toNonEmptySetOrNone should round trip" {
+    checkAll(Arb.nonEmptySet(Arb.int())) { nonEmptySet ->
+      nonEmptySet.elements.toNonEmptySetOrNone() shouldBe nonEmptySet.some()
+    }
+  }
+})
```
serras, line 24: I am missing a few properties here, like:
- calling `nonEmptySetOrNull` on an empty collection fails,
- adding an element already present doesn't change the set.

serras (changes requested):
Nice job! I think this just need a few more tests to ensure that the implementation is correct even if somebody else changes or tries to optimize it even further.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/NonEmptySetTest.kt
ibado: Okay, I'll add them :slightly_smiling_face:

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
ibado: I've changed the use of `rest` a little bit. Please take a look

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
```diff
@@ -0,0 +1,34 @@
+package arrow.core
+
+public class NonEmptySet<out T>(
+  private val first: T,
+  private val rest: Set<T>,
+) : AbstractSet<T>() {
+
+  public val elements: Set<T> by lazy { setOf(first) + rest }
```
roomscape, line 8: Does the benefit of deferred initialisation here outweigh the potentially significant cost of using a lazy property?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
```diff
@@ -0,0 +1,34 @@
...
+  override fun iterator(): Iterator<T> = elements.iterator()
+
+  override fun toString(): String = "NonEmptySet(${elements.joinToString()})"
+}
+
+public fun <T> nonEmptySetOf(fist: T, vararg rest: T): NonEmptySet<Any?> =
+  NonEmptySet(fist, rest.toSet())
+
+public fun <T> Collection<T>.toNonEmptySetOrNull(): NonEmptySet<T>? =
+  firstOrNull()?.let { NonEmptySet(it, this.drop(1).toSet()) }
```
roomscape, line 31: Does this require the source collection to have a stable iteration order?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
roomscape, line 8: Should this be private?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
```diff
@@ -0,0 +1,34 @@
...
+
+public class NonEmptySet<out T>(
+  private val first: T,
+  private val rest: Set<T>,
+) : AbstractSet<T>() {
+
+  public val elements: Set<T> by lazy { setOf(first) + rest }
+
+  public operator fun plus(other: NonEmptySet<@UnsafeVariance T>): NonEmptySet<T> =
+    NonEmptySet(this.first, rest + other.elements)
```
roomscape, line 11: Is this overload necessary, or is the `plus(Set)` function sufficient for both cases?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
```diff
@@ -0,0 +1,34 @@
...
+
+  override val size: Int
+    get() = elements.size
+
+  override fun iterator(): Iterator<T> = elements.iterator()
+
+  override fun toString(): String = "NonEmptySet(${elements.joinToString()})"
+}
+
+public fun <T> nonEmptySetOf(fist: T, vararg rest: T): NonEmptySet<Any?> =
```
roomscape, line 27: Small typo here: ~fist~ -> first

roomscape, line 27: Is the return type correct here? Should it be `T` instead of `Any?`

serras:
@ibado I'm wondering whether the `first` + `rest` separation is really needed apart from the constructor. More concretely, I'm proposing to make `NonEmptySet` simply a wrapper for `Set`, but only expose the right constructor:

```kotlin
@JvmInline public value class NonEmptySet<out T> private constructor(
  val elements: Set<T>
): AbstractSet<T> {
  public constructor(first: T, varargs rest: T): this(setOf(first) + rest.toSet())
}
```

You can still overload the `plus` operator to return `NonEmptySet`, but you already know that's the case if one of the arguments is `NonEmptySet`.

```kotlin
  public operator fun plus(s: Set<@UnsafeVariance T>): NonEmptySet<T> =
    NonEmptySet(elements + s)
```

This has the added benefit of completely removing the wrapper at runtime, since it's a value class.

@roomscape what do you think?

roomscape:
> @roomscape what do you think?

I think you're right, the `first`/`rest` separation may not be needed outside of the constructor. An efficient implementation will always need to create a `Set` to implement the required operations. Including the `first` item alongside the others in that set makes the implementation simpler, and I can't actually think of any functionality that would benefit from keeping one item separate.

In that case, your suggestion of a simple wrapper class works very well. Interface delegation would be a good fit:

```kotlin
class NonEmptySet<out T> private constructor(elements: Set<T>): Set<T> by elements
```

I like the idea of making it an inline class too, but I'm cautious about the implications, and I think the benefits would only be slight. Value classes can use interface delegation, so that would work. But value classes do seem to have some quirks, and in particular I think generic parameters are still experimental. It might be safer to leave that for a future enhancement.

ibado:
I uploaded a simplify version following the approach of @roomscape since I think is nice to delegate the "set" capabilities to the wrapped set. Not sure about the inline value class.. It's up to you.
Please take a look and LMK :slightly_smiling_face: 
P.S: the overrides for plus are necessary for returning back a `NonEmptySet`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
ibado: Not sure. I mean, the order is not relevant for a set (unless is a sorted set), right?
I think your concern is that this function may not return the exact same output every time, right?

roomscape: I think this code relies on an assumption that the element dropped by `drop(1)` will be the same element that was returned by `firstOrNull`. That would require the iterator to return its elements in the same order each time. Although that's probably true for the common standard library collections, I don't think collection iterators are required to guarantee it.

ibado: I get your point, maybe this is better then:
```kotlin
firstOrNull()?.let { NonEmptySet(it, minus(it).toSet()) }
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
```diff
@@ -0,0 +1,31 @@
...
+
+  public operator fun plus(set: Set<@UnsafeVariance T>): NonEmptySet<T> = NonEmptySet(elements + set)
+
+  public operator fun plus(element: @UnsafeVariance T): NonEmptySet<T> = NonEmptySet(elements + element)
+
+  override fun toString(): String = "NonEmptySet(${this.joinToString()})"
+
+  override fun equals(other: Any?): Boolean {
+    return elements == other
+  }
```
serras, line 17: Shouldn't we account for types here?

```suggestion
  override fun equals(other: Any?): Boolean = when(other) {
    is Set<*> -> elements == other
    else -> false
  }
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
```diff
@@ -0,0 +1,31 @@
...
+
+public class NonEmptySet<out T> private constructor(
+  private val elements: Set<T>
+) : Set<T> by elements {
+
+  public constructor(first: T, rest: Set<T>): this(setOf(first) + rest)
+
+  public operator fun plus(set: Set<@UnsafeVariance T>): NonEmptySet<T> = NonEmptySet(elements + set)
+
+  public operator fun plus(element: @UnsafeVariance T): NonEmptySet<T> = NonEmptySet(elements + element)
```
serras, line 11: There are other operations which you can implement more efficiently or with better types in `NonEmptySet`:
- `isEmpty() = false`, `isNonEmpty() = true`
- `any() = true` (the version which doesn't check a predicate)
- for `map`, you can implement a version which returns `NonEmptyList`

serras (approved):

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
ibado: You got it!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
ibado: `map` and `isEmpty` done! `isNotEmtpy()` and `any()` use `isEmpty` behind the scenes :slightly_smiling_face:

serras: Lovely! Thanks for putting all this love in the PR :)

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
```diff
@@ -0,0 +1,39 @@
...
+  public constructor(first: T, rest: Set<T>) : this(setOf(first) + rest)
+
+  public operator fun plus(set: Set<@UnsafeVariance T>): NonEmptySet<T> =
+    NonEmptySet(elements + set)
+
+  public operator fun plus(element: @UnsafeVariance T): NonEmptySet<T> =
+    NonEmptySet(elements + element)
+
+  public fun <R> map(transform: (@UnsafeVariance T) -> R): NonEmptySet<R> =
+    NonEmptySet(elements.map { transform(it) }.toSet())
```
hoc081098, line 16: How about `elements.mapTo(mutableSetOf(), transform)`

ibado: Yeap, that's more efficient. Thanks!

serras:
@ibado I've been working on a similar PR for `NonEmptyList`, and I've realized that you need to override `equals` if you want `NonEmptySet` to be comparable with `Set`. [This commit](https://github.com/arrow-kt/arrow/pull/2911/files#diff-86c169842280a4bff022b715f79323553f8a1e2b91c4fbbc9cf155eeb7a91e20R157-R164) shows how.

Once you do this, you can really change to a value class, which removes any wrapper at runtime:

```kotlin
@JvmInline public value class NonEmptySet<out T> private constructor(
  private val elements: Set<T>
) : Set<T> by elements {
```

ibado:
Now `:arrow-core:compileTestKotlinJvm` is failing :upside_down_face: Looks that also your PR. Something about IR

serras:
@ibado I think the commit I've just pushed should fix the problem (it's just a different way to write the generator)

roomscape (approved):

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
```diff
@@ -0,0 +1,44 @@
...
+  @Suppress("RESERVED_MEMBER_INSIDE_VALUE_CLASS")
+  override fun equals(other: Any?): Boolean = when (other) {
+    is NonEmptySet<*> -> elements == other.elements
+    else -> elements == other
+  }
+
+  @Suppress("RESERVED_MEMBER_INSIDE_VALUE_CLASS")
+  override fun hashCode(): Int {
+    return elements.hashCode()
+  }
```
nomisRev, line 34: Nit: expression body.

```suggestion
  override fun hashCode(): Int =
    elements.hashCode()
```

nomisRev (approved):
Looks great @ibado! Thank you and congrats on your first contribution 🙌 🥳

Small styling nit.

>  I think this just need a few more tests to ensure that the implementation is correct

This would also be nice, you can run `./gradlew koverHtmlReport` to check the coverage of `NonEmptySet`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
ibado: Done! :)

serras:
@ibado congrats on your first contribution! Thanks for your work here, and for ensuring the best quality for this new implementation :)

ibado:
Thanks! Happy to help 😄

PR #2902, opened 2023-01-24
Add Saga type to Arrow Fx Resilience

nomisRev:
Deprecates https://github.com/nomisRev/Saga, and moves it to the new Arrow Fx Resilience library.


raulraja (approved):
Great to see Saga being part of Arrow Fx Resilience, thanks @nomisRev !

arrow-libs/fx/arrow-fx-resilience/src/commonMain/kotlin/arrow/fx/resilience/Saga.kt
```diff
@@ -0,0 +1,178 @@
...
+ * By doing so we can guarantee that any transactional like operations made by the [Saga] will
+ * guarantee that it results in the correct state.
+ */
+public inline fun <A> saga(noinline block: suspend SagaScope.() -> A): Saga<A> = block
+
+/** Create a lazy [Saga] that will only run when the [Saga] is invoked. */
+public fun <A> saga(
+  action: suspend SagaActionStep.() -> A,
+  compensation: suspend (A) -> Unit
+): Saga<A> = saga { saga(action, compensation) }
```
serras, line 93: This code is a bit puzzling. Is there any nice naming to separate the `saga` that takes `SagaActionStep` from the one that takes `SagaScope`?

serras (approved):

i-walker (approved):

arrow-libs/fx/arrow-fx-resilience/src/commonMain/kotlin/arrow/fx/resilience/Saga.kt
nomisRev: Hmm, I'm not sure I can think of a better name 😅
The goal here was similar then with `Resource`, but for `Resource` we not have a different name for the DSL version and the top-level _lazy_ version. I.e. for `Resource`:

```kotlin
val res: Resource<A> = resource(acquire, release)

val res2: Resource<A> = resource {
  install(acquire, release)
}
```

and for Saga we currently have:
```kotlin
val saga: Saga<A> = saga(action, compensate)
val saga2: Saga<A> = saga {
  saga(action, compensate)
}
```

The reason for `SagaActionStep`, and we do the same for `Resource` in `acquire`, is to disallow `bind()` being called from within the `action` or `acquire` step at compile-time.

So following the `Resource` pattern, then the DSL version probably needs a different name but I am drawing a black on what can be a good name for that operation 🤔

franciscodr (approved):
Thanks, @nomisRev! 🙌

PR #2904, opened 2023-01-25
Github Actions improvements, Revert karma config, and add KotestProjectConfig

nomisRev:
In #2874 we increased property test iterations from `250` to `1000` [here](https://github.com/arrow-kt/arrow/pull/2874/files#diff-85bf56e89ebb44c64dfc77cd98e361d8568c1250af227e24763f78053e0b71ccL9).

We then later had to add `karma.config` to fix timeouts on JS platform, since running 1000 iterations was taking so long that the browsers considered the tests timing out.

This PR reverts the `karma.config`, and uses `KotestProjectConfig` to set iterations back to `250`. This was done because I noticed some flakiness issues on JVM using `1000` iterations on Github Actions 😞


nomisRev:
`check` is still flaky with this change 😞

@serras Whenever I check the failed `check` task I see an enormous amount of log output related to `classpath` of Optics KSP tests. Do you know if this is something we can disable, or configure? Or is this out of our control? All the output seems unnecessary to always be logged, it seems like `INFO` level logging.

.github/workflows/pull_request.yml
```diff
@@ -102,20 +102,10 @@ jobs:
...
-      - name: compileTestKotlin
-        uses: gradle/gradle-build-action@v2
-        with:
-          arguments: compileTestKotlin
-
       - name: check
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: check -x linuxX64Test -x jsTest
+          arguments: knitCheck apiCheck jvmTest --scan
```
JavierSegoviaCordoba, line 108: I would move `jvmTest` to its own one to keep consistency.

build.gradle.kts
```diff
@@ -16,6 +16,14 @@ allprojects {
...
     maven(url = "https://oss.sonatype.org/content/repositories/snapshots/")
   }
+
+  afterEvaluate {
+    tasks {
+      withType<Test>().configureEach {
+        maxParallelForks = 1
+      }
+    }
+  }
```
JavierSegoviaCordoba: Do we need `afterEvaluate`?

JavierSegoviaCordoba (approved):
I would only change the `jvmTest` part

PR #2905, opened 2023-01-26
Update Circuit Breaker & remove Arrow Fx Coroutines dependency from Resilience

nomisRev:
- [x] Switch to TimeMark instead of timeMillis
- [x] Introduce OpeningStrategy
- [x] Add SlidingWindow OpeningStrategy
- [x] Remove dependency on Bracket / Arrow Fx Coroutines from Arrow Fx Resilience


nomisRev (commented):
@franciscodr looks great! 👏 👏 Thanks for tacking this on 🙏 (I cannot approve my won PR 😅).

**nit:** I think `500_000` iterations for JVM is really excessive for stack-safety test 😂
Can't remember where that number came from. We used to have it around 20_000 for JVM/Native, and 5000 for JS IIRC..

**nit:** _common_ native code can also be put in `nativeMain`, and they don't have to repeat it for every platform.

We can look into both nits at a later time, they're not-blocking for this PR IMO.

franciscodr (approved):

arrow-libs/fx/arrow-fx-resilience/src/commonTest/kotlin/arrow/fx/resilience/CircuitBreakerTest.kt
```diff
@@ -1,274 +1,314 @@
...
+    when (val s = cb.state()) {
+      is CircuitBreaker.State.HalfOpen -> {
+        assertEquals(resetTimeout, s.resetTimeout)
       }
 
-      // Rejects all other tasks in HalfOpen
-      shouldThrow<CircuitBreaker.ExecutionRejected> { cb.protectOrThrow { throw dummy } }
-      shouldThrow<CircuitBreaker.ExecutionRejected> { cb.protectOrThrow { throw dummy } }
+      else -> fail("Invalid state: Expect CircuitBreaker.State.HalfOpen but found $s")
+    }
```
nomisRev, line 264: **Nit:** This pattern seems to come back a lot, we can create a small DSL for it. Something like.

```kotlin
inline fun reified <A, reified B: A> assert(expected: A, block: B.(b: B) -> Unit): Unit =
  if(expected is B) block(expected, expected) else fail("Expected ${B::class.simpleName} but found $expected)
```

nomisRev (commented):
Looking great @franciscodr. Awesome work, thank you 🙌
Just a couple of nits, and suggestions ☺️

arrow-libs/fx/arrow-fx-resilience/src/commonMain/kotlin/arrow/fx/resilience/CircuitBreaker.kt
```diff
@@ -699,10 +673,63 @@ private constructor(
...
+    is Count -> copy(failuresCount = 0)
+    else -> this
+  }
+    internal abstract fun shouldOpen(): Boolean
+    internal abstract fun trackFailure(failureAt: TimeMark): OpeningStrategy
+
+    public data class Count(
+      val maxFailures: Int,
+      val failuresCount: Int = 0,
+      val lastFailureAt: TimeMark? = null
```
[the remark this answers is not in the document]
franciscodr: Nope, I kept it in case we wanted to use it in the future, but we can remove it for now.

arrow-libs/fx/arrow-fx-resilience/src/commonMain/kotlin/arrow/fx/resilience/CircuitBreaker.kt
```diff
@@ -176,161 +177,125 @@ private constructor(
...
       is Closed -> {
         when (result) {
           is Either.Right -> {
-            if (curr.failures == 0) result.value
-            else { // In case of success, must reset the failures counter!
-              val update = Closed(0)
-              if (!state.compareAndSet(curr, update)) markOrResetFailures(result) // retry?
-              else result.value
-            }
+            result.value
```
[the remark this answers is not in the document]
franciscodr: Yes, good catch!

arrow-libs/fx/arrow-fx-resilience/src/commonTest/kotlin/arrow/fx/resilience/CircuitBreakerTest.kt
franciscodr: Adding the receiver `B` to the lambda parameter called `block` was causing this assertion to fail:
```
assert(cb.state()) { s: CircuitBreaker.State.Open -> assertEquals(resetTimeout * exponentialBackoffFactor, s.resetTimeout) }
```
The reason is that the assertion was reading the `resetTimeout` and `exponentialBackoffFactor` properties from the `B` value instead of those provided by the test suite.

arrow-libs/fx/arrow-fx-resilience/src/commonMain/kotlin/arrow/fx/resilience/CircuitBreaker.kt
```diff
@@ -212,7 +212,13 @@ private constructor(
       is Closed -> {
         when (result) {
           is Either.Right -> {
-            result.value
+            val openingStrategy = state.get().openingStrategy
+            if (openingStrategy is OpeningStrategy.Count && openingStrategy.failuresCount == 0)
+              result.value
+            else {
+              state.compareAndSet(curr, Closed(openingStrategy.resetFailuresCount()))
+              result.value
```
nomisRev, line 220: This is not a correct atomic update, `compareAndSet` returns `false` a parallel call already updated the `atomic` in-between. You need to _loop_ until `compareAndSet` returns `true`, the pattern in this data type is to use _tailrec recursion_ to do so. Alternatively you could also use `atomic.loop { }` with `return`.

The original code also used something similar to this:

```suggestion
              if(!state.compareAndSet(curr, Closed(openingStrategy.resetFailuresCount()))) markOrResetFailures(reset) // retry
              else result.value
```

arrow-libs/fx/arrow-fx-resilience/src/commonMain/kotlin/arrow/fx/resilience/CircuitBreaker.kt
```diff
@@ -724,11 +726,19 @@ private constructor(
 
       override fun trackFailure(failureAt: TimeMark): OpeningStrategy =
         if (failures.size < maxFailures + 1) copy(failures = failures + failureAt)
-        else copy(failures = failures.drop(1) + failureAt)
+        else copy(failures = failures.slide(failureAt))
+
+      private fun List<TimeMark>.slide(timeMark: TimeMark): List<TimeMark> =
+        buildList(size) {
+          forEachIndexed { index, mark ->
+            if (index == 1) Unit else add(mark)
```
nomisRev, line 734: This was my bad, whoops. `drop(1)` removes `index == 0`.

```suggestion
            if (index == 0) Unit else add(mark)
```

nomisRev (commented):
2 small nits:

1. Incorrect usage of `compareAndSet`
2. Mistake in my previous suggestion.

arrow-libs/fx/arrow-fx-resilience/src/commonMain/kotlin/arrow/fx/resilience/CircuitBreaker.kt
franciscodr: Good catch. It was a last-minute change so I totally forgot to add the retry piece

PR #2910, opened 2023-01-27
[WIP] Introduce Schedule encoding

nomisRev:
New encoding for Schedule that gets rid of the dependent type simulation, and uses recursive functions instead.


nomisRev:
Going to raise a different WIP.

PR #2911, opened 2023-01-30
Implement NonEmptyList using value class

serras:
After the discussion in #2899, I've tried to implement `NonEmptyList` using a value class wrapping over `List`. This means that most of `List` is implemented by delegation, instead of using `AbstractList`, and for those who need an overload returning `NonEmptyList`, the implementation now is just a call to the underlying `List` one.

Something we lose is good interoperability with Java. Method names in value classes as [mangled](https://kotlinlang.org/docs/inline-classes.html#mangling), and this seems to happen even for extension methods. Furthermore, at runtime `NonEmptyList` is *literally* `List`, which can be noticed in the return type of methods like `nonEmptyListOf`.

Since the binary API changes, it may make sense to push this PR to `arrow-2` instead of `main`.


serras:
I've rebased the PR against `arrow-2`, as agreed in the comments

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -182,46 +188,42 @@ public class NonEmptyList<out A>(
...
 
   public companion object {
 
+    @JvmStatic @JvmName("of")
+    public fun <A> of(first: A, vararg tail: A): NonEmptyList<A> =
+      NonEmptyList(listOf(first) + tail)
+
+    @JvmStatic @JvmName("ofOrNull")
+    public fun <A> ofOrNull(list: Iterable<A>): NonEmptyList<A>? =
+      list.firstOrNull()?.let { NonEmptyList(list.toList()) }
```
nomisRev, line 225: Weren't these constructors deprecated in 1.x.x in favour of `nonEmptyListOf` to be in line with `listOf` etc from Kotlin Std?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -230,7 +232,7 @@ public class NonEmptyList<out A>(
         "import arrow.core.toOption"
       )
     )
-    @JvmStatic
+    @JvmStatic @JvmName("fromList")
     public fun <A> fromList(l: List<A>): Option<NonEmptyList<A>> =
       if (l.isEmpty()) None else Some(NonEmptyList(l))
```
nomisRev, line 237: I thought these constructors were also deprecated in favour of `toEmptyListOrNone()` and `arrow.core.toNonEmptyListOrNull`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -241,8 +243,8 @@ public class NonEmptyList<out A>(
         "import arrow.core.toNonEmptyListOrNull"
       )
     )
-    @JvmStatic
-    public fun <A> fromListUnsafe(l: List<A>): NonEmptyList<A> =
+    @JvmStatic @JvmName("fromListUnsafe")
+    public inline fun <A> fromListUnsafe(l: List<A>): NonEmptyList<A> =
       NonEmptyList(l)
```
nomisRev, line 248: Is there any good reason to keep this API? 🤔

nomisRev (commented):
Seems this PR brings back a couple constructors that were previously deprecated.

arrow-libs/core/arrow-core/src/jvmTest/java/arrow/core/NonEmptyListUsage.java
```diff
@@ -1,23 +1,22 @@
...
+                NonEmptyList.of(1, 2, 3, 4, 5)
         );
-        NonEmptyList<Integer> flatten = flatten(nonEmptyListOf(
-                nonEmptyListOf(1, 2),
-                nonEmptyListOf(3, 4)
-        ));
+        // List<Integer> flatten = flatten(nonEmptyListOf(
+        //         nonEmptyListOf(1, 2),
+        //         nonEmptyListOf(3, 4)
+        // ));
```
franciscodr, line 18: Should we remove these commented lines?

franciscodr (approved):

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
serras: These were added back so that they could be called from Java, because mangling make it otherwise impossible. We could also remove them.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
serras: I think it's important to give a escape hatch in case somebody "knows what there're doing". But they could always use the version which may return null and do `!!` afterwards.

nomisRev: > But they could always use the version which may return null and do !! afterwards.

Yes, that is actually what I was thinking of instead of explicitly exposing an API for it. We don't expose any escape hatches anywhere else in Kotlin since you can always quite easily explicitly walk around it in the few cases you need to.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
nomisRev: 🤔 I would be in favour to remove them. Optimising for Java seems strange, especially since on the Java side it just returns `List` anyway.

PR #2913, opened 2023-02-02
[PROPOSAL] Option API deprecation, and preparation for 2.x.x

franciscodr:
This PR proposes a smaller API surface for Option without compromising its functionality or feature set.
The goal of these changes is to offer APIs that are more powerful and more idiomatic to use in Kotlin. Following the signatures and naming of some other well-known APIs from Kotlin Std, or KotlinX Coroutines

# Option (API Size 70 -> 21)

### Keep
- fold

### Remove
- crosswalk: replace with `map { f(it) }`
- crosswalkMap: replace with `fold( { emptyMap() }, { a -> f(a).mapValues { Some(it.value) } })`
- crosswalkNull: replace with `map { value -> f(value)?.let { Some(it) } }.orNull()`
- foldLeft: replace with `fold`
- foldMap: replace with `fold`
- reduceOrNull: replace with `map { value -> operation(initial(value), value) }.orNull()`
- reduceRightEvalOrNull: replace with `fold({ Eval.now(null) },  { value -> operation(value, Eval.now(initial(value))) })`
- separateEither: replace with `fold({ None to None }) { either -> either.fold<Pair<Option<A>, Option<B>>>({ Some(it) to None }, { None to Some(it) }) }`
- separateValidated: replace with `fold({ None to None }) { validated -> validated.fold<Pair<Option<A>, Option<B>>>({ Some(it) to None }, { None to Some(it) }) }`
- unite: replace with `map { iterable -> iterable.fold(MA) }`
- uniteEither: replace with `flatMap { either -> either.fold({ None }, { Some(it) }) }`
- uniteValidated: replace with `flatMap { validated -> validated.fold({ None }, ::Some) }`
- All the traverse methods can be replaced by a simple fold, or even simpler by orNull, getOrElse in many cases.
  - sequence (Iterable)
  - sequenceEither
  - sequence (Either)
  - sequenceValidated
  - sequence (Validated)
  - traverse (Iterable)
  - traverse Either
  - traverse (Either)
  - traverseValidated
  - traverse (Validated)

### New
- onSome: evaluate a fire-and-forget action if the value is defined
- onNone: evaluate a fire-and-forget action if the value is empty

### Keep
- map: transform the value if defined
- filter: keep the value if it holds the predicate. Otherwise, return `None`
- filterNot: keep the value if it doesn't hold the predicate. Otherwise, return `None`
- flatMap: transform the value if defined with Option + flatten
- flatten: flatten nested option values
- unzip: convert an option value containing a pair into a pair with two option values
- widen: conveniently change the generic parameters to a superclass type

### Remove
- all: replace with `fold({ false }, predicate)` or `map(predicate).getOrElse{ false }`
- ensure: use instead DSL + ensure, flatMap + if-else
- exists: replace with `fold({ true }, predicate)` or `map(predicate).getOrElse{ true }`
- findOrNull: replace with `getOrNull()?.takeIf(predicate)`
- pairLeft: replace with `map { left to it }`
- pairRight: replace with `map { it to right }`
- replicate: replace with `map { List(n) { it } }` // There's an alternative with Monoid
- ​​tap -> replace with `onDefined` to be more in line with Kotlin Std
- tapNone -> replace with `onEmpty` to be more in line with Kotlin Std
- unit: replace with `Some(Unit)`
- void: replace with `map {}`

### Replace
- zip: the original zip implementation will be replaced with inline option + bind

### Remove
- handleError: replace with `getOrElse { f(Unit) }`
- handleErrorWith: replace with `orElse { f(Unit) }`
- redeem: replace with `map(fb).orElse { Some(fe(Unit)) }`
- redeemWith: replace with `flatMap(fb).orElse(fe)`
- rethrow: replace with `flatMap { it.fold({ None }, { a -> Some(a) }) }`

### Remove
- align: prefer using a simple fold, or when expression
- padZip: prefer using a simple fold, or when expression
- salign: prefer using a simple fold, or when expression
- unalign: prefer using a when expression

### New
- getOrNull: extract the value or return null

### Keep
- compareTo: keep as Comparable operator
- getOrElse: extract or provide a fallback value if empty
- isEmpty: check if the value is not defined
- isNotEmpty: check if the value is defined
- orElse: return the option value if defined. Otherwise, return the provided fallback value
- toEither: convert the option value as Either.Right if defined. Otherwise, return the provided fallback value as Either.Left
- toList: convert the option value as a singleton list if defined. Otherwise, return an empty list
- toMap: convert the option value containing a pair as a singleton map if defined. Otherwise, return an empty map
- T?.toOption(): convert a nullable value into `None` if it's null. Otherwise, return `Some` containing the value

### Remove
- and: replace with `flatMap { value }`
- Boolean.maybe: replace with `if (this) { Some(f()) } else { None }`
- combine: will be covered by accumulating zip behavior in 2.x.x
- combineAll: replace with `fold(Monoid.option(MA))`
- filterIsInstance: replace with `flatMap { when (it) { is B -> Some(it) else -> None } }`
- isDefined: duplicate - use `isNotEmpty`
- mapNotNull: replace with `flatMap { fromNullable(f(it)) }`
- nonEmpty: duplicate - use `isNotEmpty`
- or: replace with `orElse`
- orNull: replace with `getOrNull`


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -679,18 +679,30 @@ public sealed class Option<out A> {
...
+  )
   public inline fun <K, V> crosswalkMap(f: (A) -> Map<K, V>): Map<K, Option<V>> =
     when (this) {
       is None -> emptyMap()
       is Some -> f(value).mapValues { Some(it.value) }
     }
 
+  @Deprecated(
+    NicheAPI + "Prefer using the Option DSL or fold",
+    ReplaceWith("fold<Option<B>?>({ null }) { value -> f(value)?.let(::Some) }")
```
nomisRev, line 704:
```suggestion
    ReplaceWith(
      "getOrNull()?.let { value -> Some(f(value)) }",
      "arrow.core.Some", "arrow.core.Option"
    )
```

Assumes `getOrNull` is added to `Option<A>` similar to we did for `Either`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -767,10 +779,18 @@ public sealed class Option<out A> {
       is None -> null
     }
 
+  @Deprecated(
+    NicheAPI + "Prefer when or fold instead",
+    ReplaceWith("MB.run { this.fold({ empty() }) {a -> empty().combine(f(a)) } }")
```
nomisRev, line 784:
```suggestion
    ReplaceWith("MB.run { this.fold({ empty() }) { a -> empty().combine(f(a)) } }")
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -795,12 +815,20 @@ public sealed class Option<out A> {
       )
     }
 
+  @Deprecated(
+    NicheAPI + "Prefer when or fold instead",
+    ReplaceWith("fold({ null }) { value -> operation(initial(value), value) }")
```
nomisRev, line 820:
```suggestion
    ReplaceWith("getOrNull()?.let { value -> operation(initial(value), value)  }")
```

Not sure if this is simpler 🤔

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -679,18 +679,30 @@ public sealed class Option<out A> {
...
+  )
   public inline fun <B> crosswalk(f: (A) -> Option<B>): Option<Option<B>> =
     when (this) {
       is None -> this
       is Some -> f(value).map { Some(it) }
     }
 
+  @Deprecated(
+    NicheAPI + "Prefer using the Option DSL or fold",
+    ReplaceWith("fold<Map<K, Option<V>>>({ emptyMap() }) { value -> f(value).mapValues { Some(it.value) } }")
```
nomisRev, line 694:
```suggestion
    ReplaceWith(
      "fold<Map<K, Option<V>>>({ emptyMap() }) { value -> f(value).mapValues { Some(it.value) } }",
      "arrow.core.Some"
    )
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -679,18 +679,30 @@ public sealed class Option<out A> {
   public inline fun all(predicate: (A) -> Boolean): Boolean =
     fold({ true }, predicate)
 
+  @Deprecated(
+    NicheAPI + "Prefer using the Option DSL or fold",
+    ReplaceWith("fold<Option<Option<B>>>({ None }) { value -> f(value).map(::Some) }")
```
nomisRev, line 684:
```suggestion
    ReplaceWith(
      "fold<Option<Option<B>>>({ None }) { value -> f(value).map(::Some) }",
      "arrow.core.None",
      "arrow.core.Some"
    )
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -795,12 +815,20 @@ public sealed class Option<out A> {
...
+  )
   public inline fun <B> reduceOrNull(initial: (A) -> B, operation: (acc: B, A) -> B): B? =
     when (this) {
       is None -> null
       is Some -> operation(initial(value), value)
     }
 
+  @Deprecated(
+    NicheAPI + "Prefer when or fold instead",
+    ReplaceWith("fold({ Eval.now(null) }) { value -> operation(value, Eval.now(initial(value))) }")
```
nomisRev, line 830:
```suggestion
    ReplaceWith(
      "fold({ Eval.now(null) }) { value -> operation(value, Eval.now(initial(value))) }",
      "arrow.core.Eval"
    )
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -813,11 +841,23 @@ public sealed class Option<out A> {
...
   @OptIn(ExperimentalTypeInference::class)
   @OverloadResolutionByLambdaReturnType
   public inline fun <B> traverse(fa: (A) -> Iterable<B>): List<Option<B>> =
     fold({ emptyList() }, { a -> fa(a).map { Some(it) } })
 
+  @Deprecated(
+    NicheAPI + "Prefer using the Option DSL, or explicit fold or when",
+    ReplaceWith(
+      "fold({ Right(None) }) { a -> fa(a).map(::Some) }",
+      "arrow.core.Either.Right")
```
nomisRev, line 859:
```suggestion
      "arrow.core.Either.Right",
      "arrow.core.None",
      "arrow.core.Some"
    )
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -830,6 +870,12 @@ public sealed class Option<out A> {
   public inline fun <AA, B> traverseEither(fa: (A) -> Either<AA, B>): Either<AA, Option<B>> =
     traverse(fa)
 
+  @Deprecated(
+    NicheAPI + "Prefer using the Option DSL, or explicit fold or when",
+    ReplaceWith(
+      "fold({ Valid(None) }) { a -> fa(a).map(::Some) }",
+      "arrow.core.Valid")
```
nomisRev, line 877:
```suggestion
      "arrow.core.Valid",
      "arrow.core.None",
      "arrow.core.Some"
    )
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -985,38 +1031,75 @@ public fun <A> Option<A>.salign(SA: Semigroup<A>, b: Option<A>): Option<A> =
...
 
 @Deprecated("sequenceEither is being renamed to sequence to simplify the Arrow API", ReplaceWith("sequence()", "arrow.core.sequence"))
 public fun <A, B> Option<Either<A, B>>.sequenceEither(): Either<A, Option<B>> =
   sequence()
 
+@Deprecated(
+  NicheAPI + "Prefer using the Option DSL, or explicit fold or when",
+  ReplaceWith(
+    "fold({ Right(None) }) { a -> fa(a).map(::Some) }",
+    "arrow.core.Either.Right")
```
nomisRev, line 1088:
```suggestion
    "arrow.core.Either.Right",
    "arrow.core.None",
    "arrow.core.Some"
  )
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -985,38 +1031,75 @@ public fun <A> Option<A>.salign(SA: Semigroup<A>, b: Option<A>): Option<A> =
...
 
 @Deprecated("sequenceValidated is being renamed to sequence to simplify the Arrow API", ReplaceWith("sequence()", "arrow.core.sequence"))
 public fun <A, B> Option<Validated<A, B>>.sequenceValidated(): Validated<A, Option<B>> =
   sequence()
 
+@Deprecated(
+  NicheAPI + "Prefer using the Option DSL, or explicit fold or when",
+  ReplaceWith(
+    "fold({ Valid(None) }) { a -> fa(a).map(::Some) }",
+    "arrow.core.Valid")
```
nomisRev, line 1101:
```suggestion
    "arrow.core.Valid",
    "arrow.core.None",
    "arrow.core.Some"
  )
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -1033,16 +1116,37 @@ public inline fun <A, B, C> Option<C>.unalign(f: (C) -> Ior<A, B>): Pair<Option<
...
 public fun <A> Option<Iterable<A>>.unite(MA: Monoid<A>): Option<A> =
   map { iterable ->
     iterable.fold(MA)
   }
 
+@Deprecated(
+  NicheAPI + "Prefer using the Option DSL or explicit flatMap",
+  ReplaceWith(
+    "flatMap<B> { either -> either.fold<Option<B>>({ None }, ::Some) }",
+    "arrow.core.Option", "arrow.core.Some"
```
nomisRev, line 1135:
```suggestion
    "arrow.core.Option", "arrow.core.Some", "arrow.core.None"
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -985,38 +1031,75 @@ public fun <A> Option<A>.salign(SA: Semigroup<A>, b: Option<A>): Option<A> =
...
-public fun <A, B> Option<Validated<A, B>>.separateValidated(): Pair<Option<A>, Option<B>> {
-  val asep = flatMap { gab -> gab.fold({ Some(it) }, { None }) }
-  val bsep = flatMap { gab -> gab.fold({ None }, { Some(it) }) }
-  return asep to bsep
-}
+@Deprecated(
+  NicheAPI + "Prefer using the Option DSL, or explicit fold or when",
+  ReplaceWith(
+    "fold({ None to None }) { validated -> validated.fold<Pair<Option<A>, Option<B>>>({ Some(it) to None }, { None to Some(it) }) }",
+    "arrow.core.None", "arrow.core.Some"
```
nomisRev, line 1059:
```suggestion
    "arrow.core.None", "arrow.core.Some", "arrow.core.Option"
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -985,38 +1031,75 @@ public fun <A> Option<A>.salign(SA: Semigroup<A>, b: Option<A>): Option<A> =
...
-public fun <A, B> Option<Either<A, B>>.separateEither(): Pair<Option<A>, Option<B>> {
-  val asep = flatMap { gab -> gab.fold({ Some(it) }, { None }) }
-  val bsep = flatMap { gab -> gab.fold({ None }, { Some(it) }) }
-  return asep to bsep
-}
+@Deprecated(
+  NicheAPI + "Prefer using the Option DSL, or explicit fold or when",
+  ReplaceWith(
+    "fold({ None to None }) { either -> either.fold<Pair<Option<A>, Option<B>>>({ Some(it) to None }, { None to Some(it) }) }",
+    "arrow.core.None", "arrow.core.Some"
```
nomisRev, line 1038:
```suggestion
    "arrow.core.None", "arrow.core.Some", "arrow.core.Option"
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -1033,16 +1116,37 @@ public inline fun <A, B, C> Option<C>.unalign(f: (C) -> Ior<A, B>): Pair<Option<
...
 public fun <A, B> Option<Either<A, B>>.uniteEither(): Option<B> =
   flatMap { either ->
     either.fold({ None }, { b -> Some(b) })
   }
 
+@Deprecated(
+  NicheAPI + "Prefer using the Option DSL or explicit flatMap",
+  ReplaceWith(
+    "flatMap<B> { validated -> validated.fold<Option<B>>({ None }, ::Some) }",
+    "arrow.core.Option", "arrow.core.Some"
```
nomisRev, line 1147:
```suggestion
    "arrow.core.Option", "arrow.core.Some", "arrow.core.None"
```

nomisRev (commented):
Looks good to me so far. I think it was missing some imports in the `ReplaceWith` 🤔

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
franciscodr: `Some(f(value))` wouldn't be correct because `f` returns a nullable value, so we could potentially return `Some(null)` instead of the expected `null`.

franciscodr:
@nomisRev Do we want to deprecate the `zip` method and all its variants with the inline option DSL? So far, I kept the methods and only changed the implementation to use the DSL, but I saw you're deprecating the `zip` method for the `Either` type.

nomisRev:
> Do we want to deprecate the zip method and all its variants with the inline option DSL? 

Yes, we agreed that `fa.zip(fb) { a, b -> transform(a, b) }` doesn't really offer any benefits over `option { transform(fa.bind(), fb.bind()) }`.

The latter is only 4 characters longer than the former, and doesn't suffer from the arity-n problem.

nomisRev:
This decision can still be challenged of course. @myuwono would be great to get your review here when it's ready for review. He's one of the biggest users of `Option` in combination with Spring WebFlux, and can perhaps short-circuit some things if needed.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -1290,32 +1301,31 @@ public fun <A> Option<Option<A>>.flatten(): Option<A> =
...
-  return map(fb).handleError(fe)
+  return map(fb).recover { fe(Unit) }
 }
 
 @Deprecated(
   NicheAPI + "Prefer using the Option DSL or explicit flatMap with orElse",
   ReplaceWith(
-    "flatMap(fb).orElse(fe)",
-    "arrow.core.orElse"
+    "flatMap(fb).recover { fe(Unit).bind() }",
```
nomisRev, line 1319: This can also be replace with `fold({ fe(Unit) }, fb)`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -1290,32 +1301,31 @@ public fun <A> Option<Option<A>>.flatten(): Option<A> =
 @Deprecated(
   NicheAPI + "Prefer using the Option DSL or explicit map with orElse",
   ReplaceWith(
-    "map(fb).orElse { Some(fe(Unit)) }",
-    "arrow.core.Some",
-    "arrow.core.orElse"
+    "map(fb).recover { fe(Unit) }",
```
nomisRev, line 1304: `fold({ fe(Unit) }, fb).some()` 🤔

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
franciscodr: I have been thinking about the original implementation for `redeemWith`, which doesn't match the `fold` behavior. Keeping in mind the current implementation:
```
flatMap(fb).handleErrorWith(fe)
```
If the function `fb` returns `None`, we can still recover and return the result of the function `fe`. However, if we use `fold`, we would return `None`. For example:
```
fun fe() = Some("recovered from None")
fun fb(a: A) = None

Some("hello").flatMap(fb).recover { fe(Unit).bind() } // Some("recovered from None")
Some("hello").fold({ fe(Unit) }, fb) // None
```
`redeem` does behave as the `fold` function, so maybe the original implementation was wrong, and we should fix it. What do you think, @nomisRev?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
franciscodr: Yeah, makes sense to me

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
nomisRev: The original inspiration comes from `MonadError` in Cats, which defines it like this: `fa.redeemWith(fe, fs) <-> fa.attempt.flatMap(_.fold(fe, fs))`. In this case `fe` cannot recover from a failure in `fe`. 

Which is in line with the original implementation (for `Either`):

```kotlin
  when (this) {
    is Left -> fa(this.value)
    is Right -> fb(this.value)
  }
```

However, I think we should not change the behavior of functions that are people are currently using in production. So we should keep the behavior the same for `Option` which was `flatMap(fb).handleErrorWith(fe)`.

This shows more proof that these APIs are a bit confusing, and reducing API surface will simplify Arrow a lot. `Option` and `Either` were not consistent in `redeemWith`.

So I am marking this comment as resolved, and the deprecation for `Option` should stay `flatMap(fb).recover { fe(Unit).bind() }` to remain consistent with current behavior.

franciscodr: Yes, it makes sense to keep the original implementation, even if it doesn't follow the expected behavior. As you've mentioned, people might be using it in production as it is.

serras (approved):
LGTM

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -980,14 +1183,33 @@ public inline fun <A> Option<A>.orElse(alternative: () -> Option<A>): Option<A>
...
 public infix fun <T> Option<T>.or(value: Option<T>): Option<T> = if (isEmpty()) {
   value
 } else {
   this
 }
 
+@Deprecated(
+  RedundantAPI + "Prefer using fromNullable",
+  ReplaceWith(
+    "Option.fromNullable(this)",
```
myuwono, line 1199: Why is this recommended? we use toOption as the primary function for transforming nullable. This isn't consistent with the other recommendation for NonEmptyList where we recommend the extension function on List.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -1029,6 +1255,14 @@ public inline fun <A> Option<A>.ensure(error: () -> Unit, predicate: (A) -> Bool
 /**
  * Returns an Option containing all elements that are instances of specified type parameter [B].
  */
+@Deprecated(
+  NicheAPI + "Prefer using option DSL or flatMap",
```
myuwono, line 1259: This aligns with stdlib for list.filterIsInstance. let's not deprecate this.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
myuwono: I propose we keep this as well. It's a nice extension function which aligns well with the rest of arrow similarly to .some and .none.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -796,10 +925,14 @@ public sealed class Option<out A> {
    *   none.exists { it > 10 }      // Result: false
    * }
    * ```
-   * <!--- KNIT example-option-21.kt -->
+   * <!--- KNIT example-option-24.kt -->
    *
    * @param predicate the predicate to test
    */
+  @Deprecated(
```
myuwono, line 932: Exists also aligns with stdlib for list. I vote for keeping this.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
nomisRev: Right, sorry.

i-walker (approved):

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
nomisRev: Is this an API that is commonly used in your projects @myuwono? It's an API I personally find kind-of strange on `Option`, even on `List` I think it's not commonly used 🤔

I am having a hard time thinking of a good use-case. I hope that `Option<Any>` is an extremely rare case 😁
Otherwise inheritance rules should apply? 🤔

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
nomisRev: It's an API that was deprecated for `Either` after community review on the PR, https://github.com/arrow-kt/arrow/pull/2830.

We can consider keeping it, but it seemed like a rare case. In that case we should revert the deprecation on `Either`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -1009,6 +1231,10 @@ public fun <A> Iterable<Option<A>>.combineAll(MA: Monoid<A>): Option<A> =
 public fun <A> Option<A>.combineAll(MA: Monoid<A>): A =
   getOrElse { MA.empty() }
 
+@Deprecated(
+  RedundantAPI + "Prefer if-else statement inside option DSL, or replace with explicit flatMap",
+  ReplaceWith("this.flatMap { b -> b.takeIf(predicate)?.let(::Some) ?: None.also(error) }")
```
nomisRev, line 1229: A pity we cannot do more complex rewrites... DSL based code looks so much better 😁 

```kotlin
option {
  ensureNotNull(bind().takeIf(predicate))
}.onNone(error)
```

nomisRev: Do we need to add imports here?

```suggestion
  ReplaceWith("this.flatMap { b -> b.takeIf(predicate)?.let(::Some) ?: None.also(error) }", "arrow.core.Some", "arrow.core.None")
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -824,8 +957,12 @@ public sealed class Option<out A> {
...
    * }
    * ```
-   * <!--- KNIT example-option-22.kt -->
+   * <!--- KNIT example-option-25.kt -->
    */
+  @Deprecated(
+    NicheAPI + "Prefer Kotlin nullable syntax instead",
+    ReplaceWith("getOrNull()?.takeIf(predicate)")
+  )
   public inline fun findOrNull(predicate: (A) -> Boolean): A? {
```
nomisRev, line 966: Pretty crazy.. these APIs predate the introduction of `takeIf` into Kotlin Std. Arrow is getting old, happy it's getting a new lack of paint 😍

nomisRev (approved):
Couple of thoughts, and nits. We should do a pass for the contracts in a subsequent PR, I think we might have missed some.

I'm going to approve, we can keep discussion open with @myuwono and make adjustments later on ☺️

For anyone reading this, anything deprecated now can be challenged before 2.x.x release. Or can be re-introduced if needed. Please provide and share use-cases, and production usage ☺️

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
franciscodr: Yep. I think so. Good catch!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -621,19 +680,35 @@ public sealed class Option<out A> {
    */
   public abstract fun isEmpty(): Boolean
```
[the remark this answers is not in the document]
franciscodr: Done. I also removed the `isEmpty` methods from the `Some` and `None` classes once we implemented them in the `Option` base class.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -621,19 +680,35 @@ public sealed class Option<out A> {
...
 
-  public fun isNotEmpty(): Boolean = !isEmpty()
+  @OptIn(ExperimentalContracts::class)
+  public fun isNotEmpty(): Boolean {
+    contract { returns(true) implies (this@Option is Some<A>) }
+    return this@Option is Some<A>
+  }
```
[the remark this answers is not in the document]
franciscodr: Done

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -642,6 +717,23 @@ public sealed class Option<out A> {
...
+   *
+   * fun test() {
+   *   Some(12).getOrNull() shouldBe 12
+   *   None.getOrNull() shouldBe null
+   * }
+   * ```
+   * <!--- KNIT example-option-23.kt -->
+   */
+  public fun getOrNull(): A? = getOrElse { null }
+
```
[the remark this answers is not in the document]
franciscodr: Done

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
myuwono: Similarly to Lists, this is useful for a call that returns a sealed type @nomisRev. For instance if a database call returns `Option<Document>` where document is a sealed class with many subtypes, and we're in are interested in only one of them e.g. `Option<Document.ExternalMetadata>`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
myuwono: Exists for either might not make much sense but for option I'd argue that it actually does. Essentially you can draw parallel of an optional value being a list of at most one element or otherwise none. If this api is removed then one would argue there's a missing functionality to assert that an option has a value with a certain quality. Lists for instance has `list.any { }` and `list.all { }`. If we're deprecating exists simply because it's deprecated in either, I would argue that we should perhaps reintroduce something that has a same function with a different name.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
nomisRev: Right, you need it for `narrow`'ing. So an `Option` alternative to `as?`.
Since this only makes sense for `Option`, lets keep it 👍

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
nomisRev: > If we're deprecating exists simply because it's deprecated in either, I would argue that we should perhaps reintroduce something that has the same function with a different name.

I was actually thinking about this @myuwono, and then apply the same for `Either` but I couldn't think of a better API (or name). 😂 My main thought was that I also wanted something for in the DSL, but couldn't think of anything...

There is now `isLeft()` * `isRight()` with smart-contracts. We could also expose `isRight { }` and `isLeft { }` 🤔
Similarly we could do the same for `Option`, but they're currently named `isEmpty` and `isNotEmpty`.

How would you feel about changing `isEmpty` and `isNotEmpty` to `isNone()`/`isSome()`? And adding/replacing `exists` with `isNone { }` and `isSome { }`? Do you think that would result in a nicer API for `Option`(, and `Either`)?

It does cover more cases, whilst still keeping the API more aligned I think. They're also very in line with `onSome { }`/`onNone { }` and `onLeft { }`/`onRight { }`.

cc\\ @franciscodr

nomisRev: Thanks for your very constructive feedback btw! 🙏 It's really valuable.

nomisRev (changes requested):
Temporarily blocking this PR pending the ongoing conversation with @myuwono.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
franciscodr: I'm also up to rename `isEmpty` and `isNotEmpty` to `isNone` and `isSome`, respectively, to align them with Either API.

Regarding `isNone { }` and `isSome { }`, what would those functions do? Evaluate the block only if the emptiness/non-emptiness predicate holds?

nomisRev: Oh, `isNone { }` doesn't make sense 😅 Something like this.

```kotlin
   public inline fun <A> Option<A>.isSome(predicate: (A) -> Boolean): Boolean {
    contract {
      callsInPlace(predicate, EXACTLY_ONCE)
      returns(true) implies (this@Option is Some<A>)
      returns(false) implies (this@Option is None)
    }
    return this is Some<A> && predicate(value)
  }
```

And for `Either` alongside `isLeft()` and `isRight`:

```kotlin
  public inline fun <A, B> Either<A, B>.isLeft(predicate: (A) -> Boolean): Boolean {
    contract {
      callsInPlace(predicate, EXACTLY_ONCE)
      returns(true) implies (this@Either is Left<A>)
      returns(false) implies (this@Either is Right<B>)
    }
    return this@Either is Left<A> && predicate(value)
  }

  public inline fun <A, B> Either<A, B>.isRight(predicate: (B) -> Boolean): Boolean {
    contract {
      callsInPlace(predicate, EXACTLY_ONCE)
      returns(true) implies (this@Either is Right<B>)
      returns(false) implies (this@Either is Left<A>)
    }
    return this@Either is Right<B> && predicate(value)
  }
```

WDYT @franciscodr @myuwono

myuwono: That's a really great idea!

franciscodr: I like your proposal, @nomisRev. I sometimes had to double-check if `exists` returns `true`  when the option value was equal to `None` or on the contrary. Having a function called `isSome` make it more evident that the function will only return `true` if the option contains a value and that value satisfies the predicate.

arrow-libs/core/arrow-core/api/arrow-core.api
```diff
@@ -867,7 +867,6 @@ public final class arrow/core/NonFatalOrThrowKt {
 
 public final class arrow/core/None : arrow/core/Option {
 	public static final field INSTANCE Larrow/core/None;
-	public fun isEmpty ()Z
```
nomisRev, line 870: @franciscodr sorry for dragging this PR on so long 🙈
We need to revert the change we made to `abstract fun isEmpty` in order to not break binary compatibility. The improvements we made in `isSome`.

franciscodr: No need to apologize, @nomisRev. Thanks for your thorough review of this PR 😄

nomisRev (approved):
Thank you @franciscodr for this awesome PR 🙏 👏 🥳 And thanks @myuwono for the great feedback and brainstorming the new API 🎉

PR #2916, opened 2023-02-03
Backport zipOrAccumulate for Either

nomisRev:


serras (approved):
Even with contracts, awesome!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -1408,6 +1411,476 @@ public sealed class Either<out A, out B> {
...
+
+
+    public inline fun <E, A, B, Z> zipOrAccumulate(
+      combine: (E, E) -> E,
+      a: Either<E, A>,
+      b: Either<E, B>,
+      transform: (A, B) -> Z,
+    ): Either<E, Z> {
+      contract { callsInPlace(transform, InvocationKind.AT_MOST_ONCE) }
+      return zipOrAccumulate(combine, a, b, unit, unit, unit, unit, unit, unit, unit, unit) { a, bb, _, _, _, _, _, _, _, _ ->
```
franciscodr, line 1423: Should we rename the lambda parameter to `aa` as we do with the rest of the parameters? That would avoid some confusion. Same for the rest of the `zipOrAccumulate` functions below.
```suggestion
      return zipOrAccumulate(combine, a, b, unit, unit, unit, unit, unit, unit, unit, unit) { aa, bb, _, _, _, _, _, _, _, _ ->
```

nomisRev: Yes, thank you for catching this @franciscodr 🙏 I updated the code.

franciscodr (approved):
Thanks, @nomisRev!

Issue #2894, opened 2023-01-19
Increase test coverage Arrow Core


nomisRev:
This ticket tracks classes that are lacking in test coverage. All of this code has been stable for a long time, but it's a good way to get familiar with the Arrow codebase.

If you're looking to contribute to Arrow, and want a small piece of work feel free to pick any of the items listed below and add missing tests for the `.kt` file. If you're not sure if a function, or code branch, is missing coverage in a certain file you can run `./gradlew koverHtmlReport` and open the resulting report.

# Arrow Core
  - [ ] [currying.kt](https://github.com/arrow-kt/arrow/blob/main/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/currying.kt) missing tests for almost all functions.
  - [ ] [Either.kt]](https://github.com/arrow-kt/arrow/blob/main/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt) only has 50% coverage.
  - [ ] [Ior.kt]](https://github.com/arrow-kt/arrow/blob/main/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Ior.kt) only has 70% coverage.
  - [ ] [Iterable.kt]](https://github.com/arrow-kt/arrow/blob/main/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt) only has 50% coverage.
  - [ ] [MapKt.kt]](https://github.com/arrow-kt/arrow/blob/main/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt) only has 16% coverage.
  - [ ] [NonEmptyList.kt]](https://github.com/arrow-kt/arrow/blob/main/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt) only has 50% coverage.
  - [ ] [Option.kt]](https://github.com/arrow-kt/arrow/blob/main/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt) only has 75% coverage.
  - [ ] [partials.kt]](https://github.com/arrow-kt/arrow/blob/main/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/partials.kt) only has 2,4% coverage.
  - [ ] Missing tests for utilities TupleX.kt

lgtout:
@raulraja @nomisRev Please assign this to me.  Thanks!

[...]

PR #2917, opened 2023-02-03
Provide test coverage for currying (#2894)

lgtout:
@nomisRev @raulraja 
I'd appreciate feedback on what I've done so far on tests for `currying`, e.g. naming, style, approach etc, so I can make any adjustments before moving on to the bulk of the work required.  
Thank you!

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/CurryingTest.kt
+++ b/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/CurryingTest.kt
@@ -1,0 +1,475 @@
+package arrow.core
+
+import io.kotest.core.spec.style.StringSpec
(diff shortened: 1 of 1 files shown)


serras:
I like the style! If possible, I would use a non-commutative operation instead of addition, since that property means that we cannot detect whether currying is passing arguments “the right way” or not.

lgtout:
Ah, good idea!  Thanks.  So, string concatenation?

serras:
> Ah, good idea! Thanks. So, string concatenation?

Or if you prefer to keep integers, you can use `x + 2 * y + 3 * z` or something similar.

lgtout:
@serras @nomisRev Do you have a preference for integers vs strings or chars?

serras:
Not really; I just mentioned integers so you didn't have to change too much of your current file contents.

PR #2918, opened 2023-02-03
Fix annoying warnings in IntelliJ

serras:
This PR is mostly for those developing Arrow within IntelliJ. If you open any build file, you get quite some errors, which can be easily suppressed. This PR does exactly so.

In addition, calls to `forUseAtConfigurationTime` are removed. This method is [deprecated](https://docs.gradle.org/current/userguide/upgrading_version_7.html), and should just be dropped.

With this, one can open the project in IntelliJ without spurious errors.

### DIFF
--- a/arrow-libs/core/arrow-annotations/build.gradle.kts
+++ b/arrow-libs/core/arrow-annotations/build.gradle.kts
@@ -1,3 +1,5 @@
+@file:Suppress("DSL_SCOPE_VIOLATION")
+
 plugins {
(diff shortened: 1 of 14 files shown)


JavierSegoviaCordoba:
2023 IDEA build fixes the DSL SCOPE issue, we could remove it soon

JavierSegoviaCordoba (approved):

franciscodr (approved):

PR #2919, opened 2023-02-06
`zipOrAccumulate` for `Raise`

serras:
Supersedes #2880, targeting `main` instead of `arrow-2`.

This PR adds *both* `zipOrAccumulate` and `mapOrAccumulate` to `Raise`, since the latter was not available in the 1.x branch.


franciscodr (approved):
Thanks. @serras

i-walker (approved):

nomisRev (changes requested):
Couple of comments, many of the current signatures won't support `suspend` as they're defined now.

Should we expose these APIs until arity-9 like we do for `Either`?

PR #2920, opened 2023-02-06
Apply suggestions for `zipOrAccumulate`

serras:
This PR applies the suggestions by @nomisRev in #2919, which I merged too quickly 😅

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
@@ -797,16 +797,16 @@ public sealed class Either<out A, out B> {
 
   public fun isLeft(): Boolean {
     contract {
-      returns(true) implies (this@Either is Left<*>)
-      returns(false) implies (this@Either is Right<*>)
+      returns(true) implies (this@Either is Left<A>)
+      returns(false) implies (this@Either is Right<B>)
     }
     return this@Either is Left<A>
   }
 
   public fun isRight(): Boolean {
     contract {
-      returns(true) implies (this@Either is Right<*>)
-      returns(false) implies (this@Either is Left<*>)
+      returns(true) implies (this@Either is Right<B>)
+      returns(false) implies (this@Either is Left<A>)
     }
     return this@Either is Right<B>
   }
@@ -970,7 +970,7 @@ public sealed class Either<out A, out B> {
     contract {
       callsInPlace(action, InvocationKind.AT_MOST_ONCE)
     }
-    return also { if (it.isRight()) action(it.value as B) }
+    return also { if (it.isRight()) action(it.value) }
   }
   
   /**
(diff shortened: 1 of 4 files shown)


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -797,16 +797,16 @@ public sealed class Either<out A, out B> {
 
   public fun isLeft(): Boolean {
     contract {
-      returns(true) implies (this@Either is Left<A>)
-      returns(false) implies (this@Either is Right<B>)
+      returns(true) implies (this@Either is Left<*>)
+      returns(false) implies (this@Either is Right<*>)
```
nomisRev, line 801: Why `*` instead of the type arguments? 🤔

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -0,0 +1,570 @@
...
+public inline fun <R, A, B, C> Raise<R>.zipOrAccumulate(
+  semigroup: Semigroup<@UnsafeVariance R>,
+  @BuilderInference action1: Raise<R>.() -> A,
+  @BuilderInference action2: Raise<R>.() -> B,
+  @BuilderInference block: Raise<R>.(A, B) -> C
+): C {
+  contract {
+    callsInPlace(action1, InvocationKind.EXACTLY_ONCE)
+    callsInPlace(action2, InvocationKind.EXACTLY_ONCE)
+  }
```
nomisRev, line 39: Right, now we can also add `contract` 🥳

nomisRev (commented):
I am a bit confused by the unrelated change in `Either.kt`. Did this slip in?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
serras: This is unrelated, indeed, but it turns out that the Kotlin compiler doesn't work well with contract with type arguments. If your types are named *exactly* like the ones in the contract, then everything is fine. But in the code the result has types `Either<R, A>` and `Either<R, B>`, and the compiler was becoming terribly confused. So I "fixed" the contracts in `Either`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
serras: I think 50% of the code starts with `callsInPlace` xD

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
nomisRev: It seems to have broken the smart-cast in `onLeft`, so does that not work anymore? That seems crucial functionality, should we report a bug? 🤔

serras: This PR can be implemented without that change, so I've reverted it

nomisRev (approved):
Thank you @serras 🙏 🙌

franciscodr (approved):

PR #2912, opened 2023-01-30
Backport Raise & error handlers

nomisRev:
# Migration Guide

See two proposal migration guides for `Either` & `Effect`.

### Replace `either { }`
​
 - Find + Replace `arrow.core.continuations.either` -> `arrow.core.raise.either`
 - Find + Replace `arrow.core.continuations.ensureNotNull` -> `arrow.core.raise.ensureNotNull`
 - Find + Replace `arrow.core.computations.either` -> `arrow.core.raise.either`
 - Find + Replace `arrow.core.computations.ensureNotNull` -> `arrow.core.raise.ensureNotNull`
​
### Replace `either.eager { }`
​
 - Find + Replace `arrow.core.continuations.either.eager` -> `arrow.core.raise.either`
   => Might introduce duplicate import for `arrow.core.raise.either`
 - Find + Replace `either.eager {` -> `either {`
​
### Replace `EffectScope`/`EagerEffectScope`
​
 - Find + Replace `arrow.core.continuations.EffectScope` -> `arrow.core.raise.Raise`
 - Find + Replace `arrow.core.continuations.EagerEffectScope` -> `arrow.core.raise.Raise`
 - Find + Replace `arrow.core.continuations.ensureNotNull` -> `arrow.core.raise.ensureNotNull`
​
## Using `Effect`
​
- Find + Replace `arrow.core.continuations.Effect` -> `arrow.core.raise.Effect`
- Find + Replace `arrow.core.continuations.ensureNotNull` -> `arrow.core.raise.ensureNotNull`
​
=> Requires manually adding missing imports for `fold`, error handlers, and all `Effect` methods since they're replaced by extension functions.
​
## Using `EagerEffect`
​
- Find + Replace `arrow.core.continuations.EagerEffect` -> `arrow.core.raise.EagerEffect`
- Find + Replace `arrow.core.continuations.ensureNotNull` -> `arrow.core.raise.ensureNotNull`
​
=> Requires manually adding missing imports for `fold`, error handlers, and all `EagerEffect` methods since they're replaced by extension functions.

PR #2923, opened 2023-02-08
[Backport] MPP No Trace & Improve debugging experience of leaked raise calls

nomisRev:
- In #2884 we added `ShiftLeakedException` to improve the debug experience, this PR applies it to https://github.com/arrow-kt/arrow/pull/2912.
- In #2869 we removed stack traces to improve performance of `shift`, this PR applies it to https://github.com/arrow-kt/arrow/pull/2912.
- Aligns `result` behavior from `arrow.core.continuations` where unexpected `Throwable` is also mapped to `Result.failure`.


franciscodr (approved):
Thanks, @nomisRev

nomisRev:
>  but I guess this is OK since this code was never released out of alpha versions

Yes, it's only important between stable versions.

Issue #2924, opened 2023-02-08
[Arrow Optics] Generated optics code does not escape Kotlin keywords in package names


hrafnthor:
#### Description:

Testing out using 'arrow-optics' in a project of mine, it immediatly caused compilation errors due to package names in the generated code containing unescaped Kotlin keywords from the originator package. 

For instance the following code is from my project
```kotlin
package `is`.hth.projectname

import arrow.optics.optics

@optics
public data class Audio(
    val url: String = "",
    val secureUrl: String = "",
    val type: String = ""
) {
    public companion object
}
```
And it generates the following:

```kotlin
package is.hth.projectname

public inline val is.hth.projectname.Audio.Companion.iso: arrow.optics.Iso<is.hth.projectname.Audio, kotlin.Triple<kotlin.String, kotlin.String, kotlin.String>> inline get() = arrow.optics.Iso(
  get = { audio: is.hth.projectname.Audio -> kotlin.Triple(audio.`url`, audio.`secureUrl`, audio.`type`) },
  reverseGet = { triple: kotlin.Triple<kotlin.String, kotlin.String, kotlin.String> -> is.hth.projectname.Audio(triple.first, triple.second, triple.third) }
)

```

In this case the keyword is being used in the package name to reference a reverse TLD.

A list of TLD's that match Kotlin's reserved keywords:
- as: American Samoa
- by: Belarus
- do: Dominican Republic
- in: India
- is: Iceland
- it: Italy

#### Versions
arrow: 1.1.5
kotlin: 1.8.0
ksp: 1.8.0-1.0.8

nomisRev:
> If my understanding is correct, would a pull request looking at solving this issue be appreciated?

Hey @hrafnthor,
Thanks for opening an issue with details reproducible steps! A PR would absolutely by greatly appreciated 🙏

hrafnthor:
I have staged a PR https://github.com/arrow-kt/arrow/pull/2925 for consideration. As this is my first contribution to the arrow project, all constructive feedback is appreciated as any norms and coding styles within it are currenlty unknown to me.

After going over the [contribution](https://github.com/arrow-kt/arrow/blob/main/CONTRIBUTING.md) document, I noticed that use of `spotless` is required for contribution PRs. However it does not seem that `spotless` is configured for the project. Or perhaps I am missunderstanding?

[...]

PR #2925, opened 2023-02-08
Keyword sanitization for Lens code generation (#2924)

hrafnthor:
Kotlin keyword sanitization added to class path fields used during Lens code generation. 

Only keywords matching potential TLDs are observed.

### DIFF
--- a/arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/internals/processor.kt
+++ b/arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/internals/processor.kt
@@ -80,7 +80,7 @@ internal fun evalAnnotatedPrismElement(
   }
 
 internal val KSDeclaration.qualifiedNameOrSimpleName: String
-  get() = (qualifiedName ?: simpleName).asString()
+  get() = (qualifiedName ?: simpleName).asSanitizedString()
 
 internal fun KSClassDeclaration.sealedSubclassFqNameList(): List<String> =
   getSealedSubclasses().mapNotNull { it.qualifiedName?.asString() }.toList()
@@ -135,10 +135,10 @@ internal fun KSClassDeclaration.getConstructorTypesNames(): List<String> =
 
 internal fun KSType.qualifiedString(): String = when (declaration) {
   is KSTypeParameter -> {
-    val n = declaration.simpleName.asString().sanitizeDelimited()
+    val n = declaration.simpleName.asSanitizedString()
     if (isMarkedNullable) "$n?" else n
   }
-  else -> when (val qname = declaration.qualifiedName?.asString()?.sanitizeDelimited()) {
+  else -> when (val qname = declaration.qualifiedName?.asSanitizedString()) {
     null -> toString()
     else -> {
       val withArgs = when {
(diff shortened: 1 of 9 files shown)


serras:
Thanks for your contribution, @hrafnthor! May I suggest to use the entire set of keywords instead of a few? You can copy the list from [here](https://github.com/kopykat-kt/kopykat/blob/main/utils/kotlin-poet/src/main/kotlin/at/kopyk/poet/KotlinPoetExtensions.kt#L73-L162).

arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/internals/utils.kt
```diff
@@ -6,3 +6,103 @@ package arrow.optics.plugin.internals
...
 fun String.plusIfNotBlank(prefix: String = "", postfix: String = "") =
   if (isNotBlank()) "$prefix${this}$postfix" else this
+
+/**
+ * Sanitizes each delimited section if it matches with Kotlin reserved keywords.
+ */
+fun String.sanitizeDelimited(delimiter: String = ".", separator: String = delimiter) =
+  if (isNotBlank() && contains(delimiter)) {
+    splitToSequence(delimiter).joinToString(separator) { if (kotlinKeywords.contains(it)) "`$it`" else it }
+  } else this
```
serras, line 16: I don't think you need to consider empty and not containing delimiter as special cases, `splitToSequence` + `joinToString` should do the right thing

arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/internals/processor.kt
```diff
@@ -135,10 +135,10 @@ internal fun KSClassDeclaration.getConstructorTypesNames(): List<String> =
 
 internal fun KSType.qualifiedString(): String = when (declaration) {
   is KSTypeParameter -> {
-    val n = declaration.simpleName.asString()
+    val n = declaration.simpleName.asString().sanitizeDelimited()
     if (isMarkedNullable) "$n?" else n
   }
-  else -> when (val qname = declaration.qualifiedName?.asString()) {
+  else -> when (val qname = declaration.qualifiedName?.asString()?.sanitizeDelimited()) {
```
serras, line 141: It might read better if `asStringSanitized` was an extension of `Name` (but if you prefer this way, it's also fine)

serras (approved):
LGTM, just a few style comments

nomisRev (approved):
Thank you @hrafnthor for being so quick to fix your own bug report 🙏 👏 Really awesome!
And congrats on your first contribution to Arrow 🙌

arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/internals/processor.kt
hrafnthor: I agree, it would be more readable as well as limit the scope of the function to relevant parts i.e KSName rather than any String. Will change the extension signature and apply it where applicable.

hrafnthor:
Thank you both @nomisRev and  @serras for the quick feedback!

I've made slight modifications based on it, which does make the changes more focused and readable.

PR #2926, opened 2023-02-09
Update Kotlin/Native targets

serras:
The list of [supported targets](https://kotlinlang.org/docs/native-target-support.html) has been updated, some new ones have been added, and some others are marked as deprecation. Since they won't be removed until 1.9.20, nothing has been removed.

This PR complements https://github.com/arrow-kt/arrow-gradle-config/pull/65 and updated our GitHub actions to test in all available platforms. Note that we are limited by [GitHub runner specs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources), which only has x86_64 hosts.


nomisRev (approved):
Looks good to me! Thanks @serras 🙌 
Something is off with the android test task, but none of the candidates seem like they'd run the tests :/

> Task 'androidNativeX64Test' is ambiguous in root project 'arrow'. Candidates are: 'androidNativeX64TestBinaries', 'androidNativeX64TestKlibrary', 'androidNativeX64TestProcessResources'.

🤯

nomisRev (approved):

i-walker (approved):

JavierSegoviaCordoba (approved):

PR #2927, opened 2023-02-09
Introduce isLeft & isRight predicate API for Either

nomisRev:
In an interesting https://github.com/arrow-kt/arrow/pull/2913#discussion_r1097879312 between myself, @myuwono and @franciscodr in [Option API deprecation](https://github.com/arrow-kt/arrow/pull/2913) we came up with a set of new operators to replace `exists` in a new meaningful API that we thought was very Kotlin idiomatic.

This PR adds the relevant APIs for `Either`.


franciscodr (approved):
Thanks, @nomisRev!

i-walker (approved):
🙌🏾 Nice

serras:
I really like this API design :)

myuwono (approved):
Thank you @nomisRev this is really awesome!

Issue #2932, opened 2023-02-15
filterOrElse changes behavior for null values in 1.1.4


bwanner:
`filterOrElse` always returns a left value if the right value is null. This is a breaking change from behavior in 1.1.3.

## Example
```
// filterOrOther works as expected
val resultOther = null.right().filterOrOther({ true }) { "Error" }
println(resultOther) // Prints Either.Right(null)

// Use of filterOrElse
val resultElse = null.right().filterOrElse({ true }) { "Error" }
println(resultElse) // Prints Either.Left(Error)
```

The regression was introduced in #2830 by migrating `filterOrElse` to ensure in this [Commit](https://sourcegraph.com/github.com/arrow-kt/arrow@cf0add9aa0e0418e4ec928c65286f22aa9d3700e/-/blob/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt?L1484)

dberg:
@bwanner, thanks for raising the issue.

serras:
> TODO: Fix this for `Option`.

I've doubled checked the current `Option`, and there's no function which may suffer from a similar problem.
(I did so by looking at the code, and also by grepping `?.` to check any possible issue with nullability handling).

PR #2933, opened 2023-02-15
Fix filterOrElse to handle null values correctly.

bwanner:
PR for issue #2932

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
@@ -2069,10 +2069,10 @@ public inline fun <A, B> Either<A, B>.getOrHandle(default: (A) -> B): B =
  */
 @Deprecated(
   RedundantAPI + "Prefer if-else statement inside either DSL, or replace with explicit flatMap",
-  ReplaceWith("flatMap { b -> b.takeIf(predicate)?.right() ?: default().left() }")
+  ReplaceWith("flatMap { if (predicate(it)) Right(it) else Left(default(it)) }")
 )
 public inline fun <A, B> Either<A, B>.filterOrElse(predicate: (B) -> Boolean, default: () -> A): Either<A, B> =
-  ensure(default, predicate)
+  flatMap { if (predicate(it)) Right(it) else Left(default()) }
 
 /**
  * Returns [Right] with the existing value of [Right] if this is a [Right] and the given


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -2069,10 +2069,10 @@ public inline fun <A, B> Either<A, B>.getOrHandle(default: (A) -> B): B =
  */
 @Deprecated(
   RedundantAPI + "Prefer if-else statement inside either DSL, or replace with explicit flatMap",
-  ReplaceWith("flatMap { b -> b.takeIf(predicate)?.right() ?: default().left() }")
+  ReplaceWith("flatMap { if (predicate(it)) Right(it) else Left(default(it)) }")
 )
 public inline fun <A, B> Either<A, B>.filterOrElse(predicate: (B) -> Boolean, default: () -> A): Either<A, B> =
-  ensure(default, predicate)
+  flatMap { if (predicate(it)) Right(it) else Left(default()) }
```
nomisRev, line 2075: @bwanner, @dberg, you don't mind the deprecation? If you came across this, I am assuming you were relying on it in your codebase?

nomisRev (approved):
Good catch 🙏 Thank you for raising the issue, and so quickly following up with a PR for a fix! We should keep this issue open since this mistake is likely also made in `Option` and potentially `Ior`. Luckily this is not released yet.

i-walker (approved):

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
dberg: Hi @nomisRev, yes we were using it but we've replaced it with a similar impl that has a specialized left value that we use to represent errors. And nope, tests and @bwanner caught the breaking change before any deploys were made.

bwanner: @nomisRev - on a related note we defined `rightIfNotNull`, `rightIfNull`, `leftIfNull` in the app itself since we found it useful for the expressiveness / readability.

serras (approved):

nomisRev:
Going to merge the PR, thanks for your contribution and quick action @bwanner 🙏
For the benefit of Arrow, and open an open discussion I wanted to try and gather some feedback from you both if you don't mind 😁

> on a related note we defined rightIfNotNull, rightIfNull, leftIfNull in the app itself since we found it useful for the expressiveness / readability.

We're always open to keep them around in Arrow if they're being used often. It's the second time I hear about `leftItNotNull`, I would personally be more in favour of keeping it aligned with DSL and also recommend the DSL usage `ensureNotNull` since it comes with all Kotlin necessities like _smart-casting_.  Note that from `1.2.x` (2.x.x API, and auto-migrateable) `either` will not be split in `eager` vs `suspend` and comes as a cost-free abstraction.

```kotlin
either { // In the next minor version (with automatic migration) you won't need `suspend/eager` anymore.
  // rightIfNull
  ensure(this == null) { error() }
  
  // rightIfNotNull
  ensureNotNull(this) { error() } // this is smart-casted to non-null after this point.
  
  // leftIfNull
  val b = either.bind()
  ensureNotNull(b) error() } // b is smart-casted to non-null after this point.
}
```

If find `rightIfNull` a bit of a weird API, curious what you're using ti for. Treating `null` as a non error case, and a value is considered an error 🤔

To keep this more in line with the DSL API we could add following APIs:
  - `either.leftfNull { }` -> `either.ensureNotNull { }` 
  - `a.rightIfNotNull { }` -> `Either.ensureNotNull(a) { }`
  - `a.rightIfNull { }` -> `Either.ensure(a == null) {  }`

At least in this case we could add the same APIs for `Option`, and stay in line with the same API naming. Albeit I think the DSL is still superior.

bwanner:
>  I would personally be more in favour of keeping it aligned with DSL and also recommend the DSL usage ensureNotNull since it comes with all Kotlin necessities like smart-casting. Note that from 1.2.x (2.x.x API, and auto-migrateable) either will not be split in eager vs suspend and comes as a cost-free abstraction.

@nomisRev I agree that just `rightIf*Null` functions should be sufficient. Sometimes we have one line functions where we don't set up the `either` context, so we can't directly use ensure:

```
data class Duck(val name: String)

suspend fun fetchDuck(): Either<String, Duck> = fetchNullableDuck().rightIfNotNull { "Error" }

suspend fun ensureAllowedToFeedDuck(): Either<String, Unit> = fetchFeedDuckDenialReason().rightIfNull { "Not allowed to feed duck :-(" }
```

I like the idea of adding `ensure*` functions under `Either` directly.

PR #2935, opened 2023-02-17
Deprecation of `Semigroup` and `Monoid`

serras:
- [X] ~~`Const`~~ the whole type is being deprecated
- [x] `Ior`
- [X] `Iterable`
- [x] `NonEmptyList`
- [x] `Option`
- [x] `Pair`
- [X] `Sequence`
- [x] `map`
- [x] `predef`
- [X] ~~`Validated`~~ the whole type is being deprecated
- [X] ~~continuations/*`~~ the whole thing is being deprecated
- [x] `raise/RaiseAccumulate`
- [X] `raise/Builders`
- [X] ~~`fx/ParTraverseValidated`~~ the whole file is being deprecated


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -43,17 +44,21 @@ public inline fun <A> option(block: OptionRaise.() -> A): Option<A> {
...
+    { block(IorRaise(combineError, state, this)) },
     { e -> throw e },
     { e -> Ior.Left(state.get().getOrElse { e }) },
     { a -> state.get().fold({ Ior.Right(a) }, { Ior.Both(it, a) }) }
   )
 }
 
+@Deprecated(SemigroupDeprecation, ReplaceWith("semigroup.run { ior({ x, y -> x + y}, block) }"))
+public inline fun <E, A> ior(semigroup: Semigroup<E>, @BuilderInference block: IorRaise<E>.() -> A): Ior<E, A> =
+  semigroup.run { ior({ x, y -> x + y}, block) }
```
nomisRev, line 60: This doesn't have to be deprecated since it's not yet in the public binary.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -1395,11 +1401,15 @@ public inline fun <A, B> Option<A>.redeemWith(fe: (Unit) -> Option<B>, fb: (A) -
...
   NicheAPI + "Prefer using the Option DSL or map",
-  ReplaceWith("MA.run { this.map { List(n) { it }.fold(empty()) { acc, v -> acc + v } } }")
+  ReplaceWith("this.map { List(n) { it }.fold(empty(), combine) }")
 )
+public fun <A> Option<A>.replicate(n: Int, initial: A, combine: (A, A) -> A): Option<A> =
+  if (n <= 0) Some(initial)
+  else map { a -> List(n) { a }.fold(initial, combine) }
+
+@Deprecated(MonoidDeprecation, ReplaceWith("MA.run { replicate(n, empty()) { acc, v -> acc + v } }"))
 public fun <A> Option<A>.replicate(n: Int, MA: Monoid<A>): Option<A> = MA.run {
```
nomisRev, line 1411: Not sure the benefit is of adding this intermediate methods 🤔

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Pair.kt
```diff
@@ -13,8 +14,12 @@ public operator fun <A : Comparable<A>, B : Comparable<B>> Pair<A, B>.compareTo(
   else first
 }
 
-public fun <A, B> Pair<A, B>.combine(SA: Semigroup<A>, SB: Semigroup<B>, b: Pair<A, B>): Pair<A, B> {
+public fun <A, B> Pair<A, B>.combine(combineA: (A, A) -> A, combineB: (B, B) -> B, b: Pair<A, B>): Pair<A, B> {
   val (xa, xb) = this
   val (ya, yb) = b
-  return Pair(SA.run { xa.combine(ya) }, SB.run { xb.combine(yb) })
+  return Pair(combineA(xa, ya), combineB(xb, yb))
```
nomisRev, line 20: Should we deprecate this method too? Not sure what type class this originally came from 🤔
Does it bring extra value to Arrow? The resulting Kotlin code seems almost simpler than the method:

```kotlin
val pairA = Pair(xa, xb)
val pairB =  Pair(xb, yb)
pairB.combine(combineA, combineB, pairB)
```

vs

```kotlin
val (a, b) = Pair(xa, xb)
val (c, d) =  Pair(xb, yb)
Pair(combineA(a, b), combineB(c, d))
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
```diff
@@ -722,7 +731,7 @@ public fun <A, B> Sequence<A>.traverseOption(f: (A) -> Option<B>): Option<Sequen
 @OptIn(ExperimentalTypeInference::class)
 @OverloadResolutionByLambdaReturnType
 public fun <E, A, B> Sequence<A>.traverse(
-  semigroup: Semigroup<E>,
+  combine: (E, E) -> E,
   f: (A) -> Validated<E, B>
 ): Validated<E, List<B>> =
```
nomisRev, line 736: Another missed variant of `mapOrAccumulate`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Validated.kt
```diff
@@ -352,28 +353,32 @@ public sealed class Validated<out E, out A> {
...
+      is Invalid -> Invalid(combineError(this.value, fb.value))
+      is Valid -> Invalid(this.value)
+    }
+    is Valid -> when (fb) {
+      is Invalid -> Invalid(fb.value)
+      is Valid -> Valid(f(this.value, fb.value))
+    }
+  }
+
+@Deprecated(SemigroupDeprecation, ReplaceWith("SE.run { zip({ x, y -> x + y }, fb, ::Pair) }"))
```
nomisRev, line 371: I've deprecated both `Semigroup` and `zip` at the same time in the `Validated` PR with `    "Either.zipOrAccumulate({ a, b -> SE.run<Semigroup<E>, E> { a.combine(b) } }, toEither(), fb.toEither(), ::Pair).toValidated()"`. So we probably shouldn't add these new methods.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/map.kt
```diff
@@ -267,7 +267,7 @@ public inline fun <K, E, A, B> Map<K, A>.traverseValidated(
   traverse(semigroup, f)
 
 public inline fun <K, E, A, B> Map<K, A>.traverse(
-  semigroup: Semigroup<E>,
+  combine: (E, E) -> E,
   f: (A) -> Validated<E, B>
 ): Validated<E, Map<K, B>> =
```
nomisRev, line 272: One more `mapOrAccumulate`?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/predef.kt
```diff
@@ -24,5 +25,13 @@ internal object EmptyValue {
...
 internal fun <T> Semigroup<T>.emptyCombine(first: Any?, second: T): T =
-  if (first == EmptyValue) second else (first as T).combine(second)
+  emptyCombine(first, second) { x, y -> x.combine(y) }
+
+/**
+ * Apply the [combine] function if [first] is not empty, otherwise return [second].
+ */
+@PublishedApi
+internal inline fun <T> emptyCombine(first: Any?, second: T, combine: (T, T) -> T): T =
+  if (first == EmptyValue) second else combine(first as T, second)
```
nomisRev, line 37: This is already defined as `EmptyValue.combine` in `predef.kt`

nomisRev (changes requested):
Thank you so much @serras for taking on this work! 🙏 🙌
A couple of remarks, and suggestions. Mostly about two things:

1. I think we missed a bunch of `mapOrAccumulate` implementation where now new methods for `Validated` have been introduced. Since `mapOrAccumulate` is defined over `Raise<E>.(A) -> B` it's compatible with both `Validated` and `Either` after users refactor away from `Validated`.

2. A couple of APIs I think we should deprecate all together with the removal of `Semigroup` rather than introduce new methods that were previously derived from `Monad + Semigroup/Monoid` or `Applicative + Semigroup/Monoid`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -758,17 +773,27 @@ public fun <A, B> Iterable<A>.align(b: Iterable<B>): List<Ior<A, B>> =
...
     it.fold(::identity, ::identity) { a, b ->
-      a.combine(b)
+      combine(a, b)
     }
   }
+
+/**
+ * aligns two structures and combine them with the given [Semigroup.combine]
+ */
+@Deprecated(SemigroupDeprecation, ReplaceWith("SG.run { salign({ a, b -> a.combine(b) }, other) }"))
```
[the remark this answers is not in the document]
serras: I just tried to follow the pattern of having the `run` method always outside. But I'm happy to change it if it may bring problems.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Validated.kt
serras: Removed in the merge commit

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
serras: These were just added for consistency: since they were available with `Semigroup`, I added them with the combination functions.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Pair.kt
serras: I don't think it hurts having it there, although I agree it may not see much use...

serras:
There's one missing point for this PR, which is what should be done with the tests using `MonoidLaws` and `SemigroupLaws`. I have no good answer, the best I can think of is moving `Monoid` and `Semigroup` to `test` once 1.2.0 is out.

nomisRev:
>  MonoidLaws and SemigroupLaws.

Does it still make sense if we remove the `typeclass`? Otherwise I would say follow the same pattern as we do here. Replace with `(E, E) -> E` and `(E, (E, E) -> E)` and deprecate old ones.

That way you can still test `String::plus` and `"", String::plus` if you want to.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -758,17 +773,27 @@ public fun <A, B> Iterable<A>.align(b: Iterable<B>): List<Ior<A, B>> =
...
+public fun <A> Iterable<A>.salign(
```
i-walker, line 792: can we make this internal or this exists to not break binaries?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Pair.kt
i-walker: Also Agree the Kotlin code looks simpler

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
nomisRev: This is the currently released API, which we're deprecating in favour of the new `(A, A) -> A` API.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Pair.kt
nomisRev: > I don't think it hurts having it there, although I agree it may not see much use...

I decided to deprecate this method instead of adding a new signature. I'm not even sure where it originally came from 🤔 `BiFunctor` and `Semigroup`?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
nomisRev: We've been deprecating `replicate` in all APIs, so I've also deprecated this one and removed the new method. I added the `NichAPI` deprecation message as well.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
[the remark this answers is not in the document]
nomisRev: I've added a `SG::combine` extension method to improve most `ReplaceWith`, and moved `run { }` inside where needed. Having it outside results in issues with `ReplaceWith` since inside `run` the `this` context changes from `Iterable<A>` to `Semigroup` and thus `salign` is called on the wrong receiver.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Ior.kt
```diff
@@ -170,10 +172,11 @@ public sealed class Ior<out A, out B> {
     return fold({ c }, { f(c, it) }, { _, b -> f(c, b) })
   }
 
+  @Deprecated(MonoidDeprecation, ReplaceWith("MN.run { foldLeft(MN.empty()) { b, a -> b + f(a) } }"))
```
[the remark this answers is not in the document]
nomisRev: Reverted all `Ior` changes in favour of @gutiory's PR.

nomisRev:
@serras I updated the PR, I think my changes need a thorough reverse review from you 😅

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -840,21 +840,17 @@ public fun <A> Iterable<A>.salign(
...
-      combine(a, b)
-    }
+    it.fold(::identity, ::identity, combine)
   }
 
 /**
  * aligns two structures and combine them with the given [Semigroup.combine]
  */
-@Deprecated(SemigroupDeprecation, ReplaceWith("salign({ a, b -> SG.run { a.combine(b) } }, other)"))
+@Deprecated(SemigroupDeprecation, ReplaceWith("salign(SG::combine, other)"))
```
nomisRev, line 849:
```suggestion
@Deprecated(SemigroupDeprecation, ReplaceWith("salign(SG::combine, other)", "arrow.typeclasses.combine"))
```

nomisRev (approved):
Looks great. Awesome work @serras 🙏 👏 🙌

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -2292,19 +2292,43 @@ public operator fun <A : Comparable<A>, B : Comparable<B>> Either<A, B>.compareT
...
+  }
+
 @Deprecated(
-  RedundantAPI + "Prefer zipOrAccumulate",
-  ReplaceWith("Either.zipOrAccumulate({ a, bb -> SGA.run { a.combine(bb) }  }, this, b) { a, bb -> SGB.run { a.combine(bb) } }")
+  SemigroupDeprecation,
+  ReplaceWith(
+    "Either.zipOrAccumulate(SGA::combine, this, b, SGB::combine)",
+    "arrow.typeclasses.combine"
+  )
```
gutiory, line 2318: **thought:** 

<sub>This comment follows the [**conventionalcomments.org**](https://conventionalcomments.org/) standard</sub>
```suggestion
  SemigroupDeprecation,
  ReplaceWith("this.combine(b, {(a:A, a:A) -> A}, {(b:B, b:B) -> B})")
```

Maybe changing the deprecation expression, we can help a bit more the user with the migrations. It does not compile, but the current expression does not either.

gutiory (approved):
LGTM! 
I've just left a thought, but it's just a thought related to replacements to depreacation methods. Feel free to use it or not 😄 .

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
serras: I don't fully understand: the `ReplaceWith` expression is a copy of the implementation. I agree that it may not compile in every case, but I think this gets the developer closer to the expected changes than otherwise.

PR #2937, opened 2023-02-21
Fix Either.isLeft, Either.isRight & Option.isSome contract

### LINKED ISSUE #2936: Contract statement is incorrect in Either.isRight(predicate: ...)
In the new `isRight(predicate: ...)` fun of type `Either`, the assumptions stated in the contract is not accurate.

False does *not* correctly imply the element is of type `Left<A>`, because it could just be a `Right<A>` with not matching predicate...
    
https://github.com/arrow-kt/arrow/blob/b3ef809e6543c7ef7a0558666ef88db3497ca389/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt#L864-L871

nomisRev:
Fixes #2936 reported by @Zordid for both `Either` and `Option`.


franciscodr (approved):
Thanks, @nomisRev!

gutiory (approved):

myuwono (approved):

serras (approved):

PR #2941, opened 2023-02-22
Add exist fun to Either, for migration from Validated

### LINKED ISSUE #2938: Deprecated fun `exists` in Either was named `exist` in Validated which makes migration harder
This function was previously known as `exist`, not `exists` - so coming from 1.1.5 to 1.1.6 or 1.2.0 breaks with unresolved reference if you exchange `Validated` with `Either` types.
Maybe it should not be added to mitigate transition to `Either`.

https://github.com/arrow-kt/arrow/blob/b3ef809e6543c7ef7a0558666ef88db3497ca389/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt#L1085-L1091

Zordid:
Fixes #2938 by introducing a deprecated new `exist` fun for `Either`, just to allow migration coming from `Validated` which had this method instead of the `exists` in `Option` and `Either`.


nomisRev:
Missing a `./gradlew apiDump` commit.

nomisRev (approved):
Thank you @Zordid! And thank you for test driving the migrations, and alpha versions 🙌

serras (approved):

PR #2944, opened 2023-02-23
Add DSL markers to all bind funs in Raise

Zordid:
Found that all bind() funs in Raise.kt where missing the corresponding DslMarker annotation. They did not get colored in IntelliJ and that is extremely visible when combining resource scopes and effect scopes.


nomisRev (approved):
Thanks for the fix @Zordid 🙌 And your interests, and contributions to Arrow 😍

serras (approved):

PR #2946, opened 2023-02-23
Tracing POC

nomisRev:
Discussing some ideas for new features with @raulraja he mentioned tracing for `Effect`, so I was playing around with some ideas and come up with a small POC.

Since `raise` already relies on `CancellationException` to co-operate with the coroutine system we currently pay a performance penalty for `raise` and the stack trace creation that comes with `CancellationException` by default. This penalty is not needed, and is stack trace creation is [by default also disabled in KotlinX Coroutines](https://github.com/Kotlin/kotlinx.coroutines/blob/d88a8c227a295f91251960c8fe6b21a02ddf704b/kotlinx-coroutines-core/jvm/src/Exceptions.kt#L40). 

This PR solves the issue of _debugging_/_tracing_, and the performance penalty we currently pay for _raise_. 
We can now disable the stack trace creation, and we can ad-hoc enable it for _any_ `Effect` you compose _traced_ upon or any `Raise` program you run within a `traced` block.

```kotlin
public fun main() {
  val error = effect<String, Int> { raise("error") }
  error.traced { traced -> traced.printStackTrace() }
    .fold({ require(it == "error") }, { error("impossible") })
}
```
```text
arrow.core.continuations.RaiseCancellationException: Raised Continuation
  at arrow.core.continuations.DefaultRaise.raise(Fold.kt:77)
  at MainKtKt$main$error$1.invoke(MainKt.kt:6)
  at MainKtKt$main$error$1.invoke(MainKt.kt:6)
  at arrow.core.continuations.Raise$DefaultImpls.bind(Shift.kt:22)
  at arrow.core.continuations.DefaultRaise.bind(Fold.kt:74)
  at arrow.core.continuations.Effect__TracingKt$traced$2.invoke(Tracing.kt:46)
  at arrow.core.continuations.Effect__TracingKt$traced$2.invoke(Tracing.kt:46)
  at arrow.core.continuations.Effect__FoldKt.fold(Fold.kt:92)
  at arrow.core.continuations.Effect.fold(Unknown Source)
  at MainKtKt.main(MainKt.kt:8)
  at MainKtKt.main(MainKt.kt)
```

# Notes

- Since the `Effect` runtime is shared amongst DSL such as `either`, `result`, `option`, `ior`, etc the `traced` operator can be added, or used for all these DSLs, and types. Including any of the DSL methods that result in shifting such as `ensure`, `ensureNotNull`, `bind`, etc.

- Currently only JVM is disabling stack-trace generation. Not sure if we also need to disable it for JS, or Native. (KotlinX Coroutines only disables it for JVM)

- What API would we want on `Traced<R>`? Currently we include mentions to `ShiftCancellationException` and the dynamic call to `DefaultShift#shift` which is not part of the user stack trace. Should we slightly change the API such that this is improved?


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
```diff
@@ -108,25 +108,90 @@ public inline fun <R, A, B> fold(
...
+ * This implies a performance penalty of creating a stacktrace when calling [Raise.raise],
+ * but **this only occurs** when composing `traced`.
+ * The stacktrace creation is disabled if no `traced` calls are made within the function composition.
+ */
+@ExperimentalTraceApi
+public inline fun <R, A> Raise<R>.traced(
+  @BuilderInference program: Raise<R>.() -> A,
+  trace: (traced: Traced<R>) -> Unit
+): A {
+  val itOuterTraced = this is DefaultRaise && isTraced
```
kyay10, line 155: Probably a typo intending `isOuterTraced`

nomisRev:
Thanks for your review and feedback @kyay10 🙏 

> Is there a reason to distinguish between RaiseCancellationExceptionNoTrace and RaiseCancellationException on a type level? it seems to me that a val isTraced might be sufficient, although I see that this might be an unnecessary overhead.

Okay, so this was an attempt to hide `RaiseCancellationExceptionXXX` from the public API, but in turn `CancellationExceptionNoTrace` is public so not sure if that's really better 🤔 Sadly `expect/actual` cannot be private 😞

> Also, is there a reason as to why Traced needs to exist as a class? It seems sufficient to just pass the needed information to the lambda, and Traced seems to just be delegating its methods to its properties anyways.

The reason for `Traced` was to have a bit more flexibility. Having `Trace` as an `@JvmInline value class` is perhaps better. The reason I modelled it like this is because I wanted to encapsulate the `Throwable` that gives more flexibility for the future.


I would've really liked to sanitise the trace, but I'm not sure if there is a decent way to do this for Native / JS. We could do it on minimum-effort for JVM only, it seems that is also what KotlinX Coroutines is doing.

I've played around with the changes, and made a commit applying your suggestions @kyay10. WDYT? To my surprise `RaiseCancellationException` didn't become public in the _api_ file.

kyay10:
LGTM, perhaps just fix the (very minor) typo I commented above. 

Should there also be changes to all the comprehension blocks (is that what we still call them btw?) to add a `traced` block for each e.g. `option.traced`? Or is that not useful?

nomisRev:
> LGTM, perhaps just fix the (very minor) typo I commented above.

Oh yes, missed that in the PR 👍 

> Should there also be changes to all the comprehension blocks (is that what we still call them btw?) to add a traced block for each e.g. option.traced? Or is that not useful?

This shouldn't be necessary you can use `traced({ option.bind() }) { }` and get a trace from it since `bind` short-circuit through `raise`. The biggest issue, or downside, is that it traces `Raise` and has no way of figuring out where the `Option` was created. So compared to regular exceptions the trace will point to the `throws` call and not the `Throwable#init`.

nomisRev:
@kyay10 actually I think this change is impossible without having two types 🤔 Since `fillInStackTrace` is called in the `init` method of `Throwable` it will always be uninitialised, hence `false`. There is no way to influence this so it requires two separate classes.

rafaparadela:
Task linked: [CU-865byujb8 Tracing PoC](https://app.clickup.com/t/865byujb8)

nomisRev:
> Ah that's peculiar!

It's standard behavior for the JVM. Initialisation order of https://stackoverflow.com/questions/32490714/order-of-initialization-instantiation-of-class-variables-of-derived-class-and-in.

>  It's a deep implementation detail that likely won't be modified often, so it might be okay to ignore DRY.

I think it's fine, especially in a library like Arrow. Optimise for the user facing API > optimise for less code in library.

kyay10:
Maybe Trace could be made a `value class`? I think the other commit (the one that failed the tests) had it as such.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
```diff
@@ -108,27 +108,92 @@ public inline fun <R, A, B> fold(
...
+
 /** Returns the raised value, rethrows the CancellationException if not our scope */
 @PublishedApi
 @Suppress("UNCHECKED_CAST")
 internal fun <R> CancellationException.raisedOrRethrow(raise: DefaultRaise): R =
-  if (this is RaiseCancellationException && this.raise === raise) raised as R
-  else throw this
+  when {
+    this is RaiseCancellationExceptionNoTrace && this.raise === raise -> raised as R
+    this is RaiseCancellationException && this.raise === raise -> raised as R
```
serras, line 172:
```suggestion
    this.raise === raise && (this is RaiseCancellationExceptionNoTrace || this is RaiseCancellationException) -> raised as R
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
```diff
@@ -108,27 +108,92 @@ public inline fun <R, A, B> fold(
...
+internal class DefaultRaise(@PublishedApi internal val isTraced: Boolean) : Raise<Any?> {
   private val isActive = AtomicBoolean(true)
+
   @PublishedApi
   internal fun complete(): Boolean = isActive.getAndSet(false)
-  override fun raise(r: Any?): Nothing =
-    if (isActive.value) throw RaiseCancellationException(r, this) else throw RaiseLeakedException()
+  override fun raise(r: Any?): Nothing = when {
+    isActive.value && !isTraced -> throw RaiseCancellationExceptionNoTrace(r, this)
+    isActive.value && isTraced -> throw RaiseCancellationException(r, this)
```
serras, line 185:
```suggestion
    isActive.value -> throw if (isTraced) RaiseCancellationException(r, this) else RaiseCancellationExceptionNoTrace(r, this)
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
serras:
```suggestion
  val isOuterTraced = this is DefaultRaise && isTraced
```

serras (approved):
Looks good to me :)

raulraja (approved):
❤️

nomisRev:
Thanks for the great suggestions and review @serras 🙏

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
franciscodr: This change introduces a compilation error because `raise` and `raised` properties don't exist in `CancellationException`. Those properties belong to `RaiseCancellationExceptionNoTrace` and `RaiseCancellationException`, so we need the smart cast here.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Trace.kt
```diff
@@ -0,0 +1,41 @@
...
+import kotlin.coroutines.cancellation.CancellationException
+import kotlin.jvm.JvmMultifileClass
+import kotlin.jvm.JvmName
+
+@RequiresOptIn("This API is experimental, and may change in the future.")
+public annotation class ExperimentalTraceApi
+
+/** Tracing result of `R`. Allows to inspect `R`, and the traces from where it was raised. */
+@ExperimentalTraceApi
+public class Trace(private val exception: CancellationException) {
```
kyay10, line 14:
```suggestion
@JvmInline value class Trace(private val exception: CancellationException) {
```

nomisRev: Good catch @kyay10 thank you!

nomisRev:
Thanks for fixing the build @franciscodr

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Trace.kt
```diff
@@ -0,0 +1,41 @@
...
+import kotlin.coroutines.cancellation.CancellationException
+import kotlin.jvm.JvmMultifileClass
+import kotlin.jvm.JvmName
+
+@RequiresOptIn("This API is experimental, and may change in the future.")
+public annotation class ExperimentalTraceApi
+
+/** Tracing result of `R`. Allows to inspect `R`, and the traces from where it was raised. */
+@ExperimentalTraceApi
+@JvmInline value class Trace(private val exception: CancellationException) {
```
nomisRev, line 14:
```suggestion
@JvmInline
public value class Trace(private val exception: CancellationException) {
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Trace.kt
```diff
@@ -0,0 +1,42 @@
...
+package arrow.core.raise
+
+import kotlin.coroutines.cancellation.CancellationException
+import kotlin.jvm.JvmMultifileClass
+import kotlin.jvm.JvmName
+
+@RequiresOptIn("This API is experimental, and may change in the future.")
+public annotation class ExperimentalTraceApi
+
+/** Tracing result of `R`. Allows to inspect `R`, and the traces from where it was raised. */
```
kyay10, line 12:
```suggestion
/** Tracing result. Allows to inspect the traces from where raise was called. */
```
Since Trace doesn't include the raised value anymore, the kdoc should reflect that change

arrow-libs/core/arrow-core/api/arrow-core.api
```diff
@@ -3570,10 +3570,19 @@ public final class arrow/core/raise/ResultRaise : arrow/core/raise/Raise {
...
+	public static fun equals-impl (Ljava/util/concurrent/CancellationException;Ljava/lang/Object;)Z
+	public static final fun equals-impl0 (Ljava/util/concurrent/CancellationException;Ljava/util/concurrent/CancellationException;)Z
+	public fun hashCode ()I
+	public static fun hashCode-impl (Ljava/util/concurrent/CancellationException;)I
+	public static final fun printStackTrace-impl (Ljava/util/concurrent/CancellationException;)V
+	public static final fun stackTraceToString-impl (Ljava/util/concurrent/CancellationException;)Ljava/lang/String;
+	public static final fun suppressedExceptions-impl (Ljava/util/concurrent/CancellationException;)Ljava/util/List;
+	public fun toString ()Ljava/lang/String;
+	public static fun toString-impl (Ljava/util/concurrent/CancellationException;)Ljava/lang/String;
+	public final synthetic fun unbox-impl ()Ljava/util/concurrent/CancellationException;
```
nomisRev, line 3585: Wow, binary says BOOM due to `value class` 🤯 Never realised this before

PR #2949, opened 2023-02-24
Option: deprecate or, orElse and unzip

nomisRev:
I propose to deprecate `or`, `orElse` since they overlap with `recover` and `Either` doesn't have this API.

Similarly I propose to deprecate `unzip` since I find it a rather obscure API that is unmatched in `Either`, or does this API make sense for `Option` but not for `Either`?

Additionally, in https://github.com/arrow-kt/arrow/pull/2935/files#r1117382480 I also proposed to deprecate `combine` for removal which overlaps with `option { combine(a.bind(), b.bind()) }`.


raulraja (approved):
👍

myuwono (approved):

serras (approved):

PR #2954, opened 2023-03-01
Simplify recover, and catch APIs. Add getOrElse

### LINKED ISSUE #2953: ["Request"] new function: Effect<R, A>.getOrElse()
Arrow version: 1.1.6-alpha.36

I miss something like the following:

```
suspend inline fun <R, A> Effect<R, A>.getOrElse(noinline onRaise: (R) -> A): A =
    fold(
        recover = onRaise,
        transform = ::identity
    )
```

Sample usage:

```
context(Raise<String>)
suspend fun f1() = 1

context(Raise<String>)
fun f2(): Nothing = raise("error")


runTest {
  val v1 = effect { f1() }.getOrElse { 33 }
  assertEquals(1, v1)

  val v2 = effect { f2() }.getOrElse { 33 }
  assertEquals(33, v2)
}
```

Related discusson: https://kotlinlang.slack.com/archives/C1JMF6UDV/p1677662786876089

nomisRev:
Draft changes related to #2948 to facilitate the discussion. Also closes #2953 

cc\\ @freynder

## Description about changes (copied from comments in #2948)

The lambda `Raise<R>.() -> A` allows to _bind_, or _unwrap_, `Option` and in the case `None` is encountered it will invoke the lambda. The lambda allows to _raise_ a value of `R` into the outer `interface Raise<R>`, or _fallback_ to a value of _A_.

1. The `Raise<R>` lambda receiver is redundant because we're already inside the context of `interface Raise<R>`, so even without the lambda receiver we can call `raise(r)`.
2. Without `Raise<R>`, we have `() -> A` which matches the already existing `Option` API of `getOrElse`.
3. Since this function is defined inside of an `interface`, it cannot be `inline` so it's inferior to the already existing `getOrElse` which is _inline_ and thus allows for `suspend` to pass through.

So this API can be completely replaced by `getOrElse` and simplifies the examples of the KDoc above.

```diff
suspend fun test() {
  val empty: Option<Int> = None
  either {
-    val x: Int = empty.bind { _: None -> 1 }
+    val x: Int = empty.getOrElse { 1 }
-    val y: Int = empty.bind { _: None -> raise("Something bad happened: Boom!") }
+    val y: Int = empty.getOrElse { raise("Something bad happened: Boom!") }
-    val z: Int = empty.recover { _: None ->
+    val z: Int = empty.getOrElse {
      delay(10)
      1
-    }.bind { raise("Something bad happened: Boom!") }
+   }
    x + y + z
  } shouldBe Either.Left("Something bad happened: Boom!")
}
```

As the example shows we can now call `suspend fun delay` directly into the `getOrElse` lambda, and we can also still call `raise` & co from inside the `getOrElse` lambda.

The same applies for `Result#bind`. Since none of this code has yet been released, we can still remove these APIs in favour of the simpler `inline` methods on the original APIs. Fixing this issue also requires some other small changes in [Builders.kt](https://github.com/arrow-kt/arrow/blob/main/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt).

---

There is multiple benefits to this actually, and we should review the current signatures in `arrow.core.raise.*` to see how we can simplify them.

Looking for example at `recover`:

```diff
@RaiseDSL
- public inline fun <R, E, A> Raise<R>.recover(
+ public inline fun <E, A> recover(
  @BuilderInference action: Raise<E>.() -> A,
-  @BuilderInference recover: Raise<R>.(E) -> A,
+  recover: (E) -> A,
): A {
  contract {
    callsInPlace(action, EXACTLY_ONCE)
    callsInPlace(recover, AT_MOST_ONCE)
  }
  return fold<E, A, A>({ action(this) }, { throw it }, { recover(it) }, { it })
}
```

Besides still supporting the current use-cases it also makes this function more general purpose, because you can now use it to execute `Raise<E>` based programs and _resolve_ the error completely without **having to** stick to `Raise<E>`.

---

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Raise.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Raise.kt
@@ -2,16 +2,16 @@
 @file:Suppress("DEPRECATION")
 @file:JvmMultifileClass
 @file:JvmName("RaiseKt")
+
 package arrow.core.raise
 
 import arrow.core.Either
-import arrow.core.None
-import arrow.core.Option
-import arrow.core.Some
 import arrow.core.Validated
 import arrow.core.continuations.EffectScope
 import arrow.core.identity
+import arrow.core.nonFatalOrThrow
 import arrow.core.recover
+import kotlin.coroutines.cancellation.CancellationException
 import kotlin.contracts.ExperimentalContracts
 import kotlin.contracts.InvocationKind.AT_MOST_ONCE
 import kotlin.contracts.InvocationKind.EXACTLY_ONCE
@@ -39,7 +39,7 @@ public annotation class RaiseDSL
  * ```kotlin
  * fun Raise<String>.failure(): Int = raise("failed")
  *
- * fun Raise<Nothing>.recovered(): Int =
+ * fun recovered(): Int =
  *   recover({ failure() }) { _: String -> 1 }
  * ```
  * <!--- KNIT example-raise-dsl-01.kt -->
@@ -96,11 +96,11 @@ public annotation class RaiseDSL
  *
  * fun Raise<String>.failure(): Int = raise("failed")
  *
- * fun Raise<Nothing>.recovered(): Int = recover({ failure() }) { _: String -> 1 }
+ * fun recovered(): Int = recover({ failure() }) { _: String -> 1 }
  * -->
  * ```kotlin
  * fun test() {
- *   val either = either { failure() }
+ *   val either: Either<Nothing, Int> = either { failure() }
  *     .recover { _: String -> recovered() }
  *
  *   either shouldBe Either.Right(1)
@@ -152,6 +152,7 @@ public interface Raise<in R> {
    * @see [recover] if you want to attempt to recover from any _logical failure_.
    */
   public operator fun <A> EagerEffect<R, A>.invoke(): A = invoke(this@Raise)
+
   @RaiseDSL
   public fun <A> EagerEffect<R, A>.bind(): A = invoke(this@Raise)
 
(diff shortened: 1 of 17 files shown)


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Raise.kt
```diff
@@ -331,69 +253,168 @@ public interface Raise<in R> {
...
-  @BuilderInference action: Raise<R>.() -> A,
-  @BuilderInference catch: Raise<R>.(T) -> A,
-): A {
-  contract {
-    callsInPlace(action, EXACTLY_ONCE)
-    callsInPlace(catch, AT_MOST_ONCE)
-  }
-  return catch(action) { t: Throwable -> if (t is T) catch(t) else throw t }
-}
+public inline fun <reified T : Throwable, A> catch(action: () -> A, catch: (T) -> A): A =
```
nomisRev, line 416: This also matches what we used, and needed in [Gh Alerts Project](https://github.com/47deg/gh-alerts-subscriptions-kotlin/blob/858669890b96d9be3d0984e0d2cee7cbcc8110ab/src/main/kotlin/alerts/predef.kt#L23).

Wondering if this should be on `Raise` companion similar to `Either.catch` 🤔 They're basically the duality of each-other. Similarly, does `Either.catch` need a `reified` variant?

nomisRev: > Raise sort of feels like much more of a primitive than Either
It absolutely is!

>  The name Raise.catch also just feels weird from a language standpoint
Thanks for the feedback ☺️ `catch` is also applicable outside of `Raise`, so that makes sense!

serras (approved):
Amazing simplification!

raulraja (approved):
Looks great @nomisRev !

nomisRev:
Merging this PR, all feedback and suggestions are still very welcome 🙏

PR #2956, opened 2023-03-03
Fix ctx inconsistency

nomisRev:
One of the 4 signatures in `parMap` has param name `ctx` instead of `context`. This PR fixes the inconsistency.
This code is not released yet, so there is no source compatibility issues.


serras (approved):
Oh, I miss my two-letter-variable-name days in Haskell...

nomisRev:
@serras We can do the same in our production code 🤪 In Arrow I think it's good practice to follow the same naming conventions as Kotlin Std and KotlinX Coroutines. Otherwise everything would remain `f`, `ctx`, `n`, etc 😁

franciscodr (approved):

PR #2958, opened 2023-03-03
Update Raise.xOrAccumulate

nomisRev:
This PR updates `RaiseAccumulate` DSL with changes from #2952.
 - It enables binding `Nel<E>` and `E` in the same DSL in Raise DSL
 - It optimises the implementation in Raise, and removes a lot of allocations
 - It re-uses the DSL to implement `Iterable.mapOrAccumulate`


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -1,4 +1,4 @@
-@file:OptIn(ExperimentalTypeInference::class, ExperimentalContracts::class)
```
serras, line 1: No more contracts?

nomisRev: I got confused how we should use them, because they resulted in a couple of false-negatives of "Unreachable code" 😕

In this case of `catch` we should be able to tell the compiler that the second lambda "recovers" from the `Nothing` of the first lambda. It's a bit strange since the _inferred_ returned type is `Int` of the `catch` lambda. So I guess `callsInPlace` assumes _not-wrapped in try_?

Screenshot 2023-03-06 at 18 49 39

The second one is more complex.. since `either` is also `callsInPlace` and the last statement is an exception, but it's not throwing but _forever suspending_... And it exists early due to `raise` 😅 

Screenshot 2023-03-06 at 18 43 59

Not sure how we should deal with this best. WDYT @serras?

nomisRev: I also noticed that `ResultKt` only specifies `AT_MOST_ONCE` and never `EXACTLY_ONCE`.

nomisRev:
Not sure why you closed @raulraja?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
serras: Thanks for the thorough investigation, I never noticed that before. I assumed that this would help the compiler inline or optimize the code better, but it seems that it's reaching wrong conclusions...

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -7,574 +7,443 @@ import arrow.core.mapOrAccumulate
...
+  val error = mutableListOf<R>()
+  val results = ArrayList<B>(list.collectionSizeOrDefault(10))
+  for (item in list) {
+    fold<NonEmptyList<R>, B, Unit>(
+      { transform(RaiseAccumulate(this), item) },
+      { errors -> error.addAll(errors) },
+      { results.add(it) }
+    )
+  }
+  return error.toNonEmptyListOrNull()?.let { raise(it) } ?: results
```
serras, line 444: The fact that this can be defined now once and for all in a nice way fills me with joy :)

serras (approved):

raulraja (approved):
great work @nomisRev 🙌

i-walker (approved):
Thanks @nomisRev !!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
nomisRev: I am going to merge this PR, but I am continuing my search for a clear answer on this in Kotlin Slack. Re-asked my question in #library-development, https://kotlinlang.slack.com/archives/C8C4JTXR7/p1678213597746259.

It seems that `AT_MOST_ONCE` doesn't suffer from the _false-negative_ so I am going to re-add that back before merging.

PR #2957, opened 2023-03-03
Replace parTupled with parZipOrAccumulate

nomisRev:
This week I had a discussion with @lgtout about `parTupled`. We were originally on the fence to add `parTupled` in https://github.com/arrow-kt/arrow/pull/2900 due to signature conflicts with `parZip`.

Whilst writing some examples, and documentation I encountered issues with `parTupled` where the current signatures are also not valid. It cannot be resolved without explicitely specifying `context: CoroutineContext` parameter. After some discussion with @lgtout on Kotlin Slack, https://kotlinlang.slack.com/archives/C8UK6RTHU/p1677616555972749 we came to a mutual conclusion it's maybe not worth adding it since it requires an additional 9 methods. Total of 18 methods to provide something that practically already exists in `parZip`.

Whilst working on some other documentation about _accumulating parallel errors_ I realised that we only provide an API for this over `Iterable` but not _zipping independent operations_.

This PR proposes a small convenience API to cover that use-case.

@lgtout If everyone agrees on this API are you interested in continuing this work? I only played out the arity-2 signatures, and you're already familiar with the layout of Arrow Fx Coroutines + `parZip` so I thought this might be a good ticket for you to work on if you're interested. I can close this PR, or feel free to work on this branch directly. Whatever would work better for you ☺️

PR #2959, opened 2023-03-03
Update Either.zipOrAccumulate

nomisRev:
## New Description

**Note:** Everything mentioned here is open for discussion, so please provide all feedback.

We've introduce earlier `Either.zipOrAccumulate` for `combine + Either`, `Either` & `EitherNel`. This however doesn't allow intermixing of `Either` & `EitherNel` in the same function and thus requires manual lifting in some cases. Defining all different combination would be almost impossible ((9^2)*2 = 162 methods). i.e.

```kotlin
Either.zipOrAccumulate(String::plus, "error".leftNel(), 1.right()) { x: Int, y: Int -> x + y }
Either.zipOrAccumulate(String::plus, "error".leftNel(), 1.right()) { x: Int, y: Int -> x + y }
Either.zipOrAccumulate(1.right(), "error".leftNel()) { x: Int, y: Int -> x + y }
Either.zipOrAccumulate("error".leftNel(), 1.right()) { x: Int, y: Int -> x + y }
```

Without adding the `162 methods` users need to manually do, although this is the same as before with `Validated`.

```kotlin
Either.zipOrAccumulate(String::plus, "error".leftNel(), 1.right().toEitherNel()) { x: Int, y: Int -> x + y }
Either.zipOrAccumulate(String::plus, "error".leftNel(), 1.right().toEitherNel()) { x: Int, y: Int -> x + y }
Either.zipOrAccumulate(1.right().toEitherNel(), "error".leftNel()) { x: Int, y: Int -> x + y }
Either.zipOrAccumulate("error".leftNel(), 1.right().toEitherNel()) { x: Int, y: Int -> x + y }
```

To solve this problem Arrow introduced a DSL [`zipOrAccumulate`](https://github.com/arrow-kt/arrow/pull/2958) API that allows binding both `Raise<E>` and `Raise<NonEmptyList<E>>` and thus also `Either<E, ?>` and `EitherNel<E, ?>` This serves as a replacement for the _accumulate error_ pattern of `Validated<E, ?>` and `ValidatedNel<E, ?>`.

However this doesn't allow to pass `Either` values, and thus the API for `Either` is the same as `Raise` with the exception of it getting nesting in the DSL.

```kotlin
either {
  zipOrAccumulate({ "error".leftNel().bindNel() }, { 1.right().bind() }) { x: Int, y: Int -> x + y }
} shouldBe Either.zipOrAccumulate({ "error".leftNel().bindNel() }, { 1.right().bind() }) { x: Int, y: Int -> x + y }
```

So this brings the question:
 - Do we want to introduce this DSL based API for `Either.Companion`
 - Do we want to force users to manually wrap in `either { }`
 - Or do we want to stick to the current status-quo requiring `toEitherNel()` and work value based? (Note that the current value based implementation could still be derived from the DSL one).

#### Side note: zip deprecation

Arrow is deprecating `Either.zip(fa, fb, f)` in favour of `either { f(fa.bind(), fb.bind()) }` since it doesn't suffer from the arity-n problem, and offers the exact same behavior. Since `either` is now `inline fun` in `1.2.0` and thus works with-or-without `suspend` it can be used as a drop-in replacement. This has happened so far for `Either`, `Option` and still needs to happen for [`Ior`](https://github.com/arrow-kt/arrow/pull/2928/files#diff-c5990e586847dc3242060efe9f70a256299b84edf33bdb0171486a3db9c5184bR1245), [`Result`](https://github.com/arrow-kt/arrow/issues/2970) & [`Nullable`](https://github.com/arrow-kt/arrow/issues/2967).😅 Unless we're coming back from that (?), the deprecation has not been released yet.

## Old description

This PR updates `Either.zipOrAccumulate` but sadly since we don't have context receivers yet, and have to emulate through `RaiseAccumulate` it breaks inference / `BuilderInference` in a worse way than the PR that started this effort, https://github.com/arrow-kt/arrow/pull/2952. This PR the signature with [parZipOrAccumulate](https://github.com/arrow-kt/arrow/pull/2957).

We can of-course put off this change, and keep the current signatures until we have _context-receivers_. The migration path for it is relatively simple, and it will be easily to automate. With `ReplaceWith` (or OpenRewrite).

More details below:

If we put `@BuilderInference` on the method, instead of the individual functions we get following error:

Screenshot 2023-03-03 at 14 43 56

Some related YouTrack tickets related to the IDEA message:
 - https://youtrack.jetbrains.com/issue/KT-53797
- https://youtrack.jetbrains.com/issue/KT-53740/Builder-inference-with-multiple-lambdas-leads-to-unsound-type

If we put `@BuilderInference` we get following error for every lambda that tries to bind `EitherNel<E>` instead of `Either<E>` and will require to either specify the type arguments in the function, or specify a return type so `E` can be inferred from there.
Screenshot 2023-03-03 at 16 21 37

The up sight, with context receivers it looks 🔥 😍 

Screenshot 2023-03-03 at 16 17 47


Zordid:
In your example with Context Receivers, did you mean to make the first `zipOrAccumulate` use `Raise<R>` instead of `Raise<Nel<R>>` ? Then I would understand the code using the combine lambda... :-)

i-walker (approved):

PR #2960, opened 2023-03-06
Iterable test

abendt:
this adds a couple of tests to IterableTest to improve test coverage for functions in Iterable

https://github.com/arrow-kt/arrow/issues/2894

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
+++ b/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
@@ -6,7 +6,6 @@ import arrow.core.test.ior
 import arrow.core.test.option
 import arrow.typeclasses.Semigroup
 import io.kotest.core.spec.style.StringSpec
-import io.kotest.matchers.collections.shouldBeEmpty
 import io.kotest.matchers.collections.shouldContainExactly
 import io.kotest.matchers.nulls.shouldBeNull
 import io.kotest.matchers.nulls.shouldNotBeNull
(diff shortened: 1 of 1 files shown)


abendt:
i think there is a bug in the implentation of `unalign`. How i understand it `align` should be the inverse of `unalign`: `align(unalign xs) shouldBe xs`.

However this only works if all `xs` have both values. Then the current implementation of `unalign` will split the input into two list of equal length:

```
val xs = listOf(Both(1,x), Both(2,y), Both(3,z))
val (l,r) = xs.unalign() // listOf(1,2,3) to listOf(x,y,z)
l.align(r) shouldBe xs 
```

with this its possible to `align` both lists and to get back to the initial value of `xs`.

If `xs` contains `Left` or `Right` values then the splitted lists will have differing length. This leads to elements end up in a wrong position in the list:
```
val xs = listOf(Left(1), Both(2,y), Right(z))
val (l,r) = xs.unalign() // listOf(1,2) to listOf(y,z)
l.align(r) shouldBe xs // wrong! actual value is listOf(Both(1,y), Both(2,z))
```

One way to fix that could be to let `unalign` return `Pair<List<A?>, List<B?>>`. Missing elements in the source would then be represented as `null` in the result

nomisRev:
Hmm, interesting find @abendt. This would be breaking, and would thus need to land in 2.0.0.

In a PR I am preparing for some deprecations for `Iterable.kt` I actually wanted to rename `unalign` to `seperateIor` because it's behavior matches `seperateEither` and not `unalign` as you mention here.

Not sure what is the best path forward though 🤔 We either need to introduce the newly proposed `unalign return Pair<List<A?>, List<B?>>` as `unaligned`, or postpone this change completely until `2.x.x`.  

Perhaps a separate PR against the `arrow-2` branch changing the `unalign` signature, and adding these tests there directly? Then we can update this PR with the `seperateIor` introduction, and relevant tests. WDYT? Are you interested in adding such a PR against arrow-2?

Feel free to take this change from my WIP into to your PR:

```kotlin
public fun <A, B> Iterable<Ior<A, B>>.seperateIor(): Pair<List<A>, List<B>> =
  fold(emptyList<A>() to emptyList()) { (l, r), x ->
    x.fold(
      { l + it to r },
      { l to r + it },
      { a, b -> l + a to r + b }
    )
  }

@Deprecated(
  "The current unalign function is renamed to seperateIor, and a new unalign function is going to be added to Arrow 2.0.0.",
  ReplaceWith("seperateIor()", "arrow.core.seperateIor")
)
public fun <A, B> Iterable<Ior<A, B>>.unalign(): Pair<List<A>, List<B>> = seperateIor()
```

nomisRev:
Awesome, thank you so much @abendt 🙏 👏 👏

abendt:
hey @nomisRev, some of the gitlab checks fail because of timeouts. think this is not related to my changes? any idea how to resolve them?

nomisRev:
It's unrelated to your changes @abendt, it's been a long standing issue on Gradle side and the fix has finally been merged and released in 7.6. I created a PR to bump the timeout on CI. https://github.com/arrow-kt/arrow/pull/2965

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -777,7 +777,7 @@ public inline fun <A, B, C> Iterable<A>.align(b: Iterable<B>, fa: (Ior<A, B>) ->
         first.hasNext() && second.hasNext() -> Ior.Both(first.next(), second.next())
         first.hasNext() -> first.next().leftIor()
         second.hasNext() -> second.next().rightIor()
-        else -> throw IllegalStateException("this should never happen")
+        else -> throw AssertionError("this should never happen")
```
nomisRev, line 780: **question**: Was this an unrelated change?

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
```diff
@@ -551,4 +553,87 @@ class IterableTest : StringSpec({
...
+    }
+  }
+
+  "zip is the inverse of unzip" {
+    checkAll(Arb.list(Arb.pair(Arb.int(), Arb.string()))) { xs ->
+      val (l, r) = xs.unzip()
+
+      l.zip(r) shouldBe xs
+    }
+  }
```
nomisRev, line 579: **Nit**: This is testing Kotlin Std `zip` method. I think this can be removed.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
abendt: sorry, that was an experiment. will revert

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
abendt: done

nomisRev (approved):
Awesome work @abendt! 🙌 🥳 Thank you 🙏

gutiory (approved):

PR #2968, opened 2023-03-08
Removes autobind DSL in favor of getOrElse

nomisRev:
POC based on a discussion on Slack, https://kotlinlang.slack.com/archives/C5UPMM0A0/p1678220999183369

This POC removes _autobind_ DSL syntax, and favours `getOrElse` as a replacement in most cases. Which is consistent across all data types. Updates all the tests to reflect this new usage.

Also applies hygiene to `Either` (needs to be moved to other PR if this is not accepted):
 -  Adds `leftOrNull` in `Either` (should we add it for `Effect`/`EagerEffect`?)
 - Removes duplicated `recover` <~> `catch` method, and only keeps `catch` for reified `T : Throwable`.
 - Adds `catchOrThrow` as `Either` alternative to `catch` DSL.


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Raise.kt
```diff
@@ -128,7 +165,7 @@ public interface Raise<in R> {
     fold({ raise(it) }, ::identity)
 
   @Deprecated(
-    "Use recover or effect & recover instead",
+    "Use getOrElse on Raise, Effect or EagerEffect instead.",
     ReplaceWith("effect { f() }")
```
serras, line 169: Should the replacement use `getOrElse` too?

serras (approved):
I think this is a good move, we trade off an extra call to `bind` by what I think is a more orthogonal separation of concerns.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Raise.kt
nomisRev: This is deprecating the `attempt { f() } catch { g() }` block, but it exists out of two pieces. So I am unsure what is the best thing to do here.

`f` is should be refactored to `Raise`, but that is a different piece in the user code. That is why I recommended `effect { }`. My train of thought:

 1. Refactor `attempt { f() }` to `effect { f() }` user gets rid of the `attempt` call
 2. Refactor `catch` to `getOrElse { g() }` user gets rid of the `catch` call
 3. Refactor `arrow.core.continuations.effect` to `arrow.core.raise.effect`, the lambda body should be source-compatible

franciscodr (approved):

PR #2971, opened 2023-03-09
Deprecate Map.kt traverse & niche signatures, hygiene & bindAll

nomisRev:


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/map.kt
```diff
@@ -321,227 +317,281 @@ public inline fun <K, E, A, B> Map<K, A>.mapOrAccumulate(
...
+    lefts[entry.key] = a
+    rights[entry.key] = b
+  }
+  return lefts to rights
+}
 
-public fun <K, V> Map<K, V>.getOrNone(key: K): Option<V> = this[key].toOption()
+@Suppress("UNCHECKED_CAST")
+public fun <K, V> Map<K, V>.getOrNone(key: K): Option<V> =
+  if (containsKey(key)) Some(get(key) as V) else None
```
serras, line 594: I'm wondering why you prefer `get(key) as V` as opposed to `get(key)!!`.

In fact, if we talk about style, I actually prefer something on the line of:
```suggestion
if (key in this) Some(this[key]) else None
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/map.kt
```diff
@@ -37,54 +36,50 @@ public fun <K, A, B> Map<K, A>.zip(other: Map<K, B>): Map<K, Pair<A, B>> =
...
  */
-public inline fun <Key, A, B, C> Map<Key, A>.zip(other: Map<Key, B>, map: (Key, A, B) -> C): Map<Key, C> {
-  val destination = LinkedHashMap<Key, C>(size)
-  for ((key, bb) in this) {
-    Nullable.zip(other[key]) { cc -> map(key, bb, cc) }
-      ?.let { l -> destination.put(key, l) }
+public inline fun <Key, A, B, C> Map<Key, A>.zip(other: Map<Key, B>, map: (Key, A, B) -> C): Map<Key, C> =
+  buildMap(size) {
+    this@zip.forEach { (key, bb) ->
+      nullable {
```
serras, line 54: This is an interesting use of `nullable` to essentially not do this if the key is not present in any of the maps. Took me a while to understand what was going on...

serras (approved):

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/map.kt
nomisRev: `this[key]` returns _nullable types_ and using `!!` results in `KotlinNullPointerException` with _nested null problem_ 😅 The original implementation contained a bug 😱

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/map.kt
nomisRev: I updated this code in preparation of `Nullable.zip` deprecation.

rafaparadela:
Task linked: [CU-865byuhpx Deprecate map signatures and hygiene](https://app.clickup.com/t/865byuhpx)

arrow-libs/core/arrow-core/src/jvmTest/kotlin/examples/example-map-04.kt
```diff
@@ -2,13 +2,11 @@
...
-   mapOf("1" to 1, "2" to 2).align(mapOf("1" to 1, "2" to 2, "3" to 3)) { (_, a) ->
-     "$a"
-   }
-  //sampleEnd
-  println(result)
+fun test() {
+  mapOf("1" to 1, "2" to 2)
+    .align(mapOf("1" to 1, "2" to 2, "3" to 3)) { (_, a) ->
+      "$a"
+    } shouldBe mapOf("1" to "Ior.Both(1, 1)", "2" to Ior.Both(2, 2), "3" to Ior.Right(3))
```
franciscodr, line 11: Shouldn't this code return
```
mapOf("1" to "Ior.Both(1, 1)", "2" to "Ior.Both(2, 2)", "3" to "Ior.Right(3)")
```
instead?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/map.kt
```diff
@@ -37,54 +36,50 @@ public fun <K, A, B> Map<K, A>.zip(other: Map<K, B>): Map<K, Pair<A, B>> =
...
+        put(key, map(key, bb, other[key].bind()))
+      }
+    }
   }
```
franciscodr, line 58: **Nit:** `bb` represents a value of type `A`. Should we rename it to `aa`?
```suggestion
  buildMap(size) {
    this@zip.forEach { (key, aa) ->
      nullable {
        put(key, map(key, aa, other[key].bind()))
      }
    }
  }
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/map.kt
nomisRev, line 54: @serras @franciscodr actually this all suffers from the _nested nullable problem_ 😓 The previous implementation was also not correct... I will fix this implementation first, or should we move the fix to 2.0.0 since it's a breaking change in the current semantics?

Or should we fix the bug for nested null in the current version?

serras: I think that counts as a bug, and thus should be fixed in this version.

franciscodr: Agreed

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/map.kt
```diff
@@ -246,55 +234,72 @@ public fun <K, A, B> Map<K, A>.flatMap(f: (Map.Entry<K, A>) -> Map<K, B>): Map<K
...
-  val acc = mutableMapOf<K, B>()
-  forEach { (k, v) ->
-    when (val res = f(v)) {
-      is Right -> acc[k] = res.value
-      is Left -> return@traverse res
-    }
-  }
-  return acc.right()
-}
+public inline fun <K, E, A, B> Map<K, A>.traverse(f: (A) -> Either<E, B>): Either<E, Map<K, B>> =
```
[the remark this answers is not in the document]
nomisRev: Yes, I've seen the issue with `Either.Right<Any>` in multiple places (in our tests)😞 I think it's due to _covariance_ and `Nothing` in the data class constructor. I think this is why Kotlin Std **always** requires specifying the arguments, and `Nothing` is never used anywhere.

If this constructor was `operator fun fun <A, B> Either.Right.Companion.invoke(b: B): Either<A, B> = ...` I think it would not have been a problem, but it comes with some other trade-offs like not being able to call `::Right`.

That being said, I don't think a lot of people are using this pattern in real code, since `traverse` with only `Right` is `map` without `Either`.. In case that the compiler can know `E` and `A` I don't think this `ReplaceWith` results in any problems. So maybe we can leave this as _best effort_?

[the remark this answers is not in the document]
gutiory: Sure thing. I think we can leave as it is and be *reactive* in the case some raises an issue.

serras (approved):

PR #2972, opened 2023-03-09
change unalign signature

abendt:
see discussion in

* https://github.com/arrow-kt/arrow/pull/2960#issuecomment-1457644028 
* https://github.com/arrow-kt/arrow/pull/2960#issuecomment-1457898158

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
@@ -683,11 +683,11 @@ public inline fun <A, B, C> Iterable<C>.unzip(fc: (C) -> Pair<A, B>): Pair<List<
  * ```
  * <!--- KNIT example-iterable-11.kt -->
  */
-public fun <A, B> Iterable<Ior<A, B>>.unalign(): Pair<List<A>, List<B>> =
+public fun <A, B> Iterable<Ior<A, B>>.unalign(): Pair<List<A?>, List<B?>> =
   fold(emptyList<A>() to emptyList()) { (l, r), x ->
     x.fold(
-      { l + it to r },
-      { l to r + it },
+      { l + it to r + null},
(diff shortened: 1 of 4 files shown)


abendt:
hi @nomisRev,
this is the PR that changes the signature of `unalign`

Note that i had to slightly change the tests to handle the null values in the results

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -793,11 +793,14 @@ public inline fun <A, B, C> Iterable<C>.unzip(fc: (C) -> Pair<A, B>): Pair<List<
...
  */
-@Deprecated(
-  "The current unalign function is renamed to separateIor, and a new unalign function is going to be added to Arrow 2.0.0.",
-  ReplaceWith("separateIor()", "arrow.core.separateIor")
-)
-public fun <A, B> Iterable<Ior<A, B>>.unalign(): Pair<List<A>, List<B>> = separateIor()
+public fun <A, B> Iterable<Ior<A, B>>.unalign(): Pair<List<A?>, List<B?>> =
+  fold(emptyList<A>() to emptyList()) { (l, r), x ->
+    x.fold(
+      { l + it to r + null},
```
serras, line 799: I prefer using the `Pair` constructor instead of `to` when the pair represents simply two things put together, as opposed to a map entry where one is the key and the other the value.

serras (approved):
LGTM, just left a style comment

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
abendt: i have changed it @serras

serras: Thanks! :)

Issue #2964, opened 2023-03-07
Fix deprecation replacements


gutiory:
Some of the deprecation suggestions are not working when running `Replace with` action in `Intellij`. This is a bit annoying when we want to replace the deprecated method automatically.

The list of failed replacements is:

## Either
| Method | Working | Priority |
| ----------| ----------| ----------|
| all zip | NO |
| all | YES |
| bifoldLeft | YES |
| bifoldMap | NO |
| bimap | YES |
| bitraverseNullable | NO |
| bitraverseOption | NO |
| bitraverseValidated | NO |
| bitraverse | NO |
| catchAndFlatten | YES |
| catch | YES |
| combineK | YES |
| combine | YES |
| conditionally | YES |
| contains | YES |
| ensure | YES |
| exists | YES |
| filterOrElse | YES |
| filterOrOther | YES |
| findOrNull | YES |
| foldLeft | YES |
| foldMap | NO |
| fromNullable | YES |
| getOrHandle | YES |
| handleErrorWith | YES |
| handleError | YES |
| isEmpty | YES |
| leftIfNull | YES |
| orNone | YES |
| orNull | YES |
| orNull | YES |
| redeemWith | YES |
| redeem | YES |
| replicate | YES |
| replicate | YES |
| rightIfNotNull | YES |
| rightIfNull | NO |
| sequenceNullable | YES | 
| sequenceOption | YES |
| sequenceValidated | YES |
| sequence (list) | YES |
| sequence (option) | YES |
| sequence | YES |
| tapLeft | YES |
| tap | YES |
| toEither | NO |
| traverseNullable | NO |
| traverse | NO |
| traverse | NO |
| void | YES |
| all zips | YES |
 
## Ior 
| Method | Working | Priority
| ----------| ----------| ----------| 
| foldLeft | YES | |
| foldMap | YES | |
| bifoldLeft | YES | |
| bimap | YES | |
| padNull | YES | |
| orNull | YES | |
| all | YES | |
| exists | YES | |
| findOrNull | YES | |
| isEmpty | YES | |
| isNotEmpty | YES | |
| traverseEither | YES | |
| traverseOption | YES | |
| traverseValidated | YES | |
| void | YES | |
| getOrElse | YES | |
| bisequence | NO | | // fixed
| bisequenceEither | NO | | // fixed
| bisequenceOption | NO | | // fixed
| bisequenceNullable | NO | | // fixed
| bitraverseValidated | NO | |// fixed
[...]

PR #2973, opened 2023-03-09
Fix `Either` deprecation replacement expressions

gutiory:
This PR fixes some of the deprecation `ReplaceWith` suggestions for `Either` listed in https://github.com/arrow-kt/arrow/issues/2964. I'll update the issue table once this is merged.

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
@@ -2547,7 +2547,7 @@ public fun <A, B> Either<A, Iterable<B>>.sequence(): List<Either<A, B>> =
 @Deprecated(
   "Prefer Kotlin nullable syntax inside either DSL, or replace with explicit fold",
   ReplaceWith(
-    "fold({ emptyList() }, { iterable -> iterable.map { it.right() } })",
+    "this.fold<Option<Either<A, B>>>({ None }, { iterable -> iterable.map<B, Either<A, B>> { it.right() } })",
     "arrow.core.right",
   )
 )
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
@@ -900,7 +900,7 @@ public sealed class Either<out A, out B> {
 
   @Deprecated(
     NicheAPI + "Prefer when or fold instead",
-    ReplaceWith("fold({ initial }) { rightOperation(initial, it) }")
+    ReplaceWith("this.fold<C>({ initial }) { rightOperation(initial, it) }")
   )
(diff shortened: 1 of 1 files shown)


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -2294,7 +2294,7 @@ public operator fun <A : Comparable<A>, B : Comparable<B>> Either<A, B>.compareT
 
 @Deprecated(
   RedundantAPI + "Prefer zipOrAccumulate",
-  ReplaceWith("Either.zipOrAccumulate({ a, bb -> SGA.run { a.combine(bb) }  }, this, b) { a, bb -> SGB.run { a.combine(bb) } }")
+  ReplaceWith("Either.zipOrAccumulate<A, B, B, B>({ a:A, bb:A -> a.plus(bb)}, this, b) { a:B, bb:B -> a.plus(bb) }")
```
franciscodr, line 2297:
```suggestion
  ReplaceWith("Either.zipOrAccumulate<A, B, B, B>({ a:A, bb:A -> a.plus(bb) }, this, b) { a:B, bb:B -> a.plus(bb) }")
```

franciscodr (approved):
Thanks, @gutiory!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -2294,7 +2294,7 @@ public operator fun <A : Comparable<A>, B : Comparable<B>> Either<A, B>.compareT
 
 @Deprecated(
   RedundantAPI + "Prefer zipOrAccumulate",
-  ReplaceWith("Either.zipOrAccumulate({ a, bb -> SGA.run { a.combine(bb) }  }, this, b) { a, bb -> SGB.run { a.combine(bb) } }")
+  ReplaceWith("Either.zipOrAccumulate<A, B, B, B>({ a:A, bb:A -> a.plus(bb) }, this, b) { a:B, bb:B -> a.plus(bb) }")
```
nomisRev, line 2297: Was the `SGA.run  { a.combine(bb) }` here broken? If yes, then I think this is an improvement but we should probably just use `a + b` instead of `a.plus(b)` or is that also broken? 😅

Perhaps we should also add in the description that for custom types, `plus` needs to be replaced with a custom `combine` function?

nomisRev (approved):
Really awesome @gutiory. Thank you so much 😘 One question/remark.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
gutiory: `+` works fine! Thanks @nomisRev . See [`6332005` (#2973)](https://github.com/arrow-kt/arrow/pull/2973/commits/6332005895eb1e90ac788faaf9ffb60b1d0eae41)

i-walker (approved):

PR #2978, opened 2023-03-13
Add bind alias for Sequence operation in DSL

nomisRev:
Adds `bind` as suggested by @pakoito in https://github.com/arrow-kt/arrow/discussions/2861.
Updates `sequence` deprecation to point towards DSL.


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -454,14 +477,18 @@ public inline fun <A, B> Iterable<A>.traverse(f: (A) -> Option<B>): Option<List<
...
-  ReplaceWith("sequence()", "arrow.core.sequence")
+  "Sequence is being deprecated in favor of the option DSL.",
+  ReplaceWith("let { l -> result<List<A>> { l.bind() } }", "arrow.core.raise.result")
 )
 public fun <A> Iterable<Option<A>>.sequenceOption(): Option<List<A>> =
-  sequence()
+  let { l -> option { l.bind() } }
 
+@Deprecated(
+  "Sequence is being deprecated in favor of the option DSL.",
```
raulraja, line 487: This message may be misleading to users reading it, to me it reads as...
`(The) `Sequence` (datatype) is being deprecated in favor of the option DSL.
It should say, `The function ~sequence~ is being deprecated in favor of the option DSL` or similar wording.

serras:
I'm not a fan of reusing the `bind` name for both `Iterable` and non-`Iterable` versions.

nomisRev:
> I'm not a fan of reusing the bind name for both Iterable and non-Iterable versions.

I named it `bind` simply because it didn't have a conflict (in signature) with the other `bind` 😅
If you like the API, but dislike the name we can also rename it. @pakoito originally proposed `bindAll`.

serras:
> > I'm not a fan of reusing the bind name for both Iterable and non-Iterable versions.
> 
> I named it `bind` simply because it didn't have a conflict (in signature) with the other `bind` 😅 If you like the API, but dislike the name we can also rename it. @pakoito originally proposed `bindAll`.

I would prefer that name, indeed.

nomisRev:
@serras one follow-up question, for `RaiseAccumulate` this should be overwritten and use _accumulate_ strategy for `bindAll`, right?

serras:
> @serras one follow-up question, for `RaiseAccumulate` this should be overwritten and use _accumulate_ strategy for `bindAll`, right?

That would be great! I thought that merely calling `bind` inside `RaiseAccumulate` would be enough to have this behavior, but if we need to overload `bindAll` there too, I would make the change too.

nomisRev:
Calling `bind` within the lambda still _short_circuits_ the "current op".

I.e.
```kotlin
listOf(1, 2).mapOrAccumulate {
  if(it == 1) {
     "fail-1".left().bind() // <-- short-circuits "current op"
     "fail-1".left().bind() // this is not accumulated
  } else {
    listOf("fail-2".left(), "fail-2".left()).bindAll() // both get accumulated
  }
}
```

So `bindAll` needs to have a specialised implementation since `map { it.bind() }` would result in the `1` scenario. Or am I overlooking something?

serras:
@nomisRev It makes sense to me, that due to the way types are resolved the `bind()` gets called on `Raise<E>` instead of `RaiseAccumulate<E>`. It still needed a bit of additional thinking, though, it doesn't seem entirely obvious to me. So good point that we're adding it to `RaiseAccumulate`, because that's the least surprising behavior.

nomisRev:
@serras I am not sure I entirely follow your train of though.

> It makes sense to me, that due to the way types are resolved the bind() gets called on Raise<E> instead of RaiseAccumulate<E>.

Isn't this irrelevant? There is no other way for `bind` to work? What would the semantics otherwise be? Since `bind` returns `A` it **has to** short-circuit, right? Otherwise it would have to return something like `Value` which you modelled in [Accumulate DSL](https://github.com/arrow-kt/arrow/pull/2881). & thus de-coupling the DSL completely from `Raise`.

> It still needed a bit of additional thinking, though, it doesn't seem entirely obvious to me.

Not sure I understood this comment. Is it the _short-circuit_ behavior inside of _accumulate_ that requires additional thinking? Or the fact that semantics of `bind` remain the same and `bindAll` semantics change depending on being inside `Raise` or `RaiseAccumulate`?

rafaparadela:
Task linked: [CU-865byueyj Add bind alias for Sequence operation in DSL](https://app.clickup.com/t/865byueyj)

franciscodr (approved):
Just a few typos. Thanks, @nomisRev!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Raise.kt
```diff
@@ -171,7 +171,7 @@ public interface Raise<in R> {
   /**
    * Extract the [Either.Right] value of an [Either].
    * Any encountered [Either.Left] will be raised as a _logical failure_ in `this` [Raise] context.
-   * You can wrap the [bind] call in [recover] if you want to attempt to recover from any _logical failure_.
+   * You can wrap the [bindAll] call in [recover] if you want to attempt to recover from any _logical failure_.
```
i-walker, line 174:
```suggestion
   * You can wrap the [bind] call in [recover] if you want to attempt to recover from any _logical failure_.
```

i-walker (approved):

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -346,24 +351,31 @@ public inline fun <A, B> Iterable<A>.traverseResult(f: (A) -> Result<B>): Result
   traverse(f)
 
 @Deprecated(
-  "sequenceResult is being renamed to sequence to simplify the Arrow API",
-  ReplaceWith("sequence()", "arrow.core.sequence")
+  "The sequence extension function is being deprecated in favor of the result DSL.",
+  ReplaceWith("let { l -> either<E, List<A>> { l.bindAll() } }", "arrow.core.raise.either")
```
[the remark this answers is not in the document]
nomisRev: 😅 Feeling the deadline rush. 🦶🔫

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -454,14 +475,18 @@ public inline fun <A, B> Iterable<A>.traverse(f: (A) -> Option<B>): Option<List<
...
-  sequence()
+  let { l -> option { l.bindAll() } }
 
+@Deprecated(
+  "The sequence extension function is being deprecated in favor of the option DSL.",
+  ReplaceWith("let { l -> option<List<A>> { l.bindAll() } }", "arrow.core.raise.option")
+)
 public fun <A> Iterable<Option<A>>.sequence(): Option<List<A>> =
-  traverse(::identity)
+  let { l -> option { l.bindAll() } }
```
pakoito, line 489: the let is unnecessary here, right?

pakoito:
I was wondering why I got a Participating notification :D Good to hear my suggestion made it <3

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -490,6 +489,15 @@ public open class RaiseAccumulate<Error>(
   public override fun raise(r: Error): Nothing =
     raise.raise(nonEmptyListOf(r))
 
+  @RaiseDSL
+  public inline fun <A, B> Iterable<A>.mapOrAccumulate(
+    transform: RaiseAccumulate<Error>.(A) -> B
+  ): List<B> = raise.mapOrAccumulate(this) { transform(it) }
```
pakoito, line 495: microperf: `raise.mapOrAccumulate(this, transform)`

nomisRev:
We have traverse separate but it’s called `separateEither` since we also have `separateIor`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
nomisRev:
```suggestion
  ): List<B> = raise.mapOrAccumulate(this, transform)
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
nomisRev: It's so that the `ReplaceWith` is inline with the implementation body. `let` should incur any perf penalty due to `@InlineOnly`.

PR #2981, opened 2023-03-15
Schedule as functor

serras:


serras:
Closed in favor of #2979

PR #2985, opened 2023-03-17
Deprecate Monoid and Semigroup in Optics

### LINKED ISSUE #2975: ["Request"] Update arrow-optics to work without Monoid & Semigroup

2.0.0-SNAPSHOT

What would you like to see?

Removal of Monoid and Semigroup in APIs

serras:
Fixes #2975 

This PR also includes some additional changes in the code to remove calls to deprecated methods like `bimap` and `orElse`.


arrow-libs/optics/arrow-optics/api/arrow-optics.api
```diff
@@ -100,7 +104,7 @@ public abstract interface class arrow/optics/Getter : arrow/optics/Fold {
 	public abstract fun choice (Larrow/optics/Getter;)Larrow/optics/Getter;
 	public abstract fun compose (Larrow/optics/Getter;)Larrow/optics/Getter;
 	public abstract fun first ()Larrow/optics/Getter;
-	public abstract fun foldMap (Larrow/typeclasses/Monoid;Ljava/lang/Object;Lkotlin/jvm/functions/Function1;)Ljava/lang/Object;
```
serras, line 103: Unfortunately the change of the `empty / combine` method being the "main" one in `Fold` is breaking binary compatibility. But I don't know how to make these changes otherwise; any help is welcome.

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Fold.kt
```diff
@@ -25,34 +26,41 @@ import kotlin.jvm.JvmStatic
...
+   */
+  public fun <R> foldMap(empty: R, combine: (R, R) -> R, source: S, map: (focus: A) -> R): R
+
   /**
    * Map each target to a type R and use a Monoid to fold the results
    */
-  public fun <R> foldMap(M: Monoid<R>, source: S, map: (focus: A) -> R): R
+  @Deprecated(MonoidDeprecation, ReplaceWith("foldMap(M.empty(), M::combine, source, map)", "arrow.typeclasses.combine"))
+  public fun <R> foldMap(M: Monoid<R>, source: S, map: (focus: A) -> R): R =
+    foldMap(M.empty(), M::combine, source, map)
```
nomisRev, line 39: I think the only way to do this in a non-binary breaking change is to:

 - Deprecate `foldMap` with `Monoid` but keep it abstract
 - Implement the new `foldMap` through the abstract method. (This would require creating an anonymous `Monoid`)

nomisRev: This would meant that someone who implements a custom `Fold` needs to refactor a breaking change in `2.0`, but a consumer would not face a breaking change when not relying on the deprecated code.

nomisRev (commented):
Looks good to me overall. I left a suggestion of how we can work around the binary compatibility problem, and put it off to 2.0

arrow-libs/optics/arrow-optics/src/commonTest/kotlin/arrow/optics/EveryTest.kt
```diff
@@ -39,13 +38,13 @@ class EveryTest : StringSpec({
...
         checkAll(Arb.list(Arb.int())) { ints ->
-          fold(Monoid.int(), ints) shouldBe ints.sum()
+          fold(0, { x, y -> x + y }, ints) shouldBe ints.sum()
         }
       }
 
       "asFold should behave as valid Fold: fold" {
         checkAll(Arb.list(Arb.int())) { ints ->
-          fold(Monoid.int(), ints) shouldBe ints.sum()
+          fold(0, { x, y -> x + y }, ints) shouldBe ints.sum()
```
franciscodr, line 47: **Nit:** We could use `Int::plus` instead
```suggestion
          fold(0, Int::plus, ints) shouldBe ints.sum()
```

serras: We've had bad experiences in the past with the implicit conversion from `A.(A) -> A` into `(A, A) -> A`, so I would rather leave it this way.

arrow-libs/optics/arrow-optics/src/commonTest/kotlin/arrow/optics/EveryTest.kt
```diff
@@ -39,13 +38,13 @@ class EveryTest : StringSpec({
 
       "asFold should behave as valid Fold: combineAll" {
         checkAll(Arb.list(Arb.int())) { ints ->
-          fold(Monoid.int(), ints) shouldBe ints.sum()
+          fold(0, { x, y -> x + y }, ints) shouldBe ints.sum()
```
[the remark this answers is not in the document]
serras: Removed in the next commit

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Fold.kt
serras: I think I've managed to keep ABI compatibility by:
- Keep the `Fold` interface as it was, using `Monoid`;
- Define the new `fold` and `foldMap` as extension functions, so they don't become part of the interface.

franciscodr (approved):
Looks good to me, @serras!

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Fold.kt
```diff
@@ -327,3 +305,18 @@ public interface Fold<S, A> {
       Every.tuple10()
   }
 }
+
+/**
+ * Fold using the given [empty] element and [combine].
+ */
+public fun <S, A> Fold<S, A>.fold(empty: A, combine: (A, A) -> A, source: S): A =
```
nomisRev, line 312: Can this be added as a method on `Fold`?

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Fold.kt
```diff
@@ -327,3 +305,18 @@ public interface Fold<S, A> {
...
+  foldMap(empty, combine, source, ::identity)
+
+/**
+ * Map each target to a type [R] and combine the results as a fold.
+ */
+public fun <S, A, R> Fold<S, A>.foldMap(empty: R, combine: (R, R) -> R, source: S, map: (focus: A) -> R): R =
```
nomisRev, line 318: Is this an extension because it was a conflicting API? and `@JvmName` doesn't work on `interface`?

serras: As discussed above, this is an extension so that the method is not added to the interface, which would break binary compatibility.

nomisRev: Adding methods is not a problem, only removal, since this would break compatibility for users relying on dependencies with older Arrow versions of the 1.x.x series.

serras: Oh, I misunderstood the binary compatibility then. I've moved the functions back to `Fold`, where they should logically belong.

nomisRev: It's really a JVM thing I guess, the problem is the following:

 - Library A has a dependency on Arrow 1.1.5 relies on `Fold.foldMap`
 - Project B has dependency on Library A & has dependency on Arrow 1.1.5
 - Arrow 1.2.0 removes `Fold.foldMap`
 - Project B bumps dependency to Arrow 1.2.0 => Fold.foldMap: [MethodNotFoundException](https://docs.oracle.com/javaee/7/api/javax/faces/el/MethodNotFoundException.html) at runtime.

So adding methods is non-problematic.

PR #2986, opened 2023-03-17
Recommend Either.getOrElse for handleError instead of recover

nomisRev:
Recommending `getOrElse` for `handleError` gives a better flowing of types than using `Either<Nothing, A>`,
and is more in line with the changes made in [Removes autobind DSL in favor of getOrElse](https://github.com/arrow-kt/arrow/pull/2968).

`recover` is a replacement for `handleErrorWith` and using it for `handleError` (sadly) results in poor inference, because `@BuilderInference` doesn't infer `Nothing` when `EE` is not used.


franciscodr (approved):

gutiory (approved):

franciscodr:
Is there anything pending in this PR? Or are we ok with merging it?

franciscodr:
> If anything from me is open, and 2x approved feel free to merge @franciscodr 😉

Thanks, @nomisRev I will do it! 😄

PR #2987, opened 2023-03-19
Raise - Align names parameters, arguments and generic types

nomisRev:


serras (approved):
This is great work @nomisRev! I can see how these changes are going to make `Raise` easier to use in the IDE.

PS: my personal favorite is `OtherError`, it makes so clear that two different error types are involved there; instead of `E` and `R`.

franciscodr (approved):
Thanks, @nomisRev! Code looks more clear with these changes

PR #2988, opened 2023-03-20
Selectively apply the versioning plug-in

serras:
⚠️ I'm not sure if this is a good solution, please review the PR carefully!

Since the versioning plug-in takes quite some time of the configuration ([build scan](https://scans.gradle.com/s/oh3ijspciggqs/performance/configuration?openScriptsAndPlugins=WyJjb2RlLXVuaXQtY29tLmphdmllcnNjLnNlbXZlci5ncmFkbGUucGx1Z2luIl0)), I've been toying with the idea of applying selectively, only when `VERSIONING=true` as environment variable. This takes the configuration time down to 25 seconds ([build scan](https://scans.gradle.com/s/qwuogylg55u5e/performance/configuration)), and due to better use of caching, to 2 seconds for subsequent builds.


JavierSegoviaCordoba:
The configuration time issue is fixed on new versions of the semver plugin, for example, the version `5.0.0-alpha.1`

There are breaking changes from the current version to the new one, the major one is the semver plugin needs to be applied to each project individually in order to get project isolation support. Is shouldn't be a problem as we can apply the convention plugin we already have to each project easily.

It has a settings plugin that applies the plugin to all projects too, so it is not necessary to apply it manually on each project.

serras:
Closed in favor of #3017

PR #2992, opened 2023-03-21
Fix nested nullable problem for map zip and flatMap implementations

franciscodr:
This pull request fixes a bug in the `zip` and `flatMap` implementation for `Map` caused by the nested nullable problem. For example, given the following code:
```
val mapA: Map<Int, String> = mapOf(1 to "a", 2 to "b", 3 to "c")
val mapB: Map<Int, String?> = mapOf(1 to "d", 2 to null, 3 to "f")

mapA.zip(mapB) { _, a, b -> Pair(a, b) }
```
It should return:
```
{1=(a, d), 2=(b, null), 3=(c, f)}
```
However, the current `zip` implementation was filtering out the key with a `null` value:
```
{1=(a, d), 3=(c, f)}
```


nomisRev (approved):
Awesome, thank you so much @franciscodr 🙏 🙌 
Looks great to me.

gutiory (approved):

Issue #2966, opened 2023-03-07
["Request"] Fix incorrect Sequence signatures 


nomisRev:
What version are you currently using?

1.x.x

What would you like to see?

There are a bunch of `Sequence` signatures that in their signature give the impression that they're lazy, but in reality they're _terminal_ and consume the entire `Sequence` before returning.

From my short investigation it looks like these cannot be made lazy, but perhaps we can instead update their implementation for them to return `List` instead of `Sequence`.

- `crosswalk`
- `crosswalkMap`
- `crosswalkNull`
- `separateEither`
- `unalign`
- `unzip`

This might be possible to rewrite as a lazy version:

- `unweave` (investigate if can be made lazy)

serras:
Implementations returning `List` have been added as part of #2993, and the incorrect ones marked for deprecation. We should remember to remove them in 2.0.

nomisRev:
This was taken care of by @gutiory

PR #2993, opened 2023-03-23
Deprecate `Sequence` signatures and hygiene

gutiory:
These are the main changes in this PR:
- Add deprecation to methods
- Rename some method parameters
- Deprecate methods that return `Sequence`, but that actually are terminal. This is related to [this issue](https://github.com/arrow-kt/arrow/issues/2966) 
- Deprecate other methods and fix replacements.
- Resolve deprecation in `Sequence` tests.


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
```diff
@@ -319,6 +360,24 @@ public fun <A, B> Sequence<A>.crosswalk(f: (A) -> Sequence<B>): Sequence<Sequenc
     }
   }
 
+@OverloadResolutionByLambdaReturnType
+@JvmName("crosswalkT")
+public fun <A, B> Sequence<A>.crosswalk(f: (A) -> List<B>): List<List<B>> =
```
serras, line 365: Could we maybe make this more general?

```suggestion
public fun <A, B> Sequence<A>.crosswalk(f: (A) -> Iterable<B>): List<List<B>> =
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
```diff
@@ -595,7 +682,7 @@ public fun <A> Sequence<A>.salign(
 /**
  * aligns two structures and combine them with the given [Semigroup.combine]
  */
-@Deprecated(SemigroupDeprecation, ReplaceWith("salign(other, SG::combine)", "arrow.typeclasses.combine"))
+@Deprecated(SemigroupDeprecation, ReplaceWith("salign(other, {a, b -> a + b})", "arrow.typeclasses.combine"))
```
serras, line 685: Is this correct if `SG` is not in scope?

serras (approved):
Nice job! I've left a few comments.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
gutiory: Sure!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Sequence.kt
gutiory: We talked like a couple of weeks a go, that over the 90-ish % of the cases will work with the replacement of `combine` with the `+`. For the rest, I'll add an entry to the migration guide indicating this matter.

franciscodr (approved):
Thanks, @gutiory!

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/SequenceKTest.kt
```diff
@@ -302,7 +321,24 @@ class SequenceKTest : StringSpec({
       }
     }
 
-    "filterOption should filter None" {
+    "unzipToPair should unzip values in a Pair in a Sequence of Pairs" {
+      checkAll(Arb.list(Arb.pair(Arb.string(), Arb.int()))){pairList ->
```
franciscodr, line 325:
```suggestion
      checkAll(Arb.list(Arb.pair(Arb.string(), Arb.int()))){ pairList ->
```

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/SequenceKTest.kt
```diff
@@ -302,7 +321,24 @@ class SequenceKTest : StringSpec({
...
+        val obtained = pairList.asSequence().unzipToPair()
+        val expected = pairList.unzip()
+        obtained shouldBe expected
+      }
+    }
+
+    "unzipToPair should unzip values in a Pair in a Sequence" {
+      checkAll(Arb.list(Arb.string())){strList ->
+        val obtained = strList.asSequence().unzipToPair{str -> Pair(str, str.length)}
```
franciscodr, line 334:
```suggestion
      checkAll(Arb.list(Arb.string())){ strList ->
        val obtained = strList.asSequence().unzipToPair { str -> Pair(str, str.length)}
```

franciscodr (approved):

PR #2995, opened 2023-03-24
NonEmptyList: Add bindAll and more unit tests

franciscodr:
This pull request adds the `bindAll` function to `NonEmptyList` for both `Raise` and `RaiseAccumulate` implementations. Additionally, it adds missing unit tests for the NonEmptyList API.

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
@@ -8,6 +8,7 @@ import arrow.atomic.Atomic
 import arrow.atomic.updateAndGet
 import arrow.core.Either
 import arrow.core.Ior
+import arrow.core.NonEmptyList
 import arrow.core.None
 import arrow.core.Option
 import arrow.core.Some
@@ -84,6 +85,11 @@ public class ResultRaise(private val raise: Raise<Throwable>) : Raise<Throwable>
   @JvmName("bindAllResult")
   public fun <A> Iterable<Result<A>>.bindAll(): List<A> =
     map { it.bind() }
+
+  @RaiseDSL
+  @JvmName("bindAllResult")
+  public fun <A> NonEmptyList<Result<A>>.bindAll(): NonEmptyList<A> =
+    map { it.bind() }
 }
 
 public class OptionRaise(private val raise: Raise<None>) : Raise<None> by raise {
@@ -97,6 +103,11 @@ public class OptionRaise(private val raise: Raise<None>) : Raise<None> by raise 
   @RaiseDSL
   @JvmName("bindAllOption")
   public fun <A> Iterable<Option<A>>.bindAll(): List<A> =
+    map { it.bind() }
+
+  @RaiseDSL
+  @JvmName("bindAllOption")
+  public fun <A> NonEmptyList<Option<A>>.bindAll(): NonEmptyList<A> =
     map { it.bind() }
 
   @RaiseDSL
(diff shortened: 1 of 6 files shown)


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -84,6 +85,11 @@ public class ResultRaise(private val raise: Raise<Throwable>) : Raise<Throwable>
   @JvmName("bindAllResult")
   public fun <A> Iterable<Result<A>>.bindAll(): List<A> =
     map { it.bind() }
+
+  @RaiseDSL
+  @JvmName("bindAllResult")
```
serras, line 90: We may have a problem with this name clashing in the Arrow 2.0 branch, where `NonEmptyList` is defined as a value class.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Raise.kt
```diff
@@ -249,6 +250,10 @@ public interface Raise<in Error> {
   @RaiseDSL
   public fun <A> Iterable<Either<Error, A>>.bindAll(): List<A> =
     map { it.bind() }
+
+  @RaiseDSL
+  public fun <A> NonEmptyList<Either<Error, A>>.bindAll(): NonEmptyList<A> =
```
serras, line 255: If we are including `NonEmptyList`, should we also include `Set` and other types of collections?

serras (approved):

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Raise.kt
franciscodr: Yes, I thought about adding `NonEmptySet` too today

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
nomisRev: Not sure how it would differ when it being a `value class` 🤔 From Kotlin AST point-of-view it's a different type, and from JVM binary it has a different name.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -502,6 +502,9 @@ public open class RaiseAccumulate<Error>(
   override fun <A> Iterable<Either<Error, A>>.bindAll(): List<A> =
     mapOrAccumulate { it.bind() }
 
+  override fun <A> NonEmptyList<Either<Error, A>>.bindAll(): NonEmptyList<A> =
+    requireNotNull(mapOrAccumulate { it.bind() }.toNonEmptyListOrNull())
```
nomisRev, line 506: Which `mapOrAccumulate` is this resolving too if we have to wrap it in `requireNotNull`

franciscodr: It uses the `mapOrAccumulate` defined for `Iterable` located in [the line](https://github.com/arrow-kt/arrow/pull/2995/files/75564ed5fa19db1f9b7a92b9541cc826dafe1c62#diff-faa31a2c3b82d4997f537af6af5e04db3cfa2fdd591098b85bf1bb6deb9f21b6R497). I could also change the implementation to:
```
requireNotNull(raise.mapOrAccumulate(this) { it.bind() }.toNonEmptyListOrNull())
```

nomisRev: Ah, I thought there was also one defined for `NonEmptyList`. My bad.

nomisRev (approved):
Looks great 🙌 Thank you @franciscodr.

PR #2998, opened 2023-03-27
let memoize cache null result values

abendt:
the current implementation of memoize does not cache null result values. 
In effect the memoized function will be called more than once for a given set of parameters, if the function returns null for that set. 

this MR changed the cache in MemoizedHandler to ensure null values are also stored.


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/memoization.kt
```diff
@@ -124,13 +126,13 @@ private class MemoizedHandler<F, in K : MemoizedCall<F, R>, out R>(val f: F) {
         // No value is set yet (race condition)
         if (bb == null) {
           // Value was set, return the value
-          if (cache.compareAndSet(old, old + Pair(k, b))) return b
+          if (cache.compareAndSet(old, old + Pair(k, Option.fromNullable(b)))) return b
```
nomisRev, line 129:
```suggestion
          if (cache.compareAndSet(old, old + Pair(k, Some(b)))) return b
```
Shouldn't this be `Some(b)`, and instead of using `Map.get` in line 125 and 120 it should use `getOrNone`.
That way it can pattern match of `Option`:
 - `None` meaning the value wasn't present for the key `K`.
 - `Some` if a value was found for `k`, and return it.

serras:
@abendt I don't fully understand why you need an additional `Option` here. I agree that the current approach is wrong, because `cache.get()[k]` returns `null` for both a non-existing key, and a key whose value is `null`. However, you can fix that problem by using `containsKey`, which returns `true` even if the value is `null`.

serras:
@abendt I don't fully understand why you need an additional `Option` here. I agree that the current approach is wrong, because `cache.get()[k]` returns `null` for both a non-existing key, and a key whose value is `null`. However, you can fix that problem by using `containsKey`, which returns `true` even if the value is `null`.

nomisRev:
Yes, @serras is right. It would indeed by better to just us `containsKey` instead of nesting into `Option`.

serras:
We were not confident of introducing a map in general in `arrow-atomic` (as done in #3002), so I've re-opened this issue and pushed the use of `containsKey` (in the form of `in` check) instead of `get`.

nomisRev (approved):
Looks good to me. Thanks for the fix @abendt & @serras 🙏

PR #3002, opened 2023-03-27
Memoization via `AtomicMap`

serras:
The PR #2998 by @abendt fixes memoization for handling nullable values correctly. The work is wonderful, but when reviewing it two things came to my mind:
1. This pattern of a map which can be `null` and supports concurrency is very useful in general,
2. There's a [`ConcurrentMap` in the JVM](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentMap.html) which we could use to provide faster access in that platform.

So I've taken the ideas in @abendt's PR (and also the new tests) and created this new version with an `AtomicMap` for `arrow-atomic`.

The crux of the implementation is using a special `Null` object to represent `null` within the `ConcurrentMap`, because the implementation in the JVM doesn't support storing `null` values. This is similar to what we do in other parts of Arrow with `EMPTY_VALUE`.

### DIFF
--- a/arrow-libs/core/arrow-atomic/src/jvmMain/kotlin/arrow/atomic/AtomicMap.kt
+++ b/arrow-libs/core/arrow-atomic/src/jvmMain/kotlin/arrow/atomic/AtomicMap.kt
@@ -16,7 +16,7 @@ public actual class AtomicMap<K, V>: Map<K, V> {
    * If the specified [key] is not already associated with a value, associates it with the given [value].
    * Otherwise, applies the [combine] value to merge those.
    */
-  public actual fun merge(key: K, value: V, combine: (old: V, new: V) -> V) {
+  public actual fun merge(key: K, value: V, combine: (oldValue: V, newValue: V) -> V) {
     map.merge(key, value ?: Null) { old, new ->
       combine(unwrapNull(old), unwrapNull(new)) ?: Null
     }
--- a/arrow-libs/core/arrow-atomic/src/commonMain/kotlin/arrow/atomic/AtomicMap.kt
+++ b/arrow-libs/core/arrow-atomic/src/commonMain/kotlin/arrow/atomic/AtomicMap.kt
@@ -5,7 +5,7 @@ public expect class AtomicMap<K, V> public constructor() : Map<K, V> {
    * If the specified [key] is not already associated with a value, associates it with the given [value].
    * Otherwise, applies the [combine] value to merge those.
    */
-  public fun merge(key: K, value: V, combine: (old: V, new: V) -> V)
+  public fun merge(key: K, value: V, combine: (oldValue: V, newValue: V) -> V)
 
   /**
    * Replaces the specified [key] with [newValue], only if it was mapped to the given [oldValue].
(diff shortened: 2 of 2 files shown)


arrow-libs/core/arrow-atomic/src/jsMain/kotlin/arrow/atomic/AtomicMap.kt
```diff
@@ -0,0 +1,3 @@
+package arrow.atomic
+
+public actual typealias AtomicMap<K, V> = DefaultAtomicMap<K, V>
```
nomisRev, line 3: This can actually be regular mutable map, no? 🤔

arrow-libs/core/arrow-atomic/src/jvmMain/kotlin/arrow/atomic/AtomicMap.kt
```diff
@@ -0,0 +1,70 @@
+package arrow.atomic
+
+import java.util.concurrent.ConcurrentHashMap
+
+public actual class AtomicMap<K, V>: Map<K, V> {
+  // used to signal that we have a 'null' as value
+  public object Null
+
+  private val map: ConcurrentHashMap<K, Any> = ConcurrentHashMap()
```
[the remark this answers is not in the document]
serras: According to the JVM docs,

> A [Map](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html) providing thread safety and atomicity guarantees.
> [..] retrieval operations do not entail locking, and there is not any support for locking the entire table in a way that prevents all access.

If I understood correctly, using `ConcurrentHashMap` allows us to have different threads updating different keys (for example, if we are memoizing and parallelizing Fibonacci) without locking, as opposed to a single concurrent map.

arrow-libs/core/arrow-atomic/src/jsMain/kotlin/arrow/atomic/AtomicMap.kt
serras: But then I don't know how to implement the operations like `putIfAbsent` ':)

nomisRev: Since JS is single threaded, can it not just be `if(!containsKey(k)) put(k, v)`? 🤔

serras: Sorry, I didn't explain my problem correctly. The problem is that I don't know how to use `expect`/`actual` in that case. If I do:
```kotlin
public actual typealias AtomicMap<K, V> = HashMap<K, V>
```
then I get the following weird error:
```
Actual typealias 'AtomicMap' has no corresponding expected declaration
The following declaration is incompatible because some supertypes are missing in the actual declaration:
* public final expect class AtomicMap<K, V> : Map<K, V> defined in arrow.atomic in file AtomicMap.kt
```

arrow-libs/core/arrow-atomic/src/jvmMain/kotlin/arrow/atomic/AtomicMap.kt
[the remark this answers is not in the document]
serras: Fair enough. Since this came from #2998, maybe I can make the corresponding changes there and call it a day?

PR #3003, opened 2023-03-28
Remove `Monoid` and `Semigroup` combine from replacements

gutiory:
This PR replace any method that belongs to deprecated `Monoid` or `Semigroup` that are in the deprecation replacements.

~~The last commit just deprecates the `foldMap` method in `Fold` interface. I'd need extra 👀 on this, because a bunch of other interfaces implement this interface. Basically, I've done the following:~~
- ~~Deprecated the `foldMap` implementation that has a `Monoid` parameter~~
- ~~Remove the `foldMap` implementation that does not have the `Monoid`, but that creates one and used the deprecated `foldMap`. This makes this method abstract.~~
- ~~Every interface that uses `Fold` implements now the abstract `foldMap`.~~

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
@@ -1201,21 +1201,30 @@ public sealed class Either<out A, out B> {
 
   @Deprecated(
     NicheAPI + "Prefer using the Either DSL, or explicit fold or when",
-    ReplaceWith("fold({ emptyList() }, { fa(it).map(::Right) })")
+    ReplaceWith(
+      "fold({ listOf(it.left()) }, { fa(it).map(::Right) })", 
+      "arrow.core.Either.Right",
+      "arrow.core.Either.left"
+    )
   )
   @OptIn(ExperimentalTypeInference::class)
   @OverloadResolutionByLambdaReturnType
   public inline fun <C> traverse(fa: (B) -> Iterable<C>): List<Either<A, C>> =
-    fold({ emptyList() }, { fa(it).map(::Right) })
+    fold({ listOf(it.left()) }, { fa(it).map(::Right) })
 
   @Deprecated(
     NicheAPI + "Prefer using the Either DSL, or explicit fold or when",
-    ReplaceWith("fold({ None }, { right -> fa(right).map(::Right) })")
+    ReplaceWith(
+      "fold({ Some(it.left()) }, { right -> fa(right).map(::Right) })",
+      "arrow.core.Either.Right",
+      "arrow.core.Some",
+      "arrow.core.left"
+    )
   )
   @OptIn(ExperimentalTypeInference::class)
   @OverloadResolutionByLambdaReturnType
   public inline fun <C> traverse(fa: (B) -> Option<C>): Option<Either<A, C>> =
-    fold({ None }, { right -> fa(right).map(::Right) })
+    fold({ Some(it.left()) }, { right -> fa(right).map(::Right) })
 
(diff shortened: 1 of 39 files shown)


serras:
> This PR replace any method that belongs to deprecated `Monoid` or `Semigroup` that are in the deprecation replacements.
> 
> The last commit just deprecates the `foldMap` method in `Fold` interface. I'd need extra 👀 on this, because a bunch of other interfaces implement this interface. Basically, I've done the following:
> 
>     * Deprecated the `foldMap` implementation that has a `Monoid` parameter
> 
>     * Remove the `foldMap` implementation that does not have the `Monoid`, but that creates one and used the deprecated `foldMap`. This makes this method abstract.
> 
>     * Every interface that uses `Fold` implements now the abstract `foldMap`.

Sorry if this was not clear from the original PR, but we decided against taking that route. The reason is that we break binary compatibility. The breaking change of actually making `foldMap` with `combine` and `empty` the "main" version is for 2.0.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -904,7 +904,7 @@ public sealed class Either<out A, out B> {
 
   @Deprecated(
     NicheAPI + "Prefer when or fold instead",
-    ReplaceWith(" fold({ MN.empty() }, f)")
+    ReplaceWith(" fold({ empty }, f)")
```
serras, line 907: I am a bit confused by these replacements. This won't result in correct code, right?

gutiory: Yes, it won't work. I'm writing an entry in the arrow web migration guide just to explain this. It's the only way to get rid of the `Monoid.empty` deprecated method.

franciscodr (commented):
Regarding the Monoid `empty` value, I see we name it differently in some `ReplaceWith` expressions, like `empty`, `initial`, or `initialValue`. We could choose a unique identifier and use it in all the examples for consistency.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Option.kt
```diff
@@ -1417,7 +1417,7 @@ public inline fun <A, B> Option<A>.redeemWith(fe: (Unit) -> Option<B>, fb: (A) -
 
 @Deprecated(
   NicheAPI + "Prefer using the Option DSL or map",
-  ReplaceWith("map { a -> List(n) { a }.fold(MA.empty(), MA::combine) }", "arrow.typeclasses.combine")
+  ReplaceWith("this.map { a -> List(n) { a }.fold(intialValue){a1, a2 -> a1 + a2} }")
```
franciscodr, line 1420:
```suggestion
  ReplaceWith("this.map { a -> List(n) { a }.fold(initialValue){a1, a2 -> a1 + a2} }")
```

gutiory:
> Regarding the Monoid `empty` value, I see we name it differently in some `ReplaceWith` expressions, like `empty`, `initial`, or `initialValue`. We could choose a unique identifier and use it in all the examples for consistency.

Absolutely. I've mixed some names from `Monoid.empty()` and others from the `fold` parameter. 
What about to rename all with the name of the `fold` parameter we're suggesting to be replaced with?

franciscodr:
> > Regarding the Monoid `empty` value, I see we name it differently in some `ReplaceWith` expressions, like `empty`, `initial`, or `initialValue`. We could choose a unique identifier and use it in all the examples for consistency.
> 
> Absolutely, @franciscodr . I've mixed some names from `Monoid.empty()` and others from the `fold` parameter. What about to rename all with the name of the `fold` parameter we're suggesting to be replaced with?

Yeah, makes sense to me

franciscodr (approved):
Thanks, @gutiory!

PR #2999, opened 2023-03-27
let Mapk 'zipN with nullable' tests use Arb.map2 to generate values

abendt:
This PR builds on https://github.com/arrow-kt/arrow/pull/2980

there are a couple of MapK tests that test zipN with nullables. The tests construct the required testdata in the test body from the passed in arguments.

The PR https://github.com/arrow-kt/arrow/pull/2980 has introduced an Arb that can be used to generate Maps that share a number of keys. However this Arb did not yet support null values.

In this PR the mentioned Arb is modified so it can also be used to generate Maps that have null values. This Arb is then used in the 'zipN with nullable' tests to replace the testdata setup in the test body.

> https://github.com/arrow-kt/arrow/pull/2980#issuecomment-1482758449
@serras I have added a fixed RandomSource to ensure the tests for the Arb won't be flaky.

Issue #2908, opened 2023-01-27
Investigate 25 slowest test in Arrow


nomisRev:
This is a list of the 25 slowest tests in Arrow reported by Gradle Scan. Feel free to pick any of these tests and comment below which one you're working on. A lot of these tests can probably be refactored to run a lot faster without impacting coverage.

- [ ]  flatMapMapKTest:arrow-core:iosX64Test | PASSED | 37.424s
- [ ] repeatAsFlow is stack-safeScheduleTest:arrow-fx-coroutines:iosX64Test | PASSED | 17.292s
- [ ] ior bindIorSpec.Concurrent - arrow:arrow-core:iosX64Test | PASSED | 13.377s
- [ ] Iso Law: compose modifyMapTest:arrow-optics:iosX64Test | PASSED | 10.871s
- [ ] (1) Lens law: compose modifyNonEmptyListTest:arrow-optics:iosX64Test | PASSED | 10.426s
- [ ] (1) Lens law: set getNonEmptyListTest:arrow-optics:iosX64Test | PASSED | 10.008s
- [ ] Iso Law: compose modifyMapTest:arrow-optics:watchosX64Test | PASSED | 9.866s
- [ ] (1) Lens law: is set idempotentNonEmptyListTest:arrow-optics:iosX64Test | PASSED | 8.184s
- [ ] can align sequences - 3SequenceKTest:arrow-core:iosX64Test | PASSED | 7.150s 
- [ ] can align sequences - 1SequenceKTest:arrow-core:iosX64Test | PASSED | 7.075s
- [ ] DSL is generated for complex model with Every()arrow.optics.plugin.DSLTests:arrow-optics-ksp-plugin:test | PASSED | 39.170s
- [ ] should be stack safe for successful async tasksarrow.fx.coroutines.CircuitBreakerTest:arrow-fx-coroutines:jvmTest | PASSED | 11.974s
- [ ] should be stack safe for successful async tasksarrow.fx.resilience.CircuitBreakerTest:arrow-fx-resilience:jvmTest | PASSED | 11.136s
- [ ] should return IOError when no responsearrow.retrofit.adapter.either.networkhandling.NetworkEitherCallAdapterTestSuite:arrow-core-retrofit:test | PASSED | 10.889s
- [ ] DSL is generated for complex model with At()arrow.optics.plugin.DSLTests:arrow-optics-ksp-plugin:test | PASSED | 10.293s
[...]

PR #3005, opened 2023-03-29
arrow-core: speedup 'Arb.string()' related slow tests

abendt:
Some of the tests in arrow-core run slow on the non jvm platforms (https://github.com/arrow-kt/arrow/issues/2908)

One reason seems to be the use of `Arb.string()` especially when used inside arbs that generate collections (map, list, nonemptylist) 

See
* https://github.com/arrow-kt/arrow/pull/2999#issuecomment-1487382990
* https://github.com/arrow-kt/arrow/pull/2999#issuecomment-1488061474

This MR replaces `Arb.string()` with `Arb.int()` in some tests to speed up the execution time


abendt:
Before https://scans.gradle.com/s/ogbwpswsaig4q/tests/overview

image

After https://scans.gradle.com/s/i6grm52lc7v6u/tests/overview

image

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/continuations/IorSpec.kt
```diff
@@ -32,11 +33,11 @@ class IorSpec :
     }
 
     "Concurrent - arrow.ior bind" {
-      checkAll(Arb.list(Arb.string()).filter(List<String>::isNotEmpty)) { strs ->
+      checkAll(Arb.list(Arb.int(), 1 .. 100)) { xs ->
```
nomisRev, line 36:
```suggestion
      checkAll(Arb.nonEmptyList(Arb.int())) { xs ->
```

nomisRev (approved):
Looks awesome @abendt. Just two small nits (probably missing proper imports, so suggestions is going to break).

abendt:
> A bit related, but also unrelated to this. I am a bit curious if running the tests on latest JDK would increase build times as well. IIRC only building the binaries in the `publish` step need to run in JDK 11? Related tweet, https://twitter.com/EsaFirm/status/1641083869083602944?s=20

Sure, that could prohably improve build times. More ideas:
* configuration one of the gradle plugins is really slow. This makes it very painful to run tests in IntelliJ. You can also see that in the build scan: https://scans.gradle.com/s/i6grm52lc7v6u/performance/configuration. I commented out the plugin locally (https://github.com/arrow-kt/arrow/blob/main/build.gradle.kts#L34)
* In other projects i have used the gradle build cache with good results

serras:
> Sure, that could prohably improve build times. More ideas:
> 
> * gradle configuration takes a long time because of one of the gradle plugins. This makes it very painful to run tests in IntelliJ. You can also see that in the build scan: https://scans.gradle.com/s/i6grm52lc7v6u/performance/configuration. I had to comment out the plugin locally (https://github.com/arrow-kt/arrow/blob/main/build.gradle.kts#L34)

We actually are working on improving that in https://github.com/arrow-kt/arrow/pull/2988, although we may fix it once and for all in the Gradle plug-in we use https://github.com/arrow-kt/arrow-gradle-config/pull/69

franciscodr (approved):
Thanks for helping us improve the build times, @abendt I appreciate the time you're putting on this!

PR #3006, opened 2023-03-30
Creating a new module for Arrow Resilience

franciscodr:
This pull request moves the Arrow Resilience module out from Arrow FX because there are no dependencies between both modules.

As this will be a new module, there is no reason to keep the deprecated methods (already existing in the Arrow FX Coroutines module), so we are removing them in this pull request too.

Additionally, we have changed the type of the `resetTimeout` and `maxResetTimeout` parameters to `kotlin.time.Duration` for the `CircuitBreaker` class.


rafaparadela:
Task linked: [CU-865c1pd4w Move Arrow Resilience out from Arrow FX](https://app.clickup.com/t/865c1pd4w)

nomisRev (approved):
Thanks @franciscodr. Looks good to me!

PR #3007, opened 2023-03-30
Create action to auto update a PR when destination branch is pushed

gutiory:
This PR creates a Github Action that uses [autoupdate](https://github.com/marketplace/actions/auto-update) to keep all opened PRs up-to-date whenever a push is done to the PR destination branch

### DIFF
--- a/.github/workflows/auto-update-pr.yaml
+++ b/.github/workflows/auto-update-pr.yaml
@@ -1,0 +1,14 @@
+name: auto-update-pr
+on:
+  push: {}
+jobs:
+  autoupdate:
+    name: Auto update PR
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: docker://chinthakagodawita/autoupdate-action:v1
+        env:
+          GITHUB_TOKEN: "${{ secrets.TOKEN_GITHUB_ACTION }}"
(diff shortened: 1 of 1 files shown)


serras:
Is there a way to restrict this to *not* auto-update `arrow-2`? That's a difficult thing to do most of the times, so it would be preferrable to always merge there by hand.

nomisRev:
>  PR_FILTER: "ready_for_review"

@serras I guess this is only applied to _ready for review_ PRs. So it would filter out `arrow-2`.

Looks great to me @gutiory! Thank you.

nomisRev (approved):

.github/workflows/auto-update-pr.yaml
```diff
@@ -0,0 +1,14 @@
+name: auto-update-pr
+on:
+  push: {}
+jobs:
+  autoupdate:
+    name: Auto update PR
+    runs-on: ubuntu-22.04
```
franciscodr, line 7: Is there a reason to tie this workflow to that Ubuntu version?
```suggestion
    runs-on: ubuntu-latest
```

franciscodr (approved):
Minor comment. Looks good to me!

PR #3013, opened 2023-03-31
Bring back `Iso`

serras:
After discussing it with @nomisRev, it seems that we were too eager in removing `Iso` from `arrow-optics`. This PR:
- Brings back the `Iso` type, but under the new hierarchy;
- Adds code to `arrow-optics-ksp` to generate `Iso`s for `value class`es.


nomisRev (approved):
Thanks @serras

PR #3014, opened 2023-03-31
Move functions to arrow functions

nomisRev:
This is a source-compatible migration of `curried`, `uncurried`, `partial`, `memoise`, and `andThen` to make them explicit _opt-in_ in 2.0.


serras (approved):

FooIbar:
It seems arrow functions is not included in BOM.

nomisRev:
Hey @FooIbar,
Are you relying on 2.0.0-SNAPSHOT?

This change won’t be released for a long time. Which functions are you using of this change?

FooIbar:
> Hey @FooIbar,
> 
> Thank you for testing 2.0.0-SNAPSHOT, and reporting this. I’ll fix the BOM!
> 
> Strange `andThen` shouldn’t have gotten removed, it also moved to `arrow-functions`.
> 
> What is the reason for your `exclude` workaround for arrow-core? 🤔

They are still here https://github.com/arrow-kt/arrow/blob/arrow-2/arrow-libs/core/arrow-core/src/jvmMain/kotlin/arrow/core/composition-jvm.kt.

So it'll cause duplicate class error.
`Duplicate class arrow.core.AndThen0 found in modules arrow-core-jvm-2.0.0-SNAPSHOT (io.arrow-kt:arrow-core-jvm:2.0.0-SNAPSHOT:20230331.160414-65) and arrow-functions-jvm-2.0.0-SNAPSHOT (io.arrow-kt:arrow-functions-jvm:2.0.0-SNAPSHOT:20230331.160414-5)`

PR #3015, opened 2023-03-31
Result & Sequence missing deprecation and revert post-fixed names

nomisRev:
This PR adds missing deprecations for `Result`, and `Sequence`.

As discussed on Slack it also reverts the `Sequence` methods that were duplicated with post fixes, in favour of breaking the API in 2.0 from `Sequence` to `List` since those are mostly source-compatible for most intents and purposes. This allows us to keep the original names, instead of names with postfixes after 2.0


franciscodr (approved):
Looks good to me. I think we need to update the knit files, don't we?

gutiory (approved):

PR #3016, opened 2023-03-31
Fix sanitization of keywords

### LINKED ISSUE #2996: ["BUG"]: Expecting a top level declaration
If the package contains reserved keywords like `internal` or `data`, it will throw this error:
image

The error is gone after I deleted all of the backtick (`) character

serras:
Fixes #2996


nomisRev (approved):
Thank you @serras 🙌

PR #3023, opened 2023-04-03
Fix kotest duplicate testnames warning

abendt:
There are some kotest warnings about duplicated test names: `WARN: Duplicated test name ... . To disable this message, set DuplicateTestNameMode to Silent.`

This MR fixes them by adding a name param to the Laws classes.


nomisRev (approved):
Thank you again @abendt to look into these things  🙌

PR #3040, opened 2023-04-13
add some Iterable tests (#2894)

jsoizo:
This is an additional test to increase coverage of Iterable.
I will add a few more except for the ones with @deprecated annotations.


nomisRev (approved):
Looks great! Thank you @jsoizo 🙌 

> I will add a few more except for the ones with @deprecated annotations.

Do you want to add them in this PR, or another one? Whatever works best for you is fine ☺️

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
```diff
@@ -11,12 +11,7 @@ import io.kotest.matchers.nulls.shouldBeNull
 import io.kotest.matchers.nulls.shouldNotBeNull
 import io.kotest.property.Arb
 import io.kotest.matchers.shouldBe
-import io.kotest.property.arbitrary.boolean
-import io.kotest.property.arbitrary.int
-import io.kotest.property.arbitrary.list
-import io.kotest.property.arbitrary.orNull
-import io.kotest.property.arbitrary.pair
-import io.kotest.property.arbitrary.string
+import io.kotest.property.arbitrary.*
```
nomisRev, line 14: We don't use `*` imports in the Arrow codebase.

jsoizo: > We don't use * imports in the Arrow codebase.

I've put that back in.

> Do you want to add them in this PR, or another one?

I will create another one.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -1265,6 +1265,21 @@ public fun <A> Iterable<A>.fold(MA: Monoid<A>): A =
 public fun <A, B> Iterable<A>.foldMap(MB: Monoid<B>, f: (A) -> B): B =
   fold(MB.empty()) { acc, a -> MB.run { acc.combine(f(a)) } }
 
+/**
+ * Applies function [f] to each element
+ * and returns a list for the applied result Iterable<B>.
```
jsoizo, line 1270: I am not sure if this explanation is adequate and would be happy to receive a revised proposal.
The same goes for crosswalkMap and crosswalkNull.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -1287,6 +1317,21 @@ public fun <A, K, V> Iterable<A>.crosswalkMap(f: (A) -> Map<K, V>): Map<K, List<
...
+ * import io.kotest.matchers.shouldBe
+ *
+ * fun test() {
+ *   val ints = listOf(1, 2)
+ *   val res = ints.crosswalkNull { i -> if (i % 2 == 0) "x${i}" else null }
+ *   res shouldBe listOf("x2")
+ * }
+ * ```
+ */
 public fun <A, B> Iterable<A>.crosswalkNull(f: (A) -> B?): List<B>? =
```
jsoizo, line 1335: Passing `emptyList()` as the first argument to fold will ensure that `bs` is always non-null, so the results of `Ior.fromNullables` and `croswalkNull` will also be non-null. If you expect `Iterable<A>` to be empty or the result of applying the function `f` to it will all be null, it seems better that the first argument of fold is also null.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -1265,6 +1265,21 @@ public fun <A> Iterable<A>.fold(MA: Monoid<A>): A =
...
+ * and returns a list for the applied result Iterable<B>.
+ *
+ * ```kotlin
+ * import arrow.core.crosswalk
+ * import io.kotest.matchers.shouldBe
+ *
+ * fun test() {
+ *   val ints = listOf(1, 2)
+ *   val res = ints.crosswalk { i -> listOf("a${i}", "b${i}", "c${i}") }
+ *   res shouldBe listOf(listOf("a2", "a1"), listOf("b2", "b1"), listOf("c2", "c1"))
```
jsoizo, line 1279: The third argument to `ior.fold`, `listOf(l) + r`, makes the list appear to be in reverse order of the original list. I would think that `r + listOf(l)` would be a better behavior, but what do you think?

```kotlin
f(a).align(bs) { ior ->
      ior.fold(
        { listOf(it) },
        ::identity,
        { l, r -> listOf(l) + r }
      )
    }
```

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
jsoizo: @nomisRev いくつかのテストとkDocを追加したので再度レビューいただけますでしょうか?
また、crosswalkの実装で気になった点があり質問をコメントで残しております。

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
jsoizo: @serras Thank you for your reply.

> Does this make sense?

I have the same understanding as you have written.

> Does this coincide with the behavior here?

However, the behavior seems not to coincides with the understandings.
The first and third case of following tests fails.

```kotlin
class CrosswalkNullTest : FunSpec({
    test("return null if f(it) returns null") {
        val list = listOf(1, 2, 3)
        val result = list.crosswalkNull { null }
        // This test fails because result is emptyList()
        result shouldBe null
    }

    test("return emptyList() if Iterable is empty") {
        val list = emptyList<Int>()
        val result = list.crosswalkNull { it.toString() }
        // This test passes
        result shouldBe emptyList()
    }

    test("return non-empty if Iterable is non-empty and f(it) returns not-null") {
        val list = listOf(1, 2, 3)
        val result = list.crosswalkNull { it.toString() }
        // This test fails because result is listOf("3", "2", "1")
        result shouldBe listOf("1", "2", "3")
    }
})
```

serras:
@jsoizo I've talked with @nomisRev, and we've agreed that it makes sense to fix the different `crosswalk` implementations, based on your findings in this PR.

Issue #3054, opened 2023-05-15
["Request"] CyclicBarrier cancellation/reset


HSAR:
What version are you currently using?
1.1.5

What would you like to see?
`java.util.concurrent.CyclicBarrier` supports interrupting awaited threads by offering a `reset()` method that interrupts all awaiting threads with a CancellationException. Something similar would be very useful in the implementation offered in this library: when called, it should interrupt all coroutines currently awaiting and cancel them.

HSAR:
I have had an attempt at implementing this, see below. I can submit a PR and figure out the tests if that would be useful.

I added a boolean to the `State` class and check it at the start of the loop call during `await()`.

```
class CancellableCyclicBarrier(private val capacity: Int) {

    private data class State(
        val awaiting: Int,
        val epoch: Long,
        val reset: Boolean,
        val unblock: CompletableDeferred<Unit>
    )

    private val state: AtomicRef<State> = AtomicRef(State(capacity, 0, false, CompletableDeferred()))

    fun reset() {
        state.update { original -> original.copy(reset = true) }
    }

    /**
     * When [await] is called the function will suspend until the required number of coroutines have reached the barrier.
     * Once the [capacity] of the barrier has been reached, the coroutine will be released and continue execution.
     */
    suspend fun await() {
        state.loop { original ->
            if (original.reset) {
                throw CancellationException("CyclicBarrier was reset.")
            }

            val (awaiting, epoch, reset, unblock) = original
            val awaitingNow = awaiting - 1
            if (awaitingNow == 0
                && state.compareAndSet(original, State(capacity, epoch + 1, reset, CompletableDeferred()))) {
                unblock.complete(Unit)
                return
            } else if (state.compareAndSet(original, State(awaitingNow, epoch, reset, unblock))) {
                return try {
                    unblock.await()
                } catch (cancelled: CancellationException) {
                    state.update { s -> if (s.epoch == epoch) s.copy(awaiting = s.awaiting + 1) else s }
                    throw cancelled
                }
            }
        }
    }
}

```

nomisRev:
Hey @HSAR,

Such functionality would be great if we can support more use-cases from `java.util.concurrent.CyclicBarrier`!

We'd need to write some test to check for edge-cases. I need to look at this a couple of times more to figure out if we're missing anything 😅 On first sight it would seem to me that `reset` needs to be checked first in the `when` block.

And can `compareAndCycleState` not result in a race condition in the `reset` function? 🤔

[...]

PR #3055, opened 2023-05-17
Add reset and barrierAction to CyclicBarrier.

HSAR:
Both reset() and getNumberWaiting() are methods that exist on java.util.concurrent.CyclicBarrier and function in the same way.

This PR comes from discussion on issue #3054.


HSAR:
I'm opening this PR in the hope of starting a conversation about the implementation and tests, both of which I'm aware could do with some improvement.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/CyclicBarrier.kt
```diff
@@ -23,28 +24,64 @@ public class CyclicBarrier(public val capacity: Int) {
...
-        unblock.complete(Unit)
-        return
-      } else if (state.compareAndSet(original, State(awaitingNow, epoch, unblock))) {
-        return try {
-          unblock.await()
-        } catch (cancelled: CancellationException) {
-          state.update { s -> if (s.epoch == epoch) s.copy(awaiting = s.awaiting + 1) else s }
-          throw cancelled
+      when {
+        reset -> {
```
HSAR, line 70: I think that the `resetComplete` field on `State` could be removed with a call to `unblock.cancel()` but this doesn't seem to work the way I expect - other threads awaiting are not cancelled when I do this.

```
                reset && awaiting == capacity && compareAndCycleState(original) -> {
                    val cancelException = CancellationException("CyclicBarrier was reset.")
                    unblock.cancel(cancelException)
                    throw cancelException
                }
```

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/CyclicBarrier.kt
```diff
@@ -23,28 +24,64 @@ public class CyclicBarrier(public val capacity: Int) {
...
+
+  /** The number of parties currently waiting. **/
+  public val numberWaiting: Int
+    get() = capacity - state.get().awaiting
+
+  /**
+   * When called, all waiting coroutines will be cancelled with [CancellationException].
+   * When all threads have been cancelled the barrier will cycle.
+   */
+  public suspend fun reset() {
```
HSAR, line 48: I'm not happy that this method is a suspend function, and I would like to be able to remove this, using a similar logic to what currently exists to cycle the barrier when capacity is reached. My first attempt at doing that failed, however, as coroutine resumption seems to get in the way.

Further explanation - my implementation was such that I check `if (reset && awaiting == capacity) then state.update(s.copy(reset = false))`.  This doesn't work because the first coroutine to resume after `reset = true` is the oldest coroutine, which (I believe) has a view of state that is empty, so it immediately sets `reset = false`. Other coroutines then never see the reset.

Happy to be told I have no idea what I'm saying! I'm a beginner to coroutine concurrency.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/CyclicBarrier.kt
HSAR: Resolved with [89457f5](https://github.com/arrow-kt/arrow/pull/3055/commits/89457f5bd5c34ad21476fa024c4e95b0a8631475).

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/CyclicBarrier.kt
HSAR: Resolved with [89457f5](https://github.com/arrow-kt/arrow/pull/3055/commits/89457f5bd5c34ad21476fa024c4e95b0a8631475).

nomisRev:
@HSAR thank you so much for the PR! I need to make some time to properly look at these changes, sorry it's been taking so long but it's not a PR I can quickly review 😅

On first sight it looks pretty good, but I need to dig around a bit and perhaps write some tests for myself. Concurrency is hard 😅

HSAR:
No problem. Keep me posted if/when you have questions or comments!

nomisRev:
Hey @HSAR,

There are some tests hanging on CI, and also for me locally. Not sure if you have time to look into it, but I am also taking a look at it whilst reviewing the code.

HSAR:
Hi @nomisRev - happy to take a look. I had thought that everything was passing when I left it last time, but I am very unfamiliar with this test framework and it is possible that I wasn't running it correctly.

nomisRev:
@HSAR I think we lost some history due to force-push, because I remember it passing at some time. You should be able to run the tests from IDEA. I think you can ignore `checkAll` for these tests too if you want. Or write them in `Kotlin-test` if you're more familiar with that. I'd prefer to have better tests, having some style difference in them is low-tech debt in favour of better tests IMO.

I can always help you refactor them to Kotest if really need to.

HSAR:
I expected to be able to run the tests from IDEA, but that doesn't seem to be the case (at least, not by default), which rather puzzled me. I'm personally only familiar with JUnit, but I am perfectly happy with any test framework if I can get it to run properly in my IDE. Are there extra steps I need to follow?

Regarding the review, I restructured the code more than once to try and avoid problems with the reset action (which has turned out to be really rather tricky). This one works in many of scenarios but I don't think it is watertight - I am worried for instance about StateMachine::startNextEpoch which seems like it is not nearly atomic enough.

nomisRev:
Hey @HSAR,

I am taking a stab at it from scratch, to see how far I can get it. That's easier for me to get my head around it 😅 I'll keep you posted today or tomorrow, so we can discuss code in further detail.

To run Kotest you should install the IDEA Kotest plugin, but it sometimes doesn't behave perfectly. I can for example not run individual tests, only entire test files which is kind-of unfortunate. I've contemplated refactoring to JUnit Kotlin-test to leverage things like WASM, but that seems like a too big of an effort for now.

HSAR:
I might hold off trying to fix the tests, as it seems entirely possible we might junk the current implementation anyway. Happy to chat code if it would help.

nomisRev:
Hey @HSAR,

I played a little bit around with the implementation and came up with #3076  What do you think? Not sure what the best approach is, because I don't want to throw out your effort. I think you should get credit for leading this in the git history.

HSAR:
I like the code inside `::reset` and the use of a specific CancellationException, both are a lot tidier. A few things that do come to mind, though:
* The epoch doesn't advance when a reset happens. That is of course part of the internal logic and state of CyclicBarrier so is acceptable from an external API point of view, I just thought it might be useful to point out. The two State possibilities should, I presume, protect from any thread-unsafe-ness.
* Keeping the `barrierAction` feature would be quite desirable: my implementation used it as a neat way of implementing the reset but that was not the main reason it existed, which was both because it was a useful feature and because it was part of the original java.util.concurrent CyclicBarrer.

As for the git history, the code is yours and so should be the git blame ;)
Jokes aside I won't lie, it would be nice to be recognised for contribution. Perhaps if you agree with my second point then I can add the `barrierAction` and thus obtain a little credit.

nomisRev:
Hey @HSAR,
I am closing my other PR. I merged the changes and tried to keep as much as your changes as possible to give you proper credit for driving this effort 🙏 It was not easy 😅

Sorry the entire process was so extraordinarily long.. Please do a thorough review of all my work, and I think with a review from @serras or @franciscodr we should be good to merg this work 🙌

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/CyclicBarrier.kt
```diff
@@ -16,37 +17,143 @@ import kotlinx.coroutines.CompletableDeferred
...
-          unblock.await()
-        } catch (cancelled: CancellationException) {
-          state.update { s -> if (s.epoch == epoch) s.copy(awaiting = s.awaiting + 1) else s }
-          throw cancelled
+    state.loop { state ->
+      when (state) {
+        is Awaiting -> {
+          val (awaiting, epoch, unblock) = state
+          val awaitingNow = awaiting - 1
+          println("awaitingNow: $awaitingNow")
```
HSAR, line 99: Might have missed removing this debug statement.

HSAR:
Hey @nomisRev,

I was a bit confused because I had pushed a different, more-conservative merge not too long ago. LGTM apart from one comment which I've left inline.

I'm glad to have contributed! Thanks for your time looking into the issue and understanding the problems.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/CyclicBarrier.kt
```diff
@@ -16,37 +17,143 @@ import kotlinx.coroutines.CompletableDeferred
...
   }
-  
-  private data class State(val awaiting: Int, val epoch: Long, val unblock: CompletableDeferred<Unit>)
-  
-  private val state: AtomicRef<State> = AtomicRef(State(capacity, 0, CompletableDeferred()))
-  
+
+  private sealed class State {
+    abstract val epoch: Long
+  }
```
serras, line 33:
```suggestion
  private interface State {
    val epoch: Long
  }
```

arrow-libs/fx/arrow-fx-coroutines/src/commonTest/kotlin/arrow/fx/coroutines/CyclicBarrierSpec.kt
```diff
@@ -9,10 +9,7 @@ import io.kotest.property.Arb
 import io.kotest.property.arbitrary.constant
 import io.kotest.property.arbitrary.int
 import io.kotest.property.checkAll
-import kotlinx.coroutines.CompletableDeferred
-import kotlinx.coroutines.CoroutineStart
-import kotlinx.coroutines.cancelAndJoin
-import kotlinx.coroutines.launch
+import kotlinx.coroutines.*
```
serras, line 12: This needs to be changed to a full list, otherwise Spotless will not like it

arrow-libs/fx/arrow-fx-coroutines/src/commonTest/kotlin/arrow/fx/coroutines/CyclicBarrierSpec.kt
```diff
@@ -21,54 +18,99 @@ class CyclicBarrierSpec : StringSpec({
...
-  
+
+  "executes runnable once full" {
+    checkAll(Arb.constant(Unit)) {
+      var barrierRunnableInvoked = false
+      val barrier = CyclicBarrier(2) { barrierRunnableInvoked = true }
+      parZip({ barrier.await() }, { barrier.await() }) { _, _ -> }
+      barrier.capacity shouldBe 2
+      barrierRunnableInvoked shouldBe true
+    }
```
serras, line 51: No need to wrap in `checkAll` if you're not using the generator at all

```suggestion
    var barrierRunnableInvoked = false
    val barrier = CyclicBarrier(2) { barrierRunnableInvoked = true }
    parZip({ barrier.await() }, { barrier.await() }) { _, _ -> }
    barrier.capacity shouldBe 2
    barrierRunnableInvoked shouldBe true
```

serras (approved):
Nice job!

nomisRev:
Thank you for the awesome work @HSAR! Congrats on your first contribution 🙌 Very impressive work! 👏

HSAR:
Happy to have my contribution accepted. Thanks for your part in this too!

PR #3058, opened 2023-05-25
Update versions

nomisRev:
1.2.0-RC has been out for almost two months now, and we've only received positive reactions.
Several people have also shared that getting rid of deprecated code was quite easy, so let's plan the stable release for Arrow 1.2.0.

As a first step let's update the dependencies in Arrow to latest version.


raulraja (approved):
thanks @nomisRev !

PR #3059, opened 2023-05-26
Add withError and (Eager)Effect.mapError

### LINKED ISSUE #3053: ["Request"] Error mapper utility
Since using Arrow, I tend to have a sealed class hierarchy of failures for each computation my app may execute. It is common for computations to execute other computations as a step.

For example, let's say we have some kind of parsing operation:
```kotlin
sealed class ParsingError {
    object ErrorReason1 : ParsingError()
    object ErrorReason2 : ParsingError()
}

fun parse(…): Either<ParsingError, ParsingResult> = …
```
And we have another operation, of which parsing is one of the steps:
```kotlin
sealed class ParentError {
    object ErrorReason1 : ParentError()

    // Encapsulate the child error
    data class ParsingError(val error: ParsingError) : ParentError()
}

fun parent(…): Either<ParentError, ParentResult> = either {
    ensure(…) { ParentError.ErrorReason1 }
    
    // Here, how should we call `parsing(…)` to propagate errors correctly?
}
```

### Current solution: mapLeft

```kotlin
parsing(…)
    .mapLeft(ParentError::ParsingError)
    .bind()
```

This is easy to read, but it's quite verbose, especially when context receivers will remove all other `.bind()` calls. It also only works if the parsing operation returns `Either`, or it requires one more conversion function.

### Current solution: recover

```kotlin
recover({ parsing(…).bind() }) { raise(ParentError.ParsingError(it)) }
```

Again, it's very bulky. The `recover` DSL is quite powerful, but this is such a simple case that it's mostly boilerplate.

### Proposed new feature: mapErrors

To simplify this case, it would be nice to provide a specific function for this:
```kotlin
mapErrors(ParentError::ParsingError) {
    parsing(…).bind()
}
```

> The discussion [started on Slack](https://kotlinlang.slack.com/archives/C5UPMM0A0/p1684066366615369).

kyay10:
Fixes #3053 
I'm not sure if `mapError` is the best name for it, but `withError` felt weird as an extension on `Effect`. Also, mapError's transform parameter seems too "busy" with `suspend` and `Raise`. At least the `Raise<OtherError>` could be removed, but then someone wanted to call a function inside that raises `OtherError`, they'd need to wrap in `either{ ... }.merge()` or something along those lines, and so I figured exposing the `Raise<OtherError>` might help in that case. Maybe there should be a `Raise` builder that merges the error and result automatically? This could be helpful especially if they're subclasses of each other.

@CLOVIS-AI

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/ErrorHandlers.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/ErrorHandlers.kt
@@ -116,7 +116,7 @@ public suspend inline infix fun <Error, A> Effect<Error, A>.getOrElse(recover: s
  * ```
  * <!--- KNIT example-effect-error-04.kt -->
  */
-public infix fun <Error, OtherError, A> Effect<Error, A>.mapError(@BuilderInference transform: suspend (error: Error) -> OtherError): Effect<OtherError, A> =
+public infix fun <Error, OtherError, A> Effect<Error, A>.mapError(transform: suspend (error: Error) -> OtherError): Effect<OtherError, A> =
   effect { withError({ transform(it) }) { invoke() } }
 
 public infix fun <Error, OtherError, A> EagerEffect<Error, A>.recover(@BuilderInference recover: Raise<OtherError>.(error: Error) -> A): EagerEffect<OtherError, A> =
@@ -152,5 +152,5 @@ public inline infix fun <Error, A> EagerEffect<Error, A>.getOrElse(recover: (err
  * ```
  * <!--- KNIT example-effect-error-05.kt -->
  */
-public infix fun <Error, OtherError, A> EagerEffect<Error, A>.mapError(@BuilderInference transform: (error: Error) -> OtherError): EagerEffect<OtherError, A> =
+public infix fun <Error, OtherError, A> EagerEffect<Error, A>.mapError(transform: (error: Error) -> OtherError): EagerEffect<OtherError, A> =
   eagerEffect { withError({ transform(it) }) { invoke() } }
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/ErrorHandlers.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/ErrorHandlers.kt
@@ -11,7 +11,7 @@ import kotlin.jvm.JvmName
 /**
  * Catch the raised value [Error] of the `Effect`.
  * You can either return a value a new value of [A],
- * or short-circuit the effect by raising with a value of [Error],
+ * or short-circuit the effect by raising with a value of [OtherError],
  * or raise an exception into [suspend].
  *
  * ```kotlin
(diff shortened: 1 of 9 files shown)


kyay10:
A `merge` builder looks very neat, and is dead-simple to make:
```
@RaiseDSL
public inline fun <A> merge(
  @BuilderInference block: Raise<A>.() -> A,
): A = recover(block, ::identity)

// Usage
val error = eagerEffect<Error, User> { raise(Error) }
val bool = false
val a = error.mapError { merge { if(bool) "some-failure" else raise("other-failure") } } // Raise(other-failure)
```
Maybe this should be a thing :eyes:. It's useful when you have a lot of functions that use `Raise`, and you want to also return a value of the same error that they might raise.

hoc081098:
Nice, mapError is my favorite method when using Either, Effect should have it

CLOVIS-AI:
What is the use-case for `mapError`?
- `withError`: simple 1-1 pure mapping between error types
- `recover`: error management, as complex as you'd like

`mapError`'s lambda being both `Raise<OtherError>` and `-> OtherError` seems confusing to me. `Raise<OtherError>` is already provided by `recover`, `-> OtherError` is provided by `withError`. I don't see the use case for needing both?

kyay10:
@CLOVIS-AI `mapError` is meant to be `Effect`'s analogue for `withError`. Think of it as the `Effect` version of `Either.mapLeft`. Its `transform` lambda taking `Raise<OtherError>` and returning `OtherError` is very ugly, I agree. The reason I exposed it is so that you can use functions that raise `OtherError` inside. However, with #3061, that need is met without the ugly signature. I'll update the PR with a better signature, especially because you can simulate merge with `either { ... }.merge()`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/ErrorHandlers.kt
```diff
@@ -96,6 +96,29 @@ public fun <Error, A> Effect<Error, A>.catch(): Effect<Error, Result<A>> =
...
+ *
+ * val error = effect<Error, User> { raise(Error) } // Raise(error)
+ *
+ * val a = error.mapError<Error, String, User> { error -> "some-failure" } // Raise(some-failure)
+ * val b = error.mapError<Error, String, User>(Any::toString) // Raise(Error)
+ * val c = error.mapError<Error, Nothing, User> { error -> throw RuntimeException("BOOM") } // Exception(BOOM)
+ * ```
+ * <!--- KNIT example-effect-error-04.kt -->
+ */
+public infix fun <Error, OtherError, A> Effect<Error, A>.mapError(@BuilderInference transform: suspend (error: Error) -> OtherError): Effect<OtherError, A> =
```
nomisRev, line 119: **Nit:** `@BuilderInference` is no longer being used here.

```suggestion
public infix fun <Error, OtherError, A> Effect<Error, A>.mapError(transform: suspend (error: Error) -> OtherError): Effect<OtherError, A> =
```

nomisRev (approved):
Looks good to me @kyay10! Thank you for this improvement, and @CLOVIS-AI for the original suggestions 🙏 🙌

Thanks for the very complete work, including documentation and tests 😍

franciscodr (approved):
Thanks for this amazing contribution, @kyay10. The code looks neat with the new functions 😍

serras (approved):
Very useful utility

PR #3061, opened 2023-05-28
Add merge builder for raise

kyay10:
I re-implemented some of Effect/s methods using the merge builder to check its functionality.

Note for the future: when/if union types are in the language, merge should likely be deprecated in favour of a `union` builder that has separate types for the returned and raised types (so that they're inferred separately) and returning a union of both.


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -25,7 +25,7 @@ public inline fun <Error, A> either(@BuilderInference block: Raise<Error>.() ->
   fold({ block.invoke(this) }, { Either.Left(it) }, { Either.Right(it) })
 
 public inline fun <A> nullable(block: NullableRaise.() -> A): A? =
-  fold({ block(NullableRaise(this)) }, { null }, ::identity)
+  merge { block(NullableRaise(this)) }
```
serras, line 28: I don't fully understand why this works...

kyay10: TL;DR: `nullable` is just secretly `merge` because it's taking `Raise<Nothing?>.() -> A` and returning `A?`, which is the supertype of `Nothing?` and `A`.
Note also that instead of `{ null }` in the original code, you could've had `::identity`, and hence `merge(blah)` == `fold(blah, ::identity, ::identity)`

Merge returns the supertype of the raised error and the returned value. The raised error here is `Null`, which is `typealias Null = Nothing?`, while the return type is `A`, hence their common supertype is `A?`. In other words, `nullable` is a general example of returning the common supertype of the raised and the returned (except of course that we have `NullableRaise` to provide nice binds for related types like Option due to the lack of context receiver support right now)

Side note: In the future when we have union types, this pattern would work with a lot more types, but right now, this just utilises the fact that `?` is the only natively supported union type in Kotlin. Btw, I think the design for union types prevents them from being automatically inferred (i.e. the compiler won't infer a type parameter `T` to a union type ever, but it respects explicit mentions of union types) and so a `union` builder that infers the types separately and returns a union of them would be the way to go when that happens

serras: I am amazed by the fact that you could find this (cool) approximation 🤯

kyay10: With a powerful type system, the possibilities are infinite! 🚀

serras (approved):

nomisRev (approved):

Issue #3041, opened 2023-04-13
Arrow Core 1.2.0-RC breaks any Kapt task when using Raise/Effect


miikeat:
We were already using 1.1.6-alpha.65 in one of our projects and tested the new features but when we wanted to switch to 1.2.0-RC our poject stopped building with the following error:

```
ExampleImpl.java:12: error: cannot access Raise
    public final kotlin.jvm.functions.Function2<arrow.core.raise.Raise<? super MyError>, kotlin.coroutines.Continuation<? super java.lang.String>, java.lang.Object> example() {
                                                                ^
  bad class file: /Users/kaiser/.gradle/caches/modules-2/files-2.1/io.arrow-kt/arrow-core-jvm/1.2.0-RC/815591a96f11786ae0ff746a4931fee9c0e46ac4/arrow-core-jvm-1.2.0-RC.jar(/arrow/core/raise/Raise.class)
    undeclared type variable: T
    Please remove or make sure it appears in the correct subdirectory of the classpath.
``` 

I created a minimal reproducable Example here: https://github.com/miikeat/ArrowKaptIssue

I narrowed the Issue down to the changes in alpha.73 -> alpha.74 (Diff: https://github.com/arrow-kt/arrow/compare/1.1.6-alpha.73...1.1.6-alpha.74)

And as you can see in build.gradle.kts in that project it only happens when a kapt(...) dependency is present.

```
dependencies {
    // as soon as any of the following two lines are not commented out, the above error appears with alpha.74 or higher
    kapt("com.datastax.oss:java-driver-mapper-processor:4.15.0")
    //kapt("androidx.room:room-compiler:2.5.0")

    // Diff between alpha 73 & 74: https://github.com/arrow-kt/arrow/compare/1.1.6-alpha.73...1.1.6-alpha.74
    //implementation("io.arrow-kt:arrow-core:1.1.6-alpha.73") // Last working Version
    //implementation("io.arrow-kt:arrow-core:1.1.6-alpha.74") // First Version with that Issue
    implementation("io.arrow-kt:arrow-core:1.2.0-RC")
    testImplementation(kotlin("test"))
}
```

Maybe thats a bug of Kapt but as it worked until version 1.1.6-alpha.73 I thought its probably better to post this first here!

nomisRev:
Hey @miikeat,

Thank you so much for the reproducible, and pin pointing it to the exact alpha’s 🙌
This is great, will look into it asap.

bddckr:
[...]

PR #3062, opened 2023-06-02
Change NonEmptySet type parameter name from T to A

franciscodr:
This pull request is related to the issue https://github.com/arrow-kt/arrow/issues/3041 reported by @mikeat in mid-April.

After publishing version `1.1.6-alpha.74`, Arrow started to break any Kapt task when using Raise/Effect with the following error:
```
ExampleImpl.java:12: error: cannot access Raise
    public final kotlin.jvm.functions.Function2<arrow.core.raise.Raise<? super MyError>, kotlin.coroutines.Continuation<? super java.lang.String>, java.lang.Object> example() {
                                                                ^
  bad class file: /Users/kaiser/.gradle/caches/modules-2/files-2.1/io.arrow-kt/arrow-core-jvm/1.2.0-RC/815591a96f11786ae0ff746a4931fee9c0e46ac4/arrow-core-jvm-1.2.0-RC.jar(/arrow/core/raise/Raise.class)
    undeclared type variable: T
    Please remove or make sure it appears in the correct subdirectory of the classpath.
```

Thanks to the valuable help from @serras, we discovered that the compiler was generating the wrong bytecode for the `bindAll` function defined as an extension function for `NonEmptySet`. The generated function was using an undefined type parameter `T`. You can see more details of the compiler bug in [this comment](https://github.com/arrow-kt/arrow/issues/3041#issuecomment-1573373525).

As a workaround for this compiler issue, we propose changing the type parameter used for the elements of the `NonEmptySet` from `T` to `A` to align it with the `bindAll` function.

This screenshot shows that the bytecode uses the type parameter `A` for the `Set` class now instead of `T`

Screenshot 2023-06-02 at 12 17 41

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
@@ -3,19 +3,19 @@ package arrow.core
 import kotlin.jvm.JvmInline
 
 @JvmInline
-public value class NonEmptySet<out T> private constructor(
-  private val elements: Set<T>
-) : Set<T> by elements {
+public value class NonEmptySet<out A> private constructor(
+  private val elements: Set<A>
+) : Set<A> by elements {
 
-  public constructor(first: T, rest: Set<T>) : this(setOf(first) + rest)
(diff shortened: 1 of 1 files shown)


serras (approved):

nomisRev (approved):
Thank you @serras and @franciscodr for looking into this 🙏

PR #3064, opened 2023-06-06
Add utility extension function for creating NonEmptyList/Set

nicolasfara:
This PR introduce two utilities method for creating `NonEmptyList` and `NonEmptySet` from a single value via extension function.


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -552,3 +552,5 @@ public fun <A> Iterable<A>.toNonEmptyListOrNull(): NonEmptyList<A>? =
 
 public fun <A> Iterable<A>.toNonEmptyListOrNone(): Option<NonEmptyList<A>> =
   toNonEmptyListOrNull().toOption()
+
+public fun <A> A.toNonEmptyList(): NonEmptyList<A> = nonEmptyListOf(this)
```
hoc081098, line 556: nonEmptyListOf is more readable 😁

nomisRev: Yes, that is indeed our train of thought since Kotlin Std also doesn't have `1.toList()`.

PR #3068, opened 2023-06-30
Introduce `NonEmptyCollection`

serras:
As part of [this Slack thread](https://kotlinlang.slack.com/archives/C5UPMM0A0/p1688051923191969) it seems that having a common parent for `NonEmptyList` and `NonEmptySet` could be quite useful, so we can have some common functions. For example, a function like `toNonEmptyList` can return a non-nullable type if we know that the receiver is `NonEmptyCollection`.

The definition of `NonEmptyCollection` looks as follows; those are all the functions I could think of which would benefit from the knowledge of something being non-empty.

```kotlin
public interface NonEmptyCollection<out E> : Collection<E> {
  override fun isEmpty(): Boolean = false
  public val head: E

  public operator fun plus(element: @UnsafeVariance E): NonEmptyCollection<E>
  public operator fun plus(elements: Iterable<@UnsafeVariance E>): NonEmptyCollection<E>

  public fun <B> map(transform: (E) -> B): NonEmptyCollection<B>
}
```


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyCollection.kt
```diff
@@ -0,0 +1,21 @@
...
+  public operator fun plus(elements: Iterable<@UnsafeVariance E>): NonEmptyCollection<E>
+
+  public fun <B> map(transform: (E) -> B): NonEmptyCollection<B>
+}
+
+public fun <E> NonEmptyCollection<E>.toNonEmptySet(): NonEmptySet<E> =
+  toNonEmptySetOrNull()!!
+
+public fun <E> NonEmptyCollection<E>.toNonEmptyList(): NonEmptyList<E> =
+  toNonEmptyListOrNull()!!
```
nomisRev, line 21: Shouldn't these be methods? This respects the `out` variance, right?

```suggestion
  public fun toNonEmptySet(): NonEmptySet<E> =
    toNonEmptySetOrNull()!!

  public fun toNonEmptyList(): NonEmptyList<E> =
    toNonEmptyListOrNull()!!
}
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyCollection.kt
```diff
@@ -0,0 +1,21 @@
+package arrow.core
+
+/**
+ * Common interface for collections that always have
+ * at least one element (available from [head]).
+ */
+public interface NonEmptyCollection<out E> : Collection<E> {
+  override fun isEmpty(): Boolean = false
+  public val head: E
```
nomisRev, line 9: I saw in the Slack thread you mentioned `first()`. Wondering if it's worth deprecating `head` to `first()` or if that would cause confusion with the Kotlin Std method `first` that throws `OutOfBoundsException` 🤔

nomisRev (approved):
Just one nit, if it's possible to define functions as methods I think we should prefer it. Results in better discoverability, auto-suggest, less imports, etc.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyCollection.kt
serras: The reason I didn't do this is because `first` is defined as an extension method, and I was a bit unsure of how the shadowing of those methods would work. And in fact, using the one in `kotlin.collections` is fine, we just know we don't get an exception...

Something more interesting would be overriding `firstOrNull()` with a non-nullable type, but I'm not sure how many people are in the situation of using `firstOrNull()` on `NonEmptyCollection` in the first place.

nomisRev: Ah right, those are extension functions defined in the Kotlin Std... If it's a method, then it should always take precedence over any extension functions but perhaps it's better not to conflict at all. `first` will in this case be "safe" anyway, so let's keep it as is.

We can override `firstOrNull` as an instance method, and it will take precedence over the `Iterable#firstNotNull` method from Kotlin Std.

serras: I've just pushed another commit where a bunch of functions are overridden to track for non-emptiness:
- `firstOrNull`, `lastOrNull`,
- `map`,
- `distinct` and `distinctBy`,
- `flatMap` and `zip` when the argument is also non-empty.

nomisRev:
> [Fix compilation problems from Java](https://github.com/arrow-kt/arrow/pull/3068/commits/501b5876d7f2704a5db9064bf580dcabb272c966)

🤕 Again this issue. Did we ever report this to YouTrack? And unrelated, should we have some guidelines for generics and align all of them? 🤔

raulraja (approved):
Thank you @serras !

PR #3072, opened 2023-07-03
Fix text repetition in EffectScope's deprecation message

### LINKED ISSUE #3043: Text repetition in EffectScope's deprecation message
Spotted on 1.2.0-RC, the deprecation warning when using `EffectScope` is
```
w: file:///[REDACTED] 'EffectScope<in R>' is deprecated. Use the arrow.core.raise.Raise type instead, which is more general and can be used to  and can be used to raise typed errors or _logical failures_
The Raise<R> type is source compatible, a simple find & replace of arrow.core.continuations.* to arrow.core.raise.* will do the trick.
```
which repeats `and can be used to  and can be used to raise`.

Looking at `EffectScope`'s source, it seems the duplication is already there. Most likely a failed copy/paste?

serras:
Fixes #3043


nomisRev (approved):
Thank you for the fix! 🙌

i-walker (approved):

i-walker:
Thanks Ale!

Issue #3039, opened 2023-04-11
["Request"] arrow.core.Eval deprecation guidance


404-:
What version are you currently using? 1.1.5

What would you like to see? Guidance on how to refactor existing usages of `arrow.core.Eval`

---

As far as I can tell, `Eval` is slated for deprecation. Which mechanism/pattern/technique is intended to replace it? As an example, how would one go about refactoring the following function:

```
import arrow.core.right
import arrow.core.Either
import arrow.core.Either.Left
import arrow.core.Either.Right
import arrow.core.Eval

fun <E, L, R> Eval.Companion.fold(
    elements: List<E>,
    zero: R,
    f: (R, E) -> Eval<Either<L, R>>
): Eval<Either<L, R>> =
    when (val head = elements.firstOrNull()) {
        null ->
            now(zero.right())
        else ->
            f(zero, head).flatMap { result ->
                when (result) {
                    is Left  -> now(result)
                    is Right -> fold(elements.drop(1), result.value, f)
                }
            }
    }
```

alexandru:
In the first example, you don't really need `Eval`, because it can be described as a simple [tail-recursive function](https://kotlinlang.org/docs/functions.html#tail-recursive-functions):

```kotlin
    tailrec fun <E, L, R> foldList(
        elements: List<E>,
        zero: R,
        f: (R, E) -> Either<L, R>
    ): Either<L, R> =
        when (val head = elements.firstOrNull()) {
            null ->
                zero.right()
            else ->
                when (val result = f(zero, head)) {
                    is Left  -> result
                    is Right -> foldList(elements.drop(1), result.value, f)
                }
        }
```

I am not part of the project, but just like with the deprecation of `IO`, the general idea might be to just use `suspend () -> A` instead of `Eval<A>`. 

```kotlin
typealias Eval<A> = suspend () -> A
```

A more interesting example would be a `foldRight`, BTW, because you'd need a lazy function parameter:

```kotlin
suspend fun <A, B> List<A>.foldRight(b: Eval<B>, f: suspend (A, Eval<B>) -> B): B {
    val list = this
    return when (val head = firstOrNull()) {
        null -> b()
        else ->
            f(head, suspend {
                list.drop(1).foldRight(b, f)
            })
    }
}
```

Note that this implementation might not be safe 🙂 as I have no idea if Kotlin's coroutine execution can do "TCO", and I think it matters.

alexandru:
@404- ah, got it, good sample then.

[...]

PR #3073, opened 2023-07-03
New idea for `Eval`

serras:
⚠️ _This doesn't work yet!_

Based on the discussion in #3039, I've tried to make a simplified implementation of `Eval` based on `DeepRecursiveFunction`. The idea is to use their `callRecursive` as suspension points to live in the heap.

```kotlin
public typealias Eval2<A> = suspend DeepRecursiveScope<Token, *>.(Token) -> A

public fun <A, B> Eval2<A>.flatMap(f: (A) -> Eval2<B>): Eval2<B> = { token ->
  val x = this@flatMap(token)
  callRecursive(token)
  f(x)(token)
}
```

Alas, the simplest test,

```kotlin
val mapped = Eval2Utils.now(0).map { sideEffect.increment(); it + 1 }
```

leads to a terrible error about exceeding the memory limits,

```
arrow.core.EvalTest[jvm] > should map wrapped value[jvm] FAILED
    java.lang.OutOfMemoryError: GC overhead limit exceeded
        at kotlin.ResultKt.createFailure(Result.kt:122)
        at _COROUTINE._BOUNDARY._(CoroutineDebugging.kt:46)
        at kotlinx.coroutines.TimeoutKt.withTimeoutOrNull(Timeout.kt:104)
        at io.kotest.engine.test.interceptors.InvocationTimeoutInterceptor.intercept(InvocationTimeoutInterceptor.kt:42)
        ...
```

I'm just putting it here in case somebody else wants to give it a thought and maybe make it work.

@nomisRev, @404- and @alexandru were involved in the discussion back then.

### DIFF
--- a/arrow-libs/core/arrow-core/api/arrow-core.api
+++ b/arrow-libs/core/arrow-core/api/arrow-core.api
@@ -443,6 +443,23 @@ public final class arrow/core/Eval$Now$Companion {
 	public final fun getUnit ()Larrow/core/Eval;
 }
 
+public final class arrow/core/Eval2Kt {
+	public static final fun flatMap (Lkotlin/jvm/functions/Function3;Lkotlin/jvm/functions/Function1;)Lkotlin/jvm/functions/Function3;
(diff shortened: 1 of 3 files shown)


nomisRev:
@serras closed 👀

PR #3075, opened 2023-07-05
Set up Spotless

### LINKED ISSUE #2522: ["Request"] Setup Spotless
What would you like to see?

Automate formatting on CI, if any incorrect format is encountered it should be automatically be fixed by CI if possible and commited. This needs to happen on PR.

https://github.com/stefanzweifel/git-auto-commit-action

serras:
Fixes #2522

- Adds Spotless to every project
- Automatic commit of formatted files using GitHub Actions
- First pass of `spotlessApply`

Some considerations:
- Spotless is set up to use ktlint as formatter; this matches better the style we've been using, and is the only formatter (AFAIK) that supports modern Kotlin features such as context receivers
- The `standard:filename` rule has been disabled because Knit generates files starting with a lowercase letter

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -240,3 +240,30 @@ jobs:
         with:
           commit_message: Update API files
           file_pattern: arrow-libs/**/api/*.api
+
+  spotless:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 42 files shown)


nomisRev (approved):
Oh, KtLint respects our 2-space indentation? I didn't know KtLint allowed that. Thanks for this PR @serras.

I am glad to have a code formatter again, and good it was possible to do this with minimal style changes 👍

PR #3076, opened 2023-07-05
Implement reset on CyclicBarrier

nomisRev:
Alternative implementation to #3055

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EffectScope.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/continuations/EffectScope.kt
@@ -15,7 +15,7 @@ import kotlin.experimental.ExperimentalTypeInference
 
 /** Context of the [Effect] DSL. */
 @Deprecated(
-  "Use the arrow.core.raise.Raise type instead, which is more general and can be used to  and can be used to raise typed errors or _logical failures_\n" +
+  "Use the arrow.core.raise.Raise type instead, which is more general and can be used to raise typed errors or _logical failures_\n" +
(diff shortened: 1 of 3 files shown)


nomisRev:
@HSAR continuing conversation here since it's related to the code in this PR.

The `epoch` is advancing after `reset`, it's happening here. https://github.com/arrow-kt/arrow/pull/3076/files#diff-256e04d2e123a62cadc54dd3157a1e09e8b946761f9a57835eb547eed034f684R109. So when transitioning from `Resetting` to `Awaiting` it takes the `original.epoch` from the _resetting_ barrier.

Although, maybe it's better to increment `epoch` on `reset` and carry it along 🤔

We try merging this code into yours, and then your history is still there and we can consider my changes an async pair session. I created a separate PR because I felt it might have been easier to start from scratch but I took inspiration from your work so I think this can make sense. I am fine keeping `barrierAction`, I left it out here since it didn't seem related to my current changes. I am unsure how Java implements `reset` under the hood, but felt like two separate features/PRs.

nomisRev:
With that being said I think `numberAwaiting` also makes sense to expose ☺️

HSAR:
RE epoch: Oh, I see - my bad, that rather slipped past me. I'm okay with it as-is functionally, though I would lean towards incrementing it on reset as very slightly more readable (but possibly very slightly less performant). The difference isn't huge between the two.

RE barrierAction: I understand and agree, splitting it up into separate commits is probably the right way to go.

RE numberAwaiting: In initial implementation, I thought `numberAwaiting` would be useful to expose. I'm not so certain now, especially as acting on it would not be thread-safe. Its main use would be in tests to check that the reset really has happened, but that isn't the most elegant thing ever.

Merging the history could easily be done; I would also have suggested that your solution at least inherits conceptually from mine, but not enough to insist that we keep my commit(s) in the git history. I don't believe that I can push to this branch though so you would have to conduct the merge I'm afraid!

PR #3080, opened 2023-07-05
Optics `copy` with assignment plug-in

serras:
There's a new [compiler plug-in](https://github.com/JetBrains/kotlin/commit/09d6dfc8bf1a8499f461de4227b1585c6072d0f0) (and its corresponding [Gradle plug-in](https://plugins.gradle.org/plugin/org.jetbrains.kotlin.plugin.assignment)) to overload Kotlin's assignment operator. This is quite nice for the `copy` DSL, since we can use `=` there directly.

Alas, the plug-in is not so powerful (yet), and it gets confused if `assign` is defined as an extension method inside an interface. For that reason, right now you need to use an intermediate call to `ref`,

```kotlin
lens.ref = newValue
```

which returns a `Reference<A, B>` which can later be used by the plug-in.

### DIFF
--- a/arrow-libs/optics/arrow-optics/api/arrow-optics.api
+++ b/arrow-libs/optics/arrow-optics/api/arrow-optics.api
@@ -1,10 +1,12 @@
 public abstract interface class arrow/optics/Copy {
+	public abstract fun getRef (Larrow/optics/PSetter;)Larrow/optics/Reference;
 	public abstract fun inside (Larrow/optics/PTraversal;Lkotlin/jvm/functions/Function1;)V
 	public abstract fun set (Larrow/optics/PSetter;Ljava/lang/Object;)V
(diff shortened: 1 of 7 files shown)


serras:
It's actually a bit cumbersome. You need to define an annotation for those types which support overloaded assignment, and declare it as part of the Gradle build file (as done [here](https://github.com/arrow-kt/arrow/pull/3080/files#diff-e1e4574419275299da6d9399a3f16a80d416db082227af41e6b992855b87faf4R72)). Then whenever the compiler finds `x = y`, it checks whether the type of `x` is one of those marked with the annotation, and proceeds to change the AST.

serras:
After more thought, this seems not worth the hassle. You need to include and configure a plug-in, and then you still need the `.ref` at the end.

```diff
  user.copy {
-   User.name set "Alejandro"
+   User.name.ref = "Alejandro"
  }
```

PR #3017, opened 2023-03-31
Update versions of several libraries

serras:
This PR updates several libraries:
- Kotlin to 1.8.21,
- Dokka to 1.8.10,
- KotlinX Coroutines to 1.7.0,
- `kotlin-compile-testing` to 1.5.0, aligned also with the compiler.
And more importantly, it uses an Arrow Gradle Config which also updates many of the plug-in, including the SemVer one. In my local tests the configure phase takes now around 30s. on a clean project, and 2s. after the first run.

PR #3090, opened 2023-07-10
Rollback to previous version of Arrow Gradle Config

nomisRev:
On https://github.com/arrow-kt/arrow/pull/3017 Arrow Gradle was bumped to `0.12.0-rc.3` but this seems to come with some changes in the semvar plugin that are incompatible with our current pipeline.

This PR does a rollback to the previous version, this should fix the release for 2.0.0 and allows us to release using the same pipeline as we did for 1.2.0-RC.

### DIFF
--- a/arrow-libs/core/arrow-annotations/build.gradle.kts
+++ b/arrow-libs/core/arrow-annotations/build.gradle.kts
@@ -4,7 +4,6 @@ plugins {
   id(libs.plugins.kotlin.multiplatform.get().pluginId)
   alias(libs.plugins.arrowGradleConfig.kotlin)
   alias(libs.plugins.arrowGradleConfig.publish)
-  alias(libs.plugins.arrowGradleConfig.versioning)
   alias(libs.plugins.spotless)
 }
 
--- a/arrow-libs/core/arrow-atomic/build.gradle.kts
+++ b/arrow-libs/core/arrow-atomic/build.gradle.kts
@@ -4,7 +4,6 @@ plugins {
   id(libs.plugins.kotlin.multiplatform.get().pluginId)
   alias(libs.plugins.arrowGradleConfig.kotlin)
   alias(libs.plugins.arrowGradleConfig.publish)
-  alias(libs.plugins.arrowGradleConfig.versioning)
   alias(libs.plugins.kotest.multiplatform)
   alias(libs.plugins.kotlinx.kover)
   alias(libs.plugins.spotless)
--- a/arrow-libs/core/arrow-continuations/build.gradle.kts
+++ b/arrow-libs/core/arrow-continuations/build.gradle.kts
@@ -4,7 +4,6 @@ plugins {
   id(libs.plugins.kotlin.multiplatform.get().pluginId)
   alias(libs.plugins.arrowGradleConfig.kotlin)
   alias(libs.plugins.arrowGradleConfig.publish)
-  alias(libs.plugins.arrowGradleConfig.versioning)
   alias(libs.plugins.kotlinx.kover)
   alias(libs.plugins.kotest.multiplatform)
   alias(libs.plugins.spotless)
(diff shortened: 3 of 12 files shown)


serras:
When we [updated the `arrow-gradle-plugin` to 0.12](https://github.com/arrow-kt/arrow/commit/f8c50f3779849d8f2c81c1a24320f15b81be1388) we also had to add the versioning plug-in to all of the `build.gradle.kts`. Maybe we need to undo those changes too?

JavierSegoviaCordoba:
@nomisRev which changes are incompatible?

serras:
> which changes are incompatible?

@JavierSegoviaCordoba we're experiencing problems in the [Arrow 2.0 branch](https://github.com/arrow-kt/arrow/pull/2778). Our goal is to keep producing the alphas for `main`, and at the same time produce a `2.0.0-SNAPSHOT` for new commits in the `arrow-2` branch.

JavierSegoviaCordoba:
> > which changes are incompatible?
> 
> @JavierSegoviaCordoba we're experiencing problems in the [Arrow 2.0 branch](https://github.com/arrow-kt/arrow/pull/2778). Our goal is to keep producing the alphas for `main`, and at the same time produce a `2.0.0-SNAPSHOT` for new commits in the `arrow-2` branch.

This should work, but I would need to know which is exactly the issue.

JavierSegoviaCordoba:
In order to get it working, the `main` branch should be untouched with `1.x.x` tags. The `arrow-2` branch needs to use `./gradlew whateverTask -P semver.stage=snapshot -P semver.scope=major`

serras:
> In order to get it working, the `main` branch should be untouched with `1.x.x` tags. The `arrow-2` branch needs to use `./gradlew whateverTask -P semver.stage=snapshot -P semver.scope=major`

I've pushed a commit at https://github.com/arrow-kt/arrow/pull/2778/commits/6524b172e2f60c8a5aa30b3cb9638d22642b5f38 to see how it goes.

serras:
@JavierSegoviaCordoba it keeps failing https://github.com/arrow-kt/arrow/actions/runs/5507646994/jobs/10037872768. Maybe you could have a look and help us fix this? 🙏

JavierSegoviaCordoba:
The versions are correct now:
```
semver for arrow-resilience: 2.0.0-SNAPSHOT
```

So this is no longer related to semver.

I am not sure why it is failing, but I think it can be related to `closeSonatypeStagingRepository`.

I think you can't put it when you are publishing a SNAPSHOT, as you can't close repositories there (it is not possible to do that manually, there is no UI for that AFAIK). So removing that line should fix the issue:

```
./gradlew publishToSonatype -Psemver.stage=snapshot -Psemver.scope=major
```

serras:
@JavierSegoviaCordoba still now working even though I removed the closing step :(

JavierSegoviaCordoba:
> @JavierSegoviaCordoba still now working even though I removed the closing step :(

The issue is not related to semver.

It is trying to publish the library to the MavenCentral repository instead of the snapshot repository, and that causes the bad request (MavenCentral doesn't allow publishing `-SNAPSHOT` versions)

This task only runs when it tries to publish to MavenCentral:
```
> Task :initializeSonatypeStagingRepository
Created staging repository 'ioarrow-kt-1486' at https://oss.sonatype.org/service/local/repositories/ioarrow-kt-1486/content/
```

I am not sure why this is happening tho, based on the failure the version is correct, so probably there is a random configuration/plugin modifying the version:

```
* What went wrong:
Execution failed for task ':arrow-core:publishIosArm32PublicationToSonatypeRepository'.
> Failed to publish publication 'iosArm32' to repository 'sonatype'
   > Could not PUT 'https://oss.sonatype.org/service/local/staging/deployByRepositoryId/ioarrow-kt-1486/io/arrow-kt/arrow-core-iosarm32/2.0.0-SNAPSHOT/maven-metadata.xml'. Received status code 400 from server: Bad Request
```

Has arrow migrated to a different server? Maybe the snapshot URL must be changed if you migrated to the new hosts (s01) https://github.com/gradle-nexus/publish-plugin#publishing-to-maven-central-via-sonatype-ossrh

PR #3091, opened 2023-07-10
MemoizedDeepRecursiveFunction

serras:
Based on [this conversation in `arrow-site`](https://github.com/arrow-kt/arrow-website/issues/213#issuecomment-1628545020)

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/MemoizedDeepRecursiveFunction.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/MemoizedDeepRecursiveFunction.kt
@@ -18,9 +18,8 @@ public fun <T, R> MemoizedDeepRecursiveFunction(
 ): DeepRecursiveFunction<T, R> {
   val cache = Atomic(emptyMap<T, R>())
   return DeepRecursiveFunction { x ->
-    when (x) {
-      in cache.get() -> cache.get().getValue(x)
-      else -> {
+    when (val v = cache.get()[x]) {
+      null -> {
         val result = block(x)
         cache.loop { old ->
           when (x) {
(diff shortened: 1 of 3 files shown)


AcartN (approved):
I really like the way you've utilized DeepRecursiveScope to implement it.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/MemoizedDeepRecursiveFunction.kt
```diff
@@ -0,0 +1,38 @@
...
+ * [callRecursive][DeepRecursiveScope.callRecursive] function can be used to make a recursive call
+ * to the declared function.
+ */
+public fun <T, R> MemoizedDeepRecursiveFunction(
+  block: suspend DeepRecursiveScope<T, R>.(T) -> R
+): DeepRecursiveFunction<T, R> {
+  val cache = Atomic(emptyMap<T, R>())
+  return DeepRecursiveFunction { x ->
+    when (x) {
+      in cache.get() -> cache.get().getValue(x)
```
balazstothofficial, line 22: Wouldn't it be a little more performant to use:

```
cache.get().get(x)?.let {
   ...
}
```
?

nomisRev: Yes, doing a single `get()` would probably be better 🤔 Not sure if this is correct syntax, but this could keep most of the code the same.

```
when(val map = cache.get()) {
  in map -> ...
  else -> ...
}
```

nomisRev (approved):
Thank you @serras! This looks great 🙌

nomisRev (commented):
Thank you @AcartN and @balazstothofficial for the collaboration, and reporting 🙏 I am going to merge this PR! 🙌

PR #3096, opened 2023-07-12
Fix Option.tap deprecation message.

wolpl:
I assume, the deprecation message for `Option.tap` should be `tap is being renamed to onSome` instead of `tap is being renamed to onNone`.


nomisRev (approved):
Thank you @wolpl for fixing this oversight! 🙏 👏 

This will be merged, and release soon-ish. We're planning a 1.2.1 release soon with Kotlin 1.9.0, and with another bug fix that is raised today to make a fix to the KSP plugin.

serras (approved):

PR #3099, opened 2023-07-17
Raise any type from ? and Option

serras:
From [this Slack thread](https://kotlinlang.slack.com/archives/C5UPMM0A0/p1689601177846049)

PR #3100, opened 2023-07-17
ignoreErrors for Raise

serras:
Alternative to #3099, from [this Slack thread](https://kotlinlang.slack.com/archives/C5UPMM0A0/p1689601177846049)

This adds a new `ignoreErrors` function in `nullable` and `option` which "swallows" every error. This is needed because `withError` is a bit cumbersome, and it doesn't always type as nicely.

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/raise/NullableSpec.kt
+++ b/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/raise/NullableSpec.kt
@@ -69,6 +69,22 @@ class NullableSpec : StringSpec({
     } shouldBe "1"
   }
 
+  "binding either in nullable" {
+    nullable {
+      val number = Either.Right("s".length)
+      val string = number.map(Int::toString).bind()
+      string
+    } shouldBe "1"
+  }
+
+  "binding either in nullable, ignore errors" {
+    nullable {
+      val number = Either.Right("s".length) as Either<Boolean, Int>
+      val string = ignoreErrors { number.map(Int::toString).bind() }
+      string
+    } shouldBe "1"
+  }
+
   "short circuit option" {
     nullable<Int> {
       val number = Some("s".length)
--- a/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/raise/NullableSpec.kt
+++ b/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/raise/NullableSpec.kt
@@ -79,7 +79,7 @@ class NullableSpec : StringSpec({
 
   "binding either in nullable, ignore errors" {
     nullable {
-      val number: Either<Boolean Int> = Either.Right("s".length)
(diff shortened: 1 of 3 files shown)


arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/raise/NullableSpec.kt
```diff
@@ -69,6 +69,22 @@ class NullableSpec : StringSpec({
...
+    nullable {
+      val number = Either.Right("s".length)
+      val string = number.map(Int::toString).bind()
+      string
+    } shouldBe "1"
+  }
+
+  "binding either in nullable, ignore errors" {
+    nullable {
+      val number = Either.Right("s".length) as Either<Boolean, Int>
```
nomisRev, line 82:
```suggestion
      val number: Either<Boolean Int> = Either.Right("s".length)
```

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/raise/NullableSpec.kt
```diff
@@ -69,6 +69,22 @@ class NullableSpec : StringSpec({
     } shouldBe "1"
   }
 
+  "binding either in nullable" {
+    nullable {
+      val number = Either.Right("s".length)
+      val string = number.map(Int::toString).bind()
```
nomisRev, line 75: I'm more a fan of `ignoreErrors` as @kyay10 proposed on Slack, but how is this working in this PR? 🤔

serras: By pure chance: `.bind()` in `nullable` works over `Nothing?`, and `Either.Right` constructs `Either<Nothing, A>` (so in some sense "there's no error to be ignored").

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/raise/NullableSpec.kt
```diff
@@ -69,6 +69,22 @@ class NullableSpec : StringSpec({
...
+      string
+    } shouldBe "1"
+  }
+
+  "binding either in nullable, ignore errors" {
+    nullable {
+      val number: Either<Boolean Int> = Either.Right("s".length)
```
serras, line 82:
```suggestion
      val number = Either.Right("s".length) as Either<Boolean, Int>
```

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/raise/NullableSpec.kt
nomisRev: Oh, _Right_. I missed that that was inferring to `Either<Nothing, Right>`.

nomisRev (approved):

PR #3114, opened 2023-08-16
Add iorNel builder and Ior.toIorNel function

sampengilly:


sampengilly:
Unfortunately we can't easily have something like a `.bindNel()` extension function on an `Ior` inside an `iorNel` block without either duplicating `IorRaise` (or opening it for extension). Context receivers could help here going forward. For the initial submission I've kept it simple with an additional converter function to convert an `Ior` to an `IorNel` before binding normally

sampengilly:
It will probably be necessary to build a new `IorNelRaise<Error>` type I think. If we want to be able to bind individual `Error` items within the raise block for a nicer API:

```kotlin
fun someEitherFunc() = Throwable().left()

iorNel<Throwable, Nothing> {
    someEitherFunc().bind()
    // vs
    someEitherFunc().mapLeft { it.nel() }.bind()
}

serras:
@sampengilly could you please run `./gradlew apiDump` on this branch and commit the results? The automatic `.api` updater doesn't seem to work in forks :/

serras (approved):

PR #3120, opened 2023-08-29
Inline `NonEmpty` maps

### LINKED ISSUE #3116: ["Request"] Ease NonEmptyCollection methods usage calling suspend
### What version are you currently using?

1.2.0

### What would you like to see?

Currently, the following `mapIndexedOnNonEmptyList` method get a `Suspension functions can be called only within coroutine body` compilation error, but not `mapIndexedOnStandardCollection ` :
```
suspend fun mySuspendFunction(index: Int, value: Int): Any = TODO()

suspend fun mapIndexedOnNonEmptyList() =
    nonEmptyListOf(1, 3, 7).mapIndexed { index: Int, value: Int -> mySuspendFunction(index, value) }

suspend fun mapIndexedOnStandardCollection() =
    listOf(1, 3, 7).mapIndexed { index: Int, value: Int -> mySuspendFunction(index, value) }
```

I would like to be able to use `NonEmptyCollection` methods in a suspend method definition the same way you can on standard collections.

(note that the standard collection behavior is due to **`inline`** method definition of `Iterable<T>.mapIndexed`)

### Why?

I think that it would ease `NonEmptyCollection` usage (to avoid custom extensions or to avoid switching back on standard collections).

serras:
Fixes #3116 

This overrides `map` and `mapIndexed` with `inline` definitions, so they can be used in combination with `suspend` computations. Unfortunately this cannot be provided in general within `NonEmptyCollection`, because `inline` members are not allowed in interfaces.

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
@@ -183,6 +183,10 @@ public class NonEmptyList<out A>(
   public override inline fun <B> map(transform: (A) -> B): NonEmptyList<B> =
     NonEmptyList(transform(head), tail.map(transform))
 
+  @Suppress("OVERRIDE_BY_INLINE")
+  public override inline fun <B> mapIndexed(transform: (index: Int, A) -> B): NonEmptyList<B> =
+    NonEmptyList(transform(0, head), tail.mapIndexed { ix, e -> transform(ix + 1, e) })
+
   override fun <B> flatMap(transform: (A) -> NonEmptyCollection<B>): NonEmptyList<B> =
     transform(head).toNonEmptyList() + tail.flatMap(transform)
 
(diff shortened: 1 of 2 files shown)


nomisRev:
Do we need to do the same for `flatMap` and `distinctBy`?

nomisRev:
> 2. We would break source compatibility, since now users would need to import map explicitly.

As you mentioned in the ticket, this was probably the best solution in hindsight :/

serras:
> Do we need to do the same for `flatMap` and `distinctBy`?

There you are

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -183,9 +183,18 @@ public class NonEmptyList<out A>(
   public override inline fun <B> map(transform: (A) -> B): NonEmptyList<B> =
     NonEmptyList(transform(head), tail.map(transform))
 
-  override fun <B> flatMap(transform: (A) -> NonEmptyCollection<B>): NonEmptyList<B> =
+  @Suppress("OVERRIDE_BY_INLINE")
```
raulraja, line 186: Are these suppressions safe to ignore?

raulraja (approved):
thanks @serras !

nomisRev (approved):
Thank you @serras!

i-walker (approved):

PR #3140, opened 2023-10-11
Prepare for Kotlin community build

serras:
We've been asked to become part of a large set of repositories which are tested with new versions of Kotlin. To become part of the program, we need to handle a few Gradle variables that set the Kotlin compiler and language version.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -179,7 +179,7 @@ jobs:
       - name: jvmTest (K2 enabled)
         uses: gradle/gradle-build-action@v2
         with:
-          arguments: jvmTest -Pkotlin_version=2.0.0-dev-4562 -Pkotlin_repo_url="https://maven.pkg.jetbrains.space/kotlin/p/kotlin/bootstrap" -Pkotlin_language_version=2.0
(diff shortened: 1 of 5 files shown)


nomisRev (approved):
Looks good to me @serras!

Do you think it's worth planning the rewrite from Kotest to Kotlin-test? For example for experimental targets like wasm? I've thought we might be able to automate this using OpenRewrite, but the biggest problem I see is our reliance on `Arb`. We'd need to write our own :/

Unless Kotest manages to only use "stable" compiler APIs, it's constantly going to be breaking on new versions.

serras:
> Do you think it's worth planning the rewrite from Kotest to Kotlin-test? For example for experimental targets like wasm? I've thought we might be able to automate this using OpenRewrite, but the biggest problem I see is our reliance on `Arb`. We'd need to write our own :/

I'm not really sure about what's the right path here. In any case, just building the project and running JVM tests should already help the Kotlin team, so maybe we can think about next steps later?

nomisRev:
> running JVM tests

That's not going to work either if the Kotest compiler plugin breaks against new version of Kotlin, right? 🤔 That is what I encountered in 1.9.20-Beta, and we pretty much had with every minor version since it came out :/

PR #3225, opened 2023-10-20
Update Arrow-Atomic

### LINKED ISSUE #3144: Refactor AtomicBooleanTest from Kotest Plugin to Kotlin-test runtime
Arrow is migrating from the Kotest Multiplatform plugin to Kotlin-test,
not because we don't love Kotest but because Arrow wants to support experimental targets like wasm and be buildable and testable with Kotlin EAP & DEV version such that we can help the Kotlin team prevent regression. We're continue using Kotest property testing, and Kotest assertions.

Arrow needs the help of the community to do this, and the Arrow maintainers will be more than happy to help you through the progress.

This ticket requires following changes in `AtomicBooleanTest.kt`. Here is a small list of things to do:

 1. Fork & clone the Arrow repository
 2. Checkout the `arrow-2` branch and create a new branch from it.
 3. Replace all tests with shape `"name of test {"` to camelCase/dromedaryCase `@Test fun nameOfTest() = runTest {` in the test class
 4. Remove `StringSpec` as the super class from the test class
 5. Run `./gradlew spotlessApply`
 6. Commit your changes, and push the new branch to your repo
 7. Create a PR on the Arrow repository from your branch targetting `arrow-2`.

Arrow really appreciates all the help, and looks forward to a bright future together in Kotlin :party:

Thanks in advance!

nomisRev:
Closes #3144, #3145, #3146, #3147.

Also replaces any usage of `Atomic` with `Int` to `AtomicInt`.

### DIFF
--- a/arrow-libs/core/arrow-annotations/build.gradle.kts
+++ b/arrow-libs/core/arrow-annotations/build.gradle.kts
@@ -13,6 +13,8 @@ spotless {
   }
 }
 
+apply(from = property("ANIMALSNIFFER_MPP"))
+
 kotlin {
   sourceSets {
     commonMain {
@@ -23,11 +25,6 @@ kotlin {
     jvmMain {
       dependencies {
         implementation(libs.kotlin.stdlib)
-      }
-    }
-    jvmTest {
-      dependencies {
-        runtimeOnly(libs.kotest.runnerJUnit5)
       }
     }
     jsMain {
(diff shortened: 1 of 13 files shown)


serras:
Let's get this one and #3224 merged first, since they are more or less self-contained.

Then my plan is to push a change that adds `kotlin.test` to every library `test` source set, so that we don't have 10000 different ways of adding the library. WDYT?

serras:
After quite some investigation, it seems that `runTest` doesn't work exactly as `runBlocking` in Native. But on the other hand `runBlocking` is not available in JS... So I ended up doing a "run one in Native and the other in the rest".

PR #3229, opened 2023-10-21
Refactor OptionTest to kotlin-test

### LINKED ISSUE #3174: Refactor OptionTest from Kotest Plugin to Kotlin-test runtime
Arrow is migrating from the Kotest Multiplatform plugin to Kotlin-test,
not because we don't love Kotest but because Arrow wants to support experimental targets like wasm and be buildable and testable with Kotlin EAP & DEV version such that we can help the Kotlin team prevent regression. We're continue using Kotest property testing, and Kotest assertions.

Arrow needs the help of the community to do this, and the Arrow maintainers will be more than happy to help you through the progress.

This ticket requires following changes in `OptionTest.kt`. Here is a small list of things to do:

 1. Fork & clone the Arrow repository
 2. Checkout the `arrow-2` branch and create a new branch from it.
 3. Replace all tests with shape `"name of test {"` to camelCase/dromedaryCase `@Test fun nameOfTest() = runTest {` in the test class
 4. Remove `StringSpec` as the super class from the test class
 5. Run `./gradlew spotlessApply`
 6. Commit your changes, and push the new branch to your repo
 7. Create a PR on the Arrow repository from your branch targetting `arrow-2`.

Arrow really appreciates all the help, and looks forward to a bright future together in Kotlin :party:

Thanks in advance!

milgner:
Closes #3174

Not sure about my implementation of `testLawsCommon` - but I couldn't find any way to dynamically generate tests using only kotlin-test: there is `@TestFactory` for JUnit5 dynamic tests but for simple kotlin-test, a simple `@Test` was the only way I could find.

(If you prefer an alternative approach, let me know and I'll change it accordingly.)

In order to run tests in JS mode, I also updated `yarn.lock` via `./gradlew kotlinUpgradeYarnLock`.

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/test/Laws.kt
+++ b/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/test/Laws.kt
@@ -1,17 +1,17 @@
 package arrow.core.test
 
 import io.kotest.assertions.fail
+import io.kotest.assertions.withClue
 import io.kotest.core.names.TestName
 import io.kotest.core.spec.style.StringSpec
-import io.kotest.core.spec.style.scopes.StringSpecScope
 import io.kotest.core.spec.style.scopes.addTest
-import io.kotest.core.test.TestContext
+import kotlinx.coroutines.test.runTest
 
 interface LawSet {
   val laws: List<Law>
 }
 
-data class Law(val name: String, val test: suspend TestContext.() -> Unit)
+data class Law(val name: String, val test: suspend () -> Unit)
 
 fun <A> A.equalUnderTheLaw(b: A, f: (A, A) -> Boolean = { x, y -> x == y }): Boolean =
   if (f(this, b)) true else fail("Found $this but expected: $b")
@@ -23,15 +23,21 @@ fun StringSpec.testLaws(vararg laws: List<Law>): Unit = laws
   .distinctBy { law: Law -> law.name }
   .forEach { law: Law ->
     addTest(TestName(null, law.name, false), false, null) {
-      law.test(StringSpecScope(this.coroutineContext, testCase))
+      runTest { law.test() }
     }
   }
 
-fun StringSpec.testLaws(prefix: String, vararg laws: List<Law>): Unit = laws
-  .flatMap { list: List<Law> -> list.asIterable() }
-  .distinctBy { law: Law -> law.name }
-  .forEach { law: Law ->
-    addTest(TestName(prefix, law.name, false), false, null) {
-      law.test(StringSpecScope(this.coroutineContext, testCase))
+fun testLawsCommon(lawSet: LawSet) = withClue("In $lawSet") {
+  testLawsCommon(lawSet.laws)
+}
+
+fun testLawsCommon(vararg laws: List<Law>) = runTest {
+  laws
+    .flatMap(List<Law>::asIterable)
+    .distinctBy(Law::name)
+    .forEach { law: Law ->
+      withClue("Testing ${law.name}") {
+        law.test()
+      }
     }
-  }
+}
--- a/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/OptionTest.kt
+++ b/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/OptionTest.kt
@@ -3,8 +3,7 @@ package arrow.core
 import arrow.core.raise.option
 import arrow.core.test.laws.MonoidLaws
 import arrow.core.test.option
-import arrow.core.test.testLaws
-import io.kotest.core.spec.style.StringSpec
+import arrow.core.test.testLawsCommon
 import io.kotest.matchers.shouldBe
 import io.kotest.matchers.shouldNotBe
 import io.kotest.property.Arb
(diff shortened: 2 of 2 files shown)


arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/test/Laws.kt
```diff
@@ -6,12 +6,14 @@ import io.kotest.core.spec.style.StringSpec
...
 import io.kotest.core.test.TestContext
+import kotlinx.coroutines.test.TestScope
+import kotlinx.coroutines.test.runTest
 
 interface LawSet {
   val laws: List<Law>
 }
 
-data class Law(val name: String, val test: suspend TestContext.() -> Unit)
+data class Law(val name: String, val test: suspend TestScope.() -> Unit)
```
nomisRev, line 16: Same suggestions as in #3232

```suggestion
data class Law(val name: String, val test: suspend() -> Unit)
```

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/test/Laws.kt
```diff
@@ -6,12 +6,14 @@ import io.kotest.core.spec.style.StringSpec
 import io.kotest.core.spec.style.scopes.StringSpecScope
 import io.kotest.core.spec.style.scopes.addTest
 import io.kotest.core.test.TestContext
+import kotlinx.coroutines.test.TestScope
```
nomisRev, line 9:
```suggestion
```

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/test/Laws.kt
```diff
@@ -23,15 +25,14 @@ fun StringSpec.testLaws(vararg laws: List<Law>): Unit = laws
   .distinctBy { law: Law -> law.name }
   .forEach { law: Law ->
     addTest(TestName(null, law.name, false), false, null) {
-      law.test(StringSpecScope(this.coroutineContext, testCase))
+      runTest { law.test(this) }
```
nomisRev, line 28:
```suggestion
      law.test()
```

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/test/Laws.kt
```diff
@@ -23,15 +25,14 @@ fun StringSpec.testLaws(vararg laws: List<Law>): Unit = laws
...
-    addTest(TestName(prefix, law.name, false), false, null) {
-      law.test(StringSpecScope(this.coroutineContext, testCase))
-    }
-  }
+fun testLawsCommon(lawSet: LawSet) = testLawsCommon(lawSet.laws)
+fun testLawsCommon(vararg laws: List<Law>) = runTest {
+  laws
+    .flatMap(List<Law>::asIterable)
+    .distinctBy(Law::name)
+    .forEach { law: Law -> law.test(this) }
```
nomisRev, line 37:
```suggestion
    .forEach { law: Law -> law.test() }
```

nomisRev: I'm only wandering here if we should enhance the assertion message now, since it's all running as a single test now it will be impossible to figure out which failed assertions came from which `Law`.

```kotlin
runCatching { law.test() }
  .onFailure { throw AssertionError(law.name, it) }
```

nomisRev (commented):
Thanks for clarifying the changes, I am personally fine with all the laws running in a single test. I added a comment/suggestion to enhance the error message. Since now they're no longer running as separate tests.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/test/Laws.kt
milgner: Since the associated ticket didn't mention replacing the kotest assertions themselves, I added a `withClue` block which works cross-platform and now reads:

On JVM:

```
Caused by java.lang.AssertionError: In MonoidLaws(name=Option, empty=Option.None, combine=(arrow.core.Option<kotlin.Int>, arrow.core.Option<kotlin.Int>) -> arrow.core.Option<kotlin.Int>, GEN=io.kotest.property.arbitrary.TrampolineArb@2fb1ecf1, eq=(F, F) -> kotlin.Boolean)
Testing Law(name=Semigroup Laws (Option): associativity, test=() -> kotlin.Unit)
Found Option.Some(877965308) but expected: Option.Some(-2125984056) at
```

On JS:
```
Caused by AssertionError: Testing Law(name=Semigroup Laws (Option): associativity, test=function ($completion) {
      return i.invoke_pbb09x_k$($completion);
    })
Found Option.Some(-1) but expected: Option.Some(1)
```

(it seems like I can't run native tests at the moment, the executable complains about missing `libcrypt.so.1`)

milgner:
Rebased onto latest `arrow-2` branch to resolve merge conflicts and address issue in law testing helpers. Made another minimal improvement to the messages but tests were green for all workflows previously so things are looking good.

nomisRev (approved):

nomisRev:
Thank you for your contribution @milgner 🙌 Great work!

PR #3240, opened 2023-10-24
Inline `AtomicBoolean`

serras:
Let's squeeze a little more performance!


nomisRev (approved):
Great improvement @serras 🙌

i-walker (approved):
🙌🏾

PR #3273, opened 2023-10-30
Update CollectionsSyntaxTests.kt

TejasMate:
#3161


serras:

PR #3277, opened 2023-10-30
Update NonFatalJvmTest.kt

TejasMate:
#3178


serras:
@TejasMate could you run `./gradlew spotlessApply` on this branch and push the results. This process is usually automated, but doesn't work with forks. Thanks in advance!

serras:

PR #3280, opened 2023-11-03
Arrow Collectors

serras:
This is an idea for a new library heavily influenced by Java's [`Collector`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collector.html) and Haskell's [`foldl` library](https://hackage.haskell.org/package/foldl/docs/Control-Foldl.html). The goal is to build complex computations over sequences of values, guaranteeing that those values are consumed only once.

For example, if you want to compute the average of a sequence in one go, you can write:

```kotlin
val average = zip(Collectors.sum, Collectors.length) { s, l -> s.toDouble() / l.toDouble() }
list.collect(average)
```

The "trick" is that collectors have a nice interface based on `map`/`contramap`/`zip` (they are applicative functors and profunctors, if we were to use fancy words). Combined with Kotlin's cold `Flow`s, they provide a nice combination.

Since this is a pattern coming from functional programming, I thought that it may be a good fit for Arrow.


arrow-libs/fx/arrow-collectors/src/commonMain/kotlin/arrow/collectors/Collectors.kt
```diff
@@ -0,0 +1,147 @@
...
+
+  /**
+   * Returns the "best" value from the value.
+   *
+   * @param selectNew Decides whether the new value is "better" than the previous best.
+   */
+  public fun <M> bestBy(
+    selectNew: (old: M, new: M) -> Boolean,
+  ): Collector<M, M?> =
+    object : CollectorI<Atomic<M?>, M, M?> {
```
raulraja, line 82: If `CollectorI` is used to build operators and it's something that would be part of what users can do perhaps we should include `@JvmOverload` constructors or factory methods in `CollectorI`'s companion so we can do construction of the possible combinations without having to implement the interface in an anonymous object.
```kotlin
public fun <T> list(): Collector<T, List<T>> =
    CollectorI(
        mutableListOf(), 
        setOf(Characteristics.IDENTITY_FINISH), 
        { current: MutableList<T>, value: T -> current.add(value) },
        ::identity
    )
```

arrow-libs/fx/arrow-collectors/src/commonMain/kotlin/arrow/collectors/Collectors.kt
```diff
@@ -0,0 +1,147 @@
...
+        }
+      }
+
+      override suspend fun finish(current: Atomic<M?>): M? = current.get()
+    }
+
+  /**
+   * Collects all the values in a list, in the order in which they are emitted.
+   */
+  public fun <T> list(): Collector<T, List<T>> =
```
raulraja, line 99: Can these become `val` properties, either single instance or computed?
Looks like the `supply` method provides a new instance each time, so I was wondering if the instance of `CollectorI` could just be cached or a singleton value.

raulraja (commented):
Looking great @serras . I realize some of the stuff we did back in the days for Arrow Query language could be easier built on top of this https://github.com/arrow-kt/arrow/pull/1079

arrow-libs/fx/arrow-collectors/src/commonMain/kotlin/arrow/collectors/Collectors.kt
serras: Unfortunately not, because `val`s cannot have generic arguments :(

hoc081098:
Great. Arrow will be more popular 😂.

arrow-libs/fx/arrow-collectors/src/commonMain/kotlin/arrow/collectors/Collectors.kt
```diff
@@ -0,0 +1,147 @@
...
+      override suspend fun accumulate(current: Atomic<M?>, value: M) {
+        current.update { old ->
+          if (old == null) value
+          else if (selectNew(old, value)) value
+          else old
+        }
+      }
+
+      override suspend fun finish(current: Atomic<M?>): M? = current.get()
+    }
```
nomisRev, line 94: I think the instance can be cached, and this suffers from nested nullability issue 🤔 

```suggestion
    ListCollector
    
   private object Empty
    
    private val ListCollector: CollectorI<Atomic<Any?>, M, M?> =
      object : CollectorI<Atomic<Any?>, M, M?> {
        override val characteristics: Set<Characteristics> = Characteristics.CONCURRENT_UNORDERED
        override suspend fun supply(): Atomic<Any?> = Atomic(Empty)
        override suspend fun accumulate(current: Atomic<Any?>, value: M) {
          current.update { old ->
            if (old == Empty) value
            else if (selectNew(old, value)) value
            else old
          }
        }

        override suspend fun finish(current: Atomic<Any?>): M? = current.get() as? M
      }
```

arrow-libs/fx/arrow-collectors/src/commonMain/kotlin/arrow/collectors/Collectors.kt
nomisRev: Idem as other comment, instance can be cached with `Any?` and casting. Same goes for everything else in this file.

arrow-libs/fx/arrow-collectors/src/commonMain/kotlin/arrow/collectors/Flow.kt
```diff
@@ -0,0 +1,104 @@
...
+public fun <T, R> Flow<T>.collectBlocking(collector: Collector<T, R>): R {
+  var final: R
+  suspend { collect(collector) }.startCoroutine(object : Continuation<R> {
+    override val context: CoroutineContext = EmptyCoroutineContext
+    override fun resumeWith(result: Result<R>) {
+      final = result.getOrThrow()
+    }
+  })
+  return final
+}
```
nomisRev, line 104: This is unsafe. Coroutines should never be ran with `EmptyCoroutineContext` (unless it's `@RestrictSuspension`, otherwise it breaks the semantics of context/thread preservation.

If anything within this `Flow` suspends, then `final` will result in an `NonInitializedException` since this is not actually blocking.

Only way to do this correctly is wrap it in `runBlocking`, but that's not available on JS... So I would personally omit this function in favor of `runBlocking { flow.collect(collector) }`.

Maybe worth adding one for `Sequence`, I think that would also be faster for `Iterable` to do `asSequence().collect(collector)` than `asFlow()` since `Sequence` has less overhead than `Flow` and runs "blocking".

arrow-libs/fx/arrow-collectors/src/jvmMain/kotlin/arrow/collectors/Collectors.kt
```diff
@@ -0,0 +1,51 @@
...
+  Collectors.concurrentMap<K, V>().contramap { (k, v) -> k to v }
+
+/**
+ * Collects all the values in a map.
+ *
+ * This [Collector] supports concurrency,
+ * so it's potential faster than [Collectors.map].
+ */
+@Suppress("UnusedReceiverParameter")
+public fun <K, V> Collectors.concurrentMap(): Collector<Pair<K, V>, ConcurrentMap<K, V>> =
```
nomisRev, line 24: Nice JVM specific addition! 🙌

arrow-libs/fx/arrow-collectors/src/jvmMain/kotlin/arrow/collectors/JvmCollector.kt
```diff
@@ -0,0 +1,37 @@
+package arrow.collectors
+
+/**
+ * Wraps a [java.util.stream.Collector] to use with [collect].
+ */
+public fun <T, R> java.util.stream.Collector<T, *, R>.asCollector(): Collector<T, R> =
+  Collectors.jvm(this)
```
nomisRev, line 7: ❤️

nomisRev (commented):
Looks awesome 🙌 👏 👏 The implementation of `Blocking` is incorrect and has some issues. If it's not clear what I meant, I would be more happy to jump on a quick call, make a commit on this branch, or clear up any doubts you have in this PR ☺️

arrow-libs/fx/arrow-collectors/src/commonMain/kotlin/arrow/collectors/Collectors.kt
serras: You're completely right. I've reworked the code in that form, and it looks waaaay better than before!

arrow-libs/fx/arrow-collectors/src/commonMain/kotlin/arrow/collectors/Collectors.kt
serras: To add a bit more, if you try to make something like `list(): Collector<T, List<T>>` into a `val`, you run into the problem that `T` appears both in `in` and `out` position, so you cannot play the trick of using `Any?` or `Nothing` like in other places.

arrow-libs/fx/arrow-collectors/src/commonMain/kotlin/arrow/collectors/Flow.kt
serras: The problem with `Sequence` is that all the operations in `Collector` may `suspend` (although none of the basic collectors does). I'm wondering whether it's worth adding a subclass of `NonSuspendCollector`s that could run without the need of a `runBlocking` nor `suspend`.

arrow-libs/fx/arrow-collectors/src/commonMain/kotlin/arrow/collectors/Collectors.kt
serras: Aha! Thanks for the trick, @nomisRev, this is fixed in a new commit

arrow-libs/fx/arrow-collectors/src/commonMain/kotlin/arrow/collectors/Flow.kt
serras: I've updated the code with `NonSuspend` variants which run on `Sequence` (or really on any `Iterator`). Now for `Iterable` you can use `collect` (non-suspend but no concurrency) or `parCollect` (suspend but uses concurrency if instructed).

arrow-libs/fx/arrow-collectors/src/commonMain/kotlin/arrow/collectors/Collect.kt
```diff
@@ -0,0 +1,150 @@
...
+    else -> started.map { collector.accumulate(accumulator, it) }
+  }
+  var completed: R
+  continued.onCompletion {
+    completed = when {
+      Characteristics.IDENTITY_FINISH in collector.characteristics -> accumulator as R
+      else -> collector.finish(accumulator)
+    }
+  }.collect { }
+  return completed
```
hoc081098, line 137: Hi, I think we can simplify this by `collect()` then returning.
It has the same the behavior.

```suggestion
  continued.collect()

  return when {
    Characteristics.IDENTITY_FINISH in collector.characteristics -> accumulator as R
    else -> collector.finish(accumulator)
  }
```

serras:
Ready for review! If this gets merged, I’ll happily write docs for the Arrow Website. And even some sort of blog post, if you want!

arrow-libs/fx/arrow-collectors/src/commonMain/kotlin/arrow/collectors/Collector.kt
```diff
@@ -0,0 +1,237 @@
...
+  ): Collector<Value, V> = of(
+    supply = { Pair(this.supply(), other.supply()) },
+    accumulate = { (currentThis, currentOther), value ->
+      this.accumulate(currentThis, value)
+      other.accumulate(currentOther, value)
+    },
+    finish = { (currentThis, currentOther) ->
+      combine(this.finish(currentThis), other.finish(currentOther))
+    },
+    characteristics = this.characteristics,
```
hoc081098, line 156: Hi @serras, I think the characteristics of the zipped collector should be the intersection of two input characteristics.
What do you think about that 🙏?

serras: I think that you're completely right!

serras: I've just pushed a fixed version. I also found that `map` was incorrect, since you were not guaranteed to have `IDENTITY_FINISH` in that case.

arrow-libs/fx/arrow-collectors/build.gradle.kts
```diff
@@ -0,0 +1,65 @@
...
+      }
+    }
+    browser {
+      testTask {
+        useKarma {
+          useChromeHeadless()
+          timeout.set(Duration.ofMinutes(1))
+        }
+      }
+    }
```
nomisRev, line 59: **Off-topic:** I think would be good if we can put this in some common configuration at some point. Still not sure what's the best way to do that in Gradle 🙃

arrow-libs/fx/arrow-collectors/src/commonMain/kotlin/arrow/collectors/Collector.kt
```diff
@@ -0,0 +1,237 @@
...
+ * The accumulation is done in three phases:
+ * - Initialization of some (mutable) accumulator ([supply])
+ * - Updating the accumulator with each value ([accumulate])
+ * - Finalize the work, and extract the final result ([finish])
+ *
+ * This interface is heavily influenced by
+ * Java's [`Collector`](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collector.html)
+ * and Haskell's [`foldl` library](https://hackage.haskell.org/package/foldl/docs/Control-Foldl.html).
+ */
+public interface CollectorI<InternalAccumulator, in Value, out Result> {
```
nomisRev, line 48: **Nit:** Not a huge fan of `CollectorI` but not sure if something else makes sense.. I guess `InternalAccumulator` is `State`? What about `StateCollector<State, in Value, out Result>`, or `StatefulCollector`? Not sure if this makes sense though, since all `Collector` have `State`. We had a similar issue with Schedule when we originally designed it, but don't think we can take the same/similar approach to solve this issue.

This is just a _nit_, feel free to ignore. For me it's also fine without any changes, just wanted to leave some thoughts.

nomisRev (approved):
Looks good to me @serras!! Thank you for the amazing work 🙌 👏 🥳

nomisRev:
Going to merge. Feel free to discuss further on my nit remark, or to ignore completely ☺️

PR #3282, opened 2023-11-04
Remove a bunch of warnings

serras:
This removes every warning in `main` source sets, and most in the Knit examples and `test`s.

### DIFF
--- a/arrow-libs/core/arrow-atomic/build.gradle.kts
+++ b/arrow-libs/core/arrow-atomic/build.gradle.kts
@@ -46,10 +46,8 @@ kotlin {
   }
 }
 
-tasks.withType<KotlinCompile>().configureEach {
-  kotlinOptions {
-    freeCompilerArgs = freeCompilerArgs + "-Xexpect-actual-classes"
-  }
+tasks.withType<KotlinCompile> {
+  kotlinOptions.freeCompilerArgs += "-Xexpect-actual-classes"
 }
 
 tasks.withType<Test> {
--- a/arrow-libs/core/arrow-core/build.gradle.kts
+++ b/arrow-libs/core/arrow-core/build.gradle.kts
@@ -58,6 +58,10 @@ kotlin {
 }
 
 // enables context receivers for Jvm Tests
+tasks.withType<KotlinCompile> {
+  kotlinOptions.freeCompilerArgs += "-Xexpect-actual-classes"
+}
+
 tasks.named<KotlinCompile>("compileTestKotlinJvm") {
   kotlinOptions.freeCompilerArgs += "-Xcontext-receivers"
 }
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
@@ -179,7 +179,7 @@ public value class NonEmptyList<out A> @PublishedApi internal constructor(
     else -> head
   }
 
-  @Suppress("OVERRIDE_BY_INLINE")
+  @Suppress("OVERRIDE_BY_INLINE", "NOTHING_TO_INLINE")
   public override inline fun distinct(): NonEmptyList<A> =
     NonEmptyList(all.distinct())
 
@@ -340,6 +340,7 @@ public fun <A> nonEmptyListOf(head: A, vararg t: A): NonEmptyList<A> =
   NonEmptyList(listOf(head) + t)
 
 @JvmName("nel")
+@Suppress("NOTHING_TO_INLINE")
 public inline fun <A> A.nel(): NonEmptyList<A> =
   NonEmptyList(listOf(this))
 
@@ -355,9 +356,11 @@ public inline fun <A, B : Comparable<B>> NonEmptyList<A>.minBy(selector: (A) -> 
 public inline fun <A, B : Comparable<B>> NonEmptyList<A>.maxBy(selector: (A) -> B): A =
   maxByOrNull(selector)!!
 
+@Suppress("NOTHING_TO_INLINE")
 public inline fun <T : Comparable<T>> NonEmptyList<T>.min(): T =
   minOrNull()!!
 
+@Suppress("NOTHING_TO_INLINE")
 public inline fun <T : Comparable<T>> NonEmptyList<T>.max(): T =
   maxOrNull()!!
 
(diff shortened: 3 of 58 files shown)


arrow-libs/core/arrow-atomic/build.gradle.kts
```diff
@@ -46,10 +46,8 @@ kotlin {
   }
 }
 
-tasks.withType<KotlinCompile>().configureEach {
-  kotlinOptions {
-    freeCompilerArgs = freeCompilerArgs + "-Xexpect-actual-classes"
-  }
+tasks.withType<KotlinCompile> {
+  kotlinOptions.freeCompilerArgs += "-Xexpect-actual-classes"
 }
```
JavierSegoviaCordoba, line 52: I highly recommend to keep the `configureEach`

```suggestion
tasks.withType<KotlinCompile>.configureEach {
  kotlinOptions.freeCompilerArgs += "-Xexpect-actual-classes"
}
```

arrow-libs/core/arrow-core/build.gradle.kts
```diff
@@ -58,6 +58,10 @@ kotlin {
 }
 
 // enables context receivers for Jvm Tests
+tasks.withType<KotlinCompile> {
+  kotlinOptions.freeCompilerArgs += "-Xexpect-actual-classes"
+}
+
```
JavierSegoviaCordoba, line 64: Same as above

arrow-libs/core/arrow-atomic/build.gradle.kts
serras: Any particular reason for this? (Gradle is all magic to me)

JavierSegoviaCordoba: Yep, `configureEach` is lazy, the other isn't. If I remember correctly, it is calling `all` under the hood.

I hope they mark all of those APIs as deprecated in Gradle 9.

arrow-libs/core/arrow-core/build.gradle.kts
serras:
```suggestion
tasks.withType<KotlinCompile>.configureEach {
  kotlinOptions.freeCompilerArgs += "-Xexpect-actual-classes"
}
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -179,7 +179,7 @@ public value class NonEmptyList<out A> @PublishedApi internal constructor(
     else -> head
   }
 
-  @Suppress("OVERRIDE_BY_INLINE")
+  @Suppress("OVERRIDE_BY_INLINE", "NOTHING_TO_INLINE")
```
nomisRev, line 182: @serras, so is this is actually inlining?

I know Kotlin Std has `@InlineOnly`, and afaik that forces inlining and removes the method from the binary. At least that's my understanding. So I am wondering, if that exists and my understanding is correct does the compiler ignore this `inline` or does it still forces inline (but keep it in the binary?)?

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
```diff
@@ -32,7 +32,7 @@ public sealed class ExitCase {
  * @see guaranteeCase for registering a handler that executes for any [ExitCase].
  */
 public suspend inline fun <A> onCancel(
-  fa: suspend () -> A,
+  fa: () -> A,
```
nomisRev, line 35: I think this is still needed.. Or what I've used to hit in the passed was a compiler bug ... 😅 Perhaps I should've suppressed with a comment at the time.

IIRC this is needed for method references, we can potentially also add a test so we can not accidentally break this in the future unless this was fixed in the compiler. For the old cases of:

```kotlin
onCancel(::awaitCancellation) { ... } 
```

This doesn't compile without `fa: suspend () -> A`.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/ParZipOrAccumulate.kt
```diff
@@ -14,44 +14,44 @@ public suspend inline fun <E, A, B, C> Raise<E>.parZipOrAccumulate(
   crossinline combine: (E, E) -> E,
   crossinline fa: suspend ScopedRaiseAccumulate<E>.() -> A,
   crossinline fb: suspend ScopedRaiseAccumulate<E>.() -> B,
-  crossinline f: suspend CoroutineScope.(A, B) -> C
+  crossinline transform: suspend CoroutineScope.(A, B) -> C
```
nomisRev, line 17: 🥇

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Traversal.kt
```diff
@@ -195,11 +195,11 @@ public interface PTraversal<S, T, A, B> {
       set: (B, B, S) -> T
     ): PTraversal<S, T, A, B> =
       object : PTraversal<S, T, A, B> {
-        override fun <R> foldMap(initial: R, combine: (R, R) -> R, s: S, f: (focus: A) -> R): R =
-          combine(f(get1(s)), f(get2(s)))
+        override fun <R> foldMap(initial: R, combine: (R, R) -> R, source: S, map: (focus: A) -> R): R =
```
nomisRev, line 198: Sorry, this is a bit nitpicky but do we have a rule of thumb for `map` vs `transform`?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
serras: I'm not sure. The error I was getting was "performance of `inline` is not significant" (even though the error is called "nothing to inline"). I think we should keep those `inline`, it feels like each step is not very significant, but in a big code this all adds up.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
serras: It definitely works now (at least in the examples and tests). The main choice for this removal was the compiler telling me that `suspend` was not super-great in an `inline` function. But we can bring it back if you feel this is breaking the way the library should be used.

PR #3284, opened 2023-11-05
Enable Kover for serialization, disable for Optics KSP

serras:
The Kover report is causing problems with the new version of KSP


build.gradle.kts
```diff
@@ -68,10 +68,11 @@ dependencies {
   kover(projects.arrowContinuations)
   kover(projects.arrowCore)
   kover(projects.arrowCoreRetrofit)
+  // kover(projects.arrowCoreSerialization)
```
nomisRev, line 71: Why is this commented?

build.gradle.kts
```diff
@@ -68,10 +68,11 @@ dependencies {
...
   kover(projects.arrowFxCoroutines)
   kover(projects.arrowFxStm)
   kover(projects.arrowOptics)
-  kover(projects.arrowOpticsKspPlugin)
+  // kover(projects.arrowOpticsKspPlugin)
```
nomisRev, line 75: I think this can be removed instead of comment?
```suggestion
```

build.gradle.kts
serras: I tried to add it for completeness, but it turns out that it causes problems with Kover (as the Optics KSP does). At the end I’ll remove both, and add a warning in the code to remember this.

build.gradle.kts
serras: I’d rather keep it alongside the comment in the corresponding section

nomisRev (approved):
Thanks @serras!

PR #3286, opened 2023-11-06
Roll back to Kover 0.7.3

serras:
This seems to solve the problem I was trying to fix in #3284, but without any hacks. We'll just need to wait until a new Kover releases.


serras:
I’m going to go ahead and merge this because it’s blocking other Pars (like #3280)

PR #3293, opened 2023-11-09
Add or fix contracts in Raise

serras:
K2 checks `callsInPlace` contracts much more thoroughly, which led to the discovery of some missing (and even one incorrect) contract.


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -32,8 +33,10 @@ import kotlin.jvm.JvmName
  * Read more about running a [Raise] computation in the
  * [Arrow docs](https://arrow-kt.io/learn/typed-errors/working-with-typed-errors/#running-and-inspecting-results).
  */
-public inline fun <Error, A> either(@BuilderInference block: Raise<Error>.() -> A): Either<Error, A> =
-  fold({ block.invoke(this) }, { Either.Left(it) }, { Either.Right(it) })
+public inline fun <Error, A> either(@BuilderInference block: Raise<Error>.() -> A): Either<Error, A> {
+  contract { callsInPlace(block, InvocationKind.EXACTLY_ONCE) }
```
nomisRev, line 37: I think we removed these on purpose, because `fun raise(error: Error): Nothing` will result in `code unreachable` after `either { }`. If I understood that is the desired behavior from the compiler POV, so I expect it to be unchanged in K2.

I think that goes for everything in `Builders.kt`. I'd need to check git history, I should've documented it in the PR.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Raise.kt
```diff
@@ -336,7 +337,13 @@ public interface Raise<in Error> {
 public inline fun <Error, A> recover(
   @BuilderInference block: Raise<Error>.() -> A,
   @BuilderInference recover: (error: Error) -> A,
-): A = fold(block, { throw it }, recover, ::identity)
+): A {
+  contract {
+    callsInPlace(block, EXACTLY_ONCE)
+    callsInPlace(recover, AT_MOST_ONCE)
+  }
```
nomisRev, line 344: Possible suffers from same issue as builders.

nomisRev (commented):
IIRC we removed the contracts from builders, and recover since it was resulting in false "unreachable".

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
serras: I've made them strict using `AT_MOST_ONCE`, which should not have the same problem (yet it enables further analysis by the compiler).

On prolem is that the semantics `EXACTLY_ONCE` are not clear: does it mean we _call_ the function, or that the function _finishes_ execution? Given that `raise` stops the execution, that means that `EXACTLY_ONCE` wouldn't be true for the second meaning. And given the warnings in K2, this is what the compiler understands too.

serras:
@nomisRev do you think it's worth keeping this PR (without the `EXACTLY_ONCE`s) or should I close it?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
nomisRev: > On prolem is that the semantics EXACTLY_ONCE are not clear: does it mean we call the function, or that the function finishes execution? Given that raise stops the execution, that means that EXACTLY_ONCE wouldn't be true for the second meaning. And given the warnings in K2, this is what the compiler understands too.

Afaik `EXACTLY_ONCE` is meant for function that are _called in place_, **and** finishes execution but finishes execution can also mean resulted in an exception. The difference lies in that `raise` is caught, so we don't return the result of `EXACTLY_ONCE` but instead we manipulate the result of `EXACTLY_ONCE` by recovering from the `raise`.

Similar for `runCatching`, it manipulates the result of the `block` lambda by capturing any results. So I think it cannot be considered as `EXACTLY_ONCE`.

I'm not entirely sure what `EXACTLY_ONCE`, or `AT_MOST_ONCE` but the issue I originally removed `EXACTLY_ONCE` for is not occurring with `AT_MOST_ONCE`.

nomisRev (approved):
I think it's definitely worth merging this PR! Thank you for looking into this @serras 🙌

PR #3294, opened 2023-11-09
Implement `fixedRate` using monotonic time source

serras:
The API we were using in `native` for `fx-coroutines` is deprecated, but fortunately there's an even better one. This also makes the tests, which were not behaving correctly under the `kotlinx-coroutines` test machinery, work again.


arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/flow.kt
```diff
@@ -209,16 +211,17 @@ public fun <A> Flow<A>.repeat(): Flow<A> =
...
 public fun <A> Flow<A>.metered(period: Long): Flow<A> =
   fixedRate(period).zip(this) { _, a -> a }
 
 @ExperimentalTime
 public fun fixedRate(
-  period: Duration,
+  periodInMillis: Long,
   dampen: Boolean = true,
-  timeStampInMillis: () -> Long = { timeInMillis() }
+  timeStamp: () -> ComparableTimeMark = { TimeSource.Monotonic.markNow() }
```
nomisRev, line 222: Would there be any benefit to taking `TimeSource` here instead? Or do we need `Monotonic` for `ComparableTimeMark`?

nomisRev (approved):
Perfect! `timeInMillis` had to be removed from Arrow Fx anyway 🙌

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/flow.kt
serras: I think we need `ComparableTimeMark` because comparison is used in at least one place. Also, the deprecation warnings from `timeInMillis` mentioned that this was the expected substitution, so I just went with it.

PR #3296, opened 2023-11-10
Add more options to memoization cache

### LINKED ISSUE #3133: ["Request"] Add basic cache size configuration option to memoize()
No specific version, came here from the Docs (https://arrow-kt.io/learn/collections-functions/memoize/)

What would you like to see?
It would be nice to be able to call `memoize()` with cache size supplied to avoid going out of memory. 

[Mnemonik](https://github.com/aballano/mnemonik) is archived, but had this possibility: 
```val memoized = ::anExpensiveFun.memoize(50)```
My current use case: I have a rather expensive function in an optimization project with Optaplanner. This function is called many times per second, so caching makes a lot of sense here. However, the underlying data also changes quite frequently and since we are operating on huge search spaces (vehicle routing), it would probably be required to keep the cache size limited. 

If it makes sense to include this feature, I'd be willing to give it a try.

### PR DESCRIPTION
**Pull request by serras**
Instead of rolling our own caches, this PR opens the door to using any library out there. Given our focus in Multiplatform, I've added a small bridge to [cache4k](https://github.com/ReactiveCircus/cache4k), which seems to have all the interesting options (size, eviction policies, and so on).

Fixes #3133

### CODE REVIEW

**Comment by serras** [comment]
@greyhairredbear do you think this PR would satisfy your needs from #3133?

**Comment by greyhairredbear** [comment]
@serras I suppose it does, at least `cache4k` does check the box for my use case. Many thanks for the implementation 👏 🙌

File: arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/MemoizedDeepRecursiveFunction.kt [hunk 1]
```diff
@@ -2,6 +2,30 @@ package arrow.core
 
 import arrow.atomic.Atomic
 import arrow.atomic.loop
+import kotlin.jvm.JvmInline
+import kotlin.jvm.JvmOverloads
+
+public interface MemoizationCache<K, V> {
+  public fun get(key: K): V?
+  public fun set(key: K, value: V): V
```
hoc081098, line 10: I think we can add `operator` modifier to `get` 🙏

```suggestion
  public operator fun get(key: K): V?
```

--> 
```
    when (val v = cache[x]) {
        ...
    }
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/MemoizedDeepRecursiveFunction.kt
```diff
@@ -2,6 +2,30 @@ package arrow.core
 
 import arrow.atomic.Atomic
 import arrow.atomic.loop
+import kotlin.jvm.JvmInline
+import kotlin.jvm.JvmOverloads
+
+public interface MemoizationCache<K, V> {
+  public fun get(key: K): V?
```
hoc081098, line 9: I think we can add `operator` modifier to `get` 🙏

```suggestion
  public operator fun get(key: K): V?
```

--> 
```
    when (val v = cache[x]) {
        ...
    }
```

serras: I actually left this one out because I was having some problems with platform-specific implementations :/

arrow-libs/core/arrow-cache4k/build.gradle.kts
```diff
@@ -0,0 +1,68 @@
+@file:Suppress("DSL_SCOPE_VIOLATION")
+
+plugins {
+  `java-library`
+  id(libs.plugins.kotlin.multiplatform.get().pluginId)
+  // alias(libs.plugins.arrowGradleConfig.kotlin)
```
nomisRev, line 6:
```suggestion
```

nit

nomisRev (approved):
This is really great work @serras! 🙌
Sorry for the unreasonable delay on the reviews 🙈

PR #3297, opened 2023-11-12
Updated zipOrAccumulate

yesitskev:
Added additional parameters to zipOrAccumulate


serras:
We used to have longer parameter counts in this kind of functions (not only `zipOrAccumulate`, also `parZip`, for example). However, we found that the binary size increased too much when compared with the usage in practice (very few people require more than 10 arguments.) Furthermore, I think that the nested usage is good enough.

@nomisRev @raulraja WDYT?

nomisRev:
I agree @serras, we originally supported up-to arity-22 but it was considered bloat by many. Would be cool if we had something like parameter packs in Swift.

I think nesting is rare enough to propose nesting, or creating a custom function in your codebase. At least we consider this for the _core_ library, we want happily help promote any library that exposes arity 10+ on top-of-Arrow.

yesitskev:
@nomisRev @serras would a happy medium be to then expose the internal API's which can facilitate extension libraries to go about creating large arity variants of some functions?

PR #3299, opened 2023-11-15
Arrow Optics ❤️ Compose

serras:
These are a few utilities which help working with `MutableState` (as in Compose) using optics.


arrow-libs/optics/arrow-optics-compose/build.gradle.kts
```diff
@@ -0,0 +1,83 @@
...
+  alias(libs.plugins.jetbrainsCompose)
+  alias(libs.plugins.android.library)
+}
+
+apply(from = property("ANIMALSNIFFER_MPP"))
+
+kotlin {
+  explicitApi()
+
+  jvm {
```
nomisRev, line 23: Off-topic: @serras how do you feel about doing this explicitly in Arrow instead of relying on the `arrowGradleConfig`?

Arrow Gradle Config came to live because of publishing, and I've in favor of replacing Arrow Gradle Config Nexus/Publish with https://github.com/vanniktech/gradle-maven-publish-plugin. I've tried it on a couple of my projects and it works better as what we have whilst supporting the same style of configuration we currently use.

serras: That sounds great (although having to define all the targets on each project seems tiresome). Maybe another thing to do in the `arrow-2` branch?

arrow-libs/optics/arrow-optics-compose/src/commonMain/kotlin/arrow/optics/Copy.kt
```diff
@@ -0,0 +1,29 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.update
+
+/**
+ * Modifies the value in this [MutableState]
+ * by applying the function [block] to the current value.
+ */
+public inline fun <T> MutableState<T>.update(block: (T) -> T) {
+  value = block(value)
+}
```
nomisRev, line 13: Oh, I thought it had `update` out of the box but it seems `androidx.compose.runtime.MutableState` is not an Atomic reference like `MutableStateFlow`.

IMO, we can keep or leave this `update` in favor of `setValue`.

nomisRev (approved):
Awesome stuff @serras! 🙌

arrow-libs/optics/arrow-optics-compose/src/commonMain/kotlin/arrow/optics/Copy.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import androidx.compose.runtime.MutableState
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.update
+
+/**
+ * Modifies the value in this [MutableState]
+ * by applying the function [block] to the current value.
+ */
+public inline fun <T> MutableState<T>.update(block: (T) -> T) {
```
hoc081098, line 11: I think we should note this method is not thread safe. It is only called on MainThread

arrow-libs/optics/arrow-optics-compose/src/commonMain/kotlin/arrow/optics/Flow.kt
```diff
@@ -0,0 +1,70 @@
...
+ */
+public fun <T, A> StateFlow<T>.optic(g: Getter<T, A>): StateFlow<A> = object : StateFlow<A> {
+  override val value: A
+    get() = g.get(this@optic.value)
+
+  override suspend fun collect(collector: FlowCollector<A>): Nothing =
+    this@optic.collect { collector.emit(g.get(it)) }
+
+  override val replayCache: List<A>
+    get() = this@optic.replayCache.map { g.get(it) }
```
hoc081098, line 31: Always contains single element.
```suggestion
    get() = listOf(value)
```

arrow-libs/optics/arrow-optics-compose/src/commonMain/kotlin/arrow/optics/Flow.kt
```diff
@@ -0,0 +1,70 @@
...
+    this@optic.tryEmit(lens.set(this@optic.value, value))
+
+  override suspend fun emit(value: A): Unit =
+    this@optic.emit(lens.set(this@optic.value, value))
+
+  override val subscriptionCount: StateFlow<Int>
+    get() = this@optic.subscriptionCount
+
+  override val replayCache: List<A>
+    get() = this@optic.replayCache.map { lens.get(it) }
```
hoc081098, line 64: Same as above 🙏

arrow-libs/optics/arrow-optics-compose/build.gradle.kts
nomisRev: > Maybe another thing to do in the arrow-2 branch?

Was wondering if we could do something interesting in `buildSrc` or something where we can define some top-level functions to avoid all the repetitive boilerplate. arrow-2 sounds good! Definitely not for 1.2.2 😅

arrow-libs/optics/arrow-optics-compose/src/commonMain/kotlin/arrow/optics/Copy.kt
serras: Thanks, I was not aware of this problem!

arrow-libs/optics/arrow-optics-compose/src/commonMain/kotlin/arrow/optics/Flow.kt
serras: I would prefer to keep this implementation, since it makes it more clear that we are just reusing the `replayCache` from the nested version.

nomisRev (approved):
Looks great @serras, thank you for this great work! I just have one small thought.

arrow-libs/optics/arrow-optics-compose/src/commonMain/kotlin/arrow/optics/Copy.kt
```diff
@@ -9,7 +10,9 @@ import kotlinx.coroutines.flow.update
  * by applying the function [block] to the current value.
  */
 public inline fun <T> MutableState<T>.update(block: (T) -> T) {
-  value = block(value)
+  Snapshot.withMutableSnapshot {
```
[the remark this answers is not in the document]
kyay10: `crossinline` would do the trick to not let suspend pass through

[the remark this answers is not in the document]
nomisRev: `crossinline` can be an even better solution here 👍

PR #3301, opened 2023-11-23
Arrow STM based on Compose

serras:
Compose has STM-like primitives in `Snapshot`. This PR implements the STM machinery over those; this should actually gives us more flexibility because the Compose `MutableStateX` implement way more operations than us.


serras:
Closing this, as it doesn't make much sense at this point

PR #3305, opened 2023-11-29
Unify SingletonRaise-esque apis and introduce impure builder for raise

kyay10:
Based on a suggestion by @CLOVIS-AI.

Not sure how Arrow handles binary compatibility issues because this introduces some. 
The name impure is not final, but I think it captures the essence of that builder well because it always returns `Unit`, so the action inside must be impure.


serras:
@kyay10 this sounds like a great way to unify the commonalities behind `Option`, `Nullable`, and similar types. However, the binary compatibility issue is a big one, since we have promised to keep it on 1.2.x.

Would you maybe have the time to apply this change over the `arrow-2` branch? There we still have leeway to make this kind of changes.

PR #3311, opened 2023-12-01
Unify SingletonRaise-esque APIs, and add impure

kyay10:
Based on a suggestion by @CLOVIS-AI. This is a followup from #3305 
The name `impure` is not final, but I think it captures the essence of that builder well because it always returns `Unit`, so the action inside must be impure.

@CLOVIS-AI gave this example of a function that uses `impure`
```kotlin
fun MutableSet<AccessRight>.addRightsForUser(user: User) = impure {
    ensure(…)
    ensure(…)

    add(AccessRight(…, user))
}
```

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
@@ -43,7 +43,7 @@ public inline fun <Error, A> either(@BuilderInference block: Raise<Error>.() -> 
  */
 public inline fun <A> nullable(block: SingletonRaise.() -> A): A? {
   contract { callsInPlace(block, InvocationKind.AT_MOST_ONCE) }
-  return merge { ignoreErrors(null, block) }
+  return merge { ignoreErrors({ null }, block) }
 }
 
 /**
@@ -70,7 +70,7 @@ public inline fun <A> result(block: ResultRaise.() -> A): Result<A> {
  */
 public inline fun <A> option(block: SingletonRaise.() -> A): Option<A> {
   contract { callsInPlace(block, InvocationKind.AT_MOST_ONCE) }
-  return merge { ignoreErrors(None, block).some() }
+  return merge { ignoreErrors({ None }, block).some() }
 }
 
 /**
@@ -136,27 +136,21 @@ public inline fun <Error, A> iorNel(noinline combineError: (NonEmptyList<Error>,
  */
 public inline fun impure(block: SingletonRaise.() -> Unit) {
   contract { callsInPlace(block, InvocationKind.AT_MOST_ONCE) }
-  return merge { ignoreErrors(Unit, block) }
+  return merge { ignoreErrors({}, block) }
 }
 
 /**
  * Implementation of [Raise] used by `ignoreErrors`.
  * You should never use this directly.
  */
-public class IgnoreErrorsRaise<N>(
-  private val raise: Raise<N>,
-  private val error: N
-) : SingletonRaise(), Raise<Any?> {
-  @RaiseDSL
-  override fun raise(): Nothing = raise.raise(error)
-
+public abstract class IgnoreErrorsRaise : SingletonRaise(), Raise<Any?> {
   @RaiseDSL
   override fun raise(r: Any?): Nothing = raise()
 
   @JvmName("recoverIgnoreErrors")
   @RaiseDSL
   public inline fun <A> recover(
-    @BuilderInference block: IgnoreErrorsRaise<*>.() -> A,
+    block: IgnoreErrorsRaise.() -> A,
     recover: () -> A,
   ): A {
     contract {
@@ -238,7 +232,7 @@ public sealed class SingletonRaise {
 
   @RaiseDSL
   public inline fun <A> recover(
-    @BuilderInference block: SingletonRaise.() -> A,
+    block: SingletonRaise.() -> A,
     recover: () -> A,
   ): A {
     contract {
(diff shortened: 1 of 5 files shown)


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -135,84 +126,106 @@ public inline fun <Error, A> iorNel(noinline combineError: (NonEmptyList<Error>,
...
+  private val error: N
+) : SingletonRaise<Any?>() {
   @RaiseDSL
-  override fun raise(r: Any?): Nothing =
-    raise.raise(error())
+  override fun raise(): Nothing =
+    raise.raise(error)
+}
+
+public sealed class SingletonRaise<in Error> : Raise<Error> {
```
serras, line 155: Is there a specific reason for including the `Error` type parameter here? I was thinking that maybe it's possible to make `SingletonRaise : Raise<Any?>`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -221,13 +234,23 @@ public class NullableRaise(private val raise: Raise<Null>) : Raise<Null> by rais
    */
   @RaiseDSL
   public inline fun <A> ignoreErrors(
-    @BuilderInference block: IgnoreErrorsRaise<Null>.() -> A,
+    @BuilderInference block: SingletonRaise<Any?>.() -> A,
   ): A {
     contract { callsInPlace(block, InvocationKind.AT_MOST_ONCE) }
-    return block(IgnoreErrorsRaise(this) { null })
+    return when (this) {
+      is IgnoreErrorsRaise<*> -> block(this)
```
serras, line 241: I'm really having trouble understanding this implementation :/

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
kyay10: I have the `Error` to restrict what types `raise(foo)` is allowed to accept. That's to match the current behaviour of `nullable`, `option` etc where they only allow raising `null` and `None` respectively. If `SingletonRaise` is `Raise<Any?> by default, that can result in unintentionally ignoring errors, and will likely be a significant breaking change. I will experiment however with perhaps some form of intersection-type shenanigans and will report back.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/raise/NullableSpec.kt
```diff
@@ -69,14 +69,6 @@ class NullableSpec {
     } shouldBe "1"
   }
 
-  @Test fun bindingEitherInNullable() = runTest {
```
kyay10, line 72: I felt justified in removing these tests because the usage of `bind()` in this way is rather pointless because `Either#merge` does the same thing, and doesn't require a `Raise` at all. A `bind` method for `Either<Nothing, A>` can be added inside `SingletonRaise`, but again, it wouldn't be using anything at all related to `raise()`, since it would simply call `Either.merge()`

serras (approved):
Thanks for this work, I think it really simplifies the implementation a lot!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -135,84 +126,129 @@ public inline fun <Error, A> iorNel(noinline combineError: (NonEmptyList<Error>,
...
-    contract { returns() implies (value != null) }
-    return ensureNotNull(value) { null }
+  public inline fun <A> recover(
+    @BuilderInference block: IgnoreErrorsRaise<*>.() -> A,
+    recover: () -> A,
+  ): A {
+    contract {
+      callsInPlace(block, InvocationKind.AT_MOST_ONCE)
+      callsInPlace(recover, InvocationKind.AT_MOST_ONCE)
+    }
```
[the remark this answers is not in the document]
kyay10: These are the same contracts that [recover has in Arrow 2](https://github.com/arrow-kt/arrow/commit/e94a9b513306e3933609887227df7ae1754a6cdf#diff-a4bdffa932e6b2f0c2f74a775c82764eb7e7fb25a3053a522cce8cb3ac978cf6R337) which were added in PR #3293, so I was replicating its semantics. Is this case different from the cases in #3293?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Raise.kt
```diff
@@ -665,6 +664,13 @@ public inline fun <Error, OtherError, A> Raise<Error>.withError(
   return recover(block) { raise(transform(it)) }
 }
 
+@RaiseDSL
+public inline fun <Error, A> Raise<Error>.ignoreErrors(value: Error, @BuilderInference block: IgnoreErrorsRaise<*>.() -> A): A {
```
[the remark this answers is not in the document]
kyay10: Looking at where other `() -> Error` functions exist, I saw that `ensure` and `ensureNotNull` call that parameter `raise`, so I'll follow that convention.

kyay10:
Not sure why this run failed. Any chance you can retrigger it? I don't seem to be able to do that. The error seems to be an uncompleted coroutine in arrow-fx-stm, which isn't related to my change at all.

serras:
@kyay10 should we close this in favor of #3328?

kyay10:
Closed in favour of #3328

PR #3315, opened 2023-12-08
chore(KProperty1.lens): Perform check for data class only once.

kyay10:
This PR does the setup work for finding the copy method of the data class and finding the relevant parameter only once, hence failing earlier if used not on a data class

This idea is based on a [video](https://www.youtube.com/watch?v=hTLUELJDaec) by @dmcg.

### DIFF
--- a/arrow-libs/optics/arrow-optics-reflect/src/main/kotlin/arrow/optics/Reflection.kt
+++ b/arrow-libs/optics/arrow-optics-reflect/src/main/kotlin/arrow/optics/Reflection.kt
@@ -4,7 +4,9 @@ import arrow.core.Either
 import arrow.core.left
 import arrow.core.right
 import kotlin.reflect.KClass
+import kotlin.reflect.KFunction
 import kotlin.reflect.KProperty1
+import kotlin.reflect.cast
 import kotlin.reflect.full.instanceParameter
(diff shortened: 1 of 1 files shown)


serras (approved):
Looks good to me. Using a closure to keep the copy is quite clever!

arrow-libs/optics/arrow-optics-reflect/src/main/kotlin/arrow/optics/Reflection.kt
```diff
@@ -54,13 +57,23 @@ public val <S, A> KProperty1<S, List<A>>.every: Every<S, A>
...
-  if (klass == null || !klass.isData || copy == null) {
+  val klass = kClass
+  val copy = klass.copyMethod
+  val fieldParam = copy.parameters.first { it.name == name }
+  val instanceParam = copy.instanceParameter!!
+  return { value, newField -> klass.cast(copy.callBy(mapOf(instanceParam to value, fieldParam to newField))) }
+}
+
+@Suppress("UNCHECKED_CAST")
+private val <S> KProperty1<S, *>.kClass: KClass<S & Any>
```
serras, line 70: I don't understand 100% why the result here is `KClass<S & Any>`, instead of merely `KClass<*>`.

PR #3316, opened 2023-12-08
Fix 3235

### LINKED ISSUE #3235: IllegalAccessError when using Raise
Arrow version: 1.2.1

I'm trying to integrate `Raise` into my test library, [Prepared](https://gitlab.com/opensavvy/prepared). Since raising usually represents an error, I want to be able to declare that a test fails if an error is raised. To help with debugging, I'd like to enable tracing in those cases (since this is only for testing, I don't mind the performance implications).

I want to be able to declare something like:
```kotlin
test("Some test") {
    failOnRaise {
        raise(5)
    }
}
```

For now, here's the implementation I have:
```kotlin
@ExperimentalTraceApi
inline fun <Failure, Success> failOnRaise(block: Raise<Failure>.() -> Success): Success {
	val result = either {
		traced(block) { trace, error ->
			throw AssertionError("An operation raised $error\n${trace.stackTraceToString()}")
				.apply {
					for (suppressed in trace.suppressedExceptions())
						addSuppressed(suppressed)
				}
		}
	}

	check(result is Either.Right) { "Impossible situation: we throw an error when the passed block raises, but it still gave us a failed either: $result" }
	return result.value
}
```

However, when I execute it, I get:
```
java.lang.AssertionError: Expected exception java.lang.AssertionError but a IllegalAccessError was thrown instead.
```
The stack trace is unhelpful (it doesn't go further than the test function).

Do you know what could cause this?

---

The entire code is available [here](https://gitlab.com/opensavvy/prepared/-/merge_requests/18).

As an aside, would it be possible to expose the underlying exception in `traced`? It would be easier to simply mark it as the `cause` of the `AssertionError`.

kyay10:
fixes #3235. See that issue for more information


nomisRev:
@kyay10 `RaiseCancellationException` is still referenced in the `e.raisedOrRethrow`. It's safe there because it's not inlined, and thus called from our module? 🤔

kyay10:
Exactly. It's fine there, just not when inlined. In fact, the reason this wasn't caught earlier is because `fold` doesn't catch `RaiseCancellationException`, it catches `CancellationException`. `traced` is the only inline method that deals with `RaiseCancellationException` directly, and I think simply no one has really tried out `traced` outside of this repo, hence why no one ran into this before.

serras:
This makes sense to me. One question, though, @kyay10: why this change and not making `RaiseCancellationException` `@PublishedApi internal`?

kyay10:
@serras I just figured making it part of the API wasn't really needed. Making `RaiseCancellationException` published would work in the same way. Note, btw, that `fold` also simply catches `CancellationException` and relies on `raisedOrRethrow` to determine whether the exception belongs to the current raise, so this change matches `fold` more closely

serras:
@kyay10 thanks for the info! I agree that it makes sense to mirror `fold`.

serras (approved):

PR #3318, opened 2023-12-12
chore(arrow-atomic): Cleanup atomic code

kyay10:
Only public api additions are internal helper functions. Some of them might be good to promote to public, but that's up to you!

### DIFF
--- a/arrow-libs/core/arrow-atomic/src/commonMain/kotlin/arrow/atomic/Atomic.kt
+++ b/arrow-libs/core/arrow-atomic/src/commonMain/kotlin/arrow/atomic/Atomic.kt
@@ -1,7 +1,7 @@
 package arrow.atomic
 
 /**
- * [Atomic] value of [A].
+ * [Atomic] value of [V].
  *
  * ```kotlin
  * import arrow.atomic.AtomicInt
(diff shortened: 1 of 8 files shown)


arrow-libs/core/arrow-atomic/src/commonMain/kotlin/arrow/atomic/Atomic.kt
```diff
@@ -44,44 +44,38 @@ public var <T> Atomic<T>.value: T
...
+  return compareAndSet(cur, upd).also { if (it) onUpdated(cur, upd) }
+}
+
+/**
+ * while (true) as an expression.
+ */
+@PublishedApi
+internal inline fun forever(block: () -> Unit): Nothing {
+  while (true) block()
 }
```
serras, line 81: Would it be possible to move this function to some other file, to stress that it's shared between all implementations?

serras (approved):
Thanks for taking the time, less repetition is always nicer :)

nomisRev (approved):
Looks good to me @kyay10.

I've personally never really cared about such code optimisations (in Arrow), since the older code was simpler to understand in exchange for 40 extra lines of code.

serras:
@kyay10 I've decided to remove `forever`, since it was leaking in the ABI, and go back to `while(true)`

PR #3326, opened 2023-12-20
Correctly escape class names in Optics KSP when a property clashes with a package name

### LINKED ISSUE #3134: Optics KSP Plugin breaks if the word “data” is part of the package
With the Arrow 1.2.1 (and possibly 1.2.0) Optics KSP plugin, one of the files seem to get an import with a broken alias. I don't quite know why it's even importing the type, as the type it imports is in the same package as the generated code. And the import alias is broken because the package it's in includes the segment `data`:

Here's the top of the generated _WorkoutHistoryPage__Optics.kt_:

```kt
package com.sats.core.`data`.workouts.models

import arrow.core.left
import arrow.core.right
import arrow.core.toOption
import com.sats.core.`data`.workouts.models.WorkoutHistoryPage as comsatscore`data`workoutsmodelsWorkoutHistoryPage
```

This results in a compilation error because of the import alias:

> WorkoutHistoryPage__Optics.kt:6:78 Expecting a top level declaration

This does work in 1.1.5, where the import is not generated. (edited)

serras:
Fixes #3134


nomisRev (approved):
Thank you @serras for fixing this issue so quickly! 🙌 🥳

raulraja (approved):

PR #3327, opened 2023-12-20
Run K2 using 2.0.0-Beta2

serras:
I think we should stop tracking the `-dev` versions of 2.0, and just use the beta's, since this is what our users are ultimately going to see

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -179,7 +179,7 @@ jobs:
       - name: jvmTest (K2 enabled)
         uses: gradle/gradle-build-action@v2
         with:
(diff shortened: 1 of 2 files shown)


nomisRev (approved):
Sounds good to me! I thought we were able to track the latest DEV/EAP and build against that to help track down regression but building against Beta2 sounds good for me ☺️

serras:
> Sounds good to me! I thought we were able to track the latest DEV/EAP and build against that to help track down regression but building against Beta2 sounds good for me ☺️

I think that made sense when this was the only way to track 2.0. However, it's becoming increasingly difficult to keep up with all the dependencies (especially some like KSP), and I think this provides a good middle term.

PR #3328, opened 2023-12-20
Alternate `SingletonRaise`

serras:
This is a variation of #3311 in which the `nullable`, `option`, and all of those directly use a `SingletonRaise` which extends `Raise<Any?>`. The consequence is that inside any of those blocks one can do `eitherThing.bind()` directly, and the error is "converted" into a `null`.

@kyay10 what do you think about this alternative design?

(The reason why I didn't do a PR against @kyay10's PR is to be able to better see what are the differences in the public API between both versions).

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
@@ -11,8 +11,21 @@ import kotlin.contracts.ExperimentalContracts
 import kotlin.contracts.InvocationKind
 import kotlin.contracts.contract
 import kotlin.experimental.ExperimentalTypeInference
+import kotlin.js.JsName
 import kotlin.jvm.JvmMultifileClass
 import kotlin.jvm.JvmName
+
+@RaiseDSL
+@JvmName("singleton")
+public inline fun <A> singleton(
+  crossinline raise: () -> A,
+  @BuilderInference block: SingletonRaise.() -> A,
+): A {
+  contract {
+    callsInPlace(block, InvocationKind.AT_MOST_ONCE)
+  }
+  return fold({ block(SingletonRaise(this)) }, { raise() }, { it })
+}
 
 /**
  * Runs a computation [block] using [Raise], and return its outcome as [Either].
@@ -26,7 +39,7 @@ import kotlin.jvm.JvmName
  */
 public inline fun <Error, A> either(@BuilderInference block: Raise<Error>.() -> A): Either<Error, A> {
   contract { callsInPlace(block, InvocationKind.AT_MOST_ONCE) }
-  return fold({ block.invoke(this) }, { Either.Left(it) }, { Either.Right(it) })
+  return fold(block, { Either.Left(it) }, { Either.Right(it) })
 }
 
 /**
@@ -43,7 +56,7 @@ public inline fun <Error, A> either(@BuilderInference block: Raise<Error>.() -> 
  */
 public inline fun <A> nullable(block: SingletonRaise.() -> A): A? {
   contract { callsInPlace(block, InvocationKind.AT_MOST_ONCE) }
-  return merge { ignoreErrors({ null }, block) }
+  return singleton({ null }, block)
 }
 
 /**
@@ -70,7 +83,7 @@ public inline fun <A> result(block: ResultRaise.() -> A): Result<A> {
  */
 public inline fun <A> option(block: SingletonRaise.() -> A): Option<A> {
   contract { callsInPlace(block, InvocationKind.AT_MOST_ONCE) }
-  return merge { ignoreErrors({ None }, block).some() }
+  return singleton({ None }, { block().some() })
 }
 
 /**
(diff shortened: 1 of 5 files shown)


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -47,12 +50,12 @@ public inline fun <Error, A> either(@BuilderInference block: Raise<Error>.() ->
  * Read more about running a [Raise] computation in the
  * [Arrow docs](https://arrow-kt.io/learn/typed-errors/working-with-typed-errors/#running-and-inspecting-results).
  *
- * @see NullableRaise.ignoreErrors By default, `nullable` only allows raising `null`.
- * Calling [ignoreErrors][NullableRaise.ignoreErrors] inside `nullable` allows to raise any error, which will be returned to the caller as if `null` was raised.
+ * @see SingletonRaise.ignoreErrors By default, `nullable` only allows raising `null`.
+ * Calling [ignoreErrors][SingletonRaise.ignoreErrors] inside `nullable` allows to raise any error, which will be returned to the caller as if `null` was raised.
  */
-public inline fun <A> nullable(block: NullableRaise.() -> A): A? {
+public inline fun <A> nullable(block: SingletonRaise<Nothing?>.() -> A): A? {
```
kyay10, line 56: Should we be allowing `raise(null)`? I struggle to see any uses for this, perhaps outside of someone having an `Either<Nothing?, R>`, but that seems rare. Should this instead be `SingletonRaise<Unit>`? I much prefer using Unit to denote that you shouldn't be raising any real values. This would also help the transition into context receivers eventually because all the APIs on `SingletonRaise` would be applicable to `Raise<Unit>`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -136,83 +139,98 @@ public inline fun <Error, A> iorNel(noinline combineError: (NonEmptyList<Error>,
...
+  public fun raise(): Nothing = raise.raise(Unit)
+
   @RaiseDSL
-  override fun raise(r: Any?): Nothing =
-    raise.raise(error())
+  override fun raise(r: E): Nothing = raise()
 
   @RaiseDSL
-  public fun ensure(value: Boolean): Unit = ensure(value) { null }
+  public fun ensure(condition: Boolean) {
```
kyay10, line 162: Maybe this can be an extension function? In fact, all methods here would ideally be extensions, but without context receivers, that can't happen yet.

serras: Indeed. Having them as part of the class also improves their discoverability, because no additional imports are needed to use them.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
serras: Keeping this one is important, because we really want the `Raise` parameter to reflect the error type, which is `None` in the case of `Option`, for example. Breaking this would also break users of Arrow 1.x with no apparent reason.

kyay10: That's fair. I'm questioning whether users of Arrow 1.x used `raise(None)` and `raise(null)`. Maybe we can add extra methods for compatibility?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
kyay10: True. Note that the corresponding functions for `Raise` are extensions though, so in fact importing `arrow.core.raise.ensure` would import both the `Raise<E>` version and the `Raise<Unit>` version, if there was to be one

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
serras: I'm not sure, but we would still be breaking compatibility. We promised that code compiled with 1.2.x would be source-compatible with 2.0 (except for deprecated methods), so we should keep (1) the `raise` with argument, (2) the fact that they are `Raise<Nothing?>` and `Raise<None>`

kyay10: Good point. `Raise<Nothing?>` and `Raise<None>` would likely show up in people's code if they were dealing with `nullable` and `option`. I think in a context world, we can have `nullable` introduce `context(Raise<Unit>, Raise<Nothing?>)` and same for `option`

kyay10 (approved):
I like how concise this implementation has become! 
It's good to note that this is considerably more involved than the other `Raise` wrappers are, and so when context parameters are available, converting `SingletonRaise` to that world will need some tweaks to its design. I think ultimately it can be replaced with contextual extensions on `Raise<Unit>`, but, as the discussions above show, some care is needed to not break reasonable code that used `Raise<None>` and similar.

serras:
@nomisRev we're now happy with the implementation. Feel free to merge if you find it ok.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/raise/OptionSpec.kt
```diff
@@ -1,9 +1,7 @@
 package arrow.core.raise
 
-import arrow.core.None
-import arrow.core.Some
-import arrow.core.some
-import arrow.core.toOption
+import arrow.core.*
```
i-walker, line 3: nit: imports

i-walker (approved):
Looks great @serras :)

nomisRev (commented):
Nice work @serras @kyay10 👏 👏

PR #3329, opened 2023-12-20
Safeguard for leaked 'raise' or 'bind'

serras:
This should be accompanied with a text in the Arrow docs.

It would be great to also detect `Deferred` or `Flow` from `kotlinx.coroutines`, but this doesn't seem like a good enough reason to add a dependency on those.


nomisRev (approved):
Oh, I really love this!! 😍 Thanks @serras, love this addition.

> It would be great to also detect Deferred or Flow from kotlinx.coroutines, but this doesn't seem like a good enough reason to add a dependency on those.

Totally agree. Still wondering if this could be generalised, and if it could be useful to have in the language as an annotation similar to `@RestrictSuspension` but in relation to escaping of operations in DSLs. That probably opens another can of worms 😅

PR #3330, opened 2023-12-20
Update kotlin-compile-testing

serras:
The version we are currently using lags quite behind the current Kotlin compiler ([here's a comment about it](https://github.com/tschuchortdev/kotlin-compile-testing/pull/394#issuecomment-1788188847)). There's a [new fork](https://github.com/zacsweers/kotlin-compile-testing) which promises compatibility with newer releases (and even has a branch for K2 support). This PR switches to that one.


nomisRev (approved):
zacsweers should be good! Thank you @serras 🙌

PR #3332, opened 2023-12-21
Separate 'arrow-eval' module

serras:
I've spent some time trying to write a good guide for #3039, and have come to the conclusion that a full replacement for `Eval` is difficult to achieve. So I've taken one of the suggestions there: move it into its own separate `arrow-eval` module. I think we can do like we've done with `arrow-resilience` and `arrow-functions`, and nudge people to move into the new module if we bring it in 1.2.2, and then remove it from `arrow-core` in 2.0.

The code is ready to be merged too in `arrow-2`, with just a few adjustments needed in the Gradle build file.


serras:
Summoning @404- and @pbtura, which reported some interest in this.

nomisRev (approved):
Perfect, thank you so much @serras 🙌

arrow-libs/core/arrow-eval/src/commonMain/kotlin/arrow/eval/Eval.kt
```diff
@@ -0,0 +1,579 @@
...
+ *
+ * The Later and Always are both lazy strategies while Now is eager.
+ * Later and Always are distinguished from each other only by
+ * memoization: once evaluated Later will save the value to be returned
+ * immediately if it is needed again. Always will run its computation
+ * every time.
+ *
+ * methods, which use an internal trampoline to avoid stack overflows.
+ * Computation done within .map and .flatMap is always done lazily,
+ * even when applied to a Now instance.
```
franciscodr, line 29: This text seems to be incomplete

franciscodr (approved):

arrow-libs/core/arrow-eval/src/commonMain/kotlin/arrow/eval/Eval.kt
serras: The file is just restored from the original version :/

PR #3333, opened 2023-12-21
Alternate `SingletonRaise`: Reduce lambda creation, and fix CCE.

kyay10:
Guided by adding `inline` wherever I can, I think this might be a better way of expressing `SingletonRaise`. The error being raised is always `Unit`, and the onus of transforming it to other singletons is on `recover`. 

The previous code failed the `raisingInIgnoreErrorsReturnsX` tests with a CCE because it would accidentally leak error values. With that in mind, `ignoreErrors` becomes just an unsafe cast. It could've instead been `block(SingletonRaise(raise))`, but that's rather pointless.

`withSingletonError` could have a better name. Perhaps `withConstantError`?


serras:
Thanks for the changes, I really like how this is going!

I still have one question: why do we need `withSingletonError`? This is just `withError` for a more specific case. My opinion is that we should just remove it.

kyay10:
I like the direction this is going as well! 
Agreed about `withSingletonError`, it should be a separate proposal later. Initially, I refactored some bits of `singleton` into `withSingletonError`, but now it's not needed anymore. It has its own merits as a `withError` with the type fixed to `Unit` basically, because it allows you to call `ensure(foo)`, `ensureNotNull(foo)`, `Option.bind()` etc by specifying the error beforehand (and so it introduces a "default error"), but it should be a separate proposal.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -221,27 +222,30 @@ public class SingletonRaise<in E>(
...
 
   /**
    * Introduces a scope where you can [bind] errors of any type,
    * but no information is saved in the [raise] case.
    */
   @RaiseDSL
-  public fun <A> ignoreErrors(
-    block: Raise<Any?>.() -> A,
+  public inline fun <A> ignoreErrors(
+    block: SingletonRaise<Any?>.() -> A,
```
serras, line 242: I'm a bit afraid of this change. If we have `Raise<Any?>`, we can directly use `ignoreErrors` with some callable reference of the right type.

kyay10: Hmmm, I haven't tested this, but wouldn't that just work? Any function that needs `Raise<Any?>` would accept `SingletonRaise<Any?>`. I'll add a test for this and see what happens.

kyay10: Added test! It functions just fine. Refining the type seems to only allow extra functionality.

serras (approved):
Awesome work!

PR #3337, opened 2023-12-23
Fix traced CCE when nested different types.

kyay10:
When traced is used in a nested manner, a `raise` call that refers to the outer traced `Raise` gets intercepted by the inner `traced`. The result of this is that a CCE gets thrown because the inner `traced` only expects errors of the type it declares. 

I think more generally any method that intercepts a `RaiseCancellationException` *cannot* function properly without the chance for CCE. This means that no methods in `Arrow` should be allowed to *reuse* a `Raise` instance and intercept its error with `raisedOrRethrow`, and instead those methods need to create their own `Raise` instance inside. The only method that does that is `traced`, but perhaps an internal note should be added to never use `raisedOrRethrow` except if you yourself created the `Raise` instance that you're using it on


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
```diff
@@ -224,17 +224,29 @@ public inline fun <Error, A> Raise<Error>.traced(
...
     val r: Error = e.raisedOrRethrow(nested)
     trace(Trace(e), r)
-    if (isOuterTraced) throw e else raise(r)
+    // If our outer Raise happens to be traced
+    // Then we want the stack trace to match the inner one
+    try {
+      raise(r)
+    } catch (rethrown: CancellationException) {
+      throw rethrown.withCause(e)
+    }
```
serras, line 240:
```suggestion
    } catch (rethrown: RaiseCancellationException) {
      throw RaiseCancellationException(rethrown.raised, rethrown.raise, e.cause ?: rethrown.cause)
    }
```

kyay10: We can't do this because of #3235 unless we `@PublishedApi` `RaiseCancellationException`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
```diff
@@ -224,17 +224,29 @@ public inline fun <Error, A> Raise<Error>.traced(
...
   }
 }
 
+@PublishedApi
+internal fun CancellationException.withCause(cause: CancellationException): CancellationException = when (this) {
```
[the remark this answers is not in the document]
kyay10: It encapsulates doing a check with the `RaiseCancellationException`, which is private. I can instead make it `@PublishedAPI internal`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
nomisRev: I've been contemplating doing this.. but I've been a bit absent last 2 months due to some personal things keeping me busy.

My rationale for making it public is because I've had a scenario or two where I wanted to differentiate between `RaiseCancellationException`, and (`Job`) `CancellationException`. It's in really low-level edge case scenarios though, but it would allow figuring out if the `CancellationException` comes from Arrow, from KotlinX Coroutines, or from somewhere else.

Perhaps it warrants `@DelicateApi` or something, like KotlinX does to force `@OptIn` so it requires "flagging" low-level code.

WDYT?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
```diff
@@ -224,17 +224,29 @@ public inline fun <Error, A> Raise<Error>.traced(
...
   trace: (trace: Trace, error: Error) -> Unit
 ): A {
-  val isOuterTraced = this is DefaultRaise && isTraced
-  val nested: DefaultRaise = if (isOuterTraced) this as DefaultRaise else DefaultRaise(true)
+  val nested = DefaultRaise(true)
   return try {
-    block.invoke(nested)
+    block(nested).also { nested.complete() }
   } catch (e: CancellationException) {
+    nested.complete()
```
nomisRev, line 231: Good catch! Now I am wondering why those weren't there... 🤔
Almost feels like I would've done that on purpose, or some really bad regression happened here 😅

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
nomisRev, line 240: I'm not entirely sure I follow the changes here... When would `raise` throw `CancellationException`?

kyay10: I think giving users that option is a good idea, especially because we already have tests confirming that `try-catch` can recover from raise. We'd have to make both Raise exceptions public thus. It's another question if we should allow inspecting the error, too.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
kyay10: I believe it's because previously this code would reuse a `DefaultRaise(true)` if it found it on the outside, and thus it didn't `complete()` it. Reusing the instance is what caused the CCE though, and so with that gone we can get better errors for leaks again.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
kyay10: The point here is to catch the `RaiseCancellationException` that `raise` will absolutely throw, and add our cause onto it to preserve the stacktrace information. Perhaps there's a clearer way to do that, but this is what I could come up with.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
nomisRev: Right, thank you for clarifying ☺️

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
nomisRev: Initially I would say don't expose it if we don't have a need for it.

I got a small WIP, that relies on some new functionality and unofficial support of Kotlin 😅
It adds a `public sealed class`, with expect/actual implementation to disable the stack traces.

I'm going to raise it as a PR targeting this one, so we can discuss further based on a code example ☺️

nomisRev:
@kyay10 I cannot target the branch of your fork from here, https://github.com/arrow-kt/arrow/pull/3349.

Also, I just realised you're not a member of the Arrow org. Would you like to be? Then you can make branches directly on Arrow, officially approve PRs, etc ☺️ You've been extremely helpful over the years, and have been very active member both here and on Slack so it seems **waaay** overdue 😅

kyay10:
@nomisRev Absolutely yes!! Thank you so, so much for the kind words! Words can't describe how grateful I am; Arrow has provided a great and welcoming environment for me to learn and grow, and hence I can only be honoured to help out whenever I can!

I'll have a look at the PR now!

nomisRev:
@kyay10 I'm so glad to hear that! ☺️

You should've received an invitation to join the Arrow org! So feel free to work directly on the Arrow repo now, etc 😉 Welcome to the team 🙌

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
nomisRev: This can now be replaced by the sealed public error for `Raise`, and then this PR is ready to be merged IMO 🥳 🙌

serras:
@nomisRev @kyay10 please check that I've merged `main` here correctly. If so, please proceed to squash and merge into `main`.

serras:
Well, I've broken the build and I don't know how to fix it... :/

kyay10:
I'll have a look in a moment! `main` I think just had significant changes to how `RaiseCancellationException` works

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
```diff
@@ -219,24 +220,37 @@ public inline fun <Error, A, B> foldUnsafe(
...
+      throw rethrown.withCause(e)
+    }
   }
 }
 
+@PublishedApi
+@DelicateRaiseApi
+internal fun Traced.withCause(cause: Traced): Traced =
+  Traced(raised, raise, cause)
+
```
nomisRev, line 250:
```suggestion
```

This is currently being added to the public API, but we're only using it once on line 241? Can we inline it there instead to avoid having it in the binary?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
```diff
@@ -219,24 +220,37 @@ public inline fun <Error, A, B> foldUnsafe(
...
+    nested.complete()
     val r: Error = e.raisedOrRethrow(nested)
     trace(Trace(e), r)
-    if (isOuterTraced) throw e else raise(r)
+    // If our outer Raise happens to be traced
+    // Then we want the stack trace to match the inner one
+    try {
+      raise(r)
+    } catch (rethrown: Traced) {
+      throw rethrown.withCause(e)
```
nomisRev, line 241:
```suggestion
      throw Traced(raised, raise, cause)
```

nomisRev (commented):
Thank you for the awesome work @kyay10! 🙌 🙏

This looks great, and I really like the clean-up with `RaiseCancellationException` in `foldUnsafe` it becomes very clear now how it works.

I would personally only like to apply my suggestion, solely because it avoids having it in the public binary.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
nomisRev: Oh... right 😓 I missed that. No, this is much better! You definitely made the right call on that one 😉
Great work @kyay10! 👏 👏 👏

nomisRev (approved):
Oh... right 😓 I missed that. No, this is much better! You definitely made the right call on that one 😉
Great work @kyay10! 👏 👏 👏

PR #3338, opened 2023-12-24
Rework IorRaise impl to use EmptyValue, and add tests

kyay10:
A secondary goal of this PR is to remove the complexity that was inside `IorRaise.recover`, and hence transition closer to an implementation that would work with contexts (with `addError` being provided as a separate context, hence allowing the normal `recover` method to do the job right).


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -304,14 +292,11 @@ public class OptionRaise(private val raise: Raise<None>) : Raise<None> by raise
...
-  private val state: Atomic<Option<Error>>,
+public abstract class IorRaise<Error> @PublishedApi internal constructor(
   private val raise: Raise<Error>,
-) : Raise<Error> {
-
-  @RaiseDSL
-  override fun raise(r: Error): Nothing = raise.raise(combine(r))
+) : Raise<Error> by raise {
+  @PublishedApi
+  internal abstract fun addError(e: Error)
```
serras, line 299: I think this function should still be called `combine`, there's no need to break API compatibility.

serras (changes requested):
This seems like a great idea; however if we are going to accept this in Arrow 1.x, there should be no removals at the API level. And to be honest, I don't fully understand we cannot have the `Atomic<Any?>` be part of `IorRaise` as we have in the current implementation.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
kyay10: I didn't want the `Atomic` to be part of `IorRaise` because it's an implementation detail that it doesn't need, same with `combine`. `IorRaise` just needs a general place to add extra errors into whenever there's an `Ior.Both`. I agree though that ABI compatibility is important, so I'll update this so the ABI is identical.
Also, the issue with exposing a `combine` is that the previous implementation erroneously used it inside `raise`, which prevents `try-catch` from recovering from it. I'll add a test for that as well. There's no real reason why `IorRaise` ever needs to read the accumulated errors so far. The `ior` builder needs to have that info, of course, but only as an implementation detail.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -306,12 +289,12 @@ public class OptionRaise(private val raise: Raise<None>) : Raise<None> by raise
  */
 public class IorRaise<Error> @PublishedApi internal constructor(
   @PublishedApi internal val combineError: (Error, Error) -> Error,
-  private val state: Atomic<Option<Error>>,
+  @PublishedApi internal val state: Atomic<Any?>,
```
serras, line 292: Would it be possible to keep this private?

kyay10: That's an interesting question... There is a subtle difference in behaviour if we keep this private (and thus implement `recover` as `ior(combineError, block).bind()`. Namely, if you `throw` inside a `recover`, the values won't get accumulated. Not sure which behaviour is more "correct", but it's important to note that the second behaviour, where throwing "rolls back" the accumulation is hard to replicate with contexts. Conceptually, `recover` should only replace the `Raise` instance you're dealing with, and nothing else, at least that's my understanding.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -6,16 +6,7 @@ package arrow.core.raise
...
-import arrow.core.Ior
-import arrow.core.IorNel
-import arrow.core.NonEmptyList
-import arrow.core.NonEmptySet
-import arrow.core.None
-import arrow.core.Option
-import arrow.core.Some
-import arrow.core.getOrElse
-import arrow.core.identity
+import arrow.core.*
```
nomisRev, line 9: No star imports.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -306,12 +289,12 @@ public class OptionRaise(private val raise: Raise<None>) : Raise<None> by raise
...
 public class IorRaise<Error> @PublishedApi internal constructor(
   @PublishedApi internal val combineError: (Error, Error) -> Error,
-  private val state: Atomic<Option<Error>>,
+  private val state: Atomic<Any?>,
   private val raise: Raise<Error>,
-) : Raise<Error> {
-
-  @RaiseDSL
-  override fun raise(r: Error): Nothing = raise.raise(combine(r))
+) : Raise<Error> by raise {
```
nomisRev, line 294: I don't think we can remove the override for `raise`, right? Don't we otherwise `combine` the error when doing `raise` now, are we? 🤔

I think a test is missing here:

```kotlin
ior(String::plus) {
  Ior.Both("Hello", Unit).bind()
  raise("World")
} shouldBe Ior.Left("Hello World")
```

nomisRev (commented):
I think removing the override for `raise` is incorrect, but besides that this looks good to me!

nomisRev:
> And to be honest, I don't fully understand we cannot have the Atomic<Any?> be part of IorRaise as we have in the current implementation.

Was this changed?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
kyay10: I have the final combine call happening inside the `ior` builder itself. This is so that try catch can recover from the error, and also so that any races in calling `raise` is only won by the call that actually reaches the builder, which is the behaviour of the rest of the builders.
That test passes (after adding a missing space) so no issues there.

removing the raise override is also good because it makes transitioning to context receivers much easier.

kyay10:
> Was this changed?

@nomisRev in a previous commit, I had removed the `State` and `combineError` in `IorRaise` in favour of a simpler `addError` method because that's all that `IorRaise` seems to need access to. However, that was a binary-breaking change. I implemented a version of that idea in arrow-context if you'd like to see it. [Here's](https://github.com/arrow-kt/arrow-context/blob/initial-prototype/src/main/kotlin/arrow/context/raise/Builders.kt#L85) the ior builder, and [here's](https://github.com/arrow-kt/arrow-context/blob/initial-prototype/src/main/kotlin/arrow/context/raise/Builders.kt#L309) the extension methods that `IorRaise` previously had (note the decomposition into `Raise` and `Accumulate`).

serras (approved):

PR #3340, opened 2023-12-28
Add attempt builder as an imperative alternative to `recover`

kyay10:
I couldn't add this in `main` because it conflicted with a deprecated method, and so it wasn't useable inside of other `Raise` instances. Perhaps if there's a better name than `attempt`, then this change can be backported.

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
@@ -256,17 +256,17 @@ public inline fun <Error, A, B, C, D, E, F, G, H, I, J> Raise<Error>.zipOrAccumu
 ): J {
   contract { callsInPlace(block, AT_MOST_ONCE) }
   var error: Any? = EmptyValue
-  val a = recover({ action1(RaiseAccumulate(this)) }) { error = combine(error, it.reduce(combine), combine); EmptyValue }
-  val b = recover({ action2(RaiseAccumulate(this)) }) { error = combine(error, it.reduce(combine), combine); EmptyValue }
-  val c = recover({ action3(RaiseAccumulate(this)) }) { error = combine(error, it.reduce(combine), combine); EmptyValue }
-  val d = recover({ action4(RaiseAccumulate(this)) }) { error = combine(error, it.reduce(combine), combine); EmptyValue }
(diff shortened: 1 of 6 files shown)


serras (approved):
Interesting addition!

nomisRev:
I am unsure about this addition. My immediate thoughts are that _early returning_ is considered a code-smell by many, and I am not sure what additional patterns this enables?

If we want to easily allow transforming the error, which is what I see mostly here perhaps we need a variant of `recover` that re-raises the error? So an alias for `recover({ foo.bind() }) { msg -> msg.toList() }`. WDYT?

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -487,19 +487,33 @@ public inline fun <Error, A, B, C, D, E, F, G, H, I, J> Raise<NonEmptyList<Error
...
+  block: RaiseAccumulate<Error>.() -> A,
+  addErrors: (NonEmptyList<Error>) -> Unit
+): Any? {
+  contract {
+    callsInPlace(block, AT_MOST_ONCE)
+    callsInPlace(addErrors, AT_MOST_ONCE)
+  }
+  val errorList = attempt { return block(RaiseAccumulate(this)) }
+  addErrors(errorList)
+  return EmptyValue
```
nomisRev, line 514: Sorry, here also really not fan of the early return 😅
I had to read this snippets too many times to understand what's going on :/

nomisRev:
Okay, thank you for clarifying. 

> Perhaps it is too much a matter of style that only individual users should decide on introducing it themselves, though, but I'll leave that decision up to you.

Well, it's up to all of us together 😉 I added arrow 2.0 tag, so lets this sink in a bit while we focus on 1.2.2.

serras:
My 2 cents here is that `attempt` could be helpful for some style of programming. However, I don't think we should rewrite the rest of the code to use `attempt`, but rather keep that isolated. I also feel the code using early returns becomes harder to understand.

serras (approved):
I think in this form this can be included in Arrow; maybe not the style we want to promote, but useful in some cases

kyay10:
I keep running into issues with this in personal projects, specifically KT-40124, so I'll turn it into a draft for now. We can revisit once that bug is fully fixed (partial mitigation has happened in 2.3, but I still ran into it in more complicated code)

PR #3349, opened 2024-01-17
Make a common public parent for raise exception

nomisRev:


kyay10:
What type of code is this PR intended to allow? merely distinguishing Raise errors from other cancellations, or is it something more? I ask because `raised` and `raise` are both internal, and I'm thinking that potentially exposing them could be useful for debugging such cases or logging them. Alternatively, we might want to add a better `toString` for the exception to include the `raised` and `raise`.

nomisRev:
Yes, the only goal was merely distinguishing Raise errors from other cancellations as we've discussed in your CCE PR.

>  I ask because raised and raise are both internal, and I'm thinking that potentially exposing them could be useful for debugging such cases or logging them. Alternatively, we might want to add a better toString for the exception to include the raised and raise.

I didn't want to prematurely add it, since there are safer and better ways to access those values. Such as `traced`, `recover`, `attempt` if its get added, etc.

It'll be easy to expose it later on, but hard/impossible to remove.

kyay10:
I like this change because it finally allows the usage intended by the `try-catch` tests. 
Should we update the tests to reflect that? I think now they're catching general `Throwable`, so perhaps we can change it to `RaiseCancellationException`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
```diff
@@ -253,18 +252,19 @@ internal class DefaultRaise(@PublishedApi internal val isTraced: Boolean) : Rais
...
-    isActive.value -> throw if (isTraced) RaiseCancellationException(r, this) else RaiseCancellationExceptionNoTrace(r, this)
+    isActive.value -> throw if (isTraced) Traced(r, this) else NoTrace(r, this)
     else -> throw RaiseLeakedException()
   }
 }
 
-/** CancellationException is required to cancel coroutines when raising from within them. */
-private class RaiseCancellationExceptionNoTrace(val raised: Any?, val raise: Raise<Any?>) :
-  CancellationExceptionNoTrace()
+public sealed class RaiseCancellationException(
```
kyay10, line 260: A tiny warning within the doc comment would likely be helpful. Something along the lines of "Note that catching this exception will swallow a raised error with no way of differentiating which `Raise` it came from"

nomisRev:
@kyay10 I added your suggestions.

The message of the exception always includes:

```kotlin
internal const val RaiseCancellationExceptionCaptured: String =
  "kotlin.coroutines.cancellation.CancellationException should never get swallowed. Always re-throw it if captured." +
    "This swallows the exception of Arrow's Raise, and leads to unexpected behavior." +
    "When working with Arrow prefer Either.catch or arrow.core.raise.catch to automatically rethrow CancellationException."
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
```diff
@@ -253,18 +252,32 @@ internal class DefaultRaise(@PublishedApi internal val isTraced: Boolean) : Rais
...
-private class RaiseCancellationExceptionNoTrace(val raised: Any?, val raise: Raise<Any?>) :
-  CancellationExceptionNoTrace()
+@MustBeDocumented
+@Retention(AnnotationRetention.BINARY)
+@RequiresOptIn(RaiseCancellationExceptionCaptured, RequiresOptIn.Level.WARNING)
+public annotation class DelicateArrowApi
 
-private class RaiseCancellationException(val raised: Any?, val raise: Raise<Any?>) : CancellationException()
+/**
+ * [RaiseCancellationException] is an _delicate_ api, and should be used with care.
```
kyay10, line 266:
```suggestion
 * [RaiseCancellationException] is a _delicate_ api, and should be used with care.
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
```diff
@@ -253,18 +252,32 @@ internal class DefaultRaise(@PublishedApi internal val isTraced: Boolean) : Rais
...
   }
 }
 
-/** CancellationException is required to cancel coroutines when raising from within them. */
-private class RaiseCancellationExceptionNoTrace(val raised: Any?, val raise: Raise<Any?>) :
-  CancellationExceptionNoTrace()
+@MustBeDocumented
+@Retention(AnnotationRetention.BINARY)
+@RequiresOptIn(RaiseCancellationExceptionCaptured, RequiresOptIn.Level.WARNING)
+public annotation class DelicateArrowApi
```
serras, line 263:
```suggestion
public annotation class DelicateRaiseApi
```

It might be the case that we have several "delicate Arrow APIs" in the future, I think it's better to be explicit about this being the `Raise` one.

arrow-libs/core/arrow-core/src/jvmMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
```diff
@@ -6,7 +6,11 @@ import kotlin.coroutines.cancellation.CancellationException
  * Inspired by KotlinX Coroutines:
  * https://github.com/Kotlin/kotlinx.coroutines/blob/3788889ddfd2bcfedbff1bbca10ee56039e024a2/kotlinx-coroutines-core/jvm/src/Exceptions.kt#L29
  */
-internal actual open class CancellationExceptionNoTrace : CancellationException(RaiseCancellationExceptionCaptured) {
+@Suppress(
+  "EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING",
+  "SEALED_INHERITOR_IN_DIFFERENT_MODULE"
+)
+internal actual class NoTrace actual constructor(raised: Any?, raise: Raise<Any?>) : RaiseCancellationException(raised, raise) {
```
serras, line 13: I am wondering whether we really need to support Android <= 6.0. Otherwise we can simplify this in a common module, instead of requiring platform-dependent implementations.

serras (approved):
Nice!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
nomisRev: Sounds good, I was in doubt between the two 😁 We have 1 instance of this already in Resource for `allocated`, but there I've re-used KotlinX's DelicateCoroutineApi which is already used in KotlinX for resources that need to be manually closed.

arrow-libs/core/arrow-core/src/jvmMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
nomisRev: We need this to be in `jvmMain` because `fillInStacktrace` is JVM-specific.

IIRC, the `stackTrace = emptyArray()` is needed in addition to disabling stacktrace creation for Android. Otherwise Android 6.0, and below will crash if the `stackTrace` is not `emptyArray` but was not created. (This is stolen from KotlinX, https://github.com/Kotlin/kotlinx.coroutines/blob/master/kotlinx-coroutines-core/jvm/src/Exceptions.kt#L43).

So removing that will still not allow us to disable stack traces in common.

kyay10 (approved):
LGTM!

PR #3350, opened 2024-01-17
Add Raise extension functions

NWuensche:
Add Extension functions to make usage of raise inside `either` block easier.

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Raise.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Raise.kt
@@ -312,6 +312,12 @@ public interface Raise<in Error> {
     is Either.Right -> value
   }
 
+  @RaiseDSL
+  public fun  Error.raise(): Nothing = raise(this)
+
+  @RaiseDSL
+  public fun  Either.Left<Error>.raise(): Nothing = this.value.raise()
+
   /**
    * Extracts all the values in the [Map], raising every [Either.Left]
    * as a _logical failure_. In other words, executed [bind] over every


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Raise.kt
```diff
@@ -312,6 +312,12 @@ public interface Raise<in Error> {
     is Either.Right -> value
   }
 
+  @RaiseDSL
+  public fun  Error.raise(): Nothing = raise(this)
+
+  @RaiseDSL
+  public fun  Either.Left<Error>.raise(): Nothing = this.value.raise()
```
serras, line 319: This one is just `bind`, but with a restricted type.

serras (commented):
I don't think this is a good idea, because it pollutes the member scope of the `Error` type. Furthermore, I don't see a big difference between `raise("blah")` and `"blah".raise()`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Raise.kt
```diff
@@ -312,6 +312,12 @@ public interface Raise<in Error> {
     is Either.Right -> value
   }
 
+  @RaiseDSL
+  public fun  Error.raise(): Nothing = raise(this)
```
kyay10, line 316: Note that the desired chaining can be achieved by using `myError.let(::raise)`

NWuensche:
> I don't think this is a good idea, because it pollutes the member scope of the `Error` type. Furthermore, I don't see a big difference between `raise("blah")` and `"blah".raise()`

Thank you for your reply. However, why does this Argument does not hold in the same way for e.g. `Any?.left()` in `Either` ? There, one could make the same argument and say that

public fun <A> A.left(): Either<A, Nothing> = Left(this)

does pollute the Either-Type and one could simply write  `this.let(::Left)`

serras:
@NWuensche there's no super-right answer. I think that the `.left` and `.right` were added mostly because `Either.Left` is a bit less discoverable, but I'm not 100% sure about that.

One difference, though, is that `.left` must be explicitly imported, whereas your `.raise` would be always in scope whenever you are "inside" of a `Raise`.

serras:
I'm going to close this, since there seems to be agreement in the team to not include this. I would suggest opening a discussion, though, so more members of the community can voice their opinions (either here or in the Slack).

nomisRev:
>  there's no super-right answer. I think that the .left and .right were added mostly because Either.Left is a bit less discoverable, but I'm not 100% sure about that.

`.left` behaves differently from the `Either.Left` constructor in terms of inference, so `this.let(::Left)` is not a valid replacement for `Either.Left`. (function vs constructor behaves differently, maybe @serras can shed more light on how this works internally for inference, perhaps this has also been fixed and this is a legacy thing by now. This has been lingering since the previous inference algorithm IIRC.).

Additionally `.left` has been a kind-of legacy method. If I were to design it again, I would followed `Result.success` and `Result.failure` from Kotlin Std.

It'll break too much code if we remove it though, and adding `Either.left` alongside it might make this more confusing so we decided to leave it as is.

PR #3352, opened 2024-01-18
Deprecate Iterable#unzip in favor of stdlib method

terminalnode:
From what I can tell the other unzip methods have no equivalent in the stdlib.

NonEmptyList is of course covered by the Iterable#unzip, but still has a use-case because of returning a pair of NonEmptyLists rather than a pair of regular Lists.

This is Iterable#unzip in kotlin 1.9.22:
```kt
public fun <T, R> Iterable<Pair<T, R>>.unzip(): Pair<List<T>, List<R>> {
    val expectedSize = collectionSizeOrDefault(10)
    val listT = ArrayList<T>(expectedSize)
    val listR = ArrayList<R>(expectedSize)
    for (pair in this) {
        listT.add(pair.first)
        listR.add(pair.second)
    }
    return listT to listR
}
```

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
@@ -21,6 +21,7 @@ import arrow.typeclasses.SemigroupDeprecation
 import arrow.typeclasses.combine
 import kotlin.experimental.ExperimentalTypeInference
 import kotlin.jvm.JvmName
+import kotlin.collections.unzip as stdlibUnzip
 
 public inline fun <B, C, D, E> Iterable<B>.zip(
   c: Iterable<C>,
@@ -849,9 +850,7 @@ public fun <A> Iterable<A>.salign(
   ReplaceWith("unzip()", "kotlin.collections.unzip")
 )
 public fun <A, B> Iterable<Pair<A, B>>.unzip(): Pair<List<A>, List<B>> =
-  fold(emptyList<A>() to emptyList()) { (l, r), x ->
-    l + x.first to r + x.second
-  }
+   stdlibUnzip()
 
 /**
  * after applying the given function unzip the resulting structure into its elements.
@@ -876,7 +875,7 @@ public fun <A, B> Iterable<Pair<A, B>>.unzip(): Pair<List<A>, List<B>> =
   ReplaceWith("map(fc).unzip()", "kotlin.collections.unzip")
 )
 public inline fun <A, B, C> Iterable<C>.unzip(fc: (C) -> Pair<A, B>): Pair<List<A>, List<B>> =
-  map(fc).unzip()
+  map(fc).stdlibUnzip()
 
 /**
  * splits a union into its component parts.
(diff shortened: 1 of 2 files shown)


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -867,6 +871,10 @@ public fun <A, B> Iterable<Pair<A, B>>.unzip(): Pair<List<A>, List<B>> =
  * <!--- KNIT example-iterable-12.kt -->
  * <!--- TEST lines.isEmpty() -->
  */
+@Deprecated(
+  "Unzip is being deprecated in favor of the standard library version.\n$NicheAPI",
+  ReplaceWith("unzip()", "kotlin.collections.unzip")
```
kyay10, line 876:
```suggestion
  ReplaceWith("map(fc).unzip()", "kotlin.collections.unzip")
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -844,6 +844,10 @@ public fun <A> Iterable<A>.salign(
  * <!--- KNIT example-iterable-11.kt -->
  * <!--- TEST lines.isEmpty() -->
  */
+@Deprecated(
+  "Unzip is being deprecated in favor of the standard library version.\n$NicheAPI",
+  ReplaceWith("unzip()", "kotlin.collections.unzip")
+)
 public fun <A, B> Iterable<Pair<A, B>>.unzip(): Pair<List<A>, List<B>> =
   fold(emptyList<A>() to emptyList()) { (l, r), x ->
```
kyay10, line 852: This should just delegate to stdlib `unzip`

kyay10:
> NonEmptyList is of course covered by the Iterable#unzip, but still has a use-case because of returning a pair of NonEmptyLists rather than a pair of regular Lists.

Perhaps the implementation should use stdlib's unzip and then call `toNonEmptyListOrNull()!!`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
terminalnode: Fixed.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
kyay10: Yes, that's exactly what I intended! I think an import alias is just the right approach here. I wonder how the `ReplaceWith` will function because of the name discrepancy though. `kotlin.collections.*` is a default import, and so perhaps we could just deprecate this with `HIDDEN` level. I'd rather wait though on another reviewer's opinion as to whether `HIDDEN` is the right approach here or not.

kyay10:
The PR checks will take care of trying the tests, don't worry!

nomisRev:
> I can't run all the tests locally for some reason (getting some random failures even with the main branch)

Might be KMP setup, like @kyay10 said, not important since CI verifies everything but if you run into this again in the future [KDoctor](https://github.com/Kotlin/kdoctor) is a great tool to help you setup your KMP development environment.

nomisRev (approved):
LGTM. Thank you @terminalnode! 🙌

PR #3356, opened 2024-01-23
Add AutoClose module

nomisRev:
Adds a new module with same support as Resource, but without requiring KotlinX or suspension.

I plan on splitting `Resource` from Arrow Fx, but we either do it here or we do it in 2.0 branch since there are some breaking changes in the Resource API. WDYT? I'm open for all ideas 😅

Some goals are to be able to use this in other libraries, and projects without having to introduce Arrow as a base dependency. 

For example `SuspendApp` could include just `AutoCloseable` and `Resource` support but it doesn't need anything else by doing this we can provide a much more elegant `SuspendApp` API, whilst keeping everything working together nicely, etc. I've had the use for this pattern quite a lot since I started adopting this DSL approach for resources.


nomisRev:
> According to our plan, this is the last non-bugfix release of the 1.x series.

From the release post, if this change is too disruptive we can also just put everything directly into 2.x.

cc\\ @serras

kyay10:
Do we need to maintain source compatibility with the pre-existing impl? Or are we free to break some signatures?

serras:
I’m a bit worried about having both `AutoCloseScope` and `Resource`, because it creates the question of what to use in each moment.

This is very similar to what I faced in `arrow-collectors`, where both suspended and non-suspended collectors were needed. For me, the best option here would be to somehow unify both concepts, although maybe we can only do that at the level of API.

About splitting `Resource`, I’m not really sure that’s a good idea. The Coroutines library is already small, since we’ve split Resilience from it. My fear is also we end up with too many libraries (as I’ve seen happening in the Haskell world) and then people find it annoying to add too many of them.

settings.gradle.kts
```diff
@@ -67,6 +67,9 @@ project(":arrow-core-serialization").projectDir = file("arrow-libs/core/arrow-co
 include("arrow-cache4k")
 project(":arrow-cache4k").projectDir = file("arrow-libs/core/arrow-cache4k")
 
+include("arrow-autocloseable")
```
serras, line 70: Nitpick: I think it's better to call the library `arrow-autoclose`, since that's the name of the scope

serras:
I've thought about it a bit more (bus rides give a lot of time for thinking), and I think we can easily explain why you need both concepts. I've noticed that `ResourceScope` extends `AutoCloseScope`, which I had missed in the previous review, and that actually sounds great.

I would go for including this in 1.2.2, and update the documentation about _Resource management_ to talk about both at the same time.

nomisRev:
> I’m a bit worried about having both AutoCloseScope and Resource, because it creates the question of what to use in each moment.

This IMO is quite simple, but I guess there is an essential piece of documentation missing.

- `AutoCloseable`, originally a Java concept, has been added to Kotlin in 1.9.x. `AutoCloseScope` is a DSL for registering `non-suspend` finalisers, or composeable `finally`. Which is capable of allowing safely unwrapping `AutoCloseable`.
- `ResourceScope` does practically the same but for `suspend` finalisers **and** cancellation.

Similar to how non-suspend function can be called from suspend function, can `AutoCloseScope` be called from `ResourceScope` but not the other way around.

Not sure if that clears it up, but I find this a reasonable and necessary split. In pure land, we would always use `Resource` but that heavily conflicts with what Java does, and thus most of the eco-system. However, I still find `Resource` necessary because in a lot of cases we need to deal with `CancellationException` in Kotlin when working with KotlinX Coroutines, etc.

> About splitting Resource, I’m not really sure that’s a good idea. The Coroutines library is already small, since we’ve split Resilience from it. My fear is also we end up with too many libraries (as I’ve seen happening in the Haskell world) and then people find it annoying to add too many of them.

It's not completely explained in my original PR comment, but my idea is not to split but rather separate `Resource` into a separate module but still export it through `arrow-fx-coroutines`. I think that covers your concern, or not?

I have a personal use-case just for `Resource` in `kotlin-kafka`, and I really do not want to rely on Arrow & Arrow Fx for a library like that.

nomisRev:
@serras saw your message to later 😁 We definitely need to include it into the docs, I will raise a PR tomorrow or the day after 👍

> @nomisRev Do we need to maintain source compatibility with the pre-existing impl? Or are we free to break some signatures?

I was planning to take a look at 2.0 first, to see what the options are. So if you have any ideas, please just jump to 2.0 and feel free to make some breaking changes there if it results in a nicer API 😉

serras:
> Do we need to maintain source compatibility with the pre-existing impl? Or are we free to break some signatures?

I would prefer that, as much as possible, `Resource` and `AutoClose` have the same API, except for `suspend` in some signatures. That would incredibly simplify the explanation of both concepts.

arrow-libs/core/arrow-autocloseable/build.gradle.kts
```diff
@@ -0,0 +1,61 @@
...
+  }
+}
+
+apply(from = property("ANIMALSNIFFER_MPP"))
+
+kotlin {
+  sourceSets {
+    commonMain {
+      dependencies {
+        implementation(libs.kotlin.stdlibCommon)
```
kyay10, line 25: nit: Isn't stdlib brought in by default

arrow-libs/core/arrow-autocloseable/src/commonMain/kotlin/arrow/AutoCloseScope.kt
```diff
@@ -0,0 +1,100 @@
...
+ * <!--- KNIT example-autocloseable-01.kt -->
+ */
+public inline fun <A> autoCloseScope(block: AutoCloseScope.() -> A): A {
+  val scope = DefaultAutoCloseScope()
+  return try {
+    block(scope).also {
+      scope.close(null)
+    }
+  } catch (e: Throwable) {
+    scope.close(e) ?: throw e
```
kyay10, line 38: A `nonFatalOrThrow()` seems apt here
```suggestion
    scope.close(e.nonFatalOrThrow()) ?: throw e
```

arrow-libs/core/arrow-autocloseable/src/commonTest/kotlin/arrow/AutoCloseTest.kt
```diff
@@ -0,0 +1,135 @@
...
+        val r = autoClose({ res }) { r, e ->
+          promise.complete(e)
+          r.shutdown()
+          throw error2
+        }
+        autoClose({ Resource() }) { _, _ -> throw error3 }
+        wasActive.complete(r.isActive())
+        throw error
+      }
+    } shouldBe error.apply {
```
kyay10, line 74: Isn't `error` already mutated inside `close`, and thus it already has those suppressed exceptions? Maybe checking `error.suppressedExceptions` could be better?

arrow-libs/core/arrow-autocloseable/src/commonTest/kotlin/arrow/AutoCloseTest.kt
```diff
@@ -0,0 +1,135 @@
...
+    shouldThrow<RuntimeException> {
+      autoCloseScope {
+        autoClose({ Resource() }) { r, e ->
+          promise.complete(e)
+          r.shutdown()
+          throw error2
+        }
+        autoClose<Int>({ throw error }) { _, _ -> }
+      }
+    } shouldBe error.apply { addSuppressed(error2) }
```
kyay10, line 99: Same as above, error already has those suppressed exceptions, so it feels strange to be adding them again. In fact, this test would succeed even if `autoCloseScope` *didn't* add suppressed exceptions, since we're always mutating `error` to have them anyways.

arrow-libs/core/arrow-autocloseable/src/commonTest/kotlin/arrow/AutoCloseTest.kt
```diff
@@ -0,0 +1,135 @@
...
+    res.isActive() shouldBe false
+  }
+
+  @OptIn(ExperimentalStdlibApi::class)
+  private class Resource : AutoCloseable {
+    private val isActive = AtomicBoolean(true)
+
+    fun isActive(): Boolean = isActive.get()
+
+    fun shutdown() {
```
kyay10, line 125: nit: same as `close()`, so should just be inlined

arrow-libs/core/arrow-autocloseable/src/jvmMain/kotlin/arrow/AutoCloseableExtensions.kt
```diff
@@ -0,0 +1,7 @@
+package arrow
+
+public fun <A : AutoCloseable> AutoCloseScope.install(autoCloseable: A): A =
```
kyay10, line 3: I'm struggling to see when this would ever get called vs the member method. Are there some weird shenanigans with how `AutoCloseable` works in common vs jvm?

kyay10 (commented):
Not very familiar with the `resource` dsl so I couldn't help much there.

arrow-libs/core/arrow-autocloseable/src/commonMain/kotlin/arrow/AutoCloseScope.kt
```diff
@@ -0,0 +1,100 @@
...
+ * <!--- KNIT example-autocloseable-02.kt -->
+ */
+public interface AutoCloseScope {
+  public fun <A> autoClose(
+    acquire: () -> A,
+    release: (A, Throwable?) -> Unit
+  ): A
+
+  @ExperimentalStdlibApi
+  public fun <A : AutoCloseable> install(autoCloseable: A): A =
```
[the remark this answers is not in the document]
nomisRev: I'm curious about your rationale.

Afaik: Micro optimisations between invokeinterface vs invokestatic on JVM, and overridable vs final.

arrow-libs/core/arrow-autocloseable/src/jvmMain/kotlin/arrow/AutoCloseableExtensions.kt
nomisRev: Oh wait. Good point. When AutoCloseable was first introduced it was not a `typealias` on `jvmMain`, so they were not actually the same type but I'm not sure if that's still the case. They were planning to fix this.

arrow-libs/core/arrow-autocloseable/build.gradle.kts
nomisRev: I don't think so, I also had to add `jvmMain` and `jsMain` to depend on `jvm` and `js` stdlib. I think there is still something broken in our Gradle setup..

In general I think it's not needed, but seems to be needed in our setup.

arrow-libs/core/arrow-autocloseable/src/jvmMain/kotlin/arrow/AutoCloseableExtensions.kt
nomisRev: Okay, so this is currently still needed. https://youtrack.jetbrains.com/issue/KT-55777/Unresolved-kotlin.AutoCloseable-in-JVM

But should removed in Kotlin 2.0

arrow-libs/core/arrow-autocloseable/src/commonTest/kotlin/arrow/AutoCloseTest.kt
nomisRev: I left them separate on purpose, to distinct between `AutoCloseable` and non-autocloseable resource.

arrow-libs/core/arrow-autocloseable/build.gradle.kts
serras: This weirdness is fixed in the 2.0 branch, there you only need to depend on the common stdlib.

Issue #2829, opened 2022-09-27
["Request"] Generate Lens for `sealed class` `abstract val`


Kernald:
What version are you currently using?
1.1.3

What would you like to see?
As per https://github.com/arrow-kt/arrow-optics/issues/5 which seems to have been lost when things were consolidated in this repo, with a sealed class of the following shape:

```
sealed class Example {
  abstract val name: String

  data class One(override val name: String, val int: Int) : Example()
  data class Two(override val name: String, val long: Long) : Example()
  data class Three(override val name: String, val double: Double) : Example()
}
```

I would love seeing a generated `val name: Lens<Example, String>`, which would look like this:

```
val name: Lens<Example, String> = Lens(
  get = { it.name },
  set = { example, name ->
    when (example) {
      is One -> example.copy(name = name)
      is Two -> example.copy(name = name)
      is Three -> example.copy(name = name)
    }
  },
)
```

This is a bit boilerplate-y, and needs updating whenever a sub-type is added. It seems like a good candidate for code generation.

serras:
Fixed by #3359, will be released on 1.2.2.

PR #3359, opened 2024-01-24
Implement sealed class Lenses

maksym-moroz:
Addresses #2829 
Implement and enable Lens generation for sealed class properties when these properties are abstract and without extension receiver, subclasses are data classes that override properties in the constructor. 
If that's not the case emit an info log message explaining why Lens is not being generated.
Additional cleanup to target logic.

All tests pass


serras:
@maksym-moroz can you run the `spotlessApply` task and re-commit?

maksym-moroz:
@serras I did just that and it didn't show any issues with build

serras:
The automatic Spotless commit was not working, I've done it manually. This should (hopefully) turn the build green.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-8.5-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.5-all.zip
```
serras, line 3: What is the reasoning for this change?

arrow-libs/optics/arrow-optics-ksp-plugin/src/test/kotlin/arrow/optics/plugin/LensTests.kt
```diff
@@ -130,4 +130,23 @@ class LensTests {
       |}
       """.compilationSucceeds()
   }
+
+  @Test
+  fun `Lens for sealed class property`() {
```
serras, line 135: I am missing three other cases for test:
1. More than one class extending the sealed one,
2. Data classes defined outside of the sealed class,
3. Sealed class with zero inheritors.

serras (changes requested):
Thanks for this great work. Not only implementing the feature, but also cleaning the code a bit on the go :)

I've left a few comments about things (mostly tests) that I think should be in the code before merging it in `main`.

PR #3360, opened 2024-01-25
Added high arity for arrow-core.api functions

yesitskev:
- Added higher arity for some functions in `arrow-core`.
  - NonEmptyLis.zip
  - Sequence.zip
  - Map.zip
  - Iterable.zip
  - memoize() + MemoizeKeys
  - Raise<NonEmptyList<Error>>.zipOrAccumulate

Certain private interfaces and objects had to be made public to accommodate these additions:
- Iterable<T>.collectionSizeOrDefault(default: Int)
- interface MemoizedCall<in F, out R>


kyay10:
I seem to recall that there's some consensus among the team to include these changes in another library instead of core.

serras:
We actually made the decision long time ago to "cut" these APIs (andd Tuple) to 10 parameters.

nomisRev:
@serras this introduces a new module. The decision was made for binary size reasons rather than maintenance. Maintenance wise this code is not very expensive.

Since we regularly get such request, I personally feel okay with having a seperate opt-in module for those that want higher arities.

serras:
Oh, I didn't realize that! Then, of course, this is more than welcome :)

My only suggestion would be _not_ to expose the private functions from `arrow-core`, since they are really implementation details. Maybe the memoization support is not so important? Since also in `arrow-2` some of those functions are now in `arrow-functions` instead...

serras:
@yesitskev is it OK if I go forward and remove the `memoize` higher-arity versions, so that we don't have to expose the internals? Anyway those APIs have been superseded by `MemoizedDeepRecursiveFunction`, so they are not so relevant anymore.

yesitskev:
@serras that's perfect. Feel free to prune whichever is necessary 👍

serras:
I've removed the memoization part, which required us exposing the internals of `memoize`. Once the build is green, I'll proceed to merge into `arrow-2` and update the docs.

PR #3364, opened 2024-01-30
Update to gradle/actions/setup-gradle@v3

serras:
According to the [release notes](https://github.com/gradle/gradle-build-action), the `gradle-build-action` is now deprecated and must be replaced with `setup-gradle` + call `./gradlew`. This PR does exactly that.

Caching has been enabled only for the `main` and `arrow-2` branches, since the docs recommends doing so only for long-lived ones.


nomisRev:
Thank you @serras. I'm going to need to do this on a lot of repos 🤕

nomisRev (approved):

PR #3367, opened 2024-02-02
Add forEachAccumulating and refactor mapOrAccumulate

kyay10:
First discussed on [Slack](https://kotlinlang.slack.com/archives/C5UPMM0A0/p1705878406791589).

This seems to be a useful primitive. I don't know whether exposing the `iterator` methods is a good idea or not. One can always use `asSequence`, but with extra wrapping.
I also added some missing `mapOrAccumulate` overloads that are refactored from other files.


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -502,6 +502,70 @@ public inline fun <Error, A, B, C, D, E, F, G, H, I, J> Raise<NonEmptyList<Error
...
+  var error = firstErrors.reduce(combine)
+  for (item in iterator) {
+    recover({
+      block(RaiseAccumulate(this), item)
+    }) {
+      error = combine(error, it.reduce(combine))
+    }
+  }
+  raise(error)
+}
```
serras, line 538: I find this code quite hard to understand. I would prefer to have some code in the style of the previous `mapOrAccumulate`

```suggestion
): Unit {
  var error: Any? = EmptyValue
  for (item in iterable) {
    fold<NonEmptyList<Error>, B, Unit>(
      { transform(RaiseAccumulate(this), item) },
      { errors -> error = combine(error, errors.reduce(combine), combine) },
      { }
    )
  }
  return if (error === EmptyValue) results else raise(unbox<Error>(error))
}
```

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -502,6 +502,70 @@ public inline fun <Error, A, B, C, D, E, F, G, H, I, J> Raise<NonEmptyList<Error
...
+    block(RaiseAccumulate(this), item)
+  }
+}) { firstError ->
+  buildList {
+    addAll(firstError)
+    for (item in iterator) {
+      recover({ block(RaiseAccumulate(this), item) }) { addAll(it) }
+    }
+  }.toNonEmptyListOrNull()!!.let(::raise)
+}
```
serras, line 567: Same as above, I would prefer to keep the simpler version without so many nested `recover`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -538,17 +593,39 @@ public inline fun <Error, A, B> Raise<Error>.mapOrAccumulate(
...
+ * and how to use it in [validation](https://arrow-kt.io/learn/typed-errors/validation/).
+ */
+@RaiseDSL
+public inline fun <Error, A, B> Raise<Error>.mapOrAccumulate(
+  sequence: Sequence<A>,
+  combine: (Error, Error) -> Error,
+  @BuilderInference transform: RaiseAccumulate<Error>.(A) -> B
+): List<B> = buildList {
+  forEachAccumulating(sequence, combine) { add(transform(it)) }
+}
```
serras, line 614: Note that this is not what one would expect from `Sequence.map`, it should rather return another `Sequence` (but it should be as easy as doing `buildSequence` instead)

serras (changes requested):
I like the idea of this API, especially because then `mapOrAccumulate` becomes simply `buildList { forEachOrAccumulate { ... } }`. However, I think that the implementation is quite too convoluted compared to the previous one (I've left a suggestion on how to fix it).

serras:
I think that we should follow the convention from the rest of the functions, and call this `forEachOrAccumulate` instead (in line with `zipOrAccumulate` and `mapOrAccumulate`). WDYT @kyay10 @nomisRev ?

serras:
> One can always use asSequence, but with extra wrapping.

Note that it's important to keep the semantics of `Sequence`, which is lazy evaluation of the elements. It's not the same to do `sequence.toList().map(f).asSequence()` that `sequence.map(f)`; the latter won't start doing any work until the sequence is inspected, whereas the former will.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
kyay10: I wanted to get rid of `EmptyValue` because it added too much complexity to the code. Also, I had hoped to provide some way to access whether or not an error has happened yet, so that `mapOrAccumulate` can stop storing successful values when it no longer needs to. Finally, I liked the fact that for the success case, which should probably be the most likely case, the code only does one allocation for the first `recover`, while the old code would allocate one everytime. It's a premature optimization yes, but I figured the code isn't too horrible. WDYT? I'll revert it back for now because this is orthogonal to the PR

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
kyay10: There's sadly no way to do `mapOrAccumulate` while returning a `Sequence` because to accumulate the errors we must do it eagerly. There's already precedent for this in the codebase with `Sequence.mapOrAccumulate` returning `Either<Error, List<A>>`. I think returning `List` emphasizes that this is a terminal operation. Crazy idea: `RaisingSequence<Error, A>` whose operations need a `Raise<Error>`.

kyay10:
wrt `forEachOrAccumulate`: `orXXX` is usually used in the stdlib as some failing strategy with some normal return value. `forEachAccumulating` sounds more `verb`-like and hence imperative, which fits well with `forEach`. However, I see that consistency might be desireable here, so I'm fine with either name.

kyay10:
> It's not the same to do sequence.toList().map(f).asSequence() that sequence.map(f); the latter won't start doing any work until the sequence is inspected, whereas the former will.

What I meant here was exposing `forEachAccumulating(iterator: Iterator<A>)`. I left it exposed because it might be useful, but we might want to make it `internal` and instead users can use the sequence overload by wrapping their iterator in a sequence.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
kyay10: It looks nicer with `attempt` btw :). I had my original implementation with that. It worked nicely because the error path is more complicated than the non-error one

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
serras: You are completely correct, I stand corrected

kyay10:
Quick note: I've used `buildXXX` extensively here, but I think there was a concern before that we should just use mutable collections to allow users to mutate if they *really* want to rather cheaply. Not sure If the take on that has changed, but just figured I should mention it

PR #3373, opened 2024-02-09
added retryRaise and retryEither functions

akotynski:
Hey, with the latest API, it is easy to write Either version of the retry function. I think it could be quite useful. 
I based the implementation on 
`public suspend fun <A> Schedule<Throwable, *>.retry(action: suspend () -> A): A`

### DIFF
--- a/arrow-libs/resilience/arrow-resilience/api/arrow-resilience.api
+++ b/arrow-libs/resilience/arrow-resilience/api/arrow-resilience.api
@@ -252,8 +252,11 @@ public final class arrow/resilience/Schedule$Decision$Done : arrow/resilience/Sc
 
 public final class arrow/resilience/ScheduleKt {
 	public static final fun retry-4AuOtiA (Lkotlin/jvm/functions/Function2;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public static final fun retry-YL6hcnA (Larrow/core/raise/Raise;Lkotlin/jvm/functions/Function2;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public static final fun retryEither-4AuOtiA (Lkotlin/jvm/functions/Function2;Lkotlin/jvm/functions/Function0;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static final fun retryOrElse-aZo8_V4 (Lkotlin/jvm/functions/Function2;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function3;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static final fun retryOrElseEither-aZo8_V4 (Lkotlin/jvm/functions/Function2;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function3;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
+	public static final fun retryRaise-4AuOtiA (Lkotlin/jvm/functions/Function2;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 }
 
 public final class arrow/resilience/common/Platform : java/lang/Enum {
--- a/arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/Schedule.kt
+++ b/arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/Schedule.kt
@@ -1,3 +1,5 @@
+@file:OptIn(ExperimentalTypeInference::class)
+
 package arrow.resilience
 
 import arrow.core.Either
@@ -7,6 +9,9 @@ import arrow.core.identity
 import arrow.core.left
 import arrow.core.merge
 import arrow.core.nonFatalOrThrow
+import arrow.core.raise.Raise
+import arrow.core.raise.either
+import arrow.core.raise.fold
 import arrow.core.right
 import arrow.core.some
 import arrow.resilience.Schedule.Companion.identity
@@ -17,6 +22,7 @@ import kotlinx.coroutines.currentCoroutineContext
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.flow.retry
+import kotlin.experimental.ExperimentalTypeInference
 import kotlin.jvm.JvmInline
 import kotlin.math.pow
 import kotlin.random.Random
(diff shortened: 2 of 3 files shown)


kyay10:
I think it might be preferable if this API used `Raise` directly instead. Could you perhaps make that change? It also seems like the `step` is only evaluated on `Left`s, so perhaps in a `Raise` version it should be evaluated on the error type only

arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/ScheduleEither.kt
```diff
@@ -0,0 +1,30 @@
...
+import kotlin.time.Duration
+
+public suspend fun <Error, Result, Output> Schedule<Error, Output>.retry(
+  action: suspend Raise<Error>.() -> Result,
+): Either<Error, Result> {
+  var step: ScheduleStep<Error, Output> = step
+
+  while (true) {
+    currentCoroutineContext().ensureActive()
+    when (val result = either { action() }) {
```
kyay10, line 18: using `recover` here is probably nicer and won't result in unnecessary `Either` wrapping

arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/ScheduleEither.kt
```diff
@@ -0,0 +1,30 @@
...
+import arrow.resilience.Schedule
+import arrow.resilience.ScheduleStep
+import kotlinx.coroutines.currentCoroutineContext
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.ensureActive
+import kotlin.time.Duration
+
+public suspend fun <Error, Result, Output> Schedule<Error, Output>.retry(
+  action: suspend Raise<Error>.() -> Result,
+): Either<Error, Result> {
```
nomisRev, line 13: I think this only works atm, because this is in a different package. See the test, it imports `import retry`.
That's not really desirable at all, the problem is going to be if this lands in `arrow.resilience` there won't be a way for the compiler to distinct between `retry` for `Throwable` and non-Throwable.

Perhaps by forcefully passing a generic parameter, but that's really not desired. So we'll probably have to rename `retry` to `retryRaise` but we can keep the one for `Either`, since that will overload correctly.

I am personally fine having both signatures, since I discussed with @serras some time ago about supporting `Either` in a similar way than `Raise`. That's why we also have `Either.zip` that works directly with `Either` values. That could also be an argument to keep traverse for `Either` though... 🤔 cc\\ @serras

Nonetheless, great work @akotynski. This was desperately missed for quite some time!

serras: Indeed, I am up for exposing both a Raise and and Either version, if possible.

About naming, we need to be conscious of our compatibility promise, so we should keep the existing function as is, and add new retryXX for the new ones.

arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/ScheduleEither.kt
serras: Another possibility is using fold, which gives a similar code pattern as matching on the result of the Raise.

akotynski: I'm not sure how to use recover here 🤔 The recover function would need to return `Result`.

either function is a shortcut for fold so it is possible to do something like this:
```kt
    fold(
      action,
      recover = {
        when (val decision = step(it)) {
          is Schedule.Decision.Continue -> {
            if (decision.delay != Duration.ZERO) delay(decision.delay)
            step = decision.step
          }

          is Schedule.Decision.Done -> return it.left()
        }
      },
      transform = { return it.right() },
    )
```
IMO it is less readable but anyway if we want to keep two signatures (retryRaise and retryEither) it will be simpler to do just:
```kt
public suspend inline fun <Error, Result, Output> Schedule<Error, Output>.retryRaise(
  @BuilderInference action: Raise<Error>.() -> Result,
): Either<Error, Result> = retryEither { either(action) }
```

arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/ScheduleEither.kt
akotynski: thanks! I renamed function and kept both signatures:
```kt
public suspend inline fun <Error, Result, Output> Schedule<Error, Output>.retryRaise(
  @BuilderInference action: Raise<Error>.() -> Result,
): Either<Error, Result>
```
```kt
public suspend inline fun <Error, Result, Output> Schedule<Error, Output>.retryEither(
  @BuilderInference action: () -> Either<Error, Result>,
): Either<Error, Result>
```

serras: I would suggest adding another one working completely within `Raise`:

```kotlin
public suspend inline fun <Error, Result, Output> Raise<Error<>.retryRaise(
  schedule: Schedule<Error, Output>,
  @BuilderInference action: Raise<Error>.() -> Result,
): Either<Error, Result>
```

so that then we would get:

```kotlin
public suspend inline fun <Error, Result, Output> Schedule<Error, Output>.retryRaise(
  @BuilderInference action: Raise<Error>.() -> Result,
): Either<Error, Result> = either { retyRaise(this@retryRaise) }

public suspend inline fun <Error, Result, Output> Schedule<Error, Output>.retryEither(
  @BuilderInference action: () -> Either<Error, Result>,
): Either<Error, Result> = this.retryRaise { action().bind() }
```

akotynski: ahh, yes, with having Raise in the context we do not need to return Either, we can just raise error :) 
I was just fixed on the extension function on Schedule class.
So the base signature will be:
```kt
public suspend inline fun <Error, Result, Output> Raise<Error>.retry(
  schedule: Schedule<Error, Output>,
  @BuilderInference action: Raise<Error>.() -> Result,
): Result
```

serras (changes requested):
It looks great to me, but I would like to ask you to add everything to the `Schedule.kt` file instead. The reason is that this way the functions go into the `ScheduleKt` class, which makes our life a bit easier when checking binary compatibility aftewards.

Thanks for your contribution, @akotynski !

arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/ScheduleEither.kt
akotynski: so at the end I used fold :)

akotynski:
sure! Moved code to Schedule.kt file 

Thank you all for your quick reply and support :)

kyay10 (approved):
Beautiful implementation!!

PR #3374, opened 2024-02-10
Stop storing transformed element when error is encountered in `mapOrAccumulate`

kyay10:
Seemed like an easy win :) If an error is encountered early-on, there's no reason why we should keep storing transformed elements (we just waste memory). Micro-benchmarking maybe is needed to figure out if this actually helps with anything. I think knowing inside `forEachAccumulating` whether there has been an error or not is likely useful outside of this anyway. Do note that 1.22 is getting finalised, so perhaps the ship has sailed for this change (it does change the API of the currently-unreleased `forEachAccumulating`)


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -506,25 +506,25 @@ public inline fun <Error, A, B, C, D, E, F, G, H, I, J> Raise<NonEmptyList<Error
 public inline fun <Error, A> Raise<Error>.forEachAccumulating(
   iterable: Iterable<A>,
   combine: (Error, Error) -> Error,
-  @BuilderInference block: RaiseAccumulate<Error>.(A) -> Unit
+  @BuilderInference block: RaiseAccumulate<Error>.(hasError: Boolean, A) -> Unit
```
nomisRev, line 509: Can we do this in a way it doesn't leak in the public implementation 😢 Leaking such low-level optimisations into the public API is a big red flag for me. I personally prefer the memory hit :/

kyay10: I figured having that accessible could be useful for callers. However, yes there's a way to do it by just wrapping every call in a `try {} catch(RaiseCancellationException)` and rethrowing it. Would that be preferable to you?

serras: As a general suggestion, I would suggest having extreme caution with `inline` methods. It's not that we are just leaking our implementation, it that it's literally _copied_ in their binary, which puts as in a extremely delicate place for compatibility.

serras: Furthermore, I agree with @nomisRev in that this exposes too much of the implementation. We shouldn't go into the rabbit hole of adding additional information to functions like this one (why only if it has an error? why not the previous error, or anything else?)

serras:
Although I'm not fond of the particular implementation, this seems like an interesting idea. Would it be possible instead to split the work of `mapOrAccumulate` into two parts? First we run usual `map` until we find an error, and if we do, then we run with the rest of the iterable using `forEachAccumulating`, but not saving any values.

In addition, @nomisRev do you remember why we make all of these functions `inline`? Doing so for `raise` seems worth it, but these functions seem big enough that `inline`ing may be worth in terms of binary size.

nomisRev:
> In addition, @nomisRev do you remember why we make all of these functions inline?

We normally make things `inline` to support `suspend` in the lambda, since now `Raise` is meant for both non-suspend and suspend.

Making things `inline` for other reasons to reduce the binary would be great! 👍

kyay10:
Closed in favour of #3376

PR #3376, opened 2024-02-16
Avoid storing more transformed elements than necessary

serras:
This is a counter-proposal for #3374 @kyay10 

This introduces a new (internal) version of `forEachAccumulating` which works differently before and after the error. That way we can simply stop `add`ing things after the first error.


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -709,7 +770,7 @@ public open class RaiseAccumulate<Error>(
 
   @RaiseDSL
   @JvmName("_mapOrAccumulate")
-  public inline fun <A, B> mapOrAccumulate(
+  public fun <A, B> mapOrAccumulate(
```
nomisRev, line 773: Without `inline` we don't support `suspend` here, and thus break "traverse" for suspend.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -547,10 +559,26 @@ public inline fun <Error, A> Raise<NonEmptyList<Error>>.forEachAccumulating(
...
+
+/**
+ * Allows to change what to do once the first error is raised.
+ * Used to provide more performant [mapOrAccumulate].
+ */
+@PublishedApi @JvmName("forEachAccumulatingDouble")
+internal inline fun <Error, A> Raise<NonEmptyList<Error>>.forEachAccumulating(
+  iterator: Iterator<A>,
+  @BuilderInference blockUntilError: RaiseAccumulate<Error>.(A) -> Unit,
+  @BuilderInference blockAfterError: RaiseAccumulate<Error>.(A) -> Unit
```
nomisRev, line 572: **Nit:** Do we really need two lambdas here? 🤔

Or was that for `JvmName`? Otherwise I am also fine using `forEachAccumulatingImpl` or something. I've started doing that in my own projects, similar how to KotlinX Coroutines does this with `produceImpl`. `flattenMergeImpl`, etc.

nomisRev (commented):
I like this approach @serras.

I think this would make the PR #3374 unnecessary, since the optimisation is applied in the base implementation. 👍

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
serras: I don't know why I've removed it, but it's now back

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
kyay10: Why not have:
```
  @BuilderInference block: RaiseAccumulate<Error>.(A) -> B,
  add: (A, B) -> Unit
```
and simply we only call `add` if we don't have any errors.
We could also just have a `Boolean` parameter instead, with similar effects.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -521,10 +520,23 @@ public inline fun <Error, A> Raise<Error>.forEachAccumulating(
...
+
+@PublishedApi
+internal inline fun <Error, A> Raise<Error>.forEachAccumulatingImpl(
+  iterator: Iterator<A>,
+  combine: (Error, Error) -> Error,
+  @BuilderInference blockUntilError: RaiseAccumulate<Error>.(A) -> Unit,
+  @BuilderInference blockAfterError: RaiseAccumulate<Error>.(A) -> Unit
 ) {
   var error: Any? = EmptyValue
+  var errorArose = false
```
kyay10, line 533: Nit: Can't this be worked out from `error == EmptyValue`? Not sure that either way is going to be faster than the other or anything

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -547,10 +559,26 @@ public inline fun <Error, A> Raise<NonEmptyList<Error>>.forEachAccumulating(
...
+ * Used to provide more performant [mapOrAccumulate].
+ */
+@PublishedApi @JvmName("forEachAccumulatingDouble")
+internal inline fun <Error, A> Raise<NonEmptyList<Error>>.forEachAccumulatingImpl(
+  iterator: Iterator<A>,
+  @BuilderInference blockUntilError: RaiseAccumulate<Error>.(A) -> Unit,
+  @BuilderInference blockAfterError: RaiseAccumulate<Error>.(A) -> Unit
 ) {
   val error: MutableList<Error> = mutableListOf()
+  var errorArose = false
```
kyay10, line 575: Nit: same as above

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -547,10 +559,26 @@ public inline fun <Error, A> Raise<NonEmptyList<Error>>.forEachAccumulating(
 public inline fun <Error, A> Raise<NonEmptyList<Error>>.forEachAccumulating(
   iterator: Iterator<A>,
   @BuilderInference block: RaiseAccumulate<Error>.(A) -> Unit
+): Unit = forEachAccumulatingImpl(iterator, block, block)
+
+/**
+ * Allows to change what to do once the first error is raised.
+ * Used to provide more performant [mapOrAccumulate].
+ */
+@PublishedApi @JvmName("forEachAccumulatingDouble")
```
kyay10, line 568: Nit: I think the JvmName is not needed.

kyay10 (changes requested):
I'm concerned about the unnecessary duplication of the lambdas of `forEachAccumulating` and `mapOrAccumulate`.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
serras: Removed!

serras:
Indeed, @kyay10's comment about size explosion was on point. I've reworked the implementation to use a version with a single function, with a Boolean with errors. (Note however that this function is only exposed for `inline` purposes, not for other users of the library)

kyay10 (approved):
Great work on this!!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -521,10 +521,19 @@ public inline fun <Error, A> Raise<Error>.forEachAccumulating(
   iterator: Iterator<A>,
   combine: (Error, Error) -> Error,
   @BuilderInference block: RaiseAccumulate<Error>.(A) -> Unit
+): Unit = forEachAccumulatingImpl(iterator, combine) { item, _ -> block(item) }
+
+@PublishedApi @JvmSynthetic
```
kyay10, line 526: Synthetic is a nice touch here! Note that this technically doesn't 100% absolve us of ABI compatibility issues according to the discussion on https://github.com/Kotlin/binary-compatibility-validator/issues/165. I wish we had access to `@InlineOnly`, then maybe BCV would consider those methods to not be part of our ABI, but alas.

PR #3389, opened 2024-03-02
NonEmptySet: change 2nd param of constructor to `Iterable`. Optimize`nonEmptySetOf` and `toNonEmptySetOrNull`

hoc081098:
- Change the second param of constructor to `Iterable` (instead of `Set`).
- Avoid unnecessary copying in `nonEmptySetOf` and `toNonEmptySetOrNull`.


serras:
I am not really sure what is the optimization here. How are you ensuring that elements are not duplicated when using `nonEmptySetOf`?

Furthermore, note that we are soon going to make `arrow-2` the new main branch. There `NonEmptySet` is defined differently, using an inline value class. It would be better if optimizations would be applied there, as we don't expect to make any further 1.2.x release.

hoc081098:
> I am not really sure what is the optimization here. How are you ensuring that elements are not duplicated when using `nonEmptySetOf`?

In constructor, we have used `setOf(first) + rest`, it returns a Set.

hoc081098:
Prev code: `res (varargs === Array) -> toSet -> setOf(first) + res` uses 2 loops 
After code : `res (varargs === Array) -> asList -> setOf(first) + res` uses 1 loops 
🙏

.idea/vcs.xml
```diff
@@ -20,4 +22,4 @@
   <component name="VcsDirectoryMappings">
     <mapping directory="" vcs="Git" />
   </component>
-</project>
+</project>
```
hoc081098, line 25:
```suggestion
</project>

```

.idea/vcs.xml
```diff
@@ -3,6 +3,8 @@
   <component name="CommitMessageInspectionProfile">
     <profile version="1.0">
       <inspection_tool class="BodyLimit" enabled="true" level="WARNING" enabled_by_default="true" />
+      <inspection_tool class="CommitFormat" enabled="true" level="WARNING" enabled_by_default="true" />
+      <inspection_tool class="CommitNamingConvention" enabled="true" level="WARNING" enabled_by_default="true" />
```
hoc081098, line 7:
```suggestion
```

hoc081098:
> * Prev code: `res (varargs === Array) -> toSet -> setOf(first) + res` uses 2 loops internally.
> * After code : `res (varargs === Array) -> asList -> setOf(first) + res` uses 1 loops internally (because asList does not copy the array to List, but returns a read-only wrapper)
>   🙏


I have targeted to `arrow-2` branch. @serras Please review it again 🙏

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
```diff
@@ -43,20 +44,26 @@ public value class NonEmptySet<out A> private constructor(
...
 
 public fun <A> nonEmptySetOf(first: A, vararg rest: A): NonEmptySet<A> =
-  NonEmptySet(first, rest.toSet())
+  NonEmptySet(first, rest.asList())
 
 public fun <A> Iterable<A>.toNonEmptySetOrNull(): NonEmptySet<A>? {
   val iter = iterator()
   if (!iter.hasNext()) return null
-  return NonEmptySet(iter.next(), Iterable { iter }.toSet())
+  return NonEmptySet.unsafeOf(iter)
```
serras, line 66: I don't think there's a need to have a `unsafeOf` if it's only used once, inline, and internal.

```suggestion
  return NonEmptySet(iter)
```

serras (changes requested):
These generalizations look quite useful. Thanks for the contribution!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptySet.kt
hoc081098: Thanks, I have just changed it to `return NonEmptySet(iter.next(), Iterable { iter })` 🙏

serras:
@hoc081098 sorry for the unnecesary noise but, would it be possible for you to rebase these changes into the new `main`? Otherwise it's impossible for us to merge properly this.

serras (approved):

PR #3390, opened 2024-03-02
optimize NonEmptyList

hoc081098:
- update the constructor:
  - store `all` instead of `head` + `tail` -> `head` and `tail` are simply derivatives of `all`
  - change the second param of ctor to `Iterable`
- replace `tail` with `all` to simplify the implementation of some functions, and also improve the perf (avoid unnecessary collection copying)


serras:
As mentioned in #3389, we are not going to release a new Arrow 1.2.x. In the `arrow-2` branch we already represent the `NonEmptyList` as a regular list, so I guess most of those optimization do not really apply.

serras:
This is very similar to the optimization we already have in `arrow-2`, so I think it's not worth pursuing it since we are going to migrate to that branch soon. WDYT @nomisRev?

PR #3394, opened 2024-03-10
Detect Raise leaks at compile-time

### LINKED ISSUE #3391: Fold restriction not allowing any Function<*> to be returned is too strict
Hi there!
Coming from Arrow 1.2.1, I was quite surprised to get some exceptions in my unit tests like this:

    java.lang.IllegalStateException: Returning a lazy computation or closure from 'fold' breaks the context scope, and may lead to 
    leaked exceptions on later execution.
    Make sure all calls to 'raise' and 'bind' occur within the lifecycle of nullable { }, either { } or similar builders.

In my case, my code creates a "Validator", which happens to be a just a function signature.
I do not leak anything from within, I simply need error detection when creating my independent validator.

I think I understand what you want to prevent, but simply disallowing return values to be a `Function<*>` is too much.
I will now rework my interface to not be a functional interface, but a normal one - just to circumvent this error. 

This worked fine with 1.2.1:

```
import arrow.core.Either
import arrow.core.getOrElse
import arrow.core.raise.Raise
import arrow.core.raise.either
import arrow.core.raise.ensure

typealias NumValidator = (Int) -> Either<String, Int>

context(Raise<String>)
fun createLessThanNumValidator(lt: Int): NumValidator {
    ensure(lt > 0) { "lt must be a positive number" }
    return { i ->
        either {
            ensure(i < lt) { "$i is not less than $lt" }
            i
        }
    }
}

fun main() {
    val lessThanThree = either { createLessThanNumValidator(3) }.getOrElse { error(it) }

    for (i in 1..10) {
        println(lessThanThree(i))
    }
}
```

The creation of a validator could have a logical error if the given threshold is not greater than 0. If it is, you get a proper validator that validates any int agains the threshold.
There is nothing wrong with the code, nothing can and will ever leak. It is just a double-stacked validation.

Outputs:
```
Either.Right(1)
Either.Right(2)
Either.Left(3 is not less than 3)
Either.Left(4 is not less than 3)
Either.Left(5 is not less than 3)
Either.Left(6 is not less than 3)
Either.Left(7 is not less than 3)
Either.Left(8 is not less than 3)
Either.Left(9 is not less than 3)
Either.Left(10 is not less than 3)
```

Going from Arrow 1.2.1 to 1.2.3 makes this code throw the error on the first line of main - in a *very* bad way, as the error is a runtime error and not a compile time error. I would strongly advise not to create runtime errors...

kyay10:
Fixes #3391

Note that `@OverloadResolutionByLambdaReturnType` is currently very limited. So limited, in fact, that it doesn't support calls with multiple lambdas, and it doesn't support calls where the lambdas have type parameter usages (other than the type parameter being the return type). Also, one has to explicitly provide a type argument of `Nothing` now when a lambda would return such, but I think that's such a niche use case that it won't show up anywhere other than our test code.
I had been pulling my hair out trying to get `@OverloadResolutionByLambdaReturnType` to work for this case, so fair warning for any of you that wish to explore this further. I think this change shouldn't break any code that previously worked (other than code of the form `nullable { throw Exception()`)

### DIFF
--- a/arrow-libs/core/arrow-core/api/arrow-core.api
+++ b/arrow-libs/core/arrow-core/api/arrow-core.api
@@ -3591,6 +3591,7 @@ public abstract interface annotation class arrow/core/raise/RaiseDSL : java/lang
 }
 
 public final class arrow/core/raise/RaiseKt {
+	public static final field PotentialRaiseLeak Ljava/lang/String;
 	public static final fun _fold (Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;)Ljava/lang/Object;
 	public static final fun _foldOrThrow (Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;)Ljava/lang/Object;
 	public static final fun _foldUnsafe (Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;)Ljava/lang/Object;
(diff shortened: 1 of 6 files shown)


serras:
I think this approach is great, @kyay10, using the built-in mechanism to highlight the errors to the developer. I have some concerns because of the use of `OverloadByLambdaReturnType`, but maybe that feature works well enough for our use case.

kyay10:
@serras Admittedly, I am scared of it as well. While tinkering I have seen it fail in any complicated-enough situation. Seemingly, for the basic declarations in this PR, everything works fine enough, but I would sleep better at night if we had maybe more tests for wild and strange usages of `nullable` et al.

nomisRev:
>  but I would sleep better at night if we had maybe more tests for wild and strange usages of nullable et al.

Yes, I can tell you you will sleep better at night after that 😂 😅

I am curios if this solves the problem seen by @Zordid, @kageru, @wfhartford.

Some more context here:
https://github.com/arrow-kt/arrow/issues/3391#issuecomment-1993744364

serras:
According to the other issue, it seems that there are valid use cases for returning a function from within a Raise context. This means that neither check (runtime or compile time) is valid, so I think we should just drop both and go back to the 1.2.1 situation.

@kyay10 @nomisRev WDYT?

nomisRev:
Agreed @serras! 

> compile time

The only possibility is probably implementing a compiler plugin, but not sure if you can truly reach 100% correctness that way. It's a problem inherit to DSLs,

PR #3395, opened 2024-03-11
Set up Klib binary compatibility validation

serras:
This is a new feature which is coming (so maybe we should wait until it's in the release). However, as multiplatform is becoming more important, I think we should also check the compatibility in those platforms.


nomisRev:
@serras I labeled this as 2.0.0. Are you okay merging/rebasing `main` into this branch if we merged arrow-2 into main?

serras:
Closed in favor of #3403

PR #3398, opened 2024-03-21
Improvements to STM + do not expose internal members

serras:
I was doing a bit of ABI cleanup, and I figured out a way to remove some duplication in the STM code, and increase the type safety (we no longer hold `Array<Any?>` but `Array<A>` in the HAMT tree).


nomisRev (approved):
Very nice work @serras 👏 👏 👏
Cleaning up the ABI is tedious... but super awesome!! ❤️

PR #3409, opened 2024-04-24
Add accumulate function to IorRaise

sampengilly:
Used to combine errors into the IorRaise state and trigger a Both output in a more convenient way. Previously the only mechanism to do this was to call `bind()` on an `Ior.Both`.

### DIFF
--- a/arrow-libs/core/arrow-core/api/arrow-core.api
+++ b/arrow-libs/core/arrow-core/api/arrow-core.api
@@ -895,6 +895,7 @@ public abstract interface annotation class arrow/core/raise/ExperimentalTraceApi
 
 public final class arrow/core/raise/IorRaise : arrow/core/raise/Raise {
 	public fun <init> (Lkotlin/jvm/functions/Function2;Ljava/util/concurrent/atomic/AtomicReference;Larrow/core/raise/Raise;)V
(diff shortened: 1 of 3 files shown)


nomisRev (approved):
Nice work @sampengilly! Looks great to me 👏 👏

sampengilly:
Rebased onto updated main branch

nomisRev:
I think you need to rerun `./gradlew apiDump`. Sadly the action to do it automatically doesn't work on forks.

sampengilly:
Thankfully there weren't many conflicts, though I see the upstream change was a big one (and was surprised to see the test framework change too).

Re-running now, for some reason I thought that manually merging it would be enough

nomisRev:
> (and was surprised to see the test framework change too).

We only changed the runner, sadly. Kotest relies on a compiler plugin, but we wanted to turn Arrow into a testing ground for new targets, and dev/EAP releases. Thanks to changing the framework runner we were able to build against dev, eap, etc version of K2 ☺️

serras:
@sampengilly would it be possible for you to rebase this work on the new `main`? This would help us a lot in getting this merged. Thanks in advance!

sampengilly:
Done :)

PR #3414, opened 2024-04-30
Allow specifying a more concrete Throwable type for retrying

nomisRev:
Small use-case I came across several times, and I stumbled across it again. This PR allows specifying concrete types for retrying. I.e.

```kotlin
Schedule.exponential<PSQLException>(...)
  .and(Schedule.recurs(5))
  .doWhile { e: PSQLExeption, _ -> e.isRetryable() }
  .retry {
    exposedTransaction { doDatabaseStuff() }
  }
```

TODO
 - [ ] Write test cases
 - [x] Broke something, investigate

### DIFF
--- a/arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/Schedule.kt
+++ b/arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/Schedule.kt
@@ -1,8 +1,10 @@
 @file:OptIn(ExperimentalTypeInference::class)
+@file:Suppress("REDUNDANT_INLINE_SUSPEND_FUNCTION_TYPE")
 
 package arrow.resilience
 
 import arrow.core.Either
+import arrow.core.NonFatal
 import arrow.core.None
 import arrow.core.Option
 import arrow.core.identity
@@ -37,8 +39,8 @@ public typealias ScheduleStep<Input, Output> =
  * A [Schedule] describes how a `suspend fun` should [retry] or [repeat].
  *
  * It's defined by a [step] function that takes an [Input] and returns a [Decision],
-* the [Decision] determines if the `suspend fun` should be [Continue] to be retried or repeated
-* (and if so, the `delay` until the next attempt),
+ * the [Decision] determines if the `suspend fun` should be [Continue] to be retried or repeated
+ * (and if so, the `delay` until the next attempt),
  * or if the [Schedule] is [Done] retrying or repeating.
  */
 @JvmInline
@@ -384,6 +386,7 @@ public value class Schedule<in Input, out Output>(public val step: ScheduleStep<
           transform(this.output, other.output),
           combineDuration(this.delay, other.delay)
         ) { this.step(it).and(other.step(it), transform, combineDuration) }
+
       else -> Done(transform(this.output, other.output))
     }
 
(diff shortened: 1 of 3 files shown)


nomisRev:
> Platform declaration clash: The following declarations have the same JVM signature

Of course... Naming is so annoying 😫

arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/Schedule.kt
```diff
@@ -431,21 +455,37 @@ public suspend fun <Input, Output> Schedule<Throwable, Output>.retryOrElse(
...
+  retryOrElseEither<Input, Output, A, Throwable>(action, orElse)
+
+/**
+ * Retries [action] only when [E] has occurred, otherwise it immediately retries the other exception.
+ * If the [Schedule] is exhausted,
+ * it will invoke [orElse] with the last [E] and the output of the [Schedule] to produce a fallback value of [A].
+ * Returns [Either] with the fallback value if the [Schedule] is exhausted, or the successful result of [action].
+ */
+@JvmName("retryOrElseEitherReified")
+public suspend inline fun <Input, Output, A, reified E : Throwable> Schedule<E, Output>.retryOrElseEither(
```
serras, line 468: We should strive to keep the `inline` functions to be just small wrappers over the real functions (otherwise every change in the implementation breaks binary compatibility). My suggestion is to have a version which takes a `KClass<E>` as additional parameter (so you don't need the `reified` anymore), and then have a version which simply calls that function with `E::class`.

arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/Schedule.kt
```diff
@@ -412,6 +417,14 @@ public value class Schedule<in Input, out Output>(public val step: ScheduleStep<
 public suspend fun <A> Schedule<Throwable, *>.retry(action: suspend () -> A): A =
   retryOrElse(action) { e, _ -> throw e }
 
+/**
+ * Retries [action] only when [E] has occurred, otherwise it immediately retries the other exception.
+ * It will throw the last [E] if the [Schedule] is exhausted, and ignores the output of the [Schedule].
+ */
+@JvmName("retryReified")
+public suspend inline fun <A, reified E : Throwable> Schedule<E, *>.retry(action: suspend () -> A): A =
```
serras, line 425: Nitpick: I think the reverse order of type arguments would look nicer on calls.

```kotlin
retry<ConnectionError, _> {
  // do something
}
```

arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/Schedule.kt
```diff
@@ -412,6 +417,14 @@ public value class Schedule<in Input, out Output>(public val step: ScheduleStep<
 public suspend fun <A> Schedule<Throwable, *>.retry(action: suspend () -> A): A =
```
serras, line 417: Are you sure that the compiler is able to infer which version of `retry` you are calling? I'm a bit concerned about problems in resolving which one you want. Maybe we can change the other one to `retryInstanceOf`? This would follow the naming conventions in Kotlin (`filterInstanceOf`...)

nomisRev: We should have tests, for this, but let's double check. I like to cover these API concerns in the tests as well, so we can also track it never breaks by accident.

serras:
Should we close this now that #3418 is merged? I don't want to step on your toes...

nomisRev:
Not stepping on my toes @serras 😘 Love collaborating together, the resulting Schedule is 😍

PR #3417, opened 2024-05-13
Alias for List for Every

nomisRev:
We've never added an `every()` aliases because we cannot define multiple (in the same package), but I think I am in favor of adding one for `List` now. We can only use one, but List is by far the most used Collection in Kotlin so I think it makes sense adding a "special" alias.

However, I found we can also work around this issue by creating the alias in different packages. The problem is JvmName name, but seems to work fine on resolution in Kotlin. Not sure if this was previously not possible, this was tested last with the old inferrer algorithm... 😅

It seems I can define:

```kotlin
package arrow.optics.dsl.list

public val <T, A> Traversal<T, List<A>>.every: Traversal<T, A>
  get() = this.compose(Every.list())
```

and 

```kotlin
package arrow.optics.dsl.sequence

public val <T, A> Traversal<T, Sequence<A>>.every: Traversal<T, A>
  get() = this.compose(Every.sequence())
```

This seems to completely remove the need to explicitly pass typeclasses instances, since the compiler can resolve the correct package, and thus the instance at the cost of defining them in different packages. WDYT @serras? This can be applied to all type classes in Optics, and users can do the same.


serras:
I am not so sure about the package-as-namespace thing... If you need to use both, then the compiler won't be able to figure it out anymore.

Maybe it's useful to define `every` for list, but also `everyInSet`, `everyInMap` for the other main collection types? I agree that the current way of working with it is by no means simple.

nomisRev:
> I am not so sure about the package-as-namespace thing... If you need to use both, then the compiler won't be able to figure it out anymore.

Huh, I tried this yesterday and it resolved fine. The example I shared with sequence, and list. To be honest I was also surprised, but then it got me thinking where I got stuck last time (couple years ago).

So my understanding was that this should resolve correctly, since it turns into a function `fun every(original: Traversal<S, List<A>>): Traversal<S, A>`, and the compiler can distinct between functions with parameter `Traversal<S, List<A>>` and `Traversal<S, Sequence<A>>`. 

Since if it can resolve in different package, it should also resolve in the same package, so JvmName should solve it in the same was as the package does.

Guess I need to write it all out, and try 😅 Feel free to take over the PR, and take a STAB (🥁) at it. I probably won't pick this up again until after KotlinConf.

PR #3418, opened 2024-05-14
Allow specifying a more concrete Throwable type for retrying, alternative

serras:
This is an alternative approach to #3414, which I think gives us better ergonomics

The main changes are:
- The `retryOrElseEither` function gets a new overload with a `KClass<E>` parameter, which is used to make the instance checks. The `reified` versions call those with `E::class`.
- The approach in #3414 was limited to the case in which you had `Schedule<E, A>` and also a `KClass<E>`. However, this shouldn't be the case, as the check for `E` is actually used to filter out the desired cases.
    - So now the functions are more polymorphic, to <E: Throw, Throw: Throwable> Schedule<Throw, A>.retry(KClass<E>)`
    - The overload resolution still works, I had to make *no* changed in the test files.


arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/Schedule.kt
```diff
@@ -406,46 +409,86 @@ public value class Schedule<in Input, out Output>(public val step: ScheduleStep<
...
+public suspend inline fun <reified E: Throw, Throw: Throwable, A> Schedule<Throw, *>.retry(
+  noinline action: suspend () -> A
+): A = retry(E::class, action)
 
 /**
- * Retries [action] using any [Throwable] that occurred as the input to the [Schedule].
+ * Retries [action] using any [E] that occurred as the input to the [Schedule].
+ * It will throw the last exception if the [Schedule] is exhausted, and ignores the output of the [Schedule].
+ */
+public suspend fun <E: Throw, Throw: Throwable, A> Schedule<Throw, *>.retry(
```
nomisRev, line 423: I am not sure I understand the approach `<E: Throw, Throw: Throwable` used. I have the feeling I am overlooking something obvious 🤔


Besides that this looks great! Wanted to update my PR to use `KCLass<E>` as we discussed on Slack (don't think I pushed yet), but I'm not sure I understand the need for this indirection in the generic definition.

Not sure I understand why this isn't the equivalent of `<E :Throwable`? 🤔

serras: The problem is in the receiver type. Let's consider both "extremes":
- If we make the receiver `Schedule<Throwable, *>`, then we can only call `retry` with that particular schedule type. But we may want to make it possible to also call it with `Schedule<IllegalStateException, *>`, for example.
- If we make the receiver `Schedule<E, *>`, then the filtering is actually useless.

This was we can call the function with any schedule with an input which is a subtype of `Throwable`, and then further filter the cases for `E` which are a subset of those. The variant with `E: Throwable, Throw: Throwable` would also compile, maybe that's the way to go?

nomisRev: Ah, okay got it. So having the indirection inside the generics, allows getting away with a single impl. That's a really awesome trick!!

To be honest, I think `<E: Throwable, Throw: Throwable>` might throw users off in Dokka, or looking at the source code. So I guess I prefer a little duplication, rather than a more complex signature from a user point of view.

This is super nitty though 😅 I like your solution better, but wondering if users would trip of seeing that signature. One of my goals, also with the Schedule refactor was to make Arrows internals simpler, and easy to understand. So I often also choose more explicit code style in Arrow when working towards 1.x.x. It's a tiny detail, but still I feel it gives quite some value.

WDYT? Curious to get lang designers opinion on that 😁

serras:
I think I got it. Actually variants using `E` everywhere is all we need!

```kotlin
fun <E: Throwable, A> Schedule<E, *>.retry(...)
```

This works without the intermediate `E: Throw: Throwable` because of the contravariance of `Schedule`. That is, since `E: Throwable`, then `Schedule<Throwable, ...>: Schedule<E, ...>`, as needed for the function to apply. I think this gives great ergonomics, as shown by the tests:
- If you use `s.retry { ... }`, the type `E` is taken from the type of `s`. So the behavior is that all the exceptions are retried.
- But you can also use `s.retry<IllegalStateException, _> { ... }` if you want a more concrete type to be filtered than the one coming from `s`. Furthermore, we ensure that `IllegalStateException` is indeed a subtype of the input to `s`, as required.

nomisRev:
> But you can also use s.retry<IllegalStateException, _> { ... } if you want a more concrete type to be filtered than the one coming from s. Furthermore, we ensure that IllegalStateException is indeed a subtype of the input to s, as required.

That's really nice ergonomics, also I have really underused `_`. I need to used it more, would also be good to use `_` more in the Arrow docs IMHO. It looks so much better.

nomisRev (approved):
Looks perfect to me! I love the new Schedule 🙌

PR #3419, opened 2024-05-14
Various small STM improvements

serras:
- More idiomatic `TVar<A>.value`
- Introduce `TVar<A>.set` function, so STM can be used with the assignment operator plugin
- Introduce `getValue`/`setValue` to allow using delegation (`var x by newTVar()` should work!)
- Improve the general implementation of `PList` for `TQueue`


arrow-libs/fx/arrow-fx-stm/src/commonMain/kotlin/arrow/fx/stm/STM.kt
```diff
@@ -312,6 +339,78 @@ public interface STM {
...
+   *     x = 20
+   *   }
+   *   //sampleEnd
+   *   println(result)
+   * }
+   * ```
+   * 
+   * <!--- KNIT example-stm-12.kt -->
+   */
+  public operator fun <A> TVar<A>.setValue(thisRef: Nothing?, property: KProperty<*>, value: A) { write(value) }
```
nomisRev, line 412: 🔥

nomisRev (approved):
Looks good to me @serras! Very nice improvements 👏 👏

PR #3420, opened 2024-05-15
Attemp to publish beta.1

nomisRev:
We updated to using Gradle Maven Publish Plugin, but the `publish` command was not doing anything even though the build was green. Nothing showed up in Sonatype, nor in Maven Central.

This update uses the older command, I am also using in other projects. It should send to sonatype, but not to Maven Central (yet). If we can get this PR to publish the artefacts, we can merge the changes to main.

CC\\ @CLOVIS-AI. Calling you in, since you volunteered to work on our Gradle mess 😅 I think this is the right time to do this, and I think we've already come to an agreement on how on Slack. So if you have time, and are still interested in doing this feel free to open a issue to discuss, a PR with proposal, or whatever you think makes sense 😉


nomisRev:
>  I don't think I'll be able to do much in the coming weeks, do you need anything soon?

@CLOVIS-AI, no expectations at all ☺️ Especially not around KotlinConf! Just wanted to get the conversation going here instead of Slack.

.github/workflows/publish.yml
```diff
@@ -40,17 +38,10 @@ jobs:
...
-        uses: actions/upload-artifact@v4
-        with:
-          name: 'reports-${{ matrix.os }}'
-          path: '**/build/reports/**'
+        run: ./gradlew assemble -Pversion=2.0.0-alpha.3 -Penable_wasm=true
 
-      - name: Publish next major SNAPSHOT version
-        run: ./gradlew -Pversion=${{ inputs.version }} publish
+      - name: Publish artifacts
+        run: ./gradlew -Pversion=2.0.0-alpha.3 publish -Penable_wasm=true
```
ILIYANGERMANOV, line 47: I think `--no-configuration-cache` fixes it (https://vanniktech.github.io/gradle-maven-publish-plugin/central/#automatic-release)

PR #3425, opened 2024-05-15
Reduce warnings

### LINKED ISSUE #219: Do a pass on compiler warnings
Run with `./gradlew test`, take a look at the warnings, and see if there's any quick win, shadowed variable, or other quick fixes that could be applied.

serras:
The goal is to make a complete pass on the code, trying to reduce any possible warning that currently arises

Fixes #219


serras:
The warnings are now minimal: just those related to usages of `@Suppress` which may not be forward-compatible. The build files have also been updated to eliminate warnings (except for the needed Dokka one).

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -582,7 +582,7 @@ public sealed class Either<out A, out B> {
    * @param ifRight transform the [Either.Right] type [B] to [C].
    * @return the transformed value [C] by applying [ifLeft] or [ifRight] to [A] or [B] respectively.
    */
-  public inline fun <C> fold(ifLeft: (left: A) -> C, ifRight: (right: B) -> C): C {
+  public inline fun <C> fold(crossinline ifLeft: (left: A) -> C, crossinline ifRight: (right: B) -> C): C {
```
kyay10, line 585: Why crossinline? I think it's fine here to do a non-local return, no? Also, this stops `suspend` from transparently passing through

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -1310,9 +1311,12 @@ public fun <A, B> Either<A, Either<A, B>>.flatten(): Either<A, B> =
  * <!--- KNIT example-either-32.kt -->
  * <!--- TEST lines.isEmpty() -->
  */
-public inline infix fun <A, B> Either<A, B>.getOrElse(default: (A) -> B): B {
+public inline infix fun <A, B> Either<A, B>.getOrElse(crossinline default: (A) -> B): B {
```
kyay10, line 1314: Same as above. I don't think `crossinline` is necessary

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Ior.kt
```diff
@@ -144,7 +144,7 @@ public sealed class Ior<out A, out B> {
    * @param fab the function to apply if this is a [Both]
    * @return the results of applying the function
    */
-  public inline fun <C> fold(fa: (A) -> C, fb: (B) -> C, fab: (A, B) -> C): C {
+  public inline fun <C> fold(crossinline fa: (A) -> C, crossinline fb: (B) -> C, crossinline fab: (A, B) -> C): C {
```
kyay10, line 147: Same here. I think it's useful to allow early-returns here. Is the K2 compiler warning about this somehow?

serras: Sorry for the noise, this was indeed a leftover of my initial attempt to remove some "leaked function" errors. At the end, the solution there was to manually rewrite some functions, because the compiler is not clever enough to see "through" other inline functions.

arrow-libs/fx/arrow-collectors/src/commonMain/kotlin/arrow/collectors/Collect.kt
```diff
@@ -115,22 +114,20 @@ internal suspend fun <A, T, R> Flow<T>.collectI(
...
-      Characteristics.UNORDERED in collector.characteristics ->
-        started.parMapUnordered(concurrency) { collector.accumulate(accumulator, it) }
 
-      else ->
-        started.parMap(concurrency) { collector.accumulate(accumulator, it) }
-    }
+  val started = this.onStart { accumulator = collector.supply() }
 
+  val continued = when {
+    collector.has(Characteristics.CONCURRENT, Characteristics.UNORDERED) ->
```
kyay10, line 121: I'd prefer
```suggestion
    Characteristics.CONCURRENT in collector && Characteristics.UNORDERED in collector ->
```
Where we define an `operator fun CollectorI.contains(x: Characteristics) = x in characteristics`

serras: I am not 100% sure... `collector` itself does not have the characteristics, this is why I decided not to use something like `contains`. But if you strongly prefer that style, I'm happy to change it.

kyay10 (approved):

arrow-libs/optics/arrow-optics/src/commonMain/kotlin/arrow/optics/Prism.kt
```diff
@@ -236,7 +236,7 @@ public interface PPrism<S, T, A, B> : POptional<S, T, A, B> {
      * [Prism] to focus on a particular subtype [B]
      */
     @JvmStatic
-    public inline fun <A: Any, B: A> instanceOf(klass: KClass<B>): Prism<A, B> = Prism(
```
nomisRev, line 239: ;)

nomisRev (approved):
Looks great! I like the simpler code in Either a lot.

PR #3387, opened 2024-03-02
fix(autoclose): close `Closeable`s in reversed order

hoc081098:
Fix #3386

PR #3430, opened 2024-05-17
Fix finalizer order for resourceScope to match autoCloseScope

kyay10:
The test here is the same one added in #3387


nomisRev (approved):
Wait, what, how!? I thought we had tests for this. I guess something weird happened in between 1.x.x and 2.x.x, I guess some regression :/
Thank you for catching it @kyay10, and cleaning up the redundant try/catch.

PR #3431, opened 2024-05-18
AutoClose and Resource rehaul/cleanup

kyay10:
This PR simplifies the implementations of `AutoCloseScope` and `ResourceScope` and allows some of their functions to be `inline`. It also makes `onRelease` not suspending.

### DIFF
--- a/arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/AutoCloseScope.kt
+++ b/arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/AutoCloseScope.kt
@@ -74,6 +74,7 @@ public inline fun <A> autoCloseScope(block: AutoCloseScope.() -> A): A = with(De
 public interface AutoCloseScope {
   public fun onClose(release: (Throwable?) -> Unit)
 
+  @ExperimentalStdlibApi
   public fun <A : AutoCloseable> install(autoCloseable: A): A =
     autoCloseable.also { onClose { autoCloseable.close() } }
 }
--- a/arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/AutoCloseScope.kt
+++ b/arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/AutoCloseScope.kt
@@ -2,7 +2,8 @@ package arrow
 
 import arrow.atomic.Atomic
 import arrow.atomic.update
-import kotlin.coroutines.cancellation.CancellationException
+import arrow.atomic.value
+import kotlin.jvm.JvmInline
 
 /**
  * The [AutoCloseScope] DSL allows for elegantly working with close-ables,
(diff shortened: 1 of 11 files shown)


nomisRev (commented):
@kyay10 amazing work like always ☺️ Very glad you're being critical of all code you come across!! 🙏

Okay, so you made the "single abstract method" installing the `onClose`, or `onRelease` finaliser. That indeed affects the `AcquireStep` in resource, as we've discussed on Slack, but this gives more flexibility.

So both these types, give us the ability to install a non-suspending finaliser, and a suspending finaliser. I think that's a fair DSL, although quite different from where we started. This is great, because in a lot of these cases I've kind-of gotten tunnel vision after so many years. 

I'd like to wait the feedback of others.

arrow-libs/core/arrow-autoclose/src/jvmMain/kotlin/arrow/throwIfFatal.kt
```diff
@@ -1,10 +1,8 @@
 package arrow
 
-import kotlin.coroutines.cancellation.CancellationException
-
 @PublishedApi
 internal actual fun Throwable.throwIfFatal(): Throwable =
   when(this) {
-    is VirtualMachineError, is ThreadDeath, is InterruptedException, is LinkageError, is CancellationException -> throw this
+    is VirtualMachineError, is ThreadDeath, is InterruptedException, is LinkageError -> throw this
```
nomisRev, line 6: Hate that this function cannot be private. I really do not want to expose it 😓

This is why I had the `CancellationException` explicitly in the try/catch, because I rather had a correct definition here, than one specific to AutoCloseable...

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
```diff
@@ -5,21 +5,26 @@ import kotlinx.coroutines.CancellationException
...
-public sealed class ExitCase {
-  public object Completed : ExitCase() {
-    override fun toString(): String =
-      "ExitCase.Completed"
-  }
-
-  public data class Cancelled(val exception: CancellationException) : ExitCase()
-  public data class Failure(val failure: Throwable) : ExitCase()
+public sealed interface ExitCase {
+  public data object Completed : ExitCase
```
nomisRev, line 9: `data object` 🥳

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
```diff
@@ -5,21 +5,26 @@ import kotlinx.coroutines.CancellationException
...
+    }
   }
 }
 
+public val ExitCase.throwableOrNull: Throwable? get() = when (this) {
+  ExitCase.Completed -> null
+  is ExitCase.Cancelled -> exception
+  is ExitCase.Failure -> failure
+}
+
```
nomisRev, line 27: Can we define this inside `ExitCase` so that we don't need an import for it? Preferably as a function I'd say, that'd be more in line with the APIs we now from `Result`.

I also think `exceptionOrNull` is a more common name, than `throwableOrNull`.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -356,21 +343,26 @@ public fun <A> resource(block: suspend ResourceScope.() -> A): Resource<A> = blo
...
-    }
-    throw ee
-  }
-  withContext(NonCancellable) {
-    scope.cancelAll(ExitCase.Completed)?.let { throw it }
-  }
-  return a
+    cancelAll(ExitCase(e))
+    error("Unreachable, cancelAll should throw the exception passed to it. Please report this bug to the Arrow team.")
+  }.also { cancelAll(Completed) }
```
nomisRev, line 352: Can we remove `also` here? I'd prefer regular function invocation after `}` if possible.

I try to avoid scoping function everywhere where they might be redundant. I also see you changed a lot of function bodies, to expressions, and rely on `also` instead. I personally feel this really degrades readability.

This is a common occurrence in this PR, I rather not have these unrelated code changes purely for style reasons. It introduces a lot of code churn for no strong reason. Also takes away the focus of this PR.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -356,21 +343,26 @@ public fun <A> resource(block: suspend ResourceScope.() -> A): Resource<A> = blo
...
+}
+
+/**
+ * Install [A] into the [ResourceScope].
+ * It's [release] function will be called with the appropriate [ExitCase] if this [ResourceScope] finishes.
+ * It results either in [ExitCase.Completed], [ExitCase.Cancelled] or [ExitCase.Failure] depending on the terminal state of [Resource] lambda.
+ */
+@ResourceDSL
+public suspend inline fun <A> ResourceScope.install(
```
nomisRev, line 361: Why are these being defined as extension functions, this breaks source compatibility with 1.x.x. Why not define this in the interface directly?

arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/AutoCloseScope.kt
```diff
@@ -61,53 +62,47 @@ import kotlin.coroutines.cancellation.CancellationException
...
   @ExperimentalStdlibApi
   public fun <A : AutoCloseable> install(autoCloseable: A): A =
-    autoClose({ autoCloseable }) { a, _ -> a.close() }
+    autoCloseable.also { onClose { autoCloseable.close() } }
 }
 
-@PublishedApi
-internal class DefaultAutoCloseScope : AutoCloseScope {
-  private val finalizers = Atomic(emptyList<(Throwable?) -> Unit>())
+public inline fun <A> AutoCloseScope.autoClose(
```
nomisRev, line 82: Idem, this introduces a source breaking change, to support suspend in the `acquire` step. If that is needed, I think users should prefer `onClose`, or `Resource` instead.

So I would prefer to revert this also.

nomisRev (changes requested):
Requesting some changes, or want to discuss some changes:
 - Quite some source compatible breaking changes, which introduces complexity for migration. I would prefer to avoid this, unless we have a strong reason to do so.
 - A lot of unrelated code was changed from bodies to expression, introducing a lot of scoped functions (also). I am personally not a fan of those, especially not of using also within a function body. Can we keep the changes smaller? I 100% support improving readability of operations for users. but I would prefer to keep them separate from actual changes.
 
 
 Thank you again for the great work @kyay10 👏 👏 Excited by these changes!

serras:
Thanks for your work on making the DSL better, @kyay10. I really like everything which removes constraints from our DSLs, so if we can have `suspend` and non-`suspend` functions when before we had only the former, that's a huge win for me!

Having said so, we **promised** to keep source compatibility between 1.2.x and 2.0 (except for the changes in Optics KSP), and some of these changes break that source compatibility. I would **really** like to keep this promise: we've already broken code in the past, and I really want to make things right this time. This means that, at the very least, the same functions need to be exposed from the same package or class they originally were.

Going also in the direction pointed by @nomisRev, maybe this could be easier to review if the fundamental changes were broken from the refactoring changes. In general in Arrow we try to not refactor things like `val` to `let` style, or vice versa, and just keep the original author's code.

kyay10:
Closing in favour of #3518, mostly to provide a clean slate since it's been a while!

PR #3435, opened 2024-05-23
Fix isos and lenses generation in case of parameters having keywords as names.

lukasz-kusek:
Data classes which use parameters that have names which are Kotlin keywords and needs to be escaped fail to compile:
```
data class LensData<T>(
    val `in`: T
) { companion object }
```

This change makes generation of isos an lenses work with such parameters.

There's an additional test added that requires some additional fix in the code which is outside of the scope of this change.

### DIFF
--- a/arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/internals/domain.kt
+++ b/arrow-libs/optics/arrow-optics-ksp-plugin/src/main/kotlin/arrow/optics/plugin/internals/domain.kt
@@ -77,6 +77,10 @@ data class Focus(
   val subclasses: List<String> = emptyList(),
   val classNameWithParameters: String? = className,
 ) {
+  val escapedParamName = paramName.plusIfNotBlank(
+    prefix = "`",
+    postfix = "`",
+  )
   val refinedArguments: List<String>
     get() = refinedType?.arguments?.filter {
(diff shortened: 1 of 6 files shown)


serras:
@lukasz-kusek may I kindly ask to check whether prism generation also suffers from this problem? Since you already know the background, it might be easy to check all cases in one go.

lukasz-kusek:
> @lukasz-kusek may I kindly ask to check whether prism generation also suffers from this problem? Since you already know the background, it might be easy to check all cases in one go.

Sure. I think it doesn't suffer from the exact same problem, but thinking about it there might be similar one. Let me see.

lukasz-kusek:
> > @lukasz-kusek may I kindly ask to check whether prism generation also suffers from this problem? Since you already know the background, it might be easy to check all cases in one go.
> 
> Sure. I think it doesn't suffer from the exact same problem, but thinking about it there might be similar one. Let me see.

As I expected it had that problem, but not for parameters only for subclass name.

lukasz-kusek:
@serras Does this look ok now?

serras:
Very nice, I was just waiting for the CI to kick in and check that everything passes. Merging now.

serras (approved):

lukasz-kusek:
@serras thanks

lkusek-cais:
@serras when's the next release planned for?

lukasz-kusek:
@serras when's the next release planned for?

PR #3438, opened 2024-05-25
AwaitScope

serras:
I am not sure this is useful at all, but this is a small module that automatically `awaitAll`s whenever you call the first `await`, so you get better characteristics out of the box.


arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/await/AwaitScope.kt
```diff
@@ -0,0 +1,72 @@
...
+): CoroutineScope by scope {
+  private val tasks: Atomic<List<Deferred<*>>> = Atomic(emptyList())
+
+  public fun <T> async(
+    context: CoroutineContext = EmptyCoroutineContext,
+    start: CoroutineStart = CoroutineStart.DEFAULT,
+    block: suspend CoroutineScope.() -> T
+  ): Deferred<T> {
+    val deferred = coroutinesAsync(context, start, block)
+    tasks.update { it + deferred }
```
kyay10, line 60: I'd like to also have a `Deferred.install` method or something because people sometimes have `XXAsync` methods that return a `Deferred`

nomisRev:
YES!!!! 😍

nomisRev (approved):

kyay10 (approved):

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/await/AwaitScope.kt
nomisRev: `install`, or more api surface..

kyay10: That's a fair criticism! I'm not sure how prevalent methods that return Deferred are, so perhaps this is okay for now...
A user can also do this manually by doing `awatAll(oneOfTheAwaitDeferreds, otherDeferred)` or by doing `async { otherDeferred.await() }`.

nomisRev: I have a informal feature request, for which I shall request an official one.

@serras Feature Request:
A Kotlin Contract for `awaitAll`, where I can safely call `.completed` or something if you call `awaitAll(deferreds)`.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/await/AwaitScope.kt
```diff
@@ -0,0 +1,72 @@
...
+    return Await(deferred)
+  }
+
+  private inner class Await<T>(
```
nomisRev, line 64: Can this be a `@JvmInline value`? Not sure if that works for `inner` 🤔

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/await/AwaitScope.kt
```diff
@@ -0,0 +1,72 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineStart
+import kotlinx.coroutines.Deferred
+import kotlinx.coroutines.awaitAll
+import kotlinx.coroutines.coroutineScope
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+
+public suspend fun <A> await(
+  block: suspend AwaitScope.() -> A
```
nomisRev, line 15:
```suggestion
public suspend fun <A> awaitAll(
  block: suspend AwaitAllScope.() -> A
```

Would this naming make more sense in relation to the KotlinX Coroutines naming?

serras: Done in the new commit

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/await/AwaitAllScope.kt
```diff
@@ -0,0 +1,72 @@
...
+ *   UserInfo(
+ *     name.await(),   // <- at this point every 'async' is 'await'ed
+ *     avatar.await()  // <- so when you reach this 'await', the value is already there
+ *   )
+ * }
+ *
+ * suspend fun loadUserInfoWithoutAwait(id: UserId): UserInfo {
+ *   val name = async { loadUserFromDb(id) }
+ *   val avatar = async { loadAvatar(id) }
+ *   awaitAll(name, avatar)  // <- this is required otherwise
```
revonateB0T, line 41: Why we need the `awaitAll` to improve concurrency? @serras 
Isn't the two coroutines(loadUser & loadAvatar) immediately dispatched according to https://github.com/Kotlin/kotlinx.coroutines/blob/1bffe67a32d9d0285320f5b23fa94bc2b5f2b92e/kotlinx-coroutines-core/common/src/Builders.common.kt#L70 ?

nomisRev: Hey @revonateB0T,
That is a great question! KotlinX Coroutines makes a distinction between `awaitAll` and `await`, which is shown by the fact that for `Iterable<Deferred<A>>` you need `awaitAll` instead of `map { it.await() }` but for `Iterable<Job>` you do not need `joinAll` but can safely use `map { it.join() }`.

I wrote some additional documentation in [a different PR](https://github.com/arrow-kt/arrow/pull/3443/files#diff-eba615458a6921b379f1477f70993424b278b3de927ece7e998c5bf7dba6f7f0R38), I should split it of and merge it already.

To recap here:

```
 * Okay, great! But doesn't a `coroutineScope { }` do the same? Let's look at some examples!
 * ```kotlin
 * suspend fun loadUserFromDb(id: UserId): String =
 *   awaitCancellation()
 *
 * suspend fun loadAvatar(id: UserId): String =
 *   throw CancellationException("DataSource got closed!")
 * ```
 * Our `loadUserFromDb` function hangs forever, stub this using `awaitCancellation()`.
 * Which means it will only finish, when our scope is cancelled, or any sibling `Job` cancels us.
 * Our `loadAvatar` functions cancels here, due to our `DataSource` getting closed for whatever reason.
 * We stub this by throwing `CancellationException`, but let's see what happens.
 * ```kotlin
 * suspend fun loadUserInfo(id: UserId): UserInfo = coroutineScope {
 *   val name = async { loadUserFromDb(id) }
 *   val avatar = async { loadAvatar(id) }
 *   UserInfo(
 *     name.await(),   // <- Waiting forever, awaitCancellation
 *     avatar.await()  // <- `CancellationException` ignored until `await` is called.
 *   )
 * }
 * ```
 * Our program hangs forever even though `CoroutineScope` saw our `CancellationException`.
 * Strange, if any exception is used besides `CancellationException` it works as expected,
 * if we call `awaitAll(name, avatar)` instead, it also works as expected,
 * if we'd use `launch` it would also work as expected!
 * So, only in this case where we use `async` for parallel operations it hangs forever.
 * This comes with a **big** downside for Arrow Core's Raise,
 * since it also relies on `CancellationException` for signalling early returning, or short-circuiting.
 *
 * However, if we use `awaitAll`, which is the DSL form of `awaitAll(name, avatar)` everything works as desired.
 *
 * ```kotlin
 * suspend fun loadUserInfoWithoutAwait(id: UserId): UserInfo = awaitAll {
 *   val name = async { loadUserFromDb(id) }
 *   val avatar = async { loadAvatar(id) }
 *   return UserInfo(
 *     name.await(), // <-- Awaits both name, and avatar
 *     avatar.await() // <-- CancellationException cancels the awaitAll, and is rethrown.
 *   )
 * }
 * ```
 *
 * **DISCLAIMER:**
 * Since `AwaitAllScope` is not part of KotlinX Coroutines,
 * you need to be careful to call the correct `async` function.
 * If you accidentally call `kotlinx.coroutines.async` instead,
 * the original `coroutineScope { }` behavior is still respected.
 * Luckily this should not pose an issue often, since async is a member of `AwaitAllScope`.
 */ * Okay, great! But doesn't a `coroutineScope { }` do the same? Let's look at some examples!
 * ```kotlin
 * suspend fun loadUserFromDb(id: UserId): String =
 *   awaitCancellation()
 *
 * suspend fun loadAvatar(id: UserId): String =
 *   throw CancellationException("DataSource got closed!")
 * ```
 * Our `loadUserFromDb` function hangs forever, stub this using `awaitCancellation()`.
 * Which means it will only finish, when our scope is cancelled, or any sibling `Job` cancels us.
 * Our `loadAvatar` functions cancels here, due to our `DataSource` getting closed for whatever reason.
 * We stub this by throwing `CancellationException`, but let's see what happens.
 * ```kotlin
 * suspend fun loadUserInfo(id: UserId): UserInfo = coroutineScope {
 *   val name = async { loadUserFromDb(id) }
 *   val avatar = async { loadAvatar(id) }
 *   UserInfo(
 *     name.await(),   // <- Waiting forever, awaitCancellation
 *     avatar.await()  // <- `CancellationException` ignored until `await` is called.
 *   )
 * }
 * ```
 * Our program hangs forever even though `CoroutineScope` saw our `CancellationException`.
 * Strange, if any exception is used besides `CancellationException` it works as expected,
 * if we call `awaitAll(name, avatar)` instead, it also works as expected,
 * if we'd use `launch` it would also work as expected!
 * So, only in this case where we use `async` for parallel operations it hangs forever.
 * This comes with a **big** downside for Arrow Core's Raise,
 * since it also relies on `CancellationException` for signalling early returning, or short-circuiting.
 *
 * However, if we use `awaitAll`, which is the DSL form of `awaitAll(name, avatar)` everything works as desired.
 *
 * ```kotlin
 * suspend fun loadUserInfoWithoutAwait(id: UserId): UserInfo = awaitAll {
 *   val name = async { loadUserFromDb(id) }
 *   val avatar = async { loadAvatar(id) }
 *   return UserInfo(
 *     name.await(), // <-- Awaits both name, and avatar
 *     avatar.await() // <-- CancellationException cancels the awaitAll, and is rethrown.
 *   )
 * }
 * ```
 *
 * **DISCLAIMER:**
 * Since `AwaitAllScope` is not part of KotlinX Coroutines,
 * you need to be careful to call the correct `async` function.
 * If you accidentally call `kotlinx.coroutines.async` instead,
 * the original `coroutineScope { }` behavior is still respected.
 * Luckily this should not pose an issue often, since async is a member of `AwaitAllScope`.
 */
```

PR #3442, opened 2024-05-27
Pattern matching DSL

serras:
This PR proposes a small DSL to describe pattern matching. This is based on three ideas:
1. Pattern matching is essentially as attempting to match several `Optional`s in a row,
2. We can build optionals that match on more than one value (that is, `Optional<S, A> + Optional<S, B> = Optional<S, Pair<A, B>>`
3. The usual way to define patterns looks a lot like function application, so we can reuse `invoke` to simulate them.

This is an example from the code. Note that `then` separates an optional that matches and the code that uses the information being matched as result.

```kotlin
val User.name: String get() = this.match {
  // Person(Name(firstName = fn), age if it < 18)
  User.person(Person.name(Name.firstName), Person.age.suchThat { it < 18 }) then { (fn, _) -> fn }
  // Person(Name(firstName = fn, lastName = ln)) -> "Sir/Madam $fn $ln"
  User.person(Person.name(Name.firstName, Name.lastName)) then { (fn, ln) -> "Sir/Madam $fn $ln" }
  // Company(name = nm, director = Name(lastName = d))
  User.company(Company.name, Company.director(Name.lastName)) then { (nm, d) -> "$nm, att. $d" }
}
```


nomisRev:
Cool stuff! Never actually used Optics for pattern matching, knew it was theoretically possible. Can we also supports  sealed classes through `Prism`?

serras:
> Can we also supports sealed classes through `Prism`?

We do! It's there in the example :P

nomisRev (approved):
SHIP IT!!

serras:
I've just pushed a new commit which I think covers both of your comments:
- The main function is now `matchOrElse`, which gets an additional argument telling what to do if no match is found;
- From those `matchOrThrow`, `matchOrRaise`, and `matchUnit` are derived.

I like the possibility of having a catch-all case _within_ the DSL, since it mirrors how one usually defines pattern matching.

```kotlin
value.match {
  User.company.address(Address.street) then { s -> ... }
  default { "Unknown address " }
}
```

nomisRev:
> I like the possibility of having a catch-all case within the DSL, since it mirrors how one usually defines pattern matching.

Yes, that looks great, but I felt that the strategy should be explicit in the function names. Thank you, great improvement!

raulraja:
Late to the party but wanted to say this looks great 👏

PR #3443, opened 2024-05-30
Arrow Fx DSL

nomisRev:
This PR proposes a **new** module for Arrow Fx Coroutines, which covers the existing behavior in Arrow Fx but in a generalised way using DSLs.

- [x] `parMapN` is covered by `AwaitAllScope`
- [x]  `Resource` already was a DSL, `Scoped` offers a similar but more lightweight DSL (no `ExitCase`), and simplified SAM.
- [ ] `raceN` will be covered by `RacingScope` (#3342)
- [ ] `parZipOrAccumulate` should be covered by a combination of `AwaitAllScope` & `Accumulating)

All the above non-DSL variants, can be implemented by their DSL variations. Simplifying the internals of Arrow Fx Coroutines. The idea behind this is future proofing, since my ambition for 3.0 is to split `Raise` and the typed error handling DSL from `Arrow Core` such that users can rely on **only** the DSL after 3.0. That should be do-able with only a single binary breakage in Arrow Core, so 3.0 should be source compatible, and 99% binary compatible once Context Parameters are released.

- [ ] Write tests
- [ ] Check all edge-cases with `AwaitAllScope`
- [ ] Implement `parZipOrAccumulate` DSL
- [ ] Finish `RacingScope`

Investigate if we need, or want, `<E>` variants when inside `Raise<E>`.


serras:
One DSL to rule them all and in the coroutines bind them?

nomisRev:
"It all began with the forging of the Great" DSLs 🤣

nomisRev:
These can also come after 2.0.0 @serras

arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/AutoCloseScope.kt
```diff
@@ -73,29 +87,11 @@ public inline fun <A> autoCloseScope(block: AutoCloseScope.() -> A): A {
...
-
-  override fun <A> autoClose(acquire: () -> A, release: (A, Throwable?) -> Unit): A =
-    try {
-      acquire().also { a ->
-        finalizers.update { it + { e -> release(a, e) } }
-      }
-    } catch (e: Throwable) {
-      throw e
-    }
+  override fun autoClose(close: (Throwable?) -> Unit) =
```
kyay10, line 93: Nit: I'd like `onClose` better to reflect `onRelease` from `ResourceScope`

arrow-libs/fx/arrow-fx-dsl/src/commonMain/kotlin/arrow/scoped/Saga.kt
```diff
@@ -0,0 +1,56 @@
...
+ * suspend fun main() = scoped {
+ *   val order = saga({ createOrder() }) { deleteOrder(it) }
+ *   val payment = saga { createPayment(order) }, ::deletePayment)
+ *   payment.awaitSuccess()
+ * }
+ * ```
+ */
+// TODO should this be implemented on ScopingScope
+//   Or should ScopingScope be used to implement a separate SagaScope??
+public suspend fun <A> ScopingScope.saga(
```
kyay10, line 46: Isn't this just `ScopingScope.resource`?

arrow-libs/fx/arrow-fx-dsl/src/commonMain/kotlin/arrow/scoped/ScopingScope.kt
```diff
@@ -0,0 +1,56 @@
...
+
+import arrow.AutoCloseScope
+import arrow.throwIfFatal
+import arrow.atomic.Atomic
+import arrow.atomic.update
+import kotlinx.coroutines.NonCancellable
+import kotlinx.coroutines.withContext
+import kotlin.coroutines.cancellation.CancellationException
+
+public interface ScopingScope : AutoCloseScope {
```
kyay10, line 14: This is identical to `ResourceScope`, no? `ScopingScope` doesn't feel like a better name IMO.

arrow-libs/fx/arrow-fx-dsl/src/commonMain/kotlin/arrow/scoped/ScopingScope.kt
```diff
@@ -0,0 +1,56 @@
...
+  public override fun autoClose(close: (Throwable?) -> Unit): Unit
+  public fun closing(block: suspend (Throwable?) -> Unit): Unit
```
kyay10, line 16: Similarly, calling this `onRelease` might make for a more seamless transition.

arrow-libs/fx/arrow-fx-dsl/src/commonMain/kotlin/arrow/scoped/ScopingScope.kt
```diff
@@ -0,0 +1,56 @@
...
+import arrow.AutoCloseScope
+import arrow.throwIfFatal
+import arrow.atomic.Atomic
+import arrow.atomic.update
+import kotlinx.coroutines.NonCancellable
+import kotlinx.coroutines.withContext
+import kotlin.coroutines.cancellation.CancellationException
+
+public interface ScopingScope : AutoCloseScope {
+  public override fun autoClose(close: (Throwable?) -> Unit): Unit
```
kyay10, line 15: this can have a default impl of: `closing { block(it) }`

kyay10:
Hey @nomisRev! Turns out I've accidentally reinvented some of your `ScopingScope` in #3518.
Do you think it'd make sense if this PR, instead of inventing a new ScopingScope, simply moves the (cleaned up) ResourceScope to this new DSL module (while also `api` importing it in fx proper so that no source or binary breakage occurs). Similarly, we can deprecate `install`, `releaseCase`, `onRelease` etc and deprecate `ExitCase` since it's equivalent to `Throwable?`.

nomisRev:
Hey @kyay10,
Thank you for all your great work! Yes, we should close this and work with the great improvements you've already made.
I have some catching up to do, I'll come back to this soon. It's non-urgent.

PR #3450, opened 2024-06-04
Add test for #3441

### LINKED ISSUE #3441: ["BUG"] Arrow Optics, `it` package in lambda

1.2.4

What would you like to see?

Use explicit lambda arguments, to prevent issues like this.
Screenshot 2024-05-27 alle 12 09 25

serras:
Fixes #3441 

The current `main` doesn't have this problem, because we use a different technique to generate prisms. This test checks not only prism, but also lens generation.


nomisRev (approved):
Thank you @serras 🙌

PR #3454, opened 2024-06-13
Delicate optics

serras:
(Raised by [this Slack conversation](https://kotlinlang.slack.com/archives/C5UPMM0A0/p1718266603329739))

This PR brings a couple of unlawful optics, namely filtering values and aggregating. These are potentially dangerous in different ways, so they need to be opted-in using `DelicateOptic`.

@nomisRev I've also renamed the `filter` functions in the `dsl` package to `filterIndex`, as that's the place they make the filtering. That way the new one can get the `filter` name, which coincides which what you expect from a collection / traversal. AFAIK that `filter` in the DSL hadn't been released yet, so should be OK changing it.


kyay10:
Can we have a test showcasing what happens when the optics are used incorrectly? Not super knowledgable about the theory, so I'd like to have examples of incorrect usage and to codify the current behaviour so that it doesn't change unexpectedly

serras:
> Can we have a test showcasing what happens when the optics are used incorrectly? Not super knowledgable about the theory, so I'd like to have examples of incorrect usage and to codify the current behaviour so that it doesn't change unexpectedly

I've added the information to the documentation. Furthermore, I've realized that it should not be a prism, but simply an optional. The example was taken from [Haskell's `optics` package](https://hackage.haskell.org/package/optics-core-0.4.1.1/docs/Optics-AffineTraversal.html).

nomisRev:
> Furthermore, I've realized that it should not be a prism, but simply an optional.

Cool, I was wondering about that given there was no "constructing" going on.

nomisRev (approved):
Very nice @serras!! 👏 👏

PR #3463, opened 2024-07-04
Fix incompatible versioning

serras:
The way we set the versions is no longer compatible with the new versions of Kotlin. And arguably, we should have always done it by setting it in the `kotlin.compilerOptions` block, instead of in each separate task.

(This was caught by early tests as part of the Kotlin User Projects run by the Kotlin Q&A team)


nomisRev (approved):
Thank you for fixing this @serras. I think this was purely poor knowledge of how to properly use Kotlin Gradle configuration, and a combination of this project (config) being so old.

PR #3466, opened 2024-07-04
Remove optional build of `optics-compose`

serras:
Hopefully the final PR to make things work again in the Kotlin User Projects of the Kotlin team. It seems we had too many `2.0` hardcoded in our build files...


nomisRev (approved):
Thank you @serras for watching over this! 🙏

PR #3480, opened 2024-08-23
Add `Nel` APIs shorthand

T45K:
there is no shorthand of `NonEmtpyList` APIs while there is `Nel`, shorthand of `NonEmptyList` itself.

in this PR, i added three APIs:
- `nelOf`: shorthand of `nonEmptyListOf`
- `toNelOrNull`: shorthand of `toNonEmptlyListOrNull`
- `toNelOrOptional`: shorthand of `toNonEmptyListOrOptional`


serras:
I am not sure of how useful this is: the longer names are much clearer to everybody reading, and if you are using an IDE, I'm not sure how many keystrokes you really save.

The only shorthand we have now is `EitherNel`, but here the main goal is to get rid of the nesting (`Either<NonEmptyListOf<...>>`) than the name per se.

T45K:
thank you for checking this PR

i added them because long API name sometimes make a length of single line long, and then hard to read (at least for me).
i define such shorthands in my project by myself, but need to define them repeatedly in different projects. it is tired task.
so i'd be happy if arrow.kt itself provided those shorthands...

if you don't agree with my idea, please close this PR...

PR #3489, opened 2024-09-17
Provide `mapOrAccumulate` for `NonEmptySet`

tPl0ch:
We are using the `NonEmptySet` abstraction pretty heavily, and are constantly missing `mapOrAccumulate` on it. While it's possible to convert to a `NonEmptyList`, run the accumulating map, and then convert back, this seems a very inefficient way to achieve this.


serras:
This looks quite useful to me, thanks! :)

serras (approved):
Sorry for the late review, thanks very much!

PR #3497, opened 2024-10-05
Flag new 2.0 APIs as experimental

serras:
Since we are not 100% sure of the API, I thought it would be useful to mark those experimental for one or two releases.


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -671,6 +671,12 @@ public inline fun <K, Error, A, B> Raise<NonEmptyList<Error>>.mapOrAccumulate(
   }
 }
 
+@RequiresOptIn(level = RequiresOptIn.Level.WARNING, message = "This API is work-in-progress and is subject to change.")
+@Retention(AnnotationRetention.BINARY)
+@Target(AnnotationTarget.FUNCTION)
+public annotation class ExperimentalRaiseAccumulateApi
```
hoc081098, line 677: I think it should be annotated with `@MustBeDocumented`

PR #3498, opened 2024-10-05
Regex-inspired optics

serras:
These add combinators like `onceOrMore` and `zeroOrMore` to traverse children to an unbounded depth.


nomisRev (approved):
Amazing @serras 👏 👏

PR #3501, opened 2024-10-08
Added common basic test cases for cache4k

Sujin1135:
I noticed that **cache4k** does not have test cases, So I thought I should create some basic ones.

**Basic Test Cases**
- Verify saved data is returned correctly.
- Verify get method should return null for an invalid key.

That's why I tried it right away. While I was working on it, I encountered an issue with java version of Gradle, So I fixed it as this file changes. (Java version 8 -> 17)

Screenshot 2024-10-09 at 2 08 18 AM

The issue related to `com.android.library` used in [root build.gradle.kts](https://github.com/arrow-kt/arrow/blob/99de6148320a4299a5aef20686a6063ca732026b/build.gradle.kts#L41) and [optics build.gradle.kts](https://github.com/arrow-kt/arrow/blob/99de6148320a4299a5aef20686a6063ca732026b/arrow-libs/optics/arrow-optics-compose/build.gradle.kts#L21).


serras:
Sorry for taking so long to merge this. It's wonderful to have people contributing test cases to ensure that we don't break things when moving forward :)

Sujin1135:
@serras Could you let me know how to join this Slack channel?

PR #3504, opened 2024-10-15
Make `Schedule` a `fun interface`

### LINKED ISSUE #3325: 1.1.5 to 1.2.1 Migration - arrow.resilience.Schedule causes 'bad class file' error in Dagger KSP
I migrated to the latest version of Arrow which made me add `io.arrow-kt:arrow-resilience` to use the Schedule class.
The import statement therefore changed from `arrow.fx.coroutines.Schedule` to `arrow.resilience.Schedule` and also `arrow.fx.coroutines.retry` to `arrow.resilience.retry`.

I'm using Dagger KSP version `2.49` and KSP version `1.9.21-1.0.16`

When I build the module I get the following error message:

```
C:\Users\Me\AndroidStudioProjects\MyApp\app\build\generated\ksp\debug\java\com\...\dependencyinjection\app\AppModule_SomeClassFactory.java:5: error: cannot access SomeClass
import com.somepackage.SomeClass;
                                ^
  bad class file: C:\Users\Me\AndroidStudioProjects\MyApp\app\build\tmp\kotlin-classes\debug\com\...\SomeClass.class
    undeclared type variable: Input
    Please remove or make sure it appears in the correct subdirectory of the classpath.
C:\Users\Me\AndroidStudioProjects\MyApp\app\build\generated\ksp\debug\java\com\...\dependencyinjection\app\AppModule_SomeClassFactory.java:32: error: cannot find symbol
public final class AppModule_SomeClassFactory implements Factory<SomeClass> {
                                                                            ^
  symbol: class SomeClass
C:\Users\Me\AndroidStudioProjects\MyApp\app\build\generated\ksp\debug\java\com\...\dependencyinjection\app\AppModule_SomeClassFactory.java:67: error: cannot find symbol
  public SomeClass get() {
         ^
  symbol:   class SomeClass
  location: class AppModule_SomeClassFactory
C:\Users\Me\AndroidStudioProjects\MyApp\app\build\generated\ksp\debug\java\com\...\dependencyinjection\app\AppModule_SomeClassFactory.java:81: error: cannot find symbol
  public static SomeClass someClass(AppModule instance,
                ^
  symbol:   class SomeClass
  location: class AppModule_SomeClassFactory
```

So actually I cannot fully migrate from 1.1.5 to 1.2.1 using the lastest APIs/import statements.

### LINKED ISSUE #3499: Arrow resilience causes an `undeclared type variable: Input` error when compiling
I used this code:
```kotlin
private val recur2 = Schedule.recurs<CallError>(3)

...
recur2.retryEither {
   searchApi.getSuggestions(...) // returns an Either<CallError, List<String>>
}
```
and I get this error when compiling:
```
error: cannot access SearchSuggestionsViewModel
    public abstract ViewModel binds(SearchSuggestionsViewModel vm);
                                    ^
  bad class file: /home/.../app/build/tmp/kotlin-classes/debug/.../searchScreens/SearchSuggestionsViewModel.class
    undeclared type variable: Input
    Please remove or make sure it appears in the correct subdirectory of the classpath.
```

serras:
Fixes #3325
Fixes #3499

It turns out that using a `value class` containing a function inside is equivalent to having just a `fun interface`. So I've reworked `Schedule` in that form, which hopefully should remove the problems we have with inlining. @dave08 @postfixNotation

This change is _breaking_, but since in 2.0 we only promised source compatibility (and not binary compatibility), this should still be OK.


serras:
PS: This also updates Kotlin to 2.1.0-Beta2, since 2.0.21 was giving a very weird compiler error.

serras:
It seems that this breaks in Native. The error is now [reported in YouTrack](https://youtrack.jetbrains.com/issue/KT-72281/Failed-to-wait-for-cache-to-be-built-with-Kotlin-Native)

arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/Schedule.kt
```diff
@@ -33,9 +30,6 @@ import kotlin.time.Duration
 import kotlin.time.Duration.Companion.ZERO
 import kotlin.time.Duration.Companion.nanoseconds
 
-public typealias ScheduleStep<Input, Output> =
```
kyay10, line 36: Nit: This technically breaks anyone who used `ScheduleStep` in their code (and similarly anyone who called `val f = Schedule.step`) but I don't think it's worth the worry

serras: I've managed to rewrite it so that `ScheduleStep` and `step` are still there :)

PR #3516, opened 2024-10-25
refactor: rename Resource.allocated to Resource.allocate

tKe:


tKe:
I'm not sure if the js test failure is anything to do with this change

serras:
The [1.2.4 branch](https://github.com/arrow-kt/arrow/blob/1.2.4/arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt#L696) shows that at that point there were two versions: `allocated` and `allocate`, with the latter being "safer". It's true that the signature correspond to the previous `allocate`, so this refactoring makes sense.

@nomisRev do you remember the history of this change? Do you think it makes sense to go back to `allocate`, as the signature matches 1.2.4?

kyay10:
Using a verb makes more sense because it *executes* the `Resource`

kyay10 (approved):
Is there a source compatibility guarantee for this? Probably not because it was declared delicate.

tKe:
The current 1.2.4 release has both `allocated` and `allocate` with the former deprecated and the latter being recommended. So this change would actually make arrow 2.0 more source compatible.

kyay10:
Oh, so this actually *brings* source compatibility! Wonderful! I'll wait on Simon to have a look, but I'm all for this change!

PR #3518, opened 2024-11-01
AutoClose and Resource cleanup 

kyay10:
This subsumes #3515. This PR can be changed to use that same approach (or if #3515 is merged, this PR can just directly use it).

I've avoided source-breaking changes here. I would've liked to make `autoClose` inline, but that'd be source-breaking.
I changed the required methods to implement for `AutoCloseScope` and `ResourceScope`, but does Arrow consider that part of its source compatibility? I'm assuming not.

Compiler complains about `resourceScope` and `use` not needing an explicitly `suspend` block anymore (because it's automatically passed-through by `inline`). However, I think that change can be source-breaking if someone was using a function reference (compiler isn't smart enough to understand that sadly), so I suppressed that warning (which is the approach taken in `Bracket` as well)

I caught an error where `autoCloseScope` wouldn't close if an early-return happened. That was easily solvable using `finally` (which I think covers every possible exit scenario? Except `suspendCoroutine<Nothing> { }` but there's no way to stop that, and it's basically like looping forever). I fixed that error for Bracket-like functions as well, and added tests for the aforementioned and for `use` and `resourceScope`

I found a logical error where `allocated` didn't have the same semantics as `resourceScope` when it came to adding suppressed errors. Now `allocated` is very dumb and clearly has the same semantics, but I still added a test case anyway.

The code uses some `also`s but they're used whenever the original code used them. However, I'd be happy to convert them to imperative style if you wish!

Some tests for bracket composing weren't testing suppressed exceptions correctly, and fixing that revealed a problem where `throwable`s were reused across tests, which doesn't work with the mutation of `suppressedExceptions`

To discuss: do we need AcquireStep still? I recall a discussion with Simon (TODO link) about where it came from (some Scala lib) and how it's likely unnecessary anymore. We can even keep it but deprecate it and remove the DSL annotation from it so its restrictions don't apply anymore


arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/AutoCloseScope.kt
```diff
@@ -63,43 +63,43 @@ import kotlin.coroutines.cancellation.CancellationException
  */
 public inline fun <A> autoCloseScope(block: AutoCloseScope.() -> A): A {
   val scope = DefaultAutoCloseScope()
+  var finalized = false
   return try {
     block(scope)
-      .also { scope.close(null) }
-  } catch (e: CancellationException) {
```
kyay10, line 69: `throwIfFatal` is only used here, so I made it not consider CancellationExceptions fatal for the sake of brevity.

arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/AutoCloseScope.kt
```diff
@@ -63,43 +63,43 @@ import kotlin.coroutines.cancellation.CancellationException
...
-    scope.close(e) ?: throw e
   } catch (e: Throwable) {
-    scope.close(e.throwIfFatal()) ?: throw e
+    finalized = true
+    scope.close(e.throwIfFatal())
+    throw e
+  } finally {
```
kyay10, line 73: Using `finally` to ensure closing even in non-local returns (or `break`s even in Kotlin 2.2)

arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/AutoCloseScope.kt
```diff
@@ -63,43 +63,43 @@ import kotlin.coroutines.cancellation.CancellationException
...
-    }
+  override fun onClose(release: (Throwable?) -> Unit) {
+    finalizers.update { it + release }
+  }
 
-  fun close(error: Throwable?): Nothing? {
-    return finalizers.get().asReversed().fold(error) { acc, function ->
-      acc.add(runCatching { function.invoke(error) }.exceptionOrNull())
+  fun close(error: Throwable?) {
+    finalizers.value.asReversed().fold(error) { acc, finalizer ->
```
kyay10, line 101: Tried to get the code closer to the style of the corresponding resource scope implementation, mainly for the sake of future readers of the code (since in essence the two are very similar)

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -303,24 +301,21 @@ public interface ResourceScope : AutoCloseScope {
...
-    return install({ a }, release)
-  }
+  public suspend infix fun <A> Resource<A>.releaseCase(release: suspend (A, ExitCase) -> Unit): A =
+    bind().also { a -> onRelease { release(a, it) } }
+
+  override fun onClose(release: (Throwable?) -> Unit): Unit = onRelease { release(it.errorOrNull) }
 
-  public suspend infix fun onRelease(release: suspend (ExitCase) -> Unit): Unit =
-    install({ }) { _, exitCase -> release(exitCase) }
+  public infix fun onRelease(release: suspend (ExitCase) -> Unit)
```
kyay10, line 318: not even marked as `suspend` anymore!

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -354,24 +349,23 @@ public fun <A> resource(block: suspend ResourceScope.() -> A): Resource<A> = blo
...
+  return try {
     action(scope)
   } catch (e: Throwable) {
-    val ex = if (e is CancellationException) ExitCase.Cancelled(e) else ExitCase.Failure(e)
-    val ee = withContext(NonCancellable) {
-      scope.cancelAll(ex, e) ?: e
+    finished = true
+    scope.cancelAll(ExitCase.ExitCase(e))
+    throw e
+  } finally {
```
kyay10, line 361: Same `finally` idea to deal with non-local control

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -354,24 +349,23 @@ public fun <A> resource(block: suspend ResourceScope.() -> A): Resource<A> = blo
...
-  val a: A = try {
+  var finished = false
+  return try {
     action(scope)
   } catch (e: Throwable) {
-    val ex = if (e is CancellationException) ExitCase.Cancelled(e) else ExitCase.Failure(e)
-    val ee = withContext(NonCancellable) {
-      scope.cancelAll(ex, e) ?: e
+    finished = true
+    scope.cancelAll(ExitCase.ExitCase(e))
```
kyay10, line 359: `cancelAll` deals with `NonCancellable` internally now, which is good because it can now throw the errors itself. By that I mean that the code:
```
withContext(NonCancellable) {
  throw someExceptionWithSuppressed
}
```
actually strips the suppressed exceptions due to internal coroutine machinery, and so we before had to throw the exception outside, but now we encapsulate all that logic in `closeAll`, which also ensures consistency in semantics between this and `allocated`

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -469,72 +463,33 @@ public fun <A> Resource<A>.asFlow(): Flow<A> =
...
-      ExitCase.Completed -> null
-      is ExitCase.Cancelled -> e.exception
-      is ExitCase.Failure -> e.failure
-    }
-    original?.apply {
-      suppressed?.let(::addSuppressed)
-    }?.let { throw it }
-  }
-  return Pair(allocated, release)
+public suspend fun <A> Resource<A>.allocated(): Pair<A, suspend (ExitCase) -> Unit> = with(ResourceScopeImpl()) {
```
kyay10, line 466: The previous code didn't correctly add up the suppressed exceptions into the new one for some reason (at least that reason isn't clear to me), so I changed it (and added the relevant test) to match the semantics of `resourceScope`

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -469,72 +463,33 @@ public fun <A> Resource<A>.asFlow(): Flow<A> =
...
-) : ResourceScope {
-  override suspend fun <A> Resource<A>.bind(): A = invoke(this@ResourceScopeImpl)
-  
-  override suspend fun <A> install(acquire: suspend AcquireStep.() -> A, release: suspend (A, ExitCase) -> Unit): A =
-    bracketCase({
-      val a = acquire(AcquireStep)
-      val finalizer: suspend (ExitCase) -> Unit = { ex: ExitCase -> release(a, ex) }
-      finalizers.update(finalizer::prependTo)
-      a
-    }, ::identity, { a, ex ->
```
kyay10, line 501: I don't know how to describe the previous code here other than that it clearly was refactored slowly until it rotted away. Basically, because the `use` function passed to `bracketCase` is `identity`, `ex` is *always* `Completed`, and hence the code in that finalizer is a no-op.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -469,72 +463,33 @@ public fun <A> Resource<A>.asFlow(): Flow<A> =
...
-  override fun <A> autoClose(acquire: () -> A, release: (A, Throwable?) -> Unit): A =
-    acquire().also { a ->
-      val finalizer: suspend (ExitCase) -> Unit = { exitCase ->
-        val errorOrNull = when (exitCase) {
-          ExitCase.Completed -> null
-          is ExitCase.Cancelled -> exitCase.exception
-          is ExitCase.Failure -> exitCase.failure
-        }
-        release(a, errorOrNull)
+  suspend fun cancelAll(exitCase: ExitCase) {
```
kyay10, line 477: This one was also refactored to look more like the auto-close impl, mainly by adding that `Throwable?.add` extension which cleans things up nicely

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -469,72 +463,33 @@ public fun <A> Resource<A>.asFlow(): Flow<A> =
...
-      if (ex != ExitCase.Completed) {
-        val e = cancelAll(ex)
-        val e2 = kotlin.runCatching { release(a, ex) }.exceptionOrNull()
-        e?.apply {
-          e2?.let(::addSuppressed)
-        }?.let { throw it }
-      }
-    })
+@PublishedApi
+internal class ResourceScopeImpl : ResourceScope {
```
kyay10, line 471: Made it not a value class to prevent API surface explosion, but it can be undone (especially if we use the #3515 technique)

tKe: Could also move the implementing of the ResourceScope to an `val scope = object : ResourceScope {...}` - effectively making this class have `val scope: ResourceScope` and `fun cancelAll`, reducing the exposed ABI

arrow-libs/core/arrow-autoclose/src/commonTest/kotlin/arrow/AutoCloseTest.kt
```diff
@@ -136,6 +135,23 @@ class AutoCloseTest {
...
+
+    run {
+      autoCloseScope {
+        val r = install(res)
+        wasActive.complete(r.isActive())
+        return@run
+      }
+    }
+
+    wasActive.await() shouldBe true
```
tKe, line 151: if the test fails, `.await()` will suspend until the 1minute timeout. It would be nice to have a `wasActive shouldHaveCompletedWith true` matcher that checks `Job.isComplete` before calling `.await()` (or `.getCompleted()`)

this also applies to a number of tests throughout (noticed when I tried breaking things)

arrow-libs/core/arrow-autoclose/src/jvmMain/kotlin/arrow/throwIfFatal.kt
```diff
@@ -1,10 +1,8 @@
 package arrow
 
-import kotlin.coroutines.cancellation.CancellationException
-
 @PublishedApi
 internal actual fun Throwable.throwIfFatal(): Throwable =
   when(this) {
-    is VirtualMachineError, is ThreadDeath, is InterruptedException, is LinkageError, is CancellationException -> throw this
+    is VirtualMachineError, is ThreadDeath, is InterruptedException, is LinkageError -> throw this
```
tKe, line 6: I understand in some places it's desirable to handle CancellationException separately but it's confusing to have two definitions of "fatal" exceptions (this vs `arrow.core.NonFatal`)
It may be clearer to explicitly handle CancellationException where required and use the `NonFatal` in failure cases.

tKe (commented):
I'm wary of removing `NonCancellable` from `install` - will that allow for a resource to be instantiated and cancellation to occur before a finalizer is registered, and therefore left dangling?

Also it may be worth including similar test/fixes to `bracket` and `guarantee` around the non-local return / finally in the same PR as an overhaul of all related constructs. Especially as when fixed they'll all likely follow the same approach of "acquire a scope, action the scope and the finalize exactly once on either error or completion" but potentially with different semantics on exception handling/throwing.

arrow-libs/core/arrow-autoclose/src/jvmMain/kotlin/arrow/throwIfFatal.kt
kyay10: That's fair, I'll add an explicit carve-out for CancellationException instead

arrow-libs/core/arrow-autoclose/src/commonTest/kotlin/arrow/AutoCloseTest.kt
kyay10: I was copying the style of the other tests in the file, but yes it might be worth it to make those tests fail-fast. I'll make the necessary changes (likely just using a `lateinit var`)

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
kyay10: Not sure how the `cancelAll` would be exposed though? Can you clarify a little more?

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -354,24 +351,25 @@ public fun <A> resource(block: suspend ResourceScope.() -> A): Resource<A> = blo
...
+    scope.cancelAll(ExitCase.ExitCase(e))
+    throw e
+  } finally {
+    if (!finished) {
+      scope.cancelAll(ExitCase.Completed)
     }
-    throw ee
   }
-  withContext(NonCancellable) {
-    scope.cancelAll(ExitCase.Completed)?.let { throw it }
```
kyay10, line 369: We should add a test for this because I'm pretty sure that suppressed exceptions would've been stripped here as well, but it should work properly now

arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/AutoCloseScope.kt
kyay10: Scratch that, I've gone back to explicitly singling out CancellationException

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
```diff
@@ -20,6 +20,14 @@ public sealed class ExitCase {
   }
 }
 
+@PublishedApi
+internal val ExitCase.errorOrNull: Throwable?
```
kyay10, line 24: Should we just make this `public`?

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
```diff
@@ -231,31 +206,35 @@ public suspend inline fun <A, B> bracketCase(
...
     }
   } catch (e: Throwable) {
-    original.addSuppressed(e.nonFatalOrThrow())
+    original?.addSuppressed(e.nonFatalOrThrow()) ?: throw e
+  }
+  original?.let { throw it }
+}
+
+@PublishedApi
+internal inline fun <R> finalizeCase(block: () -> R, finalizer: (ExitCase) -> Unit): R {
```
kyay10, line 226: Should we make this public?
Also annoyingly it duplicates the bytecode of `finalizer`. I think? this can be avoided

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -354,24 +351,15 @@ public fun <A> resource(block: suspend ResourceScope.() -> A): Resource<A> = blo
...
-    throw ee
-  }
-  withContext(NonCancellable) {
-    scope.cancelAll(ExitCase.Completed)?.let { throw it }
-  }
-  return a
+@OptIn(DelicateCoroutinesApi::class)
+@Suppress("REDUNDANT_INLINE_SUSPEND_FUNCTION_TYPE")
+public suspend inline fun <A> resourceScope(action: suspend ResourceScope.() -> A): A {
+  val (scope, cancelAll) = resource { this }.allocated()
```
kyay10, line 357: A little hacky, but it results in no new ABI added. We might instead want a function `allocateResourceScope(): Pair<ResourceScope, suspend (ExitCase) -> Unit>`

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
kyay10: I found a way to hide it completely. Not sure if making it a `value class` is a good idea though, since it'll always be boxed.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -469,69 +457,28 @@ public fun <A> Resource<A>.asFlow(): Flow<A> =
...
-        val errorOrNull = when (exitCase) {
-          ExitCase.Completed -> null
-          is ExitCase.Cancelled -> exitCase.exception
-          is ExitCase.Failure -> exitCase.failure
-        }
-        release(a, errorOrNull)
+  suspend fun cancelAll(exitCase: ExitCase) {
+    withContext(NonCancellable) {
+      finalizers.value.fold(exitCase.errorOrNull) { acc, finalizer ->
+        acc.add(runCatching { finalizer(exitCase) }.exceptionOrNull())
```
kyay10, line 473: if the finalizer fails with a fatal exception, I think we should short-circuit and throw it here, right? I'll add a test and change the behaviour. Note that `runReleaseAndRethrow` which is used for the `bracket` functions does that already

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
tKe: it seems fairly delicate with respect to exception/suppression. The only benefit over `guaranteeCase` is it's non-suspending.

kyay10: Very true. It's however exactly the right method needed to use `allocate` correctly, or to use any `release`-like method that deals with suppression and non-cancellability, but it is very niche, I'll admit

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
kyay10: Added!

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
kyay10: Added the relevant tests!

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
tKe: The original purpose of `allocate` was to break up a resource lifecycle for integrations (such as test listeners) which wouldn't be able to use `finalizeCase` for the same reason it can't just use `.use`. Gut says to keep it internal for now, and expose it should a viable usecase come about. Either that, or give it a suitable opt-in annotation.

arrow-libs/core/arrow-autoclose/src/commonTest/kotlin/arrow/AutoCloseTest.kt
kyay10: Done!

arrow-libs/core/arrow-autoclose/src/jvmMain/kotlin/arrow/throwIfFatal.kt
kyay10: Done!

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
tKe: I think the edge-case of needing this outside of the library implementation would be infrequent enough (and this method is trivial enough) to not warrant it.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
tKe: I'm not entirely certain `infix` has much use here, given `ResourceScope` is predominantly used as a receiver. I can't see anyone wanting to write `this onRelease { ... }`. It _would_ impact source compatibility but ... is it ever actually used in that way?

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
tKe: I think this is fine - if anything a comment to highlight that `this` is the `ResourceScope` for clarity but it's internal and imo understandable from the context.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
tKe: No benefit to the value class when it's being treated as a ResourceScope as it'll be boxed. I'm not sure how the `::cancelAll` reference would get represented in the byte-code, but likely it's boxed too. Simpler to reason over a simple class and instance method reference.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -303,24 +303,23 @@ public interface ResourceScope : AutoCloseScope {
...
+    bind().also { a -> onRelease { release(a) } }
 
   /** Composes a [releaseCase] action to a [Resource] value before binding. */
   @ResourceDSL
-  public suspend infix fun <A> Resource<A>.releaseCase(release: suspend (A, ExitCase) -> Unit): A {
-    val a = bind()
-    return install({ a }, release)
-  }
+  public suspend infix fun <A> Resource<A>.releaseCase(release: suspend (A, ExitCase) -> Unit): A =
+    bind().also { a -> onRelease { release(a, it) } }
```
tKe, line 318: Do we need to be concerned about `bind()` being called in a non-cancellable context here? (It wasn't previously, either) - possibly a question for @nomisRev ?

It does mean `resource { ... } release {}` acts differently to `install({ ... }) { _, _ -> }`

tKe (approved):
Not sure if my approval counts for much, but I think this is all much improved, leaner and more composed.

serras:
@tKe @kyay10 thanks very much for this great effort. Mostly for my own sake, could you make #3515 also part of this PR? That way I can get a better picture of what's going on here.

kyay10:
@serras This PR covers the features of #3515, but implements it differently, hence I didn't base it on its commits. I.e. if this is merged, #3515 is redundant, but this does more than #3515

tKe:
This covers what #3515 set out to achieve, with the same sureties on finalizers being called with non-local returns, but also simplifies and shares the previously-duplicated (and subtly different) implementations of `resourceScope` and `Resource.allocate` (the verb does make so much more sense than the past-participle `allocated`). It also extends the same fixes for non-local returns to guaranteeCase, bracketCase and autoCloseScope.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
kyay10: Agreed. Will leave as-is for now

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
kyay10: `Resource` is basically a fancy name for any function that you might call taking a `ResourceScope`, so I'd assume that either the function registers its own release (and this one here is extra for some reason?) or that the function registers some resources, but it leaves the final one for you to register yourself.
Inline usage of `resource` like that is stylistically tempting, but it feels easier to just do it imperatively.
Maybe what we actually want is for `resource` to be non-cancellable by default?

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
kyay10: I agree, but I didn't wanna touch it. Maybe some folks who don't use contexts are forced to explicitly pass around `ResourceScope`, and hence they end up calling `scope onRelease { }`

kyay10:
Hey @serras, if this PR is too big, I can split it into 2. 1 for the binary-breaking changes that would need to be merged before 2.0 is out thus, and another for the behaviour-correcting changes that can come in 2.x.

arrow-libs/core/arrow-autoclose/src/commonTest/kotlin/arrow/AutoCloseTest.kt
```diff
@@ -3,62 +3,60 @@ package arrow
...
 import kotlin.test.Test
 
-@OptIn(ExperimentalStdlibApi::class)
 class AutoCloseTest {
 
   @Test
   fun canInstallResource() = runTest {
-    val promise = CompletableDeferred<Throwable?>()
-    val wasActive = CompletableDeferred<Boolean>()
+    var throwable: Throwable? = RuntimeException("Dummy exception")
```
serras, line 16: In which situations would this `throwable` be `null`? And why is it not initialized to `null`?

arrow-libs/core/arrow-autoclose/src/commonTest/kotlin/arrow/AutoCloseTest.kt
```diff
@@ -3,62 +3,60 @@ package arrow
...
+    var wasActive = false
     val res = Resource()
 
     autoCloseScope {
       val r = autoClose({ res }) { r, e ->
-        promise.complete(e)
+        throwable = e
```
serras, line 22: Why is it OK to change here from `CompetableDeferred` to a regular `var`?

kotlin-js-store/yarn.lock
serras, line 1: Is this file needed for the commit to work?

arrow-libs/fx/arrow-fx-coroutines/src/commonTest/kotlin/arrow/fx/coroutines/BracketCaseTest.kt
```diff
@@ -206,58 +245,66 @@ class BracketCaseTest {
...
-    checkAll(10, Arb.int(), Arb.throwable(), Arb.throwable()) { n, e, e2 ->
+    checkAll(10, Arb.int(), Arb.throwableConstructor(), Arb.throwable()) { n, eConstructor, e2 ->
+      val e = eConstructor()
       Either.catch {
         bracketCase<Int, Unit>(
           acquire = { n },
           use = { throw e },
           release = { _, _ -> throw e2 }
         )
-      } shouldBe Either.Left(e + e2)
```
serras, line 221: I am very surprised by this change of behavior. Could you explain it?

serras (changes requested):
Thanks for the hard work on this!

Apart from a few clarifications, my main concern here is that some tests change their behavior, and we need to be 100% sure that this is OK.

serras:
@kyay10 sorry, I created some conflicts by merging the other two open PRs...

kyay10:
No problem with the conflicts! They should be trivial to fix up

arrow-libs/fx/arrow-fx-coroutines/src/commonTest/kotlin/arrow/fx/coroutines/BracketCaseTest.kt
kyay10: Yes sorry I noticed this while refactoring, should've probably gone in a different PR.
This is not a change of behaviour, it's just fixing a misleading test. The previous test didn't test what it claimed to.
Basically, the `plus` method did not do what it seems to do. Let me demonstrate this with (mutable) lists:
```
val e2 = ...
val e = mutableListOf<Throwable>()
e.add(e2)

e shouldBe (e.apply { add(e2) })
```
As you can see, this isn't actually testing that e2 is contained in e, in fact, this would also pass the same test:
```
val e2 = ...
val e = mutableListOf<Throwable>()

e shouldBe (e.apply { add(e2) })
```
In other words, the previous test would pass if you do:
```
Either.catch {
  bracketCase<Int, Unit>(
    acquire = { n },
    use = { throw e },
    release = { _, _ -> }
  )
} shouldBe Either.Left(e + e2)
```
because it's not actually testing inclusion because that `+` is misleading.

kyay10: Also, I had to use a constructor for throwables for the same mutability reason: the throwable might get reused across iterations of `checkAll`, and hence we actually ended up with multiple suppressed exceptions in `e`.

kyay10: I can revert this if you want and add it to a different PR because this has nothing to do with the behaviour changes in this PR

kotlin-js-store/yarn.lock
kyay10: Yeah not sure what happened there. I'll remove it

arrow-libs/core/arrow-autoclose/src/commonTest/kotlin/arrow/AutoCloseTest.kt
kyay10: The test doesn't `launch` any coroutines or anything, so the promise is unnecessary. This is in an effort to report failures faster (because if the promise.complete call was never made, the test just hangs until it times out, but now it fails fast)

arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/AutoCloseScope.kt
```diff
@@ -63,50 +64,50 @@ import kotlin.coroutines.cancellation.CancellationException
...
+    throwable = e
+    throw e
+  } finally {
+    if (throwable !is CancellationException) throwable?.throwIfFatal()
+    scope.close(throwable)
   }
 }
 
 public interface AutoCloseScope {
+  public fun onClose(release: (Throwable?) -> Unit)
```
serras, line 80: What is the benefit of exposing this additional function through this interface?

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
```diff
@@ -231,31 +206,32 @@ public suspend inline fun <A, B> bracketCase(
...
-  withContext(NonCancellable) { release(acquired, ExitCase.Completed) }
-
-  return res
+  val acquired = withContext(NonCancellable) { acquire() }
+  return guaranteeCase({ use(acquired) }) { release(acquired, it) }
 }
 
 @PublishedApi
-internal suspend inline fun runReleaseAndRethrow(original: Throwable, crossinline f: suspend () -> Unit): Nothing {
+internal suspend inline fun runReleaseAndRethrow(original: Throwable?, crossinline f: suspend () -> Unit) {
```
serras, line 214: Why has this changed to a nullable `Throwable`? In which situations would `null` be the value here?

arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/AutoCloseScope.kt
kyay10: The point was to have the same basic API as ResourceScope: `onRelease` adds a handler, so `onClose` adds a handler too. This also lets `AutoCloseScope` have a SAM

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
kyay10: In successful completion. This method basically acts as `withContext(NonCancellable) { f()` if given a null throwable (which is the "happy path" case).

serras: But why before it didn't have it? Was the handling of the happy path done somewhere else? If so, I would prefer to keep this without the `?` and inline the corresponding implementation for the `null` cases.

arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/AutoCloseScope.kt
serras: Understood. But `AutoCloseScope` is not marked as `fun interface`, right?

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
kyay10: Before, this was used directly when a throwable was `caught`. Now, we instead use `finalizeCase`. The point is to unify where the release happens one place. Actually, looking at it now, I think this method can be inlined and simplified. One second...

kyay10: Changed!

arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/AutoCloseScope.kt
kyay10: No, we can mark it as such, but it's unlikely to be of benefit. The point of having it as a SAM is more of an API surface perspective: if you understand `onClose`, you understand how the whole DSL works.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Bracket.kt
```diff
@@ -89,16 +87,19 @@ public suspend inline fun <A> guarantee(
...
-  } catch (t: Throwable) {
-    runReleaseAndRethrow(t.nonFatalOrThrow()) { finalizer(ExitCase.Failure(t)) }
+): A = finalizeCase({ fa() }) { ex ->
+  try {
+    withContext(NonCancellable) {
+      finalizer(ex)
+    }
+  } catch (e: Throwable) {
+    e.nonFatalOrThrow()
+    when (ex) {
```
kyay10, line 97: This can be written as `(ex.errorOrNull ?: throw e).addSuppressed(e)` but I went for clarity instead, so that the code now reads like "if fa completed successfully, rethrow the error from its finalizer. Otherwise, suppress that error"

arrow-libs/core/arrow-autoclose/src/commonTest/kotlin/arrow/AutoCloseTest.kt
serras: But `CompletableDeferred` has the additional feature of completing only once, which we cannot check if using a variable.

tKe: We can still get the fail-fast by additionally checking `.isCompleted shouldBe true` which could be combined in a custom matcher

kyay10: Updated!

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
serras: If this was `infix`, let's just keep it that way. No need to break source compatibility here.

arrow-libs/fx/arrow-fx-coroutines/src/commonTest/kotlin/arrow/fx/coroutines/BracketCaseTest.kt
serras: Ah, I see, this is because of the `operator plus` on `Throwable`! Now I understand, I was getting confused by how `plus` on lists actually does what you expect (create a new copy), but this is not the case for the implementation above!

serras (approved):

PR #3533, opened 2024-11-13
Change zipOrAccumulate to not use delegated properties to prevent IAE

kyay10:
Workaround for KT-73089
As per the discussion on slack. I simply made `zipOrAccumulate` not use `by` and instead use a `.value` property I added. Oh and also I made the `operator fun getValue` an inline extension method as per the optimization requirements for `getValue` functions that don't use the `KProperty` they're given. I also sealed `Value` because it didn't need to be inheritable.


StylianosGakis:
For the ones interested in previous discussion here are some links.
Kotlinlang conversation link: https://kotlinlang.slack.com/archives/C5UPMM0A0/p1731449652340769
Public kotlinlang archive channel link: https://slack-chats.kotlinlang.org/t/23415572/tried-bumping-to-2-0-0-beta-1-and-i-am-getting-a-super-weird#8eb0739e-90ef-4b6f-96b3-aa9eda361e04

PR #3559, opened 2025-01-07
Add finer-grained validation for determining the symbols to defer during symbol processing

FWest98:
This addresses an issue introduced here: https://github.com/arrow-kt/arrow/pull/3537#issuecomment-2573056265

Essentially, at the moment, processing for any candidate class for optics is "deferred" when "validation" fails - essentially when any type in the class definition is not resolved. Unfortunately, this causes issues when these types are dependent on the lenses-to-be-generated themselves, such as in the following example:

```kotlin
@optics
data class UsingLens(val field: String) {
  fun getLens() = UsingLens.field // Return type of the function is dependent on the lens
  companion object
}
```

While this class obviously has unresolved types, these do not impact code generation of optics. Ideally, we should limit the validation to only check the validity of the types involved in the generated lenses. The best place for this would probably be in the `Focus` class after generating the ADT, but I could not get this to work very neatly without changing code everywhere.

Instead, I implemented a simpler, "best effort" approach where I skip validation for irrelevant declarations/nodes, such as functions and annotations. This is not perfect, and still causes the validation to be "too strict", but might be a good initial step.

The approach suggested in the other thread - to defer classes once and actually process them on a second pass - does not work, as KSP will stop processing symbols when there is no other processor to defer symbols _to_.


serras:
Thanks for your help in fixing this!

PR #3570, opened 2025-01-25
Move integrations to main repo + new Ktor integrations

serras:
This PR does a few things (sorry, I forgot to switch branches).

It moves the Jackson module from `arrow-integrations` repo into the main repo, which makes it easier to release in the future. Alongside it reorganizes the repo, moving all the integrations modules (Retrofit, Jackson, kotlinx.serialization) into a new `integrations` folder.

Alongside those, it adds a new integration with Ktor in the form of a client plug-in. It does the same as the [built-in `HttpRetry`](https://ktor.io/docs/client-request-retry.html) but using our `Schedule` and `CircuitBreaker`.


nomisRev:
Great work @serras 👏 🙌 

I am a bit unsure of moving the Ktor integration into the main repo, but I am not opposed to it. It was still relatively small, unreleased, and unpromoted but I was quite happy with the APIs last summer. https://github.com/nomisRev/arrow-ktor/tree/main?tab=readme-ov-file. More than happy to move it here as well. What do you think?

serras:
I really like your libraries! What do you think of just using this branch, and then merge everything?

About the naming, I really like that Ktor is very clear about things being client or server oriented, so I would suggest to use:
- `arrow-resilience-ktor-client` or `arrow-ktor-client-resilience`,
- `arrow-raise-ktor-server` or `arrow-ktor-server-raise` (I think "raise" is better than "core" here because it's all about the `Raise` context).

kyay10:
Calling it raise is also good if we want to eventually have a mini-library for Raise (likely when context params are out so we can get rid of Either extensions being inside the interface)

serras:
@nomisRev I've just integrated your Ktor libraries in this repo. I've taken the code mostly as it was, but I've updated it to the newest version of Arrow and Ktor, and made everything `common` code.
- There was a problem with using a `suspend` function type directly in Ktor Client, so I've added an intermediate `fun` interface.
- The type for routes in Ktor 3 uses `RoutingContext` instead of `PipelineCall`, and this was updated.
- The Ktor server uses the common-based API for charsets, instead of using the JVM one (so now the server is also Multiplatform).

serras:
@nomisRev if you are OK with these changes, feel free to simply merge this PR into `main`.

nomisRev (approved):

nomisRev:
All good @serras! I was working on parallel on this, but guess I was too slow 🐢

PR #3573, opened 2025-01-26
Keep stack traces in `withError`

### LINKED ISSUE #3351: ["Request"] better stack traces when tracing `withError`
When using `traced` with `withError` inside, the stacktrace will always point at the `withError` lambda, and won't include any information about where the error occurred originally. Adding some way to attach that extra information would be helpful so that the traces reflect the original error location

serras:
This is an attempt to fix #3351. The idea is to "inline" the definition of `withError`, and within it carefully copy stack traces. Since the "inner exception" already contains the stack trace including any `withError`, we just need to copy it, not change it in any way.

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
@@ -190,15 +190,16 @@ public inline fun <Error, OtherError, A> Raise<Error>.withError(
   val raise = DefaultRaise(false)
   return try {
     block(raise).also { raise.complete() }
-  } catch (e: RaiseCancellationException) {
+  } catch (innerException: RaiseCancellationException) {
     raise.complete()
-    val error = transform(e.raisedOrRethrow(raise))
-    when {
-      this is RaiseAccumulate -> raise(error)
-      e is Traced -> throw Traced(error, this as Raise<Any?>, e.cause).also {
-        it.copyStacktrace(e)
+    val error = transform(innerException.raisedOrRethrow(raise))
+    try {
+      raise(error)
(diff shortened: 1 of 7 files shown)


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
```diff
@@ -152,6 +153,60 @@ public inline fun <Error, A, B> fold(
...
+public inline fun <Error, OtherError, A> Raise<Error>.withError(
+  transform: (OtherError) -> Error,
+  @BuilderInference block: Raise<OtherError>.() -> A
+): A {
+  contract {
+    callsInPlace(block, EXACTLY_ONCE)
+    callsInPlace(transform, AT_MOST_ONCE)
+  }
+  // recover({ return block(this) }) { raise(transform(it)) }
+  val raise = DefaultRaise(false)
```
kyay10, line 190: Doesn't this mean that `innerException` would always have no stack trace?

serras: Why would that be the case? As I understand the inner stack trace comes from the point where the `RaiseCancellationException` is first thrown.

kyay10: Well, the inner exception is always going to be a `NoTrace` because it's created by a `DefaultRaise(false)`. Otherwise, we would be creating a stack trace every time, which would defeat our optimizations.

serras: That is definitely true!

serras:
It was too good to be true...

PR #3576, opened 2025-02-07
Rework Gradle setup

### LINKED ISSUE #3574: Latest version 2.0.1 adds new `android.permission.READ_PHONE_STATE` to client manifest
When I was trying to update our `io.arrow-kt:arrow-core` from version `2.0.0` to version `2.0.1` we noticed thanks to a CI check an unexpected change in the Android Manifest file, requiring a new permission not used before by our app `android.permission.READ_PHONE_STATE`.

Checking into the logs we see that indeed this is coming from the Arrow dependency:

```
uses-permission#android.permission.READ_PHONE_STATE
IMPLIED from ..app/src/main/AndroidManifest.xml:1:1-85:12 reason: arrow.core has a targetSdkVersion < 4
```

I see in the [latest changelog](https://github.com/arrow-kt/arrow/compare/2.0.0...2.0.1#diff-c4b4f09fab43bff523c1ed2c893f579a676468f122186b29281491dd1b646bfcR59) that now arrow-core is declaring an andorid compileSdk, but it is missing to declare the targetSdk. Maybe due to this reason the build process infers a default 0 value, producing this behaviour in the merged manifest in the client apps. I believe that a possible solution for this bug would be to specify a targetSdk.

```
android {
  namespace = "arrow.core"
  compileSdk = libs.versions.android.compileSdk.get().toInt()
}
```

serras:
**IMPORTANT**: do not merge this PR if #3577 is going to be merged!!!!

This moves almost every configuration to the convention plug-in, so that subprojects can simply do:

```kotlin
plugins {
  id("arrow.kotlin")
}
```

Apart from the additional uniformity across all projects, this PR also:
- Sets a minimum SDK of 21, as discussed in #3574
- Uses the recommended `kotlin("stdlib")` syntax instead of having a separate dependency in the catalog (this ensures we always depend on the libraries from the corresponding Kotlin version)
- Moves AnimalSniffer configuration from its own file into the convention plug-in
- Adds Android tests to the CI pipeline

Fixes #3574


serras:
Closed in favor of #3577

PR #3577, opened 2025-02-07
Add SuspendApp to the main repo

serras:
Note: this PR is based on #3576, so if this is merged the other should be closed.

This PR brings https://github.com/arrow-kt/suspendapp to the main Arrow repo (so we can say we have a monorepo now).

In addition, it updates the implementation in `suspendapp-ktor` to use a Multiplatform equivalent of getting the current directory (based on the same code in `ktor-server-core`).


nomisRev:
Hey @serras,

I am still ironing out some issues in SuspendApp, so this might be slightly premature.

serras:
Note: @nomisRev and I have agreed to merge this and move all the issues from SuspendApp into the this repo

PR #3588, opened 2025-02-22
Update secrets to sign and publish Arrow artifacts

franciscodr:
This pull request updates the secrets necessary for signing and publishing the Arrow artifacts in Maven. This change will enable Arrow maintainers to release new versions in Maven.


nomisRev (approved):
Thank you Paco!

PR #3589, opened 2025-02-26
fix: move exit handling outside runScope

### LINKED ISSUE #3583: [SuspendApp] JVM SupendApp never terminates
_Ported from https://github.com/arrow-kt/suspendapp/issues/140, issue by @weronkagolonka_

Hello!

After upgrading to `0.5.0`, we've noticed a strange behaviour where our Kafka consumer, launched in a separate coroutine, was abruptly stopped few seconds after startup. What's worth noticing, the application itself was not terminated or crashed. The kafka coroutine is running within the same scope of the suspended app and it is polling for messages as long as the jobs is active. Initially, it seemed like simply wrapping it in a new scope would solve the issue - and it did, as the consumer was successfully polling and handling the messages. However, after a deeper investigation to find the root cause of the issue we noticed that the application would never stop, regardless of what was put in the `SuspendApp` lambda. Given this extremely simple example:

```kotlin
fun main() = SuspendApp { logger.info { "The app is running!" } }
```

We discovered a significant change in the top-level coroutine launched in the `SuspendApp`, where after executing the lambda block, the process is being exited with status code 0:

```kotlin
@OptIn(ExperimentalStdlibApi::class)
fun SuspendApp(
  context: CoroutineContext = Dispatchers.Default,
  uncaught: (Throwable) -> Unit = Throwable::printStackTrace,
  timeout: Duration = Duration.INFINITE,
  process: Process = process(),
  block: suspend CoroutineScope.() -> Unit,
): Unit =
  process.use { env ->
    env.runScope(context) {
      val job =
        launch(start = CoroutineStart.LAZY) {
          try {
            block()
            env.exit(0) // <- here
          } catch (_: SuspendAppShutdown) {} catch (e: Throwable) {
            uncaught(e)
            env.exit(-1)
          }
        }
      val unregister =
        env.onShutdown {
          withTimeout(timeout) {
            job.cancel(SuspendAppShutdown)
            job.join()
          }
        }
      job.start()
      job.join()
      unregister()
    }
  }
```

This call triggers the sequence of JVM shutdown hooks, which then executes `env.OnShutdown` lambda defined in `unregister` variable, where the job is being cancelled (and that explains why our kafka consumer was suddenly stopped and why running it in a new scope allowed the consumer to proceed). The process then waits for the cancelled job to complete, but that never happens - the application basically hangs, unless a finite timeout duration is specified, then it simply terminates with timeout exception.

My question is - what is the rationale behind exiting the process after lambda execution instead of letting the job complete and eventually calling `unregister()` as it was done in version `0.4.0`? Right now it doesn't seem like the unregistering logic is ever executed, as the `exit(0)` method is called first.

If this behaviour is expected, how should the logic of `SuspendApp` be defined to make sure that the application actually terminates at some point?

Thanks in advance for your help, for now we're skipping the upgrade of the library until we get some more insight 😄

tKe:
if it's inside runScope, on the JVM it means we have the never-terminating `processExit` call within the job that the shutdown hook is attempting to join, creating a deadlock

close #3583

### DIFF
--- a/arrow-libs/suspendapp/suspendapp/api/suspendapp.klib.api
+++ b/arrow-libs/suspendapp/suspendapp/api/suspendapp.klib.api
@@ -7,9 +7,3 @@
 
 // Library unique name: <io.arrow-kt:suspendapp>
 final fun arrow.continuations/SuspendApp(kotlin.coroutines/CoroutineContext = ..., kotlin/Function1<kotlin/Throwable, kotlin/Unit> = ..., kotlin.time/Duration = ..., kotlin.coroutines/SuspendFunction1<kotlinx.coroutines/CoroutineScope, kotlin/Unit>) // arrow.continuations/SuspendApp|SuspendApp(kotlin.coroutines.CoroutineContext;kotlin.Function1<kotlin.Throwable,kotlin.Unit>;kotlin.time.Duration;kotlin.coroutines.SuspendFunction1<kotlinx.coroutines.CoroutineScope,kotlin.Unit>){}[0]
-
-// Targets: [wasmJs]
-final fun arrow.continuations/jsExit(kotlin/Int) // arrow.continuations/jsExit|jsExit(kotlin.Int){}[0]
-
-// Targets: [wasmJs]
-final fun arrow.continuations/processOn(kotlin/Function0<kotlin.js/Promise<kotlin.js/JsAny?>>) // arrow.continuations/processOn|processOn(kotlin.Function0<kotlin.js.Promise<kotlin.js.JsAny?>>){}[0]


arrow-libs/suspendapp/suspendapp/src/commonMain/kotlin/arrow/continuations/SuspendApp.kt
```diff
@@ -17,38 +17,55 @@ import kotlinx.coroutines.*
...
           }
+        job.invokeOnCompletion {
+          unregister()
+          jobCause.complete(it)
+        }
+        job.start()
+      }
+    } catch(e: Throwable) {
+        uncaught(e)
+        throw e
```
tKe, line 59: this throw will propagate only if `forceExit` is false - as otherwise the process should be shutdown before it will go anywhere

arrow-libs/suspendapp/suspendapp/src/commonMain/kotlin/arrow/continuations/SuspendApp.kt
```diff
@@ -17,38 +17,55 @@ import kotlinx.coroutines.*
...
+ *   finishes (code 0 on success or 255 on failure)
  * @param block the lambda of the actual application.
  */
-@OptIn(ExperimentalStdlibApi::class)
 public fun SuspendApp(
   context: CoroutineContext = Dispatchers.Default,
   uncaught: (Throwable) -> Unit = Throwable::printStackTrace,
   timeout: Duration = Duration.INFINITE,
   process: Process = process(),
+  forceExit: Boolean = true,
```
tKe, line 29: this is an escape-hatch for if you want to opt-out of the `env.exit` calls

we may or may not want to keep this in the PR

tKe: actually I'll remove this for now as it's a breaking change on the API

tKe:
🤔 this is broken for JS targets, as surprisingly `runScope` doesn't "block" on JS (it simply launches the coroutine)

tKe:
I've updated with another refactor attempt that seems to be working, at least for JVM/JS/MacOS.

I've created a harness https://github.com/tKe/suspendapp-runner that automates executing the three mentioned targets with a few scenarios:
- **delay:** wait a short amount of time before "normally" completing the SuspendApp
- **wait (and signal):** wait until cancellation, send SIGINT / SIGTERM
- **fail:** throw within SuspendApp 
- **childfail:** throw within a launched job within SuspendApp

verifying that a `resourceScope`-installed "cleanup" completes successfully in all scenarios and that the expected exit codes are used.

nomisRev (commented):
Hey @tKe,

I was also looking into this but I hadn't had a chance to raise a PR. You're all too fast. I've had a lot of administration in the last month while I was trying to come back to OSS 😅 I can raise my PR tonight, but I am still doing everything in `runScope`. I was using `async` instead of `launch`, and calling `exit` only after `unregister` happened. Since there is a race happening between `exit` and the shutdown hook.

I think the `startCoroutine` with `EmptyCoroutineContext` can be problematic, no? What is `join` suspend

tKe:
I used a `CompletableDeferred` to capture the result into as the JS `runScope` isn't blocking, and returns after launching the coroutine internally, meaning I needed a way to "join" to the result. I wasn't happy with the `suspend {}.startCoroutine` either, I suppose another usage of `runScope` would have worked to allow joining on the result and subsequent handling.

The reason for bringing the exit outside of the scope entirely was to allow for any other child jobs from `block` to be completed, or rather to allow for the entire scope to complete before actioning the exit. I was finding that a `launch { error() }` within the `SuspendApp` wouldn't allow for any cleanup to occur (see my tKe/suspendapp-runner repo that I used for testing these). 

I also noticed with these tests that the JS `runScope` impl would cancel the keepAlive before any jobs launched from `block` into the scope were completed. Changing the implementation to use `withContext(innerScope.coroutineContext, block)` seemed to satisfy this without requiring the `block` to nest a `coroutineScope`. 

All of that said, your PR is much tider but still seemed to fail on the child-fail scenario.

I've played with it and found that shifting the app launch/await into a nested supervisorScope mitigates it, additionally mitigatin the need to fix runScope as above (although this is probably desirable anyway as it's a public API, and misbehaves).

tKe:
Is it worth me walking back the API changes? 

I think it's clearer to indicate that `exit(code: Int)` shouldn't ever return with `: Nothing`
And the `onShutdown` unregister function has no (current) need to be suspend (I originally removed it to allow it to be called from within the `invokeOnCompletion`, but should we add it back if it's not needed?)

arrow-libs/suspendapp/suspendapp/src/jsMain/kotlin/arrow/continuations/Enviroment.js.kt
```diff
@@ -55,13 +56,16 @@ public object JsProcess : Process {
               delay(1.hours)
             }
           }
-        runCatching { block(innerScope) }.also { keepAlive.cancelAndJoin() }.getOrThrow()
+        runCatching { withContext(context, block) }
+          .also { keepAlive.cancelAndJoin() }
+          .getOrThrow()
       }
       .startCoroutine(Continuation(EmptyCoroutineContext) {})
```
tKe, line 63: should this be `.startCoroutine(Continuation(innerScope.coroutineContext) {})`?

nomisRev: `innerScope.coroutineContext == EmptyCoroutineContext`, I guess I thought using it directly was more readily.

nomisRev (approved):
Looks great @tKe! Nice work 🙌

PR #3592, opened 2025-03-01
Test and fix for Map.combine()

### LINKED ISSUE #3591: [Bug] Map.combine() does not preserve order when delegating to the nested monoid operation
This simple unit test demonstrates the issue:

```Kotlin
    @Test
    fun testArrowMapCombine() {
        val map1 = mapOf(1 to "A", 2 to "B")
        val map2 = mapOf(1 to "lpha", 2 to "eta")

        val result = map1.combine(map2, String::plus)

        assertThat(result).isEqualTo(mapOf(1 to "Alpha", 2 to "Beta"))
    }
```

It fails with:
```
expected: {1="Alpha", 2="Beta"}
but was : {1="lphaA", 2="etaB"}
```

Affected versions: 2.0.1 and 1.2.4.

Will provide a PR with a propery-based test that fails, along with a fix that makes it pass.

hristo-stefanov:
Fixes #3591

The existing test covers monoid laws but does not verify delegation to the nested monoid's `combine` function, which is why the bug went unnoticed.

This PR provides a property-based test that verifies the property of order-preservation when `Map.combine()` delegates. The test fails with the existing implementation. The PR also includes a fix that makes it pass.

I'm happy to refine the PR to better align with the project's style and conventions.


serras:
Thanks for both finding out and fixing this bug!

hristo-stefanov:
> Thanks for both finding out and fixing this bug!

Happy to contribute to such an excellent library! Thanks for accepting my contribution, @serras.

PR #3603, opened 2025-04-03
Race until one of the tasks succeeds

serras:


arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/RaceIf.kt
```diff
@@ -0,0 +1,41 @@
...
+
+// Based on https://medium.com/@sam-cooper/custom-kotlin-coroutine-races-77161141b2ac
+
+/**
+ * Races the participants [tasks] in parallel on the [Dispatchers.Default],
+ * until one of them returns a value which satisfies [condition].
+ * The winner of the race cancels the other participants.
+ * Cancelling the operation cancels all participants.
+ */
+public suspend inline fun <A, B> raceNUtil(
```
franciscodr, line 20: Is this a typo??
```suggestion
public suspend inline fun <A, B> raceUntil(
```

franciscodr: Additionally, could we rename the file to `RaceUntil.kt`?

nomisRev:
Hey @serras,
I had some ongoing work for a more generalised racing DSL. This is close-ish to being wrapped up, but I got hold-up during my work on this. I'd prefer to see a unification of all racing needs in a singular DSL is possible.

https://github.com/arrow-kt/arrow/pull/3411

serras:
I would rather prefer to have a DSL for the whole racing, if possible. This was just low-hanging fruit we could add to the library. I'll keep it open in case we want to reconsider in the future.

nomisRev:
I should have some time soon to work on those DSLs to include in the next minor release with the ktor module. I also need to make some time for that soon.

serras:
@nomisRev what do you think about merging this, since it's a very small change requested by our users?

kyay10 (approved):
Looks reasonable enough! A more generalised version would likely be nicer, but I can see this coming in handy for now!

PR #3606, opened 2025-04-13
`Raise` with context parameters

serras:
The time has finally come to provide `Raise` with context parameters! Note that if we publish this library, users still need to [use `-Xskip-prerelease-check`](https://youtrack.jetbrains.com/issue/KT-38070) in their build file.

This PR is the start of the conversation of which functions we should expose using context-parameters-style. In this case I've decided to go with the following idea: minimize the amount of additional functions, while allowing people to freely define their functions as:

```kotlin
context(_: Raise<Something>) fun foo() = ...
```

More concretely, this implies:

- No "runner functions" (`either`, `option`, etcetera) with contextual functions. The current functions work still OK with contextual parameters, since the receiver is part of context parameter resolution,
- (Finally!) expose `bindAll` and `mapOrAccumulate` with the iterable/sequence/map being a receiver.


kyay10:
Thanks for pinging me on this! I'll have a look really soon! I had some ideas in a repo called arrow-context, but that's out-of-date now since it was using context receivers

arrow-libs/core/arrow-raise-contextual/src/commonMain/kotlin/arrow/core/raise/contextual/Raise.kt
```diff
@@ -0,0 +1,125 @@
...
+import kotlin.contracts.contract
+import kotlin.experimental.ExperimentalTypeInference
+import kotlin.jvm.JvmName
+import arrow.core.raise.Raise as CoreRaise
+import arrow.core.raise.SingletonRaise as CoreSingletonRaise
+import arrow.core.raise.ResultRaise as CoreResultRaise
+
+public typealias Raise<Error> = CoreRaise<Error>
+public typealias SingletonRaise<Error> = CoreSingletonRaise<Error>
+public typealias ResultRaise = CoreResultRaise
```
tKe, line 21: Is the purpose of these to allow consumers to use only `import arrow.core.raise.contextual.*` and avoid clashes with `arrow.core.raise.*`?

arrow-libs/core/arrow-raise-contextual/src/commonMain/kotlin/arrow/core/raise/contextual/Raise.kt
```diff
@@ -0,0 +1,125 @@
...
+}
+
+context(raise: Raise<Error>) @RaiseDSL public inline fun <Error, B : Any> ensureNotNull(value: B?, otherwise: () -> Error): B {
+  contract { returns() implies (value != null) }
+  return raise.ensureNotNull(value, otherwise)
+}
+
+context(raise: Raise<Error>) @RaiseDSL public inline fun <Error, OtherError, A> withError(
+  transform: (OtherError) -> Error,
+  @BuilderInference block: CoreRaise<OtherError>.() -> A
```
tKe, line 38:
```suggestion
  @BuilderInference block: Raise<OtherError>.() -> A
```
better to use the alias for consistency?

arrow-libs/core/arrow-raise-contextual/src/commonMain/kotlin/arrow/core/raise/contextual/RaiseAccumulate.kt
```diff
@@ -0,0 +1,56 @@
+@file:OptIn(ExperimentalTypeInference::class)
+
+package arrow.core.raise.contextual
+
+import arrow.core.NonEmptyList
+import arrow.core.raise.*
+import kotlin.experimental.ExperimentalTypeInference
+
```
tKe, line 8: Worth also bringing in a typealias for `RaiseAccumulate`?

tKe (commented):
🙌  Looks like a great first step towards supporting context-parameters. 

Is it worth taking the same approach with another module for `fx` contextual?

Worth adding in some bridges for some of the `RaiseAccumulate` methods? It's not currently possible to use:
```kotlin
context(_: RaiseAccumulate<String>)
private fun example() {
  val a by accumulating { }
}
```

(although this would require making the `getValue` operator available outside of `RaiseAccumulate`)

nomisRev:
Okay... I need to refactor my project 😂

FooIbar:
> Note that if we publish this library, users still need to [use `-Xskip-prerelease-check`](https://youtrack.jetbrains.com/issue/KT-38070) in their build file.

It seems that context parameters will not force pre-release binaries.
https://github.com/JetBrains/kotlin/blob/6f1bfe9fa11be40ff96ca01b8a55dfc5dde67f47/compiler/util/src/org/jetbrains/kotlin/config/LanguageVersionSettings.kt#L485

serras:
Since context parameters will not poison the binary, maybe we should just include these functions in `arrow-core`? WDYT @nomisRev?

nomisRev:
Yes, sounds good to me!

Zordid:
Could it be that in your contextual bridge function you did not put all of the contract definition as done by the Raise member functions? I.e.: `ensureNotNull` misses the invocation contract for the otherwise lambda.

serras:
Closed in favor of #3646

PR #3609, opened 2025-04-20
Try to configure signing in Android

serras:
This will hopefully fix the release problems we're having, as shown in this [failed GitHub workflow](https://github.com/arrow-kt/arrow/actions/runs/14438670187/job/40484090661)


tKe:
I'm not sure this is needed as the [failure](https://scans.gradle.com/s/4oso3v7igl24e/failure) seems to be from loading the wrong value rather than the value not being found/loaded correctly:

> **The :arrow-annotations:signAndroidReleasePublication task failed.**
> Could not read PGP secret key
> \> The key ID must be in a valid form (eg 00B5050F or 0x00B5050F), given value: 6CE63E2E1685215A995BB3020B99A4AB5DB59AA2

PR #3617, opened 2025-04-30
(minor): Slightly Optimize Map.getOrNone

kyay10:
I based this implementation on `getOrElseNullable`, which is an internal method in the stdlib. As you can see, it only calls `containsKey` if the value is null, which makes it slightly faster for non-nullable values, and the same speed-ish for nullable ones (conjecture, but clearly it doesn't recalculate hashcode for non-nullable values, which is good, and the only extra instruction for nullable values is just the null check. Realistically, the null check is irrelevant compared to the object allocation, but the hashcode recalculation definitely isn't, since it can be expensive depending on the data structure).


nomisRev (approved):
LGTM! Thank you @kyay10

PR #3622, opened 2025-05-06
fix: remove nullability from `OrThrow` method returns

tKe:


nomisRev (approved):
This is a non-binary breaking, right? Additionally it was only introduced in 2.1.0.

tKe:
It was only added in #3611 last week and released in 2.1.1

No change to the JVM ABI as I understand it, not sure if there's an notable impact on non-JVM targets but I expect any blast radius to be very small as I wouldn't expect much if any usage of the new `OrThrow` methods yet - especially as the nullable return reduces the benefit of the new method compared to the existing `OrNull` methods.

tKe:
I'll merge this as is - @serras feel free to revert it if you have concerns on the binary compat.

PR #3646, opened 2025-06-30
Raise with context parameters

serras:
This PR supersedes #3606. The difference is that the contextual functions are now directly in `arrow-core`, instead of creating a new module. This seems to be safe, since context parameters do not poison the binary.


mrf7:
## Resolution ambiguity in higher order functions 
I pulled in `Raise.kt` from the original PR into a project to start playing with context parameters for a better api and I ran into an issue when writing a utility higher order function that logs the result of an operation. 

If you have a function that takes a block with a raise context, then you call that function in an existing builder like `either` that provides a raise context by taking a block with a Receiver, the resolution becomes ambiguous between the context parameter version and the receiver from the outer block for things like `raise`, `ensure`, etc which can cause you to raise on the receiver in the `either` block rather than the context param of the helper function

```kotlin
context(_: Raise<E>)
inline fun <E, T> withLoggingRaise(
  logger: Logger,
  context: String,
  operation: context(Raise<E>) () -> T,
) = either {
  logger.operationStart(context)
  operation()
}.fold(
  ifLeft = {
    logger.operationFailure(context, it.toString())
    raise(it)
  },
  ifRight = {
    logger.operationSuccess(context)
    it
  },
)

val x = either {
    withLoggingRaise(logger, "operation") {
       // If you select the non context param version of this for autocomplete and/or 
       // dont explicitly import it separately, this will be the `Raise<E>.raise() 
       // variant using `this` from either block instead of context(Raise<E>) from withLoggingRaise
       // which will prevent the things in fold from being called
       raise("error") 
       // Doing this or making sure to add the correct import will raise from the `operation` block in withLoggingRaise
       com.example.mrf.raise("error") 
    }
}
```

This wont be an issue if/when context params are fully released and arrow can deprecate/remove the receiver param variant, but I'm not sure what a good solution is in the (probably lengthy) interim

UPDATE: 
Found a workaround for this. Changing the parameter from operation: context(Raise<E>) to  operation: Raise<E>.()->T  makes it so only the extension function variant is available within the lambda so it removes all the function resolution ambiguity which ensures raise(...) will always be called on the innermost available raise context. Assuming thats the reason the `withError` here uses `block: Raise<OtherError>.()` rather than `block: context(Raise<OtherError>)`. Ultimately feels like mostly a language issue of extension functions taking priority over context parameter functions when trying to decide what function to call. 

UPDATE 2: 
Turns out the base issue is mostly by design because a member function always takes precedence over any possible top level functions https://kotlinlang.slack.com/archives/C7L3JB43G/p1751473900927029?thread_ts=1751472370.793829&cid=C7L3JB43G

serras:
It turns out that to prevent ambiguity, it's better to keep a completely separate contextual-first set of functions instead of mixing receivers and context parameters. This is done in the new `arrow.core.context` package.

@mrf7 this is an issue we're working on in the Kotlin team, we're working on [new reporting](https://youtrack.jetbrains.com/issue/KT-78866) to ensure resolution works as expected.

mrf7:
> It turns out that to prevent ambiguity, it's better to keep a completely separate contextual-first set of functions instead of mixing receivers and context parameters. This is done in the new `arrow.core.context` package.
> 
> @mrf7 this is an issue we're working on in the Kotlin team, we're working on [new reporting](https://youtrack.jetbrains.com/issue/KT-78866) to ensure resolution works as expected.

Thanks for the reply and for working on this both in arrow and internally at JB. I think a compiler warning to let you know when what you might expect is being called is actually being shadowed is a great solution

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/RaiseContext.kt
```diff
@@ -0,0 +1,132 @@
...
+}
+
+context(raise: Raise<Error>) @RaiseDSL public inline fun <Error, B : Any> ensureNotNull(value: B?, otherwise: () -> Error): B {
+  contract { returns() implies (value != null) }
+  return raise.ensureNotNullExt(value, otherwise)
+}
+
+context(raise: Raise<Error>) @RaiseDSL public inline fun <Error, OtherError, A> withError(
+  transform: (OtherError) -> Error,
+  @BuilderInference block: Raise<OtherError>.() -> A
```
mrf7: Should these blocks be `context(Raise<OtherError>` instead like in the added builders functions to reduce resolution ambiguity?

serras: Yes, thanks for the catch, everything should be a context parameter.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -990,6 +990,7 @@ public open class RaiseAccumulate<Error>(
     }
   }
 
+  @Suppress("NOTHING_TO_INLINE")
   public inline operator fun <A> Value<A>.getValue(thisRef: Nothing?, property: KProperty<*>): A = value
```
tKe, line 995: This may need to be moved to allow access when RaiseAccumulate isn't available as a dispatch receiver.

serras: Unfortunately `getValue` is not supported with context parameters (yet).

serras: However, I found a way to turn it into a member of `Value<A>` itself, so we can keep the API.

kyay10: It doesn't actually need the dispatch/context, so it can just exist at the top-level
Edit: or yeah it can just be in `Value` lol!

serras:
This PR slightly touches `RaiseAccumulate.Value`, but this was still marked as experimental, so we can have that small change.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/Builders.kt
```diff
@@ -0,0 +1,24 @@
...
+import arrow.core.None
+import arrow.core.Option
+import kotlin.experimental.ExperimentalTypeInference
+import kotlin.jvm.JvmMultifileClass
+import kotlin.jvm.JvmName
+
+public inline fun <Error, A> either(@BuilderInference block: context(Raise<Error>) () -> A): Either<Error, A> =
+  arrow.core.raise.either(block)
+
+public inline fun <A> nullable(block: context(SingletonRaise<Nothing?>) () -> A): A? =
```
kyay10: Let's expand `SingletonRaise<A>` into `Raise<A>` and `Raise<Unit>`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/Builders.kt
```diff
@@ -0,0 +1,24 @@
...
+  arrow.core.raise.nullable(block)
+
+public inline fun <A> result(block: context(ResultRaise) () -> A): Result<A> =
+  arrow.core.raise.result(block)
+
+public inline fun <A> option(block: context(SingletonRaise<None>) () -> A): Option<A> =
```
kyay10: Same here.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/RaiseContext.kt
```diff
@@ -0,0 +1,132 @@
...
+import arrow.core.raise.ensureNotNull as ensureNotNullExt
+import arrow.core.raise.withError as withErrorExt
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.contract
+import kotlin.experimental.ExperimentalTypeInference
+import kotlin.jvm.JvmMultifileClass
+import kotlin.jvm.JvmName
+
+public typealias Raise<A> = arrow.core.raise.Raise<A>
+public typealias SingletonRaise<A> = arrow.core.raise.SingletonRaise<A>
```
kyay10, line 24: We no longer need `SingletonRaise<A>` and instead we should use `Raise<A>` + `Raise<Unit>`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/RaiseContext.kt
```diff
@@ -0,0 +1,132 @@
...
+public typealias ResultRaise = arrow.core.raise.ResultRaise
```
kyay10, line 25: Similarly, this should just be `Raise<Throwable>`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/RaiseContext.kt
```diff
@@ -0,0 +1,132 @@
...
+  with(raise) { bind() }
+
+@Suppress("WRONG_IMPLIES_CONDITION")
+context(raise: Raise<Error>) @RaiseDSL public fun <Error, A> Either<Error, A>.bind(): A {
+  contract { returns() implies (this@bind is Either.Right) }
+  return with(raise) { this@bind.bind() }
+}
+
+@Suppress("WRONG_IMPLIES_CONDITION")
+context(raise: SingletonRaise<Error>) @RaiseDSL public fun <Error, A> Option<A>.bind(): A {
```
kyay10, line 55: Thus this should be `raise: Raise<Unit>`. Also `Error` is superfluous here

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/RaiseContext.kt
```diff
@@ -0,0 +1,132 @@
...
+  contract { returns() implies (this@bind is Some) }
+  return with(raise) { this@bind.bind() }
+}
+
+@Suppress("WRONG_IMPLIES_CONDITION")
+context(raise: SingletonRaise<Error>) @RaiseDSL public fun <Error, A> A?.bind(): A {
```
kyay10, line 61: Same as above: use `Raise<Unit>`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/RaiseContext.kt
```diff
@@ -0,0 +1,132 @@
...
+  contract { returns() implies (this@bind != null) }
+  return with(raise) { this@bind.bind() }
+}
+
+context(raise: ResultRaise) @RaiseDSL public fun <A> Result<A>.bind(): A {
```
kyay10, line 66: Either use `Raise<Throwable>` or update the typealias as above

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/RaiseContext.kt
```diff
@@ -0,0 +1,132 @@
...
+
+context(raise: Raise<Error>) @RaiseDSL @JvmName("bindAllEither")
+public fun <Error, A> NonEmptyList<Either<Error, A>>.bindAll(): NonEmptyList<A> =
+  with(raise) { bindAll() }
+
+context(raise: Raise<Error>) @RaiseDSL @JvmName("bindAllEither")
+public fun <Error, A> NonEmptySet<Either<Error, A>>.bindAll(): NonEmptySet<A> =
+  with(raise) { bindAll() }
+
+context(raise: SingletonRaise<Error>) @RaiseDSL @JvmName("bindAllOption")
```
kyay10, line 86: Similar idea for all of these. Just replace with `Raise<Unit>`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/RaiseContext.kt
kyay10: In fact, this typealias can just be `public typealias SingletonRaise<A> = Raise<Unit>` to aid in migration

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/Builders.kt
```diff
@@ -0,0 +1,24 @@
...
+
+package arrow.core.raise.context
+
+import arrow.core.Either
+import arrow.core.None
+import arrow.core.Option
+import kotlin.experimental.ExperimentalTypeInference
+import kotlin.jvm.JvmMultifileClass
+import kotlin.jvm.JvmName
+
```
kyay10, line 17: provide `singleton`, `ior`, `iorNel`, `impure` builders too

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -171,8 +171,11 @@ public class SingletonRaise<in E>(private val raise: Raise<Unit>): Raise<E> {
 
   @RaiseDSL
   public fun <A> Option<A>.bind(): A {
-    contract { returns() implies (this@bind is Some<A>) }
-    return getOrElse { raise() }
+    contract { returns() implies (this@bind is Some) }
+    when (this) {
```
kyay10: nit: `ensure(this is Some); return value` is a bit neater!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -986,21 +986,30 @@ public open class RaiseAccumulate<Error>(
...
+  @Deprecated(message = "Deprecated in favor of member", level = DeprecationLevel.HIDDEN)
   public inline operator fun <A> Value<A>.getValue(thisRef: Nothing?, property: KProperty<*>): A = value
 
-  public sealed interface Value<out A> {
-    public val value: A
+  public sealed class Value<out A> {
+    public abstract val value: A
+
+    @Suppress("NOTHING_TO_INLINE")
+    public inline operator fun getValue(thisRef: Nothing?, property: KProperty<*>): A = value
```
kyay10, line 1001: I believe the `inline` optimization for `getValue` only works if the method is extension. Will find a link to corroborate

kyay10 (changes requested):

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/Builders.kt
serras: I don't think that's the right approach, even if conceptually true. If we do so, then a lot of functions which do not fix their parameter are going to give us resolution ambiguity.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/RaiseContext.kt
serras: `SingletonRaise` adds some additional `bind` overloads for nullable and `Option` types, so I don't think we can really do this...

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
serras: I think the error is reported for every `inline` function without any functional argument.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/Builders.kt
kyay10: Hmmm. I think then we should just use `Raise<Unit>`

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/RaiseContext.kt
kyay10: Those overloads are exactly possible with `Raise<Unit>`, we just need to add them too by using contexts.

kyay10:
I don't want to hijack your PR @serras, but I also really care about this feature. Would you mind if I made some reasonable changes directly to the branch? Especially with `SingletonRaise`, I want to put my vision for it into code, and see if you and Simon like it or not.

serras:
I am fine with comparing approaches. However, I don’t think that removing `SingletonRaise` should be done as part of this PR. Let’s first try to merge a one-to-one copy of the API and then let’s discuss whether we should remove the special Raise from the library.

kyay10:
Oh fine yeah. If one-to-one copy is the goal, then you've done it!

kyay10 (approved):

kyay10:
Just note that I'd like to relook at this and finalise our API for contexts before publishing a version and committing to binary and source compatibility (that's my main concern really. I didn't want a situation where people may have broken builds because we regret the exact signature of the API or something.

@serras if you're happy with the PR as-is then we should merge now. Alternativey we can wait for Simon to have a look. Up to you!

PR #3652, opened 2025-07-08
Make AcquireStep delegate to the calling ResourceScope

kyay10:
Also deprecate AcquireStep so it may be removed in the future

This preserves source and binary compatibility perfectly, but is a little overkill. We could instead deprecate-and-hide `install` and introduce a non-AcquireStep version of it. This would technically break source compatibility if anyone used `this` to refer to `AcquireStep`, but I find that very very unlikely. 
I know nothing about klib compatibility btw, and it's showing a break there, but I'm not sure what that entails.


arrow-libs/fx/arrow-fx-coroutines/api/arrow-fx-coroutines.klib.api
```diff
@@ -144,8 +155,6 @@ sealed class arrow.fx.coroutines/ExitCase { // arrow.fx.coroutines/ExitCase|null
     }
 }
 
-final object arrow.fx.coroutines/AcquireStep // arrow.fx.coroutines/AcquireStep|null[0]
-
```
nomisRev, line 148: I think is definitely a breaking change. There is no equivalent in the added code in klib.api.

serras:
I'm not entirely sure of what the goal of this change is, to be honest. Apart from the binary compatibility, I'm a bit concerned about making `AcquireStep` a subclass of `ResourceScope`, because now when you're working on `install` you potentially have two `ResourceScope`s in scope.

What about the (seemingly simpler) option of simply providing:

- a renamed version of the current install called `installWithAcquire`,
- make the current version hidden deprecated (so we keep binary compatibility),
- have a new version of `install` without `AcquireStep` in its signature.

This means the only people who potentially need to migrate are those using `AcquireStep` -- but that type has no members! If they really really need the receiver, then they can call `installWithAcquire`.

kyay10:
Yeah I'm realising this solution is way too overkill. In fact, the simple reality is that just removing the DSL marker annotation from `AcquireStep` is sufficient for my goal here, which is to allow registering more resources inside `install`. I talked to @nomisRev about this before, and we're rather certain that allowing that doesn't break anything.
I'll close this PR and replace it with a way simpler one.

PR #3654, opened 2025-07-10
Unmark AcquireStep to allow installing resources inside `install`

kyay10:
Also showcase this by updating a test to call `Resource.bind` inside `install`

I discussed this with @nomisRev a while ago. We're rather sure that `AcquireStep` preventing installation inside `install` is completely unnecessary.


serras:
@kyay10 could you please include a test just to ensure that no error is reported if you don't use the `AcquireStep` inside the lambda of the install?

kyay10:
I think this new test addresses your concern? It shows that nested `install` function just fine, thus `AcquireStep` doesn't block us from accessing outer `ResourceScope`

serras:
@kyay10 Yes, thanks for adding it :)

serras (approved):

PR #3656, opened 2025-07-11
Improve performance of `RaiseCancellationException` for JS platform

sgrishchenko:
### Problem

It is well-known information that a collection of stacks for `Exception` instances is an expensive operation. Actually, custom implementation of `NoTrace` for JVM platform also proves this. In this PR I tried to provide a similar hack for the JS platform.

### Implementation
To prevent stack collecting for `Error` in the JS platform, I create an instance of a plain JS object and try to pretend that it is an instance of `RaiseCancellationException` using prototype assignment. This hack should not be harmful, and all existing unit tests should pass and prove it.

### Use case
I tried to use Arrow `nullable` DSL in a code-converting tool for [Kotlin Wrappers](https://github.com/JetBrains/kotlin-wrappers). After adding some number of plugins that use `nullable` DSL I noticed some performance degradation. I assumed that the problem is caused by a stack collection in Errors, so I created a special branch [arrow-performance](https://github.com/JetBrains/kotlin-wrappers/tree/arrow-performance) to test this hypothesis and provide simutation of proposed changes. This performance degradation can be checked with command `./gradlew :kotlin-node-karakum:generateKarakumExternals`. In the `master` branch converting takes about 1 min. and 40 sec. on my machine, and in the `arrow-performance` branch it takes about 17 sec.


kyay10:
I've been curious on how to do this for a while. Even kotlinx coroutines doesn't do this on JS, only JVM, and I couldn't find any useful source online on how to do it on JS generally. Thanks for the PR! I'll have a look really soon!
I wonder if we can do this for Native and WASM too...

sgrishchenko:
@kyay10 I know about WASM platform very little :slightly_smiling_face: but I would like also to do this for this platform. If changes that I provided are helpful, maybe I will try to investigate how to achieve it later.

arrow-libs/core/arrow-core/src/jsMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
```diff
@@ -0,0 +1,17 @@
+package arrow.core.raise
+
+import js.reflect.Reflect
+
+@OptIn(DelicateRaiseApi::class)
+internal actual fun NoTrace(raised: Any?, raise: Raise<Any?>): RaiseCancellationException {
+  val surrogateError = RaiseCancellationExceptionLike(raised, raise)
```
kyay10, line 7: do we even need this, or could we just use the `js("")` insertion syntax? I'm not opposed to the plugin, but since it's just used only here, I feel we can just use raw JS instead, no?

arrow-libs/core/arrow-core/src/jsMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
```diff
@@ -0,0 +1,17 @@
...
+
+  val cancellationExceptionPrototype = RaiseCancellationException::class.js.asDynamic().prototype
+  Reflect.setPrototypeOf(surrogateError, cancellationExceptionPrototype)
```
kyay10: Since this returns a boolean, maybe add a require call around it?
```suggestion
  require(Reflect.setPrototypeOf(surrogateError, cancellationExceptionPrototype)) { "Unknown error occured while setting the prototype of constructed JS RaiseCancellationException. Please report this to the Arrow maintainers!" }
```

arrow-libs/core/arrow-core/src/jsMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
kyay10, line 7: Could we just use [Object.create](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/create) instead? seems like the ideal choice to me!

buildSrc/src/main/kotlin/arrow.kotlin.gradle.kts
```diff
@@ -165,15 +165,24 @@ if (isKotlinMultiplatform) {
       val nonJvmMain by creating { dependsOn(commonMain.get()) }
       val nonJvmTest by creating { dependsOn(commonTest.get()) }
 
+      val nonJvmAndJsMain by creating { dependsOn(commonMain.get()) }
```
kyay10: I think this should depend on `nonJvmMain`, same for the test one

arrow-libs/core/arrow-core/src/jsMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
sgrishchenko: Yes, we can, it is up to you, if you prefer raw js syntax, I am ok with it. Only one thing that can be dangerous a bit, it is that code inside js string is not statically analyzed. It may lead at least to two types of problems:
* You can quite easily make a typo inside an js block
* You can see false negative warnings in IDE about unused variables because inspections can't recognise that these variables are used in js block.

sgrishchenko: Actually, yes, it could make code even more performant. I thought that external interface will minimize a change of discrepancies in the structural contract between a plain JS object and `RaiseCancellationException`, but if you think it is worth it, we can try. Anyway, we reassign the prototype on the next step.

buildSrc/src/main/kotlin/arrow.kotlin.gradle.kts
sgrishchenko: It seems I have some strange circular dependency in tasks for WASM in this PR. I am thinking about removing this awkward  `nonJvmAndJsMain` target and adding an actual implementation for each platform. Maybe it will make code more transparent and straightforward.

arrow-libs/core/arrow-core/src/jsMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
sgrishchenko: Done

buildSrc/src/main/kotlin/arrow.kotlin.gradle.kts
sgrishchenko: So, I tried to provide an independent implementation for each platform (or just copied code one more time :slightly_smiling_face:) and I was able to get a successful run of `./gradlew jsTest` locally (which failed previously in CI). Maybe it would be better to stick with this approach? Also, there is some non-zero chance that it will be possible to provide similar hacks for other platforfac too... :slightly_smiling_face:

arrow-libs/core/arrow-core/src/jsMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
sgrishchenko: Example of implementation with `Object.create`:

```kotlin
@OptIn(DelicateRaiseApi::class)
internal actual fun NoTrace(raised: Any?, raise: Raise<Any?>): RaiseCancellationException {
  val cancellationExceptionPrototype = RaiseCancellationException::class.js.asDynamic().prototype

  val surrogateError = Object.create(cancellationExceptionPrototype)

  surrogateError.raised = raised
  surrogateError.raise = raise
  surrogateError.message = RaiseCancellationExceptionCaptured

  return surrogateError
}
```

You should be careful with this code because almost all this code is based on `dynamic` type (analog of TS `any`). Personally, I spent about half an hour to detect that I made a typo and wrote `raised` instead of `raise` in the second property assignment.

arrow-libs/core/arrow-core/api/arrow-core.klib.api
```diff
@@ -1023,3 +1023,9 @@ final suspend fun <#A: kotlin/Any?> (kotlin.coroutines/SuspendFunction1<arrow.co
 final suspend fun <#A: kotlin/Any?> (kotlin.coroutines/SuspendFunction1<arrow.core.raise/Raise<kotlin/Nothing>, #A>).arrow.core.raise/get(): #A // arrow.core.raise/get|get@kotlin.coroutines.SuspendFunction1<arrow.core.raise.Raise<kotlin.Nothing>,0:0>(){0§<kotlin.Any?>}[0]
 final suspend fun <#A: kotlin/Any?> (kotlin.coroutines/SuspendFunction1<arrow.core.raise/Raise<kotlin/Throwable>, #A>).arrow.core.raise/toResult(): kotlin/Result<#A> // arrow.core.raise/toResult|toResult@kotlin.coroutines.SuspendFunction1<arrow.core.raise.Raise<kotlin.Throwable>,0:0>(){0§<kotlin.Any?>}[0]
 final suspend inline fun <#A: kotlin/Any?, #B: kotlin/Any?> (kotlin.coroutines/SuspendFunction1<arrow.core.raise/Raise<#A>, #B>).arrow.core.raise/getOrElse(kotlin/Function1<#A, #B>): #B // arrow.core.raise/getOrElse|getOrElse@kotlin.coroutines.SuspendFunction1<arrow.core.raise.Raise<0:0>,0:1>(kotlin.Function1<0:0,0:1>){0§<kotlin.Any?>;1§<kotlin.Any?>}[0]
+
```
kyay10, line 1123: Not sure why this shows up in API. This might be a good argument against using the js plain objects plugin

kyay10 (approved):

arrow-libs/core/arrow-core/api/arrow-core.klib.api
sgrishchenko: As far as I can see, it is functions that are responsible for
* Creation JS plain object (`RaiseCancellationExceptionLike.invoke(...)` or just `RaiseCancellationExceptionLike(...)`).
* Copying JS plain object (`RaiseCancellationExceptionLike.copy(...)` like for data classes).

I made some experiments, and it seems that if your class or interface is internal, you don't have access to its companion object outside the containing module.

`RaiseCancellationExceptionLike` is declared as internal, so effectively users won't be able to call described functions. Do I understand correctly that in this case these functions should not be added to the considered file? If so, I see next risks:
* This API can be changed by the Kotlin compiler team (I think it's unlikely).
* This API can be changed if `RaiseCancellationExceptionLike` is changed, so some compatibility tests can fail, but I think these fails can be suppressed as false-negative.

If these two things bother you, I am ok to implement it using `js("")` block, just let me know.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
```diff
@@ -271,14 +271,16 @@ public annotation class DelicateRaiseApi
...
 @DelicateRaiseApi
-public sealed class RaiseCancellationException(
-  internal val raised: Any?,
+public expect sealed class RaiseCancellationException(
+  raised: Any?,
+  raise: Raise<Any?>
+) : CancellationException {
+  internal val raised: Any?
   internal val raise: Raise<Any?>
-) : CancellationException(RaiseCancellationExceptionCaptured)
```
serras, line 277: This seems very suspicious. We need to make `RaiseCancellationException` a subclass of `CancellationException` so everything works as intended with coroutines. @nomisRev WDYT?

arrow-libs/core/arrow-core/src/jsMain/kotlin/arrow/core/NonFatal.kt
serras, line 1: Why is this no longer common code? As far as I can see, the code is almost equal in every platform.

sgrishchenko: The main goal of this PR is making `RaiseCancellationException` more performant on JS platform. Unfortunately, I don't know how to do it for native and WASM platforms. For JS platform I used `unsafeCast` to ignore that created object is not fully legal instance of `RaiseCancellationException`. This trick is not harmful and quite safe only on JS platform (mostly because of prototype assignment). And actually `unsafeCast` is not available on other platforms (there is some `unsafeCast` in WASM but it is mostly for JS <-> WASM interop). So, to achieve this I provided custom implementation for JS platform (as it is done for JVM & Android). For native and WASM target I saved previous conservative and not performant implementation. Custom implementation for `RaiseCancellationException` in JS platform is required mostly because of this code:
```kotlin
  internal actual val raised: Any?
    get() = unsafeCast<RaiseCancellationExceptionLike?>()?.raised ?: _raised

  internal actual val raise: Raise<Any?>
    get() = unsafeCast<RaiseCancellationExceptionLike?>()?.raise ?: _raise
```
This code is written in this way because there is internal way to create `RaiseCancellationException` using plain JS object + prototype assignment. So in runtime, instance of `RaiseCancellationException` can be either something like this:
```js
const e = { raised: ...,  raise: ... } // plain JS object that pretends that it is instance of RaiseCancellationException
```
or something like this:
```js
const e = new RaiseCancellationException(...) // fully legal RaiseCancellationException instance
```

Kotlin/JS mangles names of class members, so if you write something like this in Kotlin:
```kotlin
class MyClass {
  fun myMethod() {}
}
```
you will get somethin like this in compiled JS:
```js
class MyClass {
  myMethod_3ty2ghruyhg87t() {}
}
```
So if I create some plain JS object `const x = { myMethod: () => {} }` and write in Kotlin `x.unsafeCast<MyClass>().myMethod()`, it will be compiled to this code `x.myMethod_3ty2ghruyhg87t()`. If I can't affect generated contract of `RaiseCancellationException` then I can affect implementation of `RaiseCancellationException` and check maybe instance reference (`this`) contains some plain unmangled fields and if it is so I return them, otherwise I use regular values that was provided in the constructor.

I hope my explanation is not very complicated. :slightly_smiling_face:

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
sgrishchenko: This line is written exactly to provide this contract:
image
Because it is `expect`, I can't provide implementation anymore. Super constructor call is implementation, so this code is written in `actual` declarations now.

kyay10:
Please have a look at #3670 when you can. It's a minimal PR that should achieve the same thing!

sgrishchenko:
Thank you for your PR #3670. 
* ~I liked the idea of using the Flyweight pattern, but I am not sure it is really safe. I tried to express my concerns in the PR comments. However, if you are really sure that semantic of `CancellationException` suppose that there won't be two instances of such exception at the same time, I can try to apply this optimization in my code too.~ (Irrelevant)
* I am really afraid of using field names that compiler generated, I think my approach is much more safe and controllable. Yes, you need to provide expext/actial for `RaiseCancellationException` too, but at least all manipulations are made explicitly.

After reviewing your PR, I applied the same configuration for Gradle to reduce the number of affected files. Sorry that I didn't get how to achieve this from the beginning.

PR #3662, opened 2025-07-16
feat: Ktor Raise call.respondOrRaise

tKe:
A much simplified Ktor/Raise integration that only looks to provide a short-circuiting response function using `Raise`.


serras:
I really like this direction of a minimal integration between Ktor and Raise 👍

tKe:
@serras @nomisRev any preference on approach between this (using `Raise<Response>` as a receiver and an intersection type `RaiseRoutingContext` in lieu of context-parameters) or embracing context parameters entirely for this API (only using `context(Raise<Response>)`, no receiver usage)?

kyay10:
Contexts are close to being fully in the language, and so I'd prefer to just go with contexts right away!

PR #3665, opened 2025-07-19
Experimental: Unify IorRaise and RaiseAccumulate APIs

kyay10:
I didn't touch `IorRaise` itself because it only has the ability to accumulate errors, but not the ability to raise the accumulated errors when necessary. The effects encompassed by `RaiseAccumulate` are precisely `Raise` and (newly added) `Accumulate`. `Accumulate` allows for accumulating a `Nel<Error>`, and gives you a delayed thunk to short-circuit the computation. It also allows you to check if any errors have been accumulated, and to short-circuit if so (so that e.g. an inner computation can accumulate errors, then you can check and short-circuit). This is used mainly in `forEachAccumulatingImpl` and `accumulating`.


serras:
The idea looks interesting, and definitely unification of `Accumulate` and `Ior` is welcome :)

On the code style, do you think some of the `object : ...` could be dropped in favor of regular classes? The reason I prefer the latter is because local classes may unadvertedly capture some piece of information, and I strongly prefer to have dependencies very clearly spelled out in the code.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
```diff
@@ -143,6 +143,44 @@ public inline fun <Error, A> iorNel(noinline combineError: (NonEmptyList<Error>,
...
+ * Read more about running a [Raise] computation in the
+ * [Arrow docs](https://arrow-kt.io/learn/typed-errors/working-with-typed-errors/#running-and-inspecting-results).
+ */
+@ExperimentalRaiseAccumulateApi
+public inline fun <Error, A> iorAccumulate(noinline combineError: (Error, Error) -> Error, block: RaiseAccumulate<Error>.() -> A): Ior<Error, A> {
+  contract { callsInPlace(block, InvocationKind.AT_MOST_ONCE) }
+  val state: Atomic<Any?> = Atomic(EmptyValue)
+  return merge {
+    val raiseAccumulated = RaiseAccumulate.Error { raise(Ior.Left(EmptyValue.unbox(state.get()))) }
+    val accumulate = object : Accumulate<Error> {
```
serras: I think this should go as part of `IorRaise`, which maybe should get an `IorRaiseAccumulate` subclass.

arrow-libs/core/arrow-core/api/android/arrow-core.api
```diff
@@ -1199,13 +1233,19 @@ public final class arrow/core/raise/Trace {
 }
 
 public final class arrow/core/raise/context/RaiseContextualKt {
-	public static final fun accumulate (Larrow/core/raise/Raise;Lkotlin/jvm/functions/Function1;)Ljava/lang/Object;
-	public static final fun accumulating (Larrow/core/raise/RaiseAccumulate;Lkotlin/jvm/functions/Function1;)Larrow/core/raise/RaiseAccumulate$Value;
```
serras, line 1203: I don't fully understand

arrow-libs/core/arrow-core/api/android/arrow-core.api
```diff
@@ -1199,13 +1233,19 @@ public final class arrow/core/raise/Trace {
...
+	public static final fun accumulate (Larrow/core/raise/Accumulate;Ljava/lang/Object;)Larrow/core/raise/RaiseAccumulate$Value;
+	public static final fun accumulate (Larrow/core/raise/Raise;Lkotlin/jvm/functions/Function2;)Ljava/lang/Object;
+	public static final fun accumulateAll-EgRvm48 (Larrow/core/raise/Accumulate;Ljava/util/List;)Larrow/core/raise/RaiseAccumulate$Value;
+	public static final fun accumulating (Larrow/core/raise/Accumulate;Lkotlin/jvm/functions/Function2;)Larrow/core/raise/RaiseAccumulate$Value;
```
serras: I don't fully understand why this went from `Function1` to `Function2` in the API.

kyay10: This is in `RaiseContextualKt` anyways, so the API hasn't been released to anyone. Here I believe it's because I made the lambda take `context(Raise<E>, Accumulate<E>)` instead of `context(RaiseAccumulate<E>)`. If you want, I can remove the context-related functions from this PR and have them for another one

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/context/RaiseAccumulateContext.kt
```diff
@@ -8,113 +8,218 @@ package arrow.core.raise.context
...
 
+@Deprecated("RaiseAccumulate has been split into Raise and Accumulate. Use the new APIs instead.")
 public typealias RaiseAccumulate<A> = arrow.core.raise.RaiseAccumulate<A>
 
+// without the (this, this) we get an internal compiler error. May be fixed in 2.2.20?
 context(raise: Raise<NonEmptyList<Error>>)
 @ExperimentalRaiseAccumulateApi @RaiseDSL public inline fun <Error, A> accumulate(
-  block: context(RaiseAccumulate<Error>) () -> A
-): A = raise.accumulateExt(block)
+  block: context(Raise<Error>, Accumulate<Error>) () -> A
```
serras: I think this paints a wrong picture of what is going on. As far as I understand, you're not supposed to use `Raise` and `Accumulate` separately, they should always come in tandem. But by making them separate context parameters, you risk resolution choosing different ones.

serras: As with the other PR about context parameters, I would prefer you separate the changes about this PR itself (unifying RaiseAccumulate and IorRaise, which I have no problems with), from the idea of making context-versions expose a slightly different APIs (for which I still hold some reservations).

kyay10: Fair enough yes! I'll separate the context stuff into another PR, likely with way more explanation and so we may discuss the design there.

arrow-libs/core/arrow-core/api/jvm/arrow-core.api
```diff
@@ -1013,12 +1044,13 @@ public class arrow/core/raise/RaiseAccumulate : arrow/core/raise/Raise {
 	public final fun mapValuesOrAccumulate (Ljava/util/Map;Lkotlin/jvm/functions/Function2;)Ljava/util/Map;
 	public fun raise (Ljava/lang/Object;)Ljava/lang/Void;
 	public final fun raiseErrors ()Ljava/lang/Void;
+	public final fun recover (Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function1;)Ljava/lang/Object;
 	public final fun withNel (Lkotlin/jvm/functions/Function1;)Ljava/lang/Object;
 }
 
 public final class arrow/core/raise/RaiseAccumulate$Error : arrow/core/raise/RaiseAccumulate$Value {
-	public fun <init> (Larrow/core/raise/RaiseAccumulate;)V
-	public final fun getRaise ()Larrow/core/raise/RaiseAccumulate;
```
kyay10, line 1021: This was an accidental exposure in the context parameters PR. It didn't get published though, so this removal here is safe.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -951,65 +968,206 @@ public open class RaiseAccumulate<Error>(
...
+  ): A {
+    contract { callsInPlace(block, AT_MOST_ONCE) }
+    return arrow.core.raise.recover({
+      block(RaiseAccumulate(this@RaiseAccumulate, ::raise))
+    }, recover)
+  }
 
   @ExperimentalRaiseAccumulateApi
-  public inline fun ensureOrAccumulate(condition: Boolean, raise: () -> Error) {
+  public inline fun ensureOrAccumulate(condition: Boolean, raise: () -> Error): Value<Unit> {
```
kyay10, line 1038: Changed the return type here to allow `by` delegation on this method like its other siblings. Note that it's experimental, so such a breaking change is fine.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -826,18 +835,20 @@ public inline fun <K, Error, A, B> Raise<NonEmptyList<Error>>.mapValuesOrAccumul
...
   block: RaiseAccumulate<Error>.() -> A
 ): A {
   contract { callsInPlace(block, EXACTLY_ONCE) }
-  val nel = RaiseAccumulate(this)
-  val result = block(nel)
-  if (nel.hasErrors()) nel.raiseErrors()
-  return result
+  return with(RaiseAccumulate(this)) {
+    block().also { latestError?.value } // raises if there are accumulated errors
+  }
```
serras, line 853: Personally, this code is much more difficult to understand to me than the previous one. `also` and `?` work here as control flow primitives because of the inherent `raise`.

```suggestion
  return with(RaiseAccumulate(this)) {
    val result = block()
    if (hasAccumulatedErrors) { latestError?.value }
    return result
  }
```

serras (approved):

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/RaceUntil.kt
```diff
@@ -0,0 +1,41 @@
+package arrow.fx.coroutines
```
serras: I'm not sure why this file shows up here...

kyay10: Updating the branch seems to have fixed it!

PR #3670, opened 2025-07-22
Speed up NoTrace on JS

kyay10:
Supercedes (and is based on) #3656, and is smaller in its collateral

On my machine, running `:arrow-core:jsNodeTest`, I'm getting 47 seconds with this PR vs ~66 seconds without. 
This now uses `Error.stackTraceLimit`, which is supported everywhere except Firefox and Dino. It's not part of the standard, but I think that's okay for a performance optimization.


serras:
@sgrishchenko may I kindly ask you for your input here? This seems like a much smaller diff than your change, which is great from the maintainability point of view, but I don't have enough expertise to thoroughly review the code.

arrow-libs/core/arrow-core/src/jsMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
```diff
@@ -0,0 +1,32 @@
+package arrow.core.raise
+
+@OptIn(DelicateRaiseApi::class)
+@Suppress(
+  "EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING",
+  "SEALED_INHERITOR_IN_DIFFERENT_MODULE"
+)
+private class NoTraceImpl(raised: Any?, raise: Raise<Any?>) : RaiseCancellationException(raised, raise)
+
+private object SpecialRaised
```
kyay10: I'm not a big fan of all the code in this file, but it can always be changed. None of it is even ABI or source exposed, so I think it's okay if it's a little messy for now.

arrow-libs/core/arrow-core/src/jsMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
```diff
@@ -0,0 +1,32 @@
...
+    .map { entry -> entry[0] as String to entry[1] }
+
+private val noTraceKeys: Pair<String, String> = run {
+  val entries = entriesOf(js("Object").getOwnPropertyDescriptors(noTrace))
+  entries.first { (key, value) -> value.value == SpecialRaised }.first to entries.first { (key, value) -> value.value == SpecialRaise }.first
+} // Seems to always just be raised_1 and raise_1, but we compute them on startup to be safe.
+
+@DelicateRaiseApi
+internal actual fun NoTrace(raised: Any?, raise: Raise<Any?>): RaiseCancellationException {
+  val descs = js("Object").getOwnPropertyDescriptors(noTrace)
```
kyay10: I really don't like this recalculation of `getOwnPropertyDescriptors`. I have a feeling it might only be `message` that we need to add, but I haven't tested that yet. This technically future-proofs it. Maybe we can calculate the descriptors once, and then we can modify the values directly on `noTrace`? I was worried that that wouldn't work if the values are not writable, but we can just force them to be writable, no? all it takes is making `RaiseCancellationException` have mutable fields, but that feels a little weird

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Fold.kt
```diff
@@ -277,8 +277,7 @@ public sealed class RaiseCancellationException(
 ) : CancellationException(RaiseCancellationExceptionCaptured)
 
 @DelicateRaiseApi
-@Suppress("EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING")
-internal expect class NoTrace(raised: Any?, raise: Raise<Any?>) : RaiseCancellationException
+internal expect fun NoTrace(raised: Any?, raise: Raise<Any?>) : RaiseCancellationException
```
kyay10, line 280: I like this anyway, since we don't need to rely on a class at all!

arrow-libs/core/arrow-core/src/androidAndJvmMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
```diff
@@ -11,11 +9,15 @@ import kotlin.coroutines.cancellation.CancellationException
   "EXPECT_ACTUAL_CLASSIFIERS_ARE_IN_BETA_WARNING",
   "SEALED_INHERITOR_IN_DIFFERENT_MODULE"
 )
-internal actual class NoTrace actual constructor(raised: Any?, raise: Raise<Any?>) : RaiseCancellationException(raised, raise) {
+internal class NoTraceImpl(raised: Any?, raise: Raise<Any?>) : RaiseCancellationException(raised, raise) {
```
sgrishchenko: Why is this class internal? I guess it might be private. I am asking because I could align my PR too. 
In `arrow-libs/core/arrow-core/src/nonJvmAndJsMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt` you used private modifier.

arrow-libs/core/arrow-core/src/jsMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
```diff
@@ -0,0 +1,32 @@
...
+  descs[noTraceKeys.first].value = raised
+  descs[noTraceKeys.second].value = raise
```
sgrishchenko: As far as I can see you use the Flyweight pattern here. But it will work only if there is only one instance of `RaiseCancellationException` in runtime at the same time. Are you sure that there is no scenario, when there will be more than one instance of `RaiseCancellationException`? E.g. I know cases when you can get an error while you already have an error, it happens with resource handling:
* [proposal-explicit-resource-management](https://github.com/tc39/proposal-explicit-resource-management): `SuppressedError` 
* Kotlin [`use`](https://github.com/JetBrains/kotlin/blob/2.2.0/libraries/stdlib/jvm/src/kotlin/io/Closeable.kt#L21): `cause.addSuppressed(closeException)`
* [`Promise.any()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/any): `AggregateError`

arrow-libs/core/arrow-core/src/jsMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
```diff
@@ -0,0 +1,32 @@
...
+private val noTrace = NoTraceImpl(raised = SpecialRaised, raise = SpecialRaise)
+
+private fun entriesOf(jsObject: dynamic): List<Pair<String, dynamic>> =
+  (js("Object").entries(jsObject) as Array<Array<dynamic>>)
+    .map { entry -> entry[0] as String to entry[1] }
+
+private val noTraceKeys: Pair<String, String> = run {
+  val entries = entriesOf(js("Object").getOwnPropertyDescriptors(noTrace))
+  entries.first { (key, value) -> value.value == SpecialRaised }.first to entries.first { (key, value) -> value.value == SpecialRaise }.first
+} // Seems to always just be raised_1 and raise_1, but we compute them on startup to be safe.
```
sgrishchenko: I think this code is quite fragile.
In Kotlin/JS this code
```kotlin
class X {
    val x = 1
    val y = 2 
}

fun main() {
    val x = X()
    println(x.x)
    println(x.y)
}
```
will be compiled to this code:
```js
  function X() {
    this.x_1 = 1;
    this.y_1 = 2;
  }
  function main() {
    var x = new X();
    println(x.x_1);
    println(x.y_1);
  }
```
But this code (pay attention for `open`):
```kotlin
open class X {
    open val x = 1
    open val y = 2 
}

fun main() {
    val x = X()
    println(x.x)
    println(x.y)
}
```
will be compiled to this one:
```js
  function X() {
    this.x_1 = 1;
    this.y_1 = 2;
  }
  protoOf(X).get_x_1mhr67_k$ = function () {
    return this.x_1;
  };
  protoOf(X).get_y_1mhr68_k$ = function () {
    return this.y_1;
  };
  function main() {
    var x = new X();
    println(x.get_x_1mhr67_k$());
    println(x.get_y_1mhr68_k$());
  }
```
So this trick reveals that:
1) The reason why Kotlin/JS expose `x_1` and `y_1` as public fields it is just lowering (technic in compiler development to reorganize produced code). Its implementation can be changed at any time without notifications, because actually `x_1` and `y_1` are private backing fields.
1) In future implementations Kotlin/JS can use JS private fields for backing fields.

arrow-libs/core/arrow-core/src/androidAndJvmMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
kyay10: My bad, it should be private lol!

arrow-libs/core/arrow-core/src/jsMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
kyay10: I don't think that's a problem? `getOwnPropertyDescriptors` should create a fresh property descriptors object each time. Even if it isn't, JS is never parallel, and the only 2 things that change about each property descriptors is raised and raise, both of which I set immediately then use the descriptors. Each created `NoTraceImpl` is still completely independent of all the other ones

arrow-libs/core/arrow-core/src/jsMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
kyay10: JS private fields is a good shout actually! I'll see what I can do about that... I think your approach is indeed immune to it

arrow-libs/core/arrow-core/src/jsMain/kotlin/arrow/core/raise/CancellationExceptionNoTrace.kt
sgrishchenko: Oh, my bad, you create a new object each time using `Object.create(...)`. My comment is irrelevant. :slightly_smiling_face:

kyay10:
Closed in favour of #3656

PR #3673, opened 2025-07-23
"At most once" computation for Eval

### LINKED ISSUE #3668: `SuspendEval.later` is not thread-safe
Hey All,

Per the title, `SuspendEval.later` is not thread-safe (or "coroutine safe", if that's a thing/term). This "issue" may be an intentional design choice, but if so, it'd be great to see an alternative that "works" when executed in parallel. Given that `suspend` functions are often used for longer-running I/O tasks, it feels appropriate that `SuspendEval.later`, in particular, would guarantee that its lambda will only ever run a single time.

My use case is that I'd like to cache the result of a lazily-evaluated API request, and the process that triggers the API request (*if* it triggers the API request, hence wanting lazy evaluation) could be executed multiple times in parallel.

Example JUnit (with kotest assertion) test for the desired behavior:

```kotlin
@Test
fun `SuspendEval later is atomic`() {
    var executionCount = 0

    val name: SuspendEval<String> =
        SuspendEval.later {
            executionCount++
            delay(1.seconds)
            "John"
        }

    runBlocking {
        Array(20) { async { name.run() } }.toList().awaitAll()
        executionCount shouldBe 1
    }
}
```

serras:
Fixes #3668

### DIFF
--- a/arrow-libs/core/arrow-eval/api/arrow-eval.klib.api
+++ b/arrow-libs/core/arrow-eval/api/arrow-eval.klib.api
@@ -74,10 +74,6 @@ sealed interface <#A: out kotlin/Any?> arrow.eval/SuspendEval { // arrow.eval/Su
     final class <#A1: out kotlin/Any?> AtMostOnce : arrow.eval/SuspendEval<#A1> { // arrow.eval/SuspendEval.AtMostOnce|null[0]
         constructor <init>(kotlin.coroutines/SuspendFunction0<#A1>) // arrow.eval/SuspendEval.AtMostOnce.<init>|<init>(kotlin.coroutines.SuspendFunction0<1:0>){}[0]
 
-        final var result // arrow.eval/SuspendEval.AtMostOnce.result|{}result[0]
-            final fun <get-result>(): arrow.core/Option<#A1> // arrow.eval/SuspendEval.AtMostOnce.result.<get-result>|<get-result>(){}[0]
-            final fun <set-result>(arrow.core/Option<#A1>) // arrow.eval/SuspendEval.AtMostOnce.result.<set-result>|<set-result>(arrow.core.Option<1:0>){}[0]
-
         final fun copy(kotlin.coroutines/SuspendFunction0<#A1> = ...): arrow.eval/SuspendEval.AtMostOnce<#A1> // arrow.eval/SuspendEval.AtMostOnce.copy|copy(kotlin.coroutines.SuspendFunction0<1:0>){}[0]
         final fun equals(kotlin/Any?): kotlin/Boolean // arrow.eval/SuspendEval.AtMostOnce.equals|equals(kotlin.Any?){}[0]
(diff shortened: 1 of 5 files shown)


serras:
@ScottHamper what do you think of this API, and the `atMostOnce` name? I didn't want to change the behavior of `later`, because people may (inadvertently) rely on their behavior on multiple on concurrent scenarios.

ScottHamper:
Love it! Commit looks great - thanks for your work, @serras

ScottHamper:
One quick follow up - [`SuspendEval.kt`, line 170](https://github.com/arrow-kt/arrow/blob/c99fd8a410ae3184cdc17d945d594de094107891/arrow-libs/core/arrow-eval/src/commonMain/kotlin/arrow/eval/SuspendEval.kt#L170) has `result` as a public `var`. Should this value be `private`? Looks like it is consistent with `Later`, but seems like a strange choice to expose mutable state. Regular, non-suspend `Eval.Later` does not expose a mutable value like this.

arrow-libs/core/arrow-eval/api/arrow-eval.klib.api
```diff
@@ -160,11 +176,12 @@ sealed class <#A: out kotlin/Any?> arrow.eval/Eval : arrow.eval/SuspendEval<#A>
 
     final class <#A1: out kotlin/Any?> Later : arrow.eval/Eval<#A1>, kotlin/Lazy<#A1> { // arrow.eval/Eval.Later|null[0]
         constructor <init>(kotlin/Function0<#A1>) // arrow.eval/Eval.Later.<init>|<init>(kotlin.Function0<1:0>){}[0]
+        constructor <init>(kotlin/LazyThreadSafetyMode, kotlin/Function0<#A1>) // arrow.eval/Eval.Later.<init>|<init>(kotlin.LazyThreadSafetyMode;kotlin.Function0<1:0>){}[0]
 
         final val value // arrow.eval/Eval.Later.value|{}value[0]
             final fun <get-value>(): #A1 // arrow.eval/Eval.Later.value.<get-value>|<get-value>(){}[0]
 
-        final fun copy(kotlin/Function0<#A1> = ...): arrow.eval/Eval.Later<#A1> // arrow.eval/Eval.Later.copy|copy(kotlin.Function0<1:0>){}[0]
+        final fun copy(kotlin/LazyThreadSafetyMode = ..., kotlin/Function0<#A1> = ...): arrow.eval/Eval.Later<#A1> // arrow.eval/Eval.Later.copy|copy(kotlin.LazyThreadSafetyMode;kotlin.Function0<1:0>){}[0]
```
kyay10, line 177: should probably be hidden by making the constructor internal. I doubt anyone was relying on this, but just in case

kyay10 (approved):
Seems pretty good! I'd maybe prefer a name like `lazy`, but naming is hard

arrow-libs/core/arrow-eval/api/arrow-eval.klib.api
serras: I just added the missing `copy` overload

PR #3677, opened 2025-07-24
Migrate to Android KMP plug-in

serras:
This is the new [recommended way](https://developer.android.com/kotlin/multiplatform/plugin#apply) to have a project with KMP with Android as one of the targets.


serras:
@kyay10 thanks for catching that! The updated PR adds tests back (although disabling Kover because there needs to be a new release with support for this plug-in).

PR #3680, opened 2025-07-26
Pass Trace transparently through withError

kyay10:
I used a default property this time around. Since we support Java 8 and up, I believe this shouldn't be binary incompatible?


arrow-libs/core/arrow-core/api/android/arrow-core.api
```diff
@@ -1101,7 +1103,6 @@ public final class arrow/core/raise/RaiseKt {
 	public static final fun toResult (Lkotlin/jvm/functions/Function2;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static final fun toResult (Lkotlin/jvm/functions/Function2;Lkotlin/jvm/functions/Function2;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 	public static final fun tolerant (Larrow/core/raise/Accumulate;Larrow/core/raise/Raise;)Larrow/core/raise/Accumulate;
-	public static final fun withCause (Larrow/core/raise/Traced;Larrow/core/raise/Traced;)Larrow/core/raise/Traced;
```
kyay10, line 1104: This was only used in experimental trace api.

PR #3682, opened 2025-07-26
Add Arrow Version Catalog module

nomisRev:
Introduced a new module for managing version catalogs for Arrow Kotlin libraries. This includes updates to `settings.gradle.kts`, a `build.gradle.kts` configuration, and a README file with setup instructions. The catalog simplifies dependency management for various Arrow libraries.

### DIFF
--- a/arrow-libs/version-catalog/README.MD
+++ b/arrow-libs/version-catalog/README.MD
@@ -1,0 +1,27 @@
+# Arrow Version Catalog
+
+## settings.gradle.kts
+
+```kotlin
+dependencyResolutionManagement {
+    repositories {
+        mavenCentral()
+    }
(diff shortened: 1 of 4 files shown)


serras:
This supersedes `arrow-stack`, right? If so, I would prefer to remove `arrow-stack` in the same PR, and then we can update the instructions in the website.

nomisRev:
> This supersedes arrow-stack, right? If so, I would prefer to remove arrow-stack in the same PR, and then we can update the instructions in the website.

I don't think this supersedes `arrow-stack` since that's the maven BOM, which is more often used by Maven users which we have.

kyay10 (approved):

PR #3683, opened 2025-07-26
Introduce `Racing` and `RaiseRacing` DSL

nomisRev:


arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Racing.kt
```diff
@@ -0,0 +1,137 @@
...
+ * @param block The block of code to execute within the racing scope.
+ * @return The result of the first racing block to complete successfully.
+ */
+public suspend fun <A> racing(
+  handleException: (suspend (context: CoroutineContext, exception: Throwable) -> Unit)? = null,
+  block: suspend RacingScope<A>.() -> Unit,
+): A {
+  val exceptionHandler: suspend (context: CoroutineContext, exception: Throwable) -> Unit =
+    handleException
+      ?: currentCoroutineContext()[CoroutineExceptionHandler]?.let { { ctx, e -> it.handleException(ctx, e) } }
```
kyay10, line 98: nit: function reference would be nicer!

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Racing.kt
```diff
@@ -0,0 +1,137 @@
...
+    }
+
+  override suspend fun race(context: CoroutineContext, block: suspend CoroutineScope.() -> A) =
+    raceOrThrow {
+      try {
+        block()
+      } catch (e: CancellationException) {
+        throw e
+      } catch (e: Throwable) {
+        handleException(currentCoroutineContext(), e/*.nonFatalOrThrow()*/)
```
kyay10, line 133: `nonFatalOrThrow` seems like a good idea here, no? If one gets an OOM, shouldn't it end everything?

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/RaiseRacing.kt
```diff
@@ -0,0 +1,202 @@
...
+import kotlinx.coroutines.currentCoroutineContext
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.selects.SelectBuilder
+import kotlinx.coroutines.selects.select
+import kotlin.coroutines.CoroutineContext
+import kotlin.coroutines.EmptyCoroutineContext
+import kotlin.time.Duration
+
+/**
```
kyay10, line 22: This may be controversial, but I think we should just YOLO and go with contexts instead of such an intersection type.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/RaiseRacing.kt
```diff
@@ -0,0 +1,202 @@
...
+public interface RaiseScope<in Error> : CoroutineScope, Raise<Error>
+
+/**
+ * A scope that allows racing multiple coroutine blocks against each other with [Raise] context receiver support.
+ * The first block to complete successfully will provide the result, and all other blocks will be cancelled.
+ *
+ * @param Error The error type that can be raised by the racing blocks.
+ * @param Result The type of value that will be returned by the racing blocks.
+ */
+public interface RaiseRacingScope<in Error, in Result> {
```
kyay10, line 37: Shouldn't this extend `RacingScope`?

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/RaiseRacing.kt
```diff
@@ -0,0 +1,202 @@
...
+): A {
+  val handleRaise: suspend (context: CoroutineContext, raised: E) -> Unit =
+    handleRaise ?: { _, _ -> }
+  val exceptionHandler: suspend (context: CoroutineContext, exception: Throwable) -> Unit =
+    handleException
+      ?: currentCoroutineContext()[CoroutineExceptionHandler]?.let { { ctx, e -> it.handleException(ctx, e) } }
+      ?: { _, e -> e.printStackTrace() }
+  return coroutineScope {
+    val job = Job(currentCoroutineContext()[Job])
+    recover({
```
kyay10, line 126: `recover(x) { raise(it) }` is isomorphic to `x()` I'm pretty sure

kyay10:
I'll have a deeper look at this later today (so apologies for the vague comments). I really think `RaiseRacingScope` might not be strictly necessary though. I think the normal `racing` with an outer Raise builder has the same exact functionality

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Racing.kt
kyay10: I guess that's up to the user to `handle` properly though.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Racing.kt
```diff
@@ -0,0 +1,137 @@
...
+    handleException
+      ?: currentCoroutineContext()[CoroutineExceptionHandler]?.let { { ctx, e -> it.handleException(ctx, e) } }
+      ?: { _, e -> e.printStackTrace() }
+  return coroutineScope {
+    val job = Job(currentCoroutineContext()[Job])
+    try {
+      select {
+        launch(start = CoroutineStart.UNDISPATCHED) {
+          val scope = SelectRacingScope(this@select, this@coroutineScope, job, exceptionHandler)
+          scope.block()
```
kyay10, line 106: It feels a little weird to potentially `suspend` within `select`

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Racing.kt
```diff
@@ -0,0 +1,137 @@
...
+ * ```
+ *
+ * @param A The type of value that will be returned by the racing blocks.
+ * @param handleException A function to handle exceptions thrown by racing blocks. If null, exceptions will be handled
+ *                       by the current coroutine context's [CoroutineExceptionHandler] or printed to stderr.
+ * @param block The block of code to execute within the racing scope.
+ * @return The result of the first racing block to complete successfully.
+ */
+public suspend fun <A> racing(
+  handleException: (suspend (context: CoroutineContext, exception: Throwable) -> Unit)? = null,
```
kyay10, line 93: Do we need explicit exception handling if the caller can just provide a `CoroutineExceptionHandler` themselves?

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Racing.kt
kyay10, line 133: nit: since `handleException` is `suspend`, it doesn't need to be passed the `currentCoroutineContext()` explicitly, it can just access it itself

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/RaiseRacing.kt
```diff
@@ -0,0 +1,202 @@
...
+
+  private fun raceImpl(
+    failOnThrow: Boolean,
+    failOnRaise: Boolean,
+    context: CoroutineContext,
+    block: suspend RaiseScope<E>.() -> A
+  ) = with(select) {
+    async(context = context + job) {
+      recover({ block() }, { error: E ->
+        if (failOnRaise) {
```
kyay10, line 186: I don't like this `raiseHandler` idea too much. The user can instead just handle it within `race`, no?

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Racing.kt
nomisRev: No, I should put this back I ported this file from another project 😅

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/RaiseRacing.kt
nomisRev: It can't because the functions won't resolve nicely without requiring different names.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Racing.kt
nomisRev: Good point, probably better to remove this weird hack to support `suspend` there.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Racing.kt
nomisRev: I felt it was such critical part of the functions behavior it required being part of the function signature for explicitness, the fallback on `currentCoroutineContext()[CoroutineExceptionHandler]` felt like the best behavior rather than just `printStacktrace`.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Racing.kt
nomisRev: It's part of the `CoroutineContextHandler` signature? 🤔

kyay10: because theirs isn't `suspend`. Maybe ours shouldn't be either then?

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/RaiseRacing.kt
nomisRev: The idea behind this DSL was to easily to race to _success_ so ignoring exceptions and typed failures. I think `recover({ block() }) { e -> /* ... */ awaitCancellation() }` is a non-trivial idiom for many which they can still do for more fine-grained control.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/RaiseRacing.kt
nomisRev: I originally had context parameters, then I changed it to `context(CoroutineScope) Raise<E>` because outer wise outer `Raise<E2>.() ->` might take precedence in explicit nested blocks without context receivers.

And then in the end I decided to ditch it all together since I already have the instance anyway to support it for non-context users 😅

nomisRev:
@kyay10 Please have a closer look when you have time 🙏

> just YOLO and go with contexts

If we really want to YOLO we should just make a context based kernel, and build the rest on top of it but that would probably turn it into Arrow 3.0 and require another year+ so we might as well wait for rich errors and see what happens 😁

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/RaiseRacing.kt
kyay10: Maybe `raceRecovering` or something that works in the normal `racing` DSL but takes a `(E) -> Unit` handler for the block. Inside the impl can be as simple as `raceOrThrow {

  recover(block) { 
    handle(it)
    awaitCancellation
  }
}
`
Is it that likely that all the different branches have the same `raise` type that one wishes to ignore? Even if, one can then locally define a `handleRaise` lambda, and pass it to each `raceRecovering`

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Racing.kt
```diff
@@ -0,0 +1,137 @@
...
+  block: suspend RacingScope<A>.() -> Unit,
+): A {
+  val exceptionHandler: suspend (context: CoroutineContext, exception: Throwable) -> Unit =
+    handleException
+      ?: currentCoroutineContext()[CoroutineExceptionHandler]?.let { { ctx, e -> it.handleException(ctx, e) } }
+      ?: { _, e -> e.printStackTrace() }
+  return coroutineScope {
+    val job = Job(currentCoroutineContext()[Job])
```
kyay10, line 101: Is there any reason not to put the `job` within `coroutineScope` directly? Only difference I can see is that the `CoroutineScope` accessible from `block` is unchanged

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Racing.kt
```diff
@@ -0,0 +1,137 @@
...
+  scope: CoroutineScope,
+  private val job: Job,
+  private val handleException: suspend (context: CoroutineContext, exception: Throwable) -> Unit
+) : RacingScope<A>, CoroutineScope by scope {
+  override suspend fun raceOrThrow(context: CoroutineContext, block: suspend CoroutineScope.() -> A) =
+    with(select) {
+      async(context = context + job, block = block).onAwait { it }
+    }
+
+  override suspend fun race(context: CoroutineContext, block: suspend CoroutineScope.() -> A) =
```
kyay10, line 126:
```kotlin
  override suspend fun race(context: CoroutineContext, block: suspend CoroutineScope.() -> A) =
    raceOrThrow {
      catch({ block() }) {
        handleException(currentCoroutineContext(), it)
        awaitCancellation()
      }
    }
```

serras:
@nomisRev @kyay10 it would be great if you could figure out what is the best API for this (between #3683 and #3684), at least for now. That way we can have it in 2.2 and let people experiment with it.

serras:
Closed in favor of #3684

PR #3701, opened 2025-08-12
Fix crosswalk ordering logic

jsoizo:
This PR addresses the crosswalk-related implementation problems discovered during the review of #3040:

- Fix ordering issue in crosswalk: Corrects the fold operation to preserve original element order instead of reversing it
- Fix crosswalkNull behavior: Ensures proper null handling and return type behavior according to the expected semantics

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
@@ -882,7 +882,7 @@ public fun <A, B> Iterable<A>.crosswalk(f: (A) -> Iterable<B>): List<List<B>> =
       ior.fold(
         { listOf(it) },
         ::identity,
-        { l, r -> listOf(l) + r }
+        { l, r -> r + l }
       )
     }
   }
@@ -893,19 +893,13 @@ public fun <A, K, V> Iterable<A>.crosswalkMap(f: (A) -> Map<K, V>): Map<K, List<
       ior.fold(
         { listOf(it) },
         ::identity,
-        { l, r -> listOf(l) + r }
+        { l, r -> r + l }
       )
     }
   }
 
 public fun <A, B> Iterable<A>.crosswalkNull(f: (A) -> B?): List<B>? =
-  fold<A, List<B>?>(emptyList()) { bs, a ->
-    Ior.fromNullables(f(a), bs)?.fold(
-      { listOf(it) },
-      ::identity,
-      { l, r -> listOf(l) + r }
-    )
-  }
+  mapNotNull(f).takeIf { it.isNotEmpty() || !this.any() }
 
 public operator fun <A : Comparable<A>> Iterable<A>.compareTo(other: Iterable<A>): Int =
   align(other) { ior -> ior.fold({ 1 }, { -1 }, { a1, a2 -> a1.compareTo(a2) }) }
--- a/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
+++ b/arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
@@ -559,9 +559,8 @@ class IterableTest {
       val transarr: List<(Int) -> Int> = incr.map { i -> { it + i } }
 
       val expected = mutableListOf<List<Int>>()
-      val arev = a.reversed()
       transarr.forEach { t ->
-        arev.map { t(it) }.takeIf { it.isNotEmpty() }
+        a.map { t(it) }.takeIf { it.isNotEmpty() }
           ?.let(expected::add)
       }
 
@@ -590,7 +589,17 @@ class IterableTest {
     checkAll(Arb.list(Arb.int(), 0..10), Arb.pair(Arb.int(1..1000), Arb.int())) { a, mod ->
       fun trans(i: Int) = if (i % mod.first == 0) i + mod.second else null
 
-      val expected = a.reversed().mapNotNull(::trans)
+      var expected: MutableList<Int>? = if (a.isEmpty()) mutableListOf() else null
+      a.forEach { n ->
+        val result = trans(n)
+        if (result != null) {
+          if (expected == null) {
+            expected = mutableListOf(result)
+          } else {
+            expected.add(result)
+          }
+        }
+      }
       a.crosswalkNull(::trans) shouldBe expected
     }
   }


arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
```diff
@@ -590,7 +589,17 @@ class IterableTest {
     checkAll(Arb.list(Arb.int(), 0..10), Arb.pair(Arb.int(1..1000), Arb.int())) { a, mod ->
       fun trans(i: Int) = if (i % mod.first == 0) i + mod.second else null
 
-      val expected = a.reversed().mapNotNull(::trans)
+      var expected: MutableList<Int>? = if (a.isEmpty()) mutableListOf() else null
```
kyay10: Why not just
```kotlin
val expected = if (a.isEmpty()) a else a.mapNotNull(::trans).takeIf { it.isNotEmpty() }
```
The logic as written took me a minute to follow, but this one liner is clearer in intent I think

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -893,19 +893,13 @@ public fun <A, K, V> Iterable<A>.crosswalkMap(f: (A) -> Map<K, V>): Map<K, List<
...
 
 public fun <A, B> Iterable<A>.crosswalkNull(f: (A) -> B?): List<B>? =
-  fold<A, List<B>?>(emptyList()) { bs, a ->
-    Ior.fromNullables(f(a), bs)?.fold(
-      { listOf(it) },
-      ::identity,
-      { l, r -> listOf(l) + r }
-    )
-  }
+  mapNotNull(f).takeIf { it.isNotEmpty() || !this.any() }
```
kyay10, line 902:
```suggestion
  mapNotNull(f).takeIf { it.isNotEmpty() || this.isEmpty() }
```

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
kyay10: I know it's very close to the implementation, but this here basically inlines `mapNotNull` and makes the code ever so slightly worse

jsoizo:
@kyay10 Thank you for reviewing. 
I have fixed both of the two you commented on.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -893,19 +893,13 @@ public fun <A, K, V> Iterable<A>.crosswalkMap(f: (A) -> Map<K, V>): Map<K, List<
       ior.fold(
         { listOf(it) },
         ::identity,
-        { l, r -> listOf(l) + r }
```
kyay10, line 896: I really would prefer not using `fold` here at all, and instead using `buildMap` and an imperative implementation. I understand that this is likely a standard implementation though, so I get why it's written this way. Maybe we can change it in another PR I guess

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
```diff
@@ -893,19 +893,13 @@ public fun <A, K, V> Iterable<A>.crosswalkMap(f: (A) -> Map<K, V>): Map<K, List<
...
 
 public fun <A, B> Iterable<A>.crosswalkNull(f: (A) -> B?): List<B>? =
-  fold<A, List<B>?>(emptyList()) { bs, a ->
-    Ior.fromNullables(f(a), bs)?.fold(
-      { listOf(it) },
-      ::identity,
-      { l, r -> listOf(l) + r }
-    )
-  }
+  mapNotNull(f).takeIf { it.isNotEmpty() || it.isEmpty() }
```
kyay10: I think you mean:
```suggestion
  mapNotNull(f).takeIf { it.isNotEmpty() || this.isEmpty() }
```
because otherwise you've just written `mapNotNull(f).takeIf { true }`

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
```diff
@@ -590,7 +589,7 @@ class IterableTest {
     checkAll(Arb.list(Arb.int(), 0..10), Arb.pair(Arb.int(1..1000), Arb.int())) { a, mod ->
       fun trans(i: Int) = if (i % mod.first == 0) i + mod.second else null
 
-      val expected = a.reversed().mapNotNull(::trans)
+      val expected = if (a.isEmpty()) a else a.mapNotNull(::trans).takeIf { it.isNotEmpty() || it.isEmpty() }
```
kyay10:
```suggestion
      val expected = if (a.isEmpty()) a else a.mapNotNull(::trans).takeIf { it.isNotEmpty() }
```
Weird that this didn't trigger test failure. We should make sure that a list of all nulls is tested too

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
```diff
@@ -590,7 +589,7 @@ class IterableTest {
     checkAll(Arb.list(Arb.int(), 0..10), Arb.pair(Arb.int(1..1000), Arb.int())) { a, mod ->
```
kyay10, line 589: The `list` should have `nulls` too. I think we have an `Arb` for that

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
jsoizo: `Iterable<T>` does not have `isEmpty` method and it is `Collection<T>` method.
So, I will make this back to `!this.any()`

kyay10: Yep my bad I didn't consider the types!

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
jsoizo: Are you suggesting that i should also take into account scenarios where the list elements might contain null values?

kyay10: Oh wait my bad! Didn't realise `trans` took care of that.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
jsoizo: fixed this at commit 7eb453ec2f56519a736cfea54c3b2ff519baa60a

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
jsoizo: I've make this back to `!this.any()` at commit https://github.com/arrow-kt/arrow/commit/7eb453ec2f56519a736cfea54c3b2ff519baa60a

kyay10 (approved):

PR #3703, opened 2025-08-14
Remove incorrect and misleading returns(null) contracts

kyay10:
First discovered on Slack by Huib Donkers.

Luckily, `returns` contracts are broken when consumed from third-party library anyways. Also, Arrow's current Kotlin version has a bug that doesn't smart case based on those contracts lol. Regardless, with that bug fixed, this code would produce a CCE:
```kotlin
val either: Either<String, Int?> = Right(null)
if (either.gon() == null) {
  val left: Either.Left<*> = either
  left.value
}
```
Simply, `returns(null)` shouldn't actually imply that the either is `Left`, since it might be `Right(null)`. When this [feature request](https://youtrack.jetbrains.com/issue/KT-43798/Cannot-check-for-instance-of-erased-type-inside-contracts-for-generics) is implemented, we can put back the contract as the more complicated `returns(null) implies (this@Either is Left || this@Either is Right<Nothing?>)` and we can update the current contract as well to be `returnsNotNull() implies (this@Either is Right<B & Any>)`.

I also cleaned up other similar contracts

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
@@ -489,8 +489,8 @@ public sealed class Either<out A, out B> {
    */
   public fun isLeft(): Boolean {
     contract {
-      returns(true) implies (this@Either is Left<A>)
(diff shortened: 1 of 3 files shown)


serras:
Thanks for catching all of those!

PR #3705, opened 2025-08-27
[KUP] Disable opt-in Kotlin compiler warnings

FenstonSingel:
It was decided to deprecate and phase out the concept (see KT-77721).


serras:
@FenstonSingel the KUP branch has been updated

PR #3707, opened 2025-08-28
Use new binary built-in validator in Kotlin 2.2

### LINKED ISSUE #3702: Move to KGP's BCV
https://kotlinlang.org/docs/gradle-binary-compatibility-validation.html
Seems like all new features will happen there anyways, so it's worth it to make the move ASAP. It's experimental though, so gradle changes may be necessary. Maybe we could hold off until 3.0 to make this move

serras:
https://kotlinlang.org/docs/gradle-binary-compatibility-validation.html

Fixes #3702


serras:
The changes in `.api` files are coming from filtering out experimental APIs from the ABI dump.

buildSrc/src/main/kotlin/arrow.kotlin.gradle.kts
```diff
@@ -29,6 +34,13 @@ val Project.requiresAndroidCoreLibraryDesugaring
 val Project.needsJava11
   get() = project.name.endsWith("-compose")
 
+val Project.noAbiValidation
```
kyay10: nit: Maybe `needsAbiValidation` is a better name (inverted of course)

kyay10 (approved):
Great that experimental APIs are filtered out now!

buildSrc/src/main/kotlin/arrow.kotlin.gradle.kts
serras: Done!

Issue #3639, opened 2025-06-10
["Request"] Analogue of Either.recover for turning right values into left values


jdsatlin:
What version are you currently using?
`2.1.2`

What would you like to see?

I have frequently found myself reaching for an ergonomic way to examine the results of a call that returns an Either, and in some additional additional situations turn a right value into a left value.

This is common when using things like Either.catch { } on an http connector or a database call, where a value might not have led to an exception, but could still be invalid for the business logic of the application, and should be remapped into a business error after being recognized. Similarly so, if you are combining multiple Eithers that might have been successes on their own, but when some combination of them comes back, there are additional validations can be done.

I know this can be achieved by binding the Eithers within a Raise context, but particularly with code that is returning an Either is often more convenient to continue along in chaining rather than switching to the Raise DSL. At least for now, public functions returning Eithers are more discoverable through autocomplete than Raise equivalents.

The closest equivalent to what I have found myself reaching for is the 'recover' function, able to continue chaining on an Either and examine a failure, and either propagate it, transform it, or turn it into a success.

With credit to @kyay10 for both helping with current ways to accomplish this action in the current version of Arrow, and for prototyping solutions.

Toy sample code that might benefit from this is here:

```
fun randomNumber(): Either<String, Int> {
    val randomNum = Random(1).nextInt(1..10)

    return if (randomNum > 5) {
        randomNum.right()
    } else {
        "The number was too low".left()
    }
}

fun failOnMoreConditionsFlatmap(): Either<String, Int> {
    return randomNumber()
        .flatMap { num ->
           either {
               ensure(num != 10) { "Number 10 also not allowed" }
               num
           }
    }
}

fun failOnMoreConditionsBind(): Either<String, Int> = either {
  randomNumber().bind().also {
    ensure(it != 10) { "Number 10 also not allowed" }
  }
}.map { it + 100 }
```

whereas the ability to deal with a similar situation through recover requires less nesting, as there's no need to create a raise dsl through an inline `either`

```
fun recoverExample(): Either<String, Int> {
    return randomNumber()
        .recover {
            val recoveryAttempt = Random(2).nextInt(1..10)
            ensure(recoveryAttempt > 5) { "tried to recover and failed" }
            recoveryAttempt
        }
        .map { it + 100 }
}
```

A solution was suggested by @kyay10 on slack

```
[...]

PR #3711, opened 2025-08-29
Introduce `Raise` in `onRightBind` and `onLeftBind`

serras:
This fixes the original goal of #3639 of being able to check for errors in the middle of a chain of calls, by allowing `raise`ing in `onRight`.

```kotlin
fun failOnMoreConditionsWithBindMap(): Either<String, Int> =
    randomNumber()
        .onRight { ensure(it != 10) { "Number 10 also not allowed" } }
        .map { it + 100 }
```

IMO this way of looking at this point requires a much smaller change to our API (just generalizing a function type). We should not try to account for all possible ways of working on `either` in every possible chain.

### DIFF
--- a/arrow-libs/core/arrow-core/api/android/arrow-core.api
+++ b/arrow-libs/core/arrow-core/api/android/arrow-core.api
@@ -74,7 +74,8 @@ public abstract class arrow/core/Either {
 	public final fun map (Lkotlin/jvm/functions/Function1;)Larrow/core/Either;
 	public final fun mapLeft (Lkotlin/jvm/functions/Function1;)Larrow/core/Either;
 	public final fun onLeft (Lkotlin/jvm/functions/Function1;)Larrow/core/Either;
-	public final fun onRight (Lkotlin/jvm/functions/Function1;)Larrow/core/Either;
+	public final synthetic fun onRight (Lkotlin/jvm/functions/Function1;)Larrow/core/Either;
+	public final fun onRight (Lkotlin/jvm/functions/Function2;)Larrow/core/Either;
 	public final fun swap ()Larrow/core/Either;
 	public final fun toIor ()Larrow/core/Ior;
 	public fun toString ()Ljava/lang/String;
--- a/arrow-libs/core/arrow-core/api/android/arrow-core.api
+++ b/arrow-libs/core/arrow-core/api/android/arrow-core.api
@@ -74,8 +74,9 @@ public abstract class arrow/core/Either {
 	public final fun map (Lkotlin/jvm/functions/Function1;)Larrow/core/Either;
 	public final fun mapLeft (Lkotlin/jvm/functions/Function1;)Larrow/core/Either;
 	public final fun onLeft (Lkotlin/jvm/functions/Function1;)Larrow/core/Either;
-	public final synthetic fun onRight (Lkotlin/jvm/functions/Function1;)Larrow/core/Either;
-	public final fun onRight (Lkotlin/jvm/functions/Function2;)Larrow/core/Either;
+	public final fun onLeftBind (Lkotlin/jvm/functions/Function2;)Larrow/core/Either;
+	public final fun onRight (Lkotlin/jvm/functions/Function1;)Larrow/core/Either;
+	public final fun onRightBind (Lkotlin/jvm/functions/Function2;)Larrow/core/Either;
 	public final fun swap ()Larrow/core/Either;
 	public final fun toIor ()Larrow/core/Ior;
 	public fun toString ()Ljava/lang/String;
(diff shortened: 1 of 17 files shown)


tKe:
It feels off that `onRight` can change the return value based on the lambda. Most/all `onBlah` methods are usually intended to be a filtered form of `also`.

It may also break any existing code that might be using `ensure`/`raise` from an outer scope.

```kotlin
either {
  getSomething()
    .onRight { ensure(it > 10) { "not enough foo" } }
    .bind()
}
```

nomisRev:
Great point @tKe! All `onXXX` functions were indeed meant as analogs like `also` and `onEach` on `Sequence` or `Flow` which was a renamed after `tap` and `tapLeft` which was strange naming.

We might be better off introducing a `dsl { }` function so people use `either { }` in a _fluent API_ DSL which I think is the point of the original request? 🤔  Makes me think of `persistentlist.mutate { }` as well.

serras:
@tKe @nomisRev Gotcha! I agree that reusing the same name may be quite confusing.

I've prepared a new proposal: `onRightBind` and `onLeftBind`, which work as `onRight` and `onLeft` except that they also give you a `Raise` context. The goal here is to provide a simple way for one-liners with `ensure` / `ensureNotNull` / other small validations, without having to go into the bigger DSL.

tKe:
Certainly it resolves the issue of existing code changing meaning, but I'm still uncertain on the `onXXX` prefix. 

Something like `.ensuring { ensure(it != 5) { "5 is right out" } }` feels better semantically, at the downside of doubling the `ensure`. 

Alternatively, would the use-case be catered for with something like 

```kotlin
  either<String, Int> { 5 }
    .mapRaise {
      ensure(it != 5) { "5 is right out" }
      it
    }
```

It's not quite the one-liner of `.onRightBind { ensure(it != 5) { "5 is right out " } }` as it requires the trailing `it` for the return (as it's a `map`) but could fit better with the existing api.

serras:
@tKe I would say that the whole point of this is making one-liners possible, otherwise the Raise DSL provides better ergonomics. This PR for me is accounting for those people who prefer a chain-of-transformation style for working with Either directly, and that only works if every step takes one line.

I fully agree on the naming side, although I cannot find a better name. Some thoughts:
- `onRightXX` and `onLeftXX` are very clear on what is matched;
- Something like `also` goes very well with what the standard library does, although definitely here the "kind" of `Either` may change. But maybe `alsoOnRight`?

tKe:
Fully with the purpose of supporting a one liner. 

I'm mainly concerned that there's usually an expectation that side-effect scope functions like `also`, and filtered `onXxx` functions tend to always return their receiver directly (or not return at all due to non-local returns/control-flow).

The only names I can come up with would be `.ensuring { }` (as above) or possibly `.filter { }`. I'm more taken with the the latter, despite it not taking a strict predicate as you would expect for a filter method, as the act of "filtering" an Either to me implies reducing the set of valid Rights, with the only way of doing so being to provide (via `raise(L)`) an replacement `Left` value.

```kotlin
  either<String, Int> { 5 }
    .filter { ensure(it != 5) { "5 is right out" } }
```

nomisRev:
Looking at the signatures, and naming, and thinking about Kotlin Std I can’t help but feel like the `map` function should include `Raise<Error>` and the other function should be named `mapError` 😅 Similar to `Flow.map` having `suspend` and Result mapCatching.

Edit: feels like I offended functor..

kyay10:
`mapCatching` is the closest example here, so maybe `mapBinding` could be good? But then we actively want an `also` style thing, so maybe `alsoBinding` or something?

Actually maybe `letBinding` and `alsoBinding` might be the way here.

kyay10:
Note that `mapCatching` is basically a direct-style flatMap. As in, it's a flatMap that also sets up a `runCatching` handler. Perhaps thus we should have like `mapEither`? Doesn't sound that good IMO.

tKe:
A DSL-enabled flatmap was what I was alluding to with this snippet (here named `mapRaise`).

> ```kotlin
>   either<String, Int> { 5 }
>     .mapRaise {
>       ensure(it != 5) { "5 is right out" }
>       it
>     }
> ```


As @serras pointed out though, the intent for this PR was a capability to have a one-liner that either changes a right to a left or leaves it untouched (unlike the above which can change the right value too).

The above I feel is effectively filtering the right in the same way as `takeIf`, except with the need to provide an alternative left value. I think the most important aspect is to ensure reading the code reflects the intent. Personally if I saw a codebase with `.filter { ensure(cond) { errorValue } }` I would know what it's doing.

I think introducing Raise to `map` in a separate PR would make a lot of sense but I expect we'd hit a similar issue with source compatibility, necessitating an alternative name such as `mapRaising`.

serras:
I think we're over-generalizing here (me being the first one), so let's figure out the actual goals here.

I think we want to expose something similar to `takeIf`, I think we should just reuse the `ensure` name, or maybe something like `filter`.

```kotlin
fun failOnMoreConditionsWithBindMap(): Either<String, Int> =
    randomNumber()
        .ensure({ it != 100 }){ "Number 10 also not allowed" }
        .map { it + 100 }
```

If we want a run a more general function (of the form `Raise<A>.(B) -> Unit`), I think we are doing _validation_.

```kotlin
fun failOnMoreConditionsWithBindMap(): Either<String, Int> =
    randomNumber()
        .validate { ensure(t != 100) { "Number 10 also not allowed" } }
        .map { it + 100 }
```

My concrete proposal is to add those two methods on `Either<A, B>`:
- `fun ensure(condition: (B) -> Boolean, failure: () - > A): Either<A, B>`
- `fun validate(block: Raise<A>.(B) -> Unit): Either<A, B>`

If we want to keep the API fully similar to `Raise`, we could also include `fun ensureNotNull(failure: () -> A): Either<A, B>`

kyay10:
I'm very happy with `validate`. Not very sure about `ensure` because of the double lambdas, but I can be easily convinced.

kyay10:
I'm very happy with `validate`. Not quite sure about `ensure` because of the double lambdas, but I can be easily convinced

tKe:
In the interest of keeping the API footprint of Arrow lean, I'd suggest adding `validate` is sufficient. It composes well with the existing `Raise` DSL and allows for both `ensure` and `ensureNotNull` (in addition to any other pertinent extensions one might have on aparticular `Raise<E>`). 

The only benefit of separate `Either<A, B>.ensure` and `Either<A, B>.ensureNotNull` functions is for those not wanting to use the `Raise` DSL. At this point I'd think `Raise` is fairly integral to arrow and should ideally be embraced.

PR #3716, opened 2025-09-01
`validate` for `Either`

### LINKED ISSUE #3639: ["Request"] Analogue of Either.recover for turning right values into left values

serras:
Fixes #3639
Closes #3711

This is the result of the discussion in #3711 (it was easier for me to create a new PR than to rollback all the changes there). This PR introduces a single `validate` function that runs a `Raise` block on `Right` values, but leaves them unchanged until something `raise`s. This seems quite effective in combination with `ensure` to write one-liners:

```kotlin
fun failOnMoreConditionsWithBindMap(): Either<String, Int> =
    randomNumber()
        .validate { ensure(it != 10) { "Number 10 also not allowed" } }
        .map { it + 100 }
```

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
@@ -795,7 +795,7 @@ public sealed class Either<out A, out B> {
    */
   public fun validate(validation: Raise<@UnsafeVariance A>.(B) -> Unit): Either<A, B> = when (this) {
     is Left -> this
-    is Right -> either { validation(value) ; value }
+    is Right -> fold({ validation(value) }, { Left(it) }, { this })
   }
 
   /**
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
@@ -9,6 +9,7 @@ import arrow.core.Either.Right
 import arrow.core.Either.Right.Companion.unit
 import arrow.core.raise.Raise
 import arrow.core.raise.either
+import arrow.core.raise.fold
 import kotlin.contracts.ExperimentalContracts
 import kotlin.contracts.InvocationKind
 import kotlin.contracts.contract
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
@@ -773,6 +773,32 @@ public sealed class Either<out A, out B> {
   public fun getOrNone(): Option<B> = fold({ None }, { Some(it) })
 
   /**
+   * Runs some [validation], expressed as a lambda that may [raise], on [Right] values.
+   * If the [validation] is successful, the value is left unchanged.
+   * The main use case is in combination with [ensure] and other validation functions.
+   *
+   * ```kotlin
+   * import arrow.core.Either
+   * import arrow.core.raise.ensure
+   * import io.kotest.matchers.shouldBe
+   *
+   * fun test() {
+   *   val one: Either<String, Int> = Either.Right(1)
+   *   one.validate { ensure(it > 0) { "negative" } } shouldBe one
+   *
+   *   val zero: Either<String, Int> = Either.Right(0)
+   *   zero.validate { ensure(it > 0) { "negative" } } shouldBe Either.Left("negative")
+   * }
+   * ```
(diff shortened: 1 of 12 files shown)


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -772,6 +772,32 @@ public sealed class Either<out A, out B> {
...
+   *   val zero: Either<String, Int> = Either.Right(0)
+   *   zero.validate { ensure(it > 0) { "negative" } } shouldBe Either.Left("negative")
+   * }
+   * ```
+   * <!--- KNIT example-either-32.kt -->
+   * <!--- TEST lines.isEmpty() -->
+   */
+  public fun validate(validation: Raise<@UnsafeVariance A>.(B) -> Unit): Either<A, B> = when (this) {
+    is Left -> this
+    is Right -> either { validation(value) ; value }
```
tKe: Very minor, but we could avoid creating new `Right` by using fold directly:

```suggestion
    is Right -> fold({ validation(value) }, { Left(it) }, { this })
```

kyay10: This should probably be `inline`

kyay10 (approved):

PR #3723, opened 2025-09-09
Delegate to underlying `toString` in `NonEmpty` collections

serras:


tKe (approved):
The only downside of delegating `toString` directly is the potential to "leak" the underlying delegate's `toString`, potentially mis-representing the type. 
That said, the only time this would be reached is if you explicitly wrapped a collection with a custom toString which requires opting in to `@PotentiallyUnsafeNonEmptyOperation`.

kyay10:
I'm actually very in support of delegating like this. `NonEmptyList` is basically a restricted/refined type. All that it does is add an extra invariant on the underlying collection (and thus provide better types for things like `first`). Thus, I don't think it should change its behaviour at all.

PR #3725, opened 2025-09-10
Add better contracts using 2.2.20

serras:
Unfortunately adding these poisons the binary, so we cannot release them yet :(


arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Either.kt
```diff
@@ -501,8 +501,8 @@ public sealed class Either<out A, out B> {
    */
   public fun isRight(): Boolean {
     contract {
-      returns(true) implies (this@Either is Right)
-      returns(false) implies (this@Either is Left)
+      returns(true) implies (this@Either is Right<B>)
+      returns(false) implies (this@Either is Left<A>)
```
nomisRev, line 505: This is so amazing 😍

kyay10: Wasn't this possible already, or am I stupid lol? Maybe I tried it on 2.2.20 as well.

serras: Before 2.2.20, the compiler was only aware of the "raw type", so you sometimes would lose some type information. In some versions including the generic parameter would even confuse the compiler, since it would create a "new `A` type parameter", instead of understanding it was a reference to the actual one -- this is the reason the current code only states `is Left`.

But in the new version you can declare contracts about [generic parameters](https://kotlinlang.org/docs/whatsnew2220.html#support-for-generics-in-contract-type-assertions), so this part can be readily improved.

kyay10: Oh that's incredible! I have some usecases for hacky DSL stuff that I'll be trying out soon then.
I thought that this was allowed when the type parameter was known because it was a subclass of a class with type parameters though. Like AFAIK this `Left` wasn't actually a raw type, it was secretly a `Left<A>` that the compiler accepts a shorthand for just like when you do `when(either) { is Left -> ... }`. Maybe I'm misremembering though...

kyay10: This compiles fine on 2.0.21 ([playground](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMi4wLjIxIiwicGxhdGZvcm0iOiJqYXZhIiwiYXJncyI6IiIsIm5vbmVNYXJrZXJzIjp0cnVlLCJ0aGVtZSI6ImlkZWEiLCJjb2RlIjoiaW1wb3J0IGtvdGxpbi5jb250cmFjdHMuKlxuZnVuIG1haW4oKSB7XG4gICAgdmFsIGZvbzogRWl0aGVyPEludCwgU3RyaW5nPiA9IEVpdGhlci5SaWdodChcImhpXCIpXG4gICAgaWYgKCFmb28uaXNMZWZ0KCkpIHtcbiAgICAgICAgcHJpbnRsbihmb28udmFsdWUpXG4gICAgfVxufVxuXG5zZWFsZWQgY2xhc3MgRWl0aGVyPG91dCBBLCBvdXQgQj4ge1xuICAgIGNsYXNzIFJpZ2h0PEI+KHZhbCB2YWx1ZTogQik6IEVpdGhlcjxOb3RoaW5nLCBCPigpXG4gICAgY2xhc3MgTGVmdDxBPjogRWl0aGVyPEEsIE5vdGhpbmc+KClcbiAgICBmdW4gaXNMZWZ0KCk6IEJvb2xlYW4ge1xuICAgICAgICBjb250cmFjdCB7XG4gICAgICAgICAgICByZXR1cm5zKHRydWUpIGltcGxpZXMgKHRoaXNARWl0aGVyIGlzIExlZnQ8QT4pXG4gICAgICAgICAgICByZXR1cm5zKGZhbHNlKSBpbXBsaWVzICh0aGlzQEVpdGhlciBpcyBSaWdodDxCPilcbiAgICAgICAgfVxuICAgICAgICByZXR1cm4gdGhpcyBpcyBMZWZ0XG4gICAgfVxufSJ9)):
```kotlin
import kotlin.contracts.*
fun main() {
    val foo: Either<Int, String> = Either.Right("hi")
    if (!foo.isLeft()) {
        println(foo.value)
    }
}

sealed class Either<out A, out B> {
    class Right<B>(val value: B): Either<Nothing, B>()
    class Left<A>: Either<A, Nothing>()
    fun isLeft(): Boolean {
        contract {
            returns(true) implies (this@Either is Left<A>)
            returns(false) implies (this@Either is Right<B>)
        }
        return this is Left
    }
}
```

kyay10: Is the idea here that `is Left<A>` and `is Left` used to cause differing behaviour pre 2.2.20, and now also cause differening behaviour, with `is Left<A>` having the behaviour we desire?
I have no problem with this change; it just feels like `is Left` and `is Left<A>` should have the same behaviour now, and anything other than that is a compiler bug.
Or are we just doing `is Left<A>` post 2.2.20 for the sake of code style?

serras:
This doesn't seem to be needed, although we should check whether there are even more niceties coming.

kyay10:
Especially `holdsIn` might be useful in some places

PR #3727, opened 2025-09-12
Fix too strict validation on optics generator

serras:
This fixes the problems reported in [this repo](https://github.com/StochasticTinkr/broken-arrow).

In particular, we were too eager in validating that references in a class with generated optics were valid. However, if they refer to the optics themselves, this validation needs to happen later.


arrow-libs/optics/arrow-optics-ksp-plugin/src/test/kotlin/arrow/optics/plugin/DSLTests.kt
```diff
@@ -246,4 +247,43 @@ class DSLTests {
       |val r = i != null
       """.evals("r" to true)
   }
+
+  @Test
+  fun `Multiple files`() {
```
StochasticTinkr, line 252: This test looks significantly different in both form AND style than the ones above it.  Seems like maybe this is either the wrong place for the test, or the wrong way to implement it.  It's not my codebase, so maybe I'm being overly pedantic, was just an observation.

serras: Yes, I had to create a new kind of tests because up to now we assumed we had only one file to compile.

PR #3736, opened 2025-09-22
Add missing `-jvm` dependencies to `arrow-stack` BOM

### LINKED ISSUE #3732: ["Request"] Consider adding `-jvm` dependencies to arrow-stack BOM

2.1.2

What would you like to see?

Hi. I would like to see dependencies with `-jvm` suffix listed in `arrow-stack` BOM.
Currently, [`arrow-stack` BOM](https://repo1.maven.org/maven2/io/arrow-kt/arrow-stack/2.1.2/arrow-stack-2.1.2.pom) sometimes includes just other BOMs instead of direct references to target jar-POMs. For example, it includes `io.arrow-kt:arrow-core` (which is just another BOM), but not `io.arrow-kt:arrow-core-jvm` (which is a regular jar library)

Why is it a problem?

Consider the following simple Maven project:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project>
....
    <dependencyManagement>
        <dependencies>
            <dependency>
                <groupId>io.arrow-kt</groupId>
                <artifactId>arrow-stack</artifactId>
                <version>2.1.2</version>
                <type>pom</type>
                <scope>import</scope>
            </dependency>
        </dependencies>
    </dependencyManagement>

    <dependencies>
        <dependency>
            <groupId>io.arrow-kt</groupId>
            <artifactId>arrow-core-jvm</artifactId>
        </dependency>
    </dependencies>

</project>
```

It won't compile, it will fail with the following error:

```text
❯ mvn clean compile
[INFO] Scanning for projects...
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'dependencies.dependency.version' for io.arrow-kt:arrow-core-jvm:jar is missing. @ line 27, column 21
 @
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]
[ERROR]   The project com.example:arrow-bom-problem:1.0-SNAPSHOT (/Users/user/Documents/sandbox/arrow-bom-problem/pom.xml) has 1 error
[ERROR]     'dependencies.dependency.version' for io.arrow-kt:arrow-core-jvm:jar is missing. @ line 27, column 21
```

Gradle also won't handle that:

```kotlin
dependencies {
    implementation(platform("io.arrow-kt:arrow-stack:2.1.2"))
    implementation("io.arrow-kt:arrow-core-jvm")
}
```

```text
> Could not resolve all files for configuration ':lib:compileClasspath'.
   > Could not find io.arrow-kt:arrow-core-jvm:.
     Required by:
         project ':lib
```

The solution to compilation issue is switching from declaration of `arrow-core-jvm` jar to declaration of `arrow-core` "BOM" (where the latter is explicitly declared in `arrow-stack` BOM, that's why it works with no explicit version provided).

The downside of declaring non-`-jvm` dependency is that:
1) there is an extra hop during dependencies resolution
2) `mvn dependency:analyze` will complain:

```
[INFO] --- dependency:3.7.0:analyze (default-cli) @ arrow-bom-problem ---
[WARNING] Used undeclared dependencies found:
[WARNING]    io.arrow-kt:arrow-core-jvm:jar:2.1.2:compile
[WARNING] Unused declared dependencies found:
[WARNING]    io.arrow-kt:arrow-core:jar:2.1.2:compile
```

Maven complains because the general best practice is:
- to explicitly declare dependencies, directly consumed by your project
- and not to declare dependencies, which are not needed (neither at compile or runtime time)

So, I would be happy to declare `arrow-core-jvm` with no build failure and no warnings from from `mvn dependency:analyze`.

For reference, here are other BOMs, that include both dependencies (with and without `-jvm` suffix):

- [Ktor BOM](https://repo1.maven.org/maven2/io/ktor/ktor-bom/3.3.0/ktor-bom-3.3.0.pom)
- [Kotlin AWS SDK BOM](https://repo1.maven.org/maven2/aws/sdk/kotlin/bom/1.5.40/bom-1.5.40.pom)

gavvvr:
Fixes #3732


serras:
Thanks very much for the fix!

PR #3737, opened 2025-09-22
Integration module for Result4k

serras:
Just a few helpers for using the `Raise` DSL with [Result4k](https://github.com/fork-handles/forkhandles/tree/trunk/result4k)

### DIFF
--- a/arrow-libs/integrations/arrow-core-result4k/build.gradle.kts
+++ b/arrow-libs/integrations/arrow-core-result4k/build.gradle.kts
@@ -1,0 +1,17 @@
+plugins {
+  id(libs.plugins.kotlin.jvm.get().pluginId)
+  id("arrow.kotlin")
+}
+
+kotlin {
+  compilerOptions.freeCompilerArgs.add("-Xcontext-parameters")
+}
+
+dependencies {
+  api(projects.arrowCore)
+  api(libs.result4k)
+
+  testImplementation(projects.arrowCore)
+  testImplementation(kotlin("test"))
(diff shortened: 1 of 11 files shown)


arrow-libs/integrations/arrow-core-result4k/build.gradle.kts
```diff
@@ -0,0 +1,17 @@
+plugins {
+  id(libs.plugins.kotlin.jvm.get().pluginId)
+  id("arrow.kotlin")
+}
+
+kotlin {
+  compilerOptions.freeCompilerArgs.add("-Xcontext-parameters")
```
tKe, line 7: 🤔 is it worth adding this to the `arrow.kotlin` convention plugin yet?

arrow-libs/integrations/arrow-core-result4k/src/main/kotlin/arrow/core/raise/Result4k.kt
```diff
@@ -0,0 +1,33 @@
+@file:OptIn(ExperimentalTypeInference::class, ExperimentalContracts::class)
```
tKe, line 2: Is it worth giving this a nicer `@file:JvmName("Result4k")`?

arrow-libs/integrations/arrow-core-result4k/build.gradle.kts
serras: Since it's still experimental, I'd rather keep it only on those projects that need it.

PR #3740, opened 2025-09-25
Resource-scoped coroutine scope helpers

tKe:
I'm not entirely certain if there's some hidden nuance on these, but the intent is to provide "unwrapped" coroutine scopes bound to the resourceScope lifecycle that could then be passed/composed into other resources.


arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -538,3 +546,28 @@ public fun <A : AutoCloseable> autoCloseable(
 ): Resource<A> = resource {
   autoCloseable(closingDispatcher, autoCloseable)
 }
+
```
kyay10, line 547: These are really neat!

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -538,3 +546,28 @@ public fun <A : AutoCloseable> autoCloseable(
...
+    ExitCase.Completed -> complete()
+  }
+}
+
+public suspend fun CompletableJob.completeWithAndJoin(exitCase: ExitCase) {
+  completeWith(exitCase)
+  return join()
+}
+
+private fun ResourceScope.coroutineScope(coroutineContext: CoroutineContext, jobCreator: (Job?) -> CompletableJob): CoroutineScope {
```
kyay10: Nit: could be `inline` I guess, but it's not that big of a deal

tKe: I think in my mind I already had when I factored it out.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -538,3 +546,28 @@ public fun <A : AutoCloseable> autoCloseable(
...
+    is ExitCase.Cancelled -> cancel(exitCase.exception)
+    is ExitCase.Failure -> completeExceptionally(exitCase.failure)
+    ExitCase.Completed -> complete()
+  }
+}
+
+public suspend fun CompletableJob.completeWithAndJoin(exitCase: ExitCase) {
+  completeWith(exitCase)
+  return join()
+}
```
serras: What is the use case of those functions? I'm a bit worried about making the API too big.

tKe: I suppose the only real use-case is if you had a `CompletableJob` from elsewhere you wanted to complete in an `onRelease` block or such.

It probably makes sense to make it private for now. I left it public as it felt useful for composition, but I can't see many reasons to create a completable job that aren't already covered by the scopes in this PR.

I did consider additionally adding something like

```kotlin
public fun ResourceScope.cancelAndJoinOnRelease(job: Job) {
  onRelease { exitCase ->
    when(exitCase) {
      is ExitCase.Cancelled -> job.cancel(exitCase.exception)
      is ExitCase.Failure -> job.cancel("Resource scope failure", exitCase.failure)
      ExitCase.Completed -> job.cancel()
    }
    job.join()
  }
}
```

which would cover tying _any_ job lifecycle (for example, kafka/channel consumer, flow producer) into the resource scope, which I've used as a pattern in a number of services. It's simple to do as a local extension in each service but seems like something that might be available out of the box for resource management.

I'll mark the `completeWith` functions as private for now, and if viable raise another PR for the `cancelAndJoinOnRelease` function for discussion.

tKe: @kyay10 do you see any value in keeping the `completeWith` and `completeWithAndJoin` helpers public?

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -538,3 +545,28 @@ public fun <A : AutoCloseable> autoCloseable(
...
+    is ExitCase.Cancelled -> cancel(exitCase.exception)
+    is ExitCase.Failure -> completeExceptionally(exitCase.failure)
+    ExitCase.Completed -> complete()
+  }
+}
+
+private suspend fun CompletableJob.completeWithAndJoin(exitCase: ExitCase) {
+  completeWith(exitCase)
+  return join()
+}
```
serras: This is more of a style thing, but may you (literally) inline this function into its (single) usage below. I think that it would made for a simpler codebase.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -538,3 +545,28 @@ public fun <A : AutoCloseable> autoCloseable(
...
+
+private inline fun ResourceScope.coroutineScope(coroutineContext: CoroutineContext, jobCreator: (Job?) -> CompletableJob): CoroutineScope {
+  val job = jobCreator(coroutineContext[Job])
+  onRelease { job.completeWithAndJoin(it) }
```
serras:
```suggestion
  onRelease { exitCase ->
    job.completeWith(exitCase)
    job.join()
  }
```

serras (changes requested):

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
tKe: Sure - I split it to mirror `Job`'s `cancel` and `cancelWithJoin` but as it's not exposed now there's no benefit.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -538,3 +544,26 @@ public fun <A : AutoCloseable> autoCloseable(
...
+    job.join()
+  }
+  return CoroutineScope(coroutineContext + job)
+}
+
+public fun ResourceScope.supervisorScope(coroutineContext: CoroutineContext = EmptyCoroutineContext): CoroutineScope =
+  coroutineScope(coroutineContext, ::SupervisorJob)
+
+public fun ResourceScope.coroutineScope(coroutineContext: CoroutineContext = EmptyCoroutineContext): CoroutineScope =
+  coroutineScope(coroutineContext, ::Job)
```
serras: The [`coroutineScope`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/coroutine-scope.html) and [`supervisorScope`](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/supervisor-scope.html) functions use continuation style. It think we should follow the same pattern, or change their names, to avoid surprises.

serras: My suggestion is to use something like `managedCoroutineScope`, `acquireCoroutineScope`, or something along those lines.

serras (changes requested):

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
tKe: I think I was relying on the difference between passing a lambda or not. I also questioned whether they should inherit the `currentCoroutineContext()` by default as `coroutineScope{}` does but decided these are analogous of the `CoroutineScope()` "constructor" and should therefore imply no default context (you can always use `coroutineScope(coroutineContext)` as needed).

Maybe `managed` prefix makes sense though as it's otherwise not as easy to distinguish either by lack of lambda or lack of capitalisation alone.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -538,3 +544,26 @@ public fun <A : AutoCloseable> autoCloseable(
...
+private inline fun ResourceScope.managedCoroutineScope(coroutineContext: CoroutineContext, jobCreator: (Job?) -> CompletableJob): CoroutineScope {
+  val job = jobCreator(coroutineContext[Job])
+  onRelease { exitCase ->
+    job.completeWith(exitCase)
+    job.join()
+  }
+  return CoroutineScope(coroutineContext + job)
+}
+
+public fun ResourceScope.managedSupervisorScope(coroutineContext: CoroutineContext = EmptyCoroutineContext): CoroutineScope =
```
kyay10: Just a suggestion. I'm worried this breaks structured concurrency because it doesn't automatically inherit the current coroutine scope. Maybe a contextual version of it could be nice to have? As in, it takes in `CoroutineScope` as a context, and uses it as the basis for its context, with extras passed in as the argument. This is analogous to `coroutineScope` and `supervisorScope`, but maybe there's something I'm missing here

tKe: I intended this to be analogous to the `CoroutineScope` "constructor" which doesn't automatically inherit context, but with its lifecycle coupled to the containing resource scope. 

Given the nature of resource scope, structured concurrency should be maintained even without the job inheritance. 

One could always pass`currentCoroutineContext()` if truly desired.

serras: I misunderstood the use case here then. I would suggest to emphasize the "constructorness" by naming it `ManagedCoroutineScope` instead. Given the potential for confusion, I also suggest to add an `OptIn` annotation, and describe in the documentation the expected use case.

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -538,3 +544,26 @@ public fun <A : AutoCloseable> autoCloseable(
...
+
+private fun CompletableJob.completeWith(exitCase: ExitCase) {
+  when (exitCase) {
+    is ExitCase.Cancelled -> cancel(exitCase.exception)
+    is ExitCase.Failure -> completeExceptionally(exitCase.failure)
+    ExitCase.Completed -> complete()
+  }
+}
+
+private inline fun ResourceScope.managedCoroutineScope(coroutineContext: CoroutineContext, jobCreator: (Job?) -> CompletableJob): CoroutineScope {
```
serras:
```suggestion
private inline fun ResourceScope.ManagedCoroutineScope(
  coroutineContext: CoroutineContext,
  jobCreator: (Job?) -> CompletableJob
): CoroutineScope {
```

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
serras:
```suggestion
public fun ResourceScope.ManagedSupervisorScope(
  coroutineContext: CoroutineContext = EmptyCoroutineContext
): CoroutineScope =
```

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Resource.kt
```diff
@@ -538,3 +544,26 @@ public fun <A : AutoCloseable> autoCloseable(
...
+    job.completeWith(exitCase)
+    job.join()
+  }
+  return CoroutineScope(coroutineContext + job)
+}
+
+public fun ResourceScope.managedSupervisorScope(coroutineContext: CoroutineContext = EmptyCoroutineContext): CoroutineScope =
+  managedCoroutineScope(coroutineContext, ::SupervisorJob)
+
+public fun ResourceScope.managedCoroutineScope(coroutineContext: CoroutineContext = EmptyCoroutineContext): CoroutineScope =
```
serras:
```suggestion
public fun ResourceScope.ManagedCoroutineScope(
  coroutineContext: CoroutineContext = EmptyCoroutineContext
): CoroutineScope =
```

PR #3759, opened 2025-10-25
Racing + Raise, redux

serras:
This is the result of discussion between @nomisRev and me about how to better integrate `Raise` and the new `Racing` DSL. It seems that the approach in this PR leads to the least amount of surprises: namely, if you start a `racing` block inside a `Raise` context it "does the right thing" and treats `raise`d values as non-successful.

The name `raceOrThrow` has been changed to `raceOrFail` to better showcase that racing works the same with exceptions and `raise`d values. This seemed like a better name than `raceOrThrowOrFail`.

This PR also add experimental annotation to the whole API, which was not but should be there since this API is new.

### DIFF
--- a/arrow-libs/fx/arrow-fx-coroutines/src/commonTest/kotlin/arrow/fx/coroutines/RaiseRacingTest.kt
+++ b/arrow-libs/fx/arrow-fx-coroutines/src/commonTest/kotlin/arrow/fx/coroutines/RaiseRacingTest.kt
@@ -22,7 +22,7 @@ import kotlin.time.Duration.Companion.milliseconds
 
 @ExperimentalRacingApi
 class RaiseRacingTest {
-  private val defaultExceptionHandler = CoroutineExceptionHandler { _, _ ->}
+  private val doNothingExceptionHandler = CoroutineExceptionHandler { _, _ ->}
 
   @Test
   fun immediateWinner() = runTest {
@@ -157,7 +157,7 @@ class RaiseRacingTest {
   }
 
   /* Note on the implementation:
-   * without the outer try/catch the test fails
+   * without the 'doNothingExceptionHandler' the test fails
    * because [runTest] overrides the default coroutine
    * exception handler, conflicting with our own usage.
    */
@@ -165,8 +165,8 @@ class RaiseRacingTest {
   fun testRaceOrRaise() = runTest {
     val result = either {
       racing {
-        race(defaultExceptionHandler) { raise("Test error") }
-        race(defaultExceptionHandler) { "success" }
+        race(doNothingExceptionHandler) { raise("Test error") }
+        race(doNothingExceptionHandler) { "success" }
       }
     }
     assertEquals("success".right(), result)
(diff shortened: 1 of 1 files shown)


arrow-libs/fx/arrow-fx-coroutines/src/commonTest/kotlin/arrow/fx/coroutines/RaiseRacingTest.kt
```diff
@@ -147,29 +147,39 @@ class RaiseRacingTest {
...
           race { awaitCancellation() }
         }
       }
     }
   }
 
+  /* Note on the implementation:
+   * without the outer try/catch the test fails
+   * because [runTest] overrides the default coroutine
+   * exception handler, conflicting with our own usage.
```
kyay10, line 162: I think you just need to pass in a coroutine exception handler into the `race` calls, no?

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Racing.kt
```diff
@@ -14,43 +14,56 @@ import kotlinx.coroutines.selects.select
...
 
-private val defaultCoroutineExceptionHandler = CoroutineExceptionHandler { _, exception -> exception.printStackTrace() }
+private val defaultCoroutineExceptionHandler =
+  CoroutineExceptionHandler { _, exception -> exception.printStackTrace() }
 
 /**
  * Races a coroutine block against other blocks in the racing scope.
- * If the block throws an exception, it will be handled by [CoroutineExceptionHandler].
+ * If the block throws an exception
+ * or raises an error through [arrow.core.raise.Raise],
```
kyay10, line 55: This is incorrect. `catch` rethrows any `RaiseCancellationException`

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/RaiseRacing.kt
```diff
@@ -48,13 +49,117 @@ public fun <Error, Result> RacingScope<Result>.raceOrThrow(
...
+ *
+ * Example 2:
+ *
+ * ```kotlin
+ * import arrow.fx.coroutines.racing
+ * import arrow.core.raise.either
+ *
+ * suspend fun main() {
+ *   val result = either<String, Int> {
+ *     racing<String> {
```
kyay10: This doesn't compile. I believe what you meant is `racing<String, Int>`

arrow-libs/fx/arrow-fx-coroutines/src/commonMain/kotlin/arrow/fx/coroutines/Racing.kt
kyay10: Now it's correct

arrow-libs/fx/arrow-fx-coroutines/src/commonTest/kotlin/arrow/fx/coroutines/RaiseRacingTest.kt
```diff
@@ -147,29 +149,34 @@ class RaiseRacingTest {
...
         racing {
-          raceOrThrow { throw IllegalStateException("Test exception") }
+          raceOrFail { throw IllegalStateException("Test exception") }
           race { awaitCancellation() }
         }
       }
     }
   }
 
+  /* Note on the implementation:
```
kyay10, line 159: Nit: comment is outdated

kyay10 (approved):

PR #3760, opened 2025-10-25
Swallow `RaiseCancellationException` inside race

kyay10:
I felt this was easier to suggest through a PR. I think this might be a better way to go about it. We swallow raises inside `race` (which matches its documentation). Thus, the functionality previously provided by `Raise.racing` is instead provided immediately by `racing`. This is important because one may be using multiple `Raise`s in a scope, and it makes little sense to call `racing { race { racing { ... } } }` nested for every `Raise` in scope. Additionally, comparing `Raise` instances by reference seems a little weak. Someone may have 2 raise instances in scope, and intentionally only do `.racing` on one, only to discover that the 2 instances happen to be the same (lexical effects vs dynamic effects).
If you wish, we can also provide a `raceOrThrow` that only swallows raises, a `raceOrRaise` that only swallows throwables, and so on.


serras:
This seems like a nice way to get the expected behavior in a much easier way. Thanks!

PR #3773, opened 2025-11-02
Enable "unused value checker"

serras:
This still requires 2.3.0-Beta2, but it's good to be forward-looking to these changes.

### DIFF
--- a/arrow-libs/core/arrow-exception-utils/src/commonMain/kotlin/arrow/core/ThrowableUtils.kt
+++ b/arrow-libs/core/arrow-exception-utils/src/commonMain/kotlin/arrow/core/ThrowableUtils.kt
@@ -1,18 +1,20 @@
 package arrow.core
 
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.contract
 import kotlin.coroutines.cancellation.CancellationException
 
 public fun Throwable?.throwIfNotNull() { if (this != null) throw this }
 
-public infix fun Throwable?.mergeSuppressed(other: Throwable?): Throwable? = when {
-  other == null -> this
-  this == null -> other
-  other is CancellationException -> {
-    this.addSuppressed(other)
-    this
+@OptIn(ExperimentalContracts::class)
+public infix fun Throwable?.mergeSuppressed(other: Throwable?): Throwable? {
+  contract {
+    returns(null) implies (this@mergeSuppressed == null && other == null)
   }
-  else -> {
-    this.addSuppressed(other.nonFatalOrThrow())
-    this
+  return when {
+    other == null -> this
+    this == null -> other
+    other is CancellationException -> this.also { addSuppressed(other) }
+    else -> this.also { addSuppressed(other.nonFatalOrThrow()) }
   }
 }
--- a/arrow-libs/integrations/arrow-core-jackson/src/main/kotlin/arrow/integrations/jackson/module/internal/ProductTypeSerializer.kt
+++ b/arrow-libs/integrations/arrow-core-jackson/src/main/kotlin/arrow/integrations/jackson/module/internal/ProductTypeSerializer.kt
@@ -1,7 +1,6 @@
 package arrow.integrations.jackson.module.internal
 
 import arrow.core.Option
-import arrow.core.Some
 import com.fasterxml.jackson.core.JsonGenerator
 import com.fasterxml.jackson.databind.SerializerProvider
 import com.fasterxml.jackson.databind.ser.std.StdSerializer
@@ -15,9 +14,8 @@ public class ProductTypeSerializer<T>(clazz: Class<T>, private val fields: List<
   override fun serialize(value: T, gen: JsonGenerator, provider: SerializerProvider) {
     gen.writeStartObject()
     for (projector in fields) {
-      when (val projected = projector.getOption(value)) {
-        is Some -> provider.defaultSerializeField(projector.fieldName, projected, gen)
-        else -> { }
+      projector.getOption(value).onSome {
+        provider.defaultSerializeField(projector.fieldName, it, gen)
       }
     }
     gen.writeEndObject()
--- a/arrow-libs/integrations/arrow-core-jackson/src/main/kotlin/arrow/integrations/jackson/module/internal/UnionTypeSerializer.kt
+++ b/arrow-libs/integrations/arrow-core-jackson/src/main/kotlin/arrow/integrations/jackson/module/internal/UnionTypeSerializer.kt
@@ -1,7 +1,6 @@
 package arrow.integrations.jackson.module.internal
 
 import arrow.core.Option
-import arrow.core.Some
 import com.fasterxml.jackson.core.JsonGenerator
 import com.fasterxml.jackson.databind.SerializerProvider
 import com.fasterxml.jackson.databind.ser.std.StdSerializer
@@ -18,9 +17,8 @@ public class UnionTypeSerializer<T>(clazz: Class<T>, private val fields: List<Pr
         "unexpected failure when attempting projection during serialization"
       }
     gen.writeStartObject()
-    when (val projected = project.getOption(value)) {
-      is Some -> provider.defaultSerializeField(project.fieldName, projected, gen)
-      else -> { }
+    project.getOption(value).onSome {
+      provider.defaultSerializeField(project.fieldName, it, gen)
     }
     gen.writeEndObject()
   }


arrow-libs/core/arrow-atomic/src/commonMain/kotlin/arrow/atomic/Atomic.kt
```diff
@@ -1,4 +1,5 @@
 @file:OptIn(ExperimentalContracts::class)
+@file:Suppress("API_NOT_AVAILABLE")
```
kyay10: I don't actually know what this does, so maybe a comment explaining what this does is good.

arrow-libs/core/arrow-atomic/src/commonMain/kotlin/arrow/atomic/Atomic.kt
```diff
@@ -38,6 +39,7 @@ public expect class Atomic<V>(initialValue: V) {
    * Compare current value with expected and set to new if they're the same. Note, 'compare' is checking
    * the actual object id, not 'equals'.
    */
+  @IgnorableReturnValue
```
kyay10: Should this be ignorable? The `set` doesn't happen if the object isn't the same. It seems like one should always check the return value thus. This isn't like `MutableSet.add` where it doesn't matter

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -1,5 +1,5 @@
 @file:OptIn(ExperimentalTypeInference::class, ExperimentalContracts::class, ExperimentalStdlibApi::class)
-@file:Suppress("API_NOT_AVAILABLE")
```
kyay10, line 2: Isn't this allowed now? I'll check, but I thought `equals` for value class landed already

arrow-libs/integrations/arrow-core-jackson/src/main/kotlin/arrow/integrations/jackson/module/internal/ProductTypeSerializer.kt
```diff
@@ -14,8 +15,9 @@ public class ProductTypeSerializer<T>(clazz: Class<T>, private val fields: List<
   override fun serialize(value: T, gen: JsonGenerator, provider: SerializerProvider) {
     gen.writeStartObject()
     for (projector in fields) {
-      projector.getOption(value).map {
-        provider.defaultSerializeField(projector.fieldName, it, gen)
+      when (val projected = projector.getOption(value)) {
```
kyay10: Nit: maybe use `onSome`

arrow-libs/core/arrow-atomic/src/commonMain/kotlin/arrow/atomic/Atomic.kt
serras: Otherwise we get a warning because we're targeting stdlib 2.1, but the annotation is not available until 2.3.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
serras: I've moved them closer to the `equals` implementation.

kyay10: I meant that we should use `-XXLanguage:+CustomEqualsInValueClasses` which I think enables this. Instead of just suppressing the error

arrow-libs/core/arrow-atomic/src/commonMain/kotlin/arrow/atomic/Atomic.kt
serras: Fixed in a newest commit

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
serras: I think this is mostly stylistic: I prefer to keep the suppression instead of enabling a very-experimental language feature.

kyay10 (approved):

arrow-libs/core/arrow-autoclose/api/android/arrow-autoclose.api
```diff
@@ -22,6 +22,6 @@ public final class arrow/DefaultAutoCloseScope : arrow/AutoCloseScope {
 }
 
 public final class arrow/ThrowIfFatalKt {
-	public static final fun throwIfFatal (Ljava/lang/Throwable;)Ljava/lang/Throwable;
```
kyay10, line 25: This is a binary breaking change...

arrow-libs/core/arrow-exception-utils/src/commonMain/kotlin/arrow/core/nonFatalOrThrow.kt
```diff
@@ -35,6 +33,9 @@ package arrow.core
  *
  */
 // https://youtrack.jetbrains.com/issue/KT-36036
-@IgnorableReturnValue
 public fun Throwable.nonFatalOrThrow(): Throwable =
```
kyay10, line 36: nit: maybe implement in the opposite way? As in, have `Throwable.throwIfFatal() { if (!NonFatal(this)) throw this }` and then `Throwable.nonFatalOrThrow() = also { throwIfFatal() }`

arrow-libs/optics/arrow-optics/src/commonTest/kotlin/arrow/optics/test/laws/Laws.kt
```diff
@@ -9,8 +11,9 @@ interface LawSet {
   val laws: List<Law>
 }
 
-data class Law(val name: String, val test: suspend TestScope.() -> Unit)
+data class Law(val name: String, val test: suspend TestScope.() -> Any?)
```
kyay10: I don't really see the point of this. `Unit` coercion works well enough, no? and the return value of `test` isn't used anywhere seemingly

arrow-libs/optics/arrow-optics/src/commonTest/kotlin/arrow/optics/test/laws/Laws.kt
```diff
@@ -9,8 +11,9 @@ interface LawSet {
...
 
+@IgnorableReturnValue
 fun <A> A.equalUnderTheLaw(b: A, f: (A, A) -> Boolean = { x, y -> x == y }): Boolean =
```
kyay10: Since this is test code, I'd rather just change this to return `Unit` since it can never return `false`

arrow-libs/resilience/arrow-resilience/src/commonTest/kotlin/arrow/resilience/ScheduleTest.kt
```diff
@@ -237,7 +239,7 @@ class ScheduleTest {
     val iterations = 20_000L
     val l = Either.catch {
       Schedule.recurs<Throwable>(iterations).retry {
-        count.updateAndGet { it + 1 }
+        val _ = count.updateAndGet { it + 1 }
```
kyay10: nit: use update instead

arrow-libs/optics/arrow-optics/src/commonTest/kotlin/arrow/optics/test/laws/Laws.kt
serras: This makes it clear that the `Law` "consumes" the value, so the unused return value checker knows of that fact.

arrow-libs/core/arrow-autoclose/api/android/arrow-autoclose.api
serras: Fixed in the latest commit

arrow-libs/optics/arrow-optics/src/commonTest/kotlin/arrow/optics/test/laws/Laws.kt
serras: That's a good idea, changed in one of the latest commits.

arrow-libs/optics/arrow-optics/src/commonTest/kotlin/arrow/optics/test/laws/Laws.kt
serras: I've fixed to make it more clear you need to give a `PropertyContext` there.

serras:
I've ended up splitting the things related to exceptions and `addSuppressed` into its own small module, to avoid code duplication. @kyay10 would you mind having a look at it, since you recently touched that code?

The whole things is binary compatible: even though some things are moved to `arrow-exceptions-utils`, we still use the same package, and it's exported through `arrow-core`. So nobody needs to change anything (it they were doing so). For the one in `arrow-autoclose`, I've created a small wrapper re-exporting it.

arrow-libs/core/arrow-exception-utils/src/commonMain/kotlin/arrow/core/nonFatalOrThrow.kt
kyay10: @serras highlighting this again, but otherwise I like the new changes!

arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/ThrowIfFatal.kt
```diff
@@ -0,0 +1,11 @@
+package arrow
+
+import arrow.core.nonFatalOrThrow
+
+@PublishedApi
+@Deprecated(
+  message = "Moved to common arrow-exception-utils module",
```
kyay10, line 7: nit: message and replaceWith don't matter for a hidden deprecated method, let alone one that's also `internal`

arrow-libs/core/arrow-core/api/android/arrow-core.api
```diff
@@ -600,14 +600,6 @@ public final class arrow/core/NonEmptySetKt {
 	public static final fun wrapAsNonEmptySetOrThrow (Ljava/util/Set;)Ljava/util/Set;
 }
 
-public final class arrow/core/NonFatalKt {
```
kyay10, line 603: Mildly annoying that the abi dump doesn't understand Gradle `api()` dependencies, but what can ya do!

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
```diff
@@ -1013,8 +1015,7 @@ public open class RaiseAccumulate<Error> @ExperimentalRaiseAccumulateApi constru
 
   @PublishedApi internal class Ok<out A>(override val value: A): Value<A>()
 
-
-  @ExperimentalRaiseAccumulateApi
+  @ExperimentalRaiseAccumulateApi @IgnorableReturnValue
   public inline fun <A> accumulating(block: RaiseAccumulate<Error>.() -> A): Value<A> {
```
kyay10, line 1019: hmmmm, I'm a bit worried about this one.
This means that `accumulating { 42 }` results in no warnings. Effectively, `accumulating` is a blackhole.
What I wish we had is some way to tell the compiler that `Value<Unit>` can be safely ignored (and so can `Value<Nothing>`, obviously), but other such ones can't.
I've had a deep think about this already, but I couldn't find a way around this beside a sister function named `accumulatingUnit` or something like that.
Things I've tried: OverloadResolutionByLambdaReturnType, kotlin.internal.* annotations.
I'm happy to discuss this later. I'm fine with this, even though it may be mildly unhelpful, and we can always remove the annotation later, no?

arrow-libs/core/arrow-exception-utils/src/commonMain/kotlin/arrow/core/ThrowableUtils.kt
```diff
@@ -0,0 +1,18 @@
+package arrow.core
+
+import kotlin.coroutines.cancellation.CancellationException
+
+public fun Throwable?.throwIfNotNull() { if (this != null) throw this }
+
+public infix fun Throwable?.mergeSuppressed(other: Throwable?): Throwable? = when {
```
kyay10: nit: contracts? No need for the new `condition implies returnsX` contracts, just the normal ones
I believe it should be like:
```kotlin
contract { returns(null) implies (this@mergeSuppressed == null && other == null)
```
Then I believe you don't need `throwIfNotNull()` later on (as long as the compiler is smart enough to understand contrapositive), and it can probably be removed then (cutting down on public API)

arrow-libs/integrations/arrow-core-jackson/src/main/kotlin/arrow/integrations/jackson/module/internal/ProductTypeSerializer.kt
kyay10: In fact, that's a behaviour change! It used to use `projected.value`, and now it uses `projected`!

arrow-libs/integrations/arrow-core-jackson/src/main/kotlin/arrow/integrations/jackson/module/internal/UnionTypeSerializer.kt
```diff
@@ -17,7 +18,10 @@ public class UnionTypeSerializer<T>(clazz: Class<T>, private val fields: List<Pr
         "unexpected failure when attempting projection during serialization"
       }
     gen.writeStartObject()
-    project.getOption(value).map { provider.defaultSerializeField(project.fieldName, it, gen) }
+    when (val projected = project.getOption(value)) {
```
kyay10: Behaviour change! `it` refers to `projected.value`, not `projected`

kyay10:
Ignore the nits, we can address them later! Just these 2 behaviour changes highlighted above are all I need.

arrow-libs/core/arrow-autoclose/src/commonMain/kotlin/arrow/ThrowIfFatal.kt
serras: I'm never sure how this is showcased in Java code (not that anyone is consuming it that way). Anyway, I think it's a good reminder for ourselves.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/RaiseAccumulate.kt
serras: As far as I know, there will be new ways to mark those functions with "ignore the return value if the return of `block` is ignored". But for now, I think this gives the best developer experience.

arrow-libs/core/arrow-exception-utils/src/commonMain/kotlin/arrow/core/ThrowableUtils.kt
serras: That was a good idea, added!

arrow-libs/integrations/arrow-core-jackson/src/main/kotlin/arrow/integrations/jackson/module/internal/UnionTypeSerializer.kt
serras: I think it's correct, `it` used to refer to `projected.value`, and that's exactly what `projected` is. Note that there's a `project` and a `projected` (yes, not that great on my side).

serras: Anyway, I've changed it to use `onSome`

arrow-libs/core/arrow-exception-utils/src/commonMain/kotlin/arrow/core/nonFatalOrThrow.kt
serras: Is this a blocker for you? I'd rather keep the code the same in this first go.

kyay10: No, not a blocker!

Issue #3777, opened 2025-11-11
["Request"] Additional context bridge functions for Raise DSL should be added


Zordid:
What version are you currently using?
2.2.0

What would you like to see?
With 2.2.0 some, but not all, bridge functions for the Raise DSL have been added in the package `arrow.core.raise.context.raise`, those include the basics `raise`, `bind`, `ensure`, etc.

But: there are more functions that still lack the appropriate context bridge functions, namely:

* `public inline fun <Error, A> Raise<NonEmptyList<Error>>.forEachAccumulating` with lots of overloads for different parameters

* `public inline fun <Error, A, B> Raise<NonEmptyList<Error>>.mapOrAccumulate` with lots of overloads for different parameters

* zipOrAccumulate...

I only listed the ones that I currently use - surely there are more.

Would be great to also see them in the next release. Maybe there's a better way than to add all of them in the `context` package?

PR #3778, opened 2025-11-15
add high arity raise accumulate context support

tibtof:
Partial implementation for #3777. Needed the high-arity support for `zipOrAccumulate` in my project, and since those overloads were already implemented locally I decided to open this PR. Added a single test covering the new overloads as a starting point; if the approach and style are acceptable, I can extend the coverage to the remaining variants.


serras:
Thanks for your work! However, the overloads you've added should go on the main `arrow-core` module, the `arrow-high-arity` module is for more than 9 parameters.

tibtof:
Redo in #3782

PR #3782, opened 2025-11-19
Additional overloads for zipOrAccumulate

tibtof:
Partial implementation for https://github.com/arrow-kt/arrow/issues/3777. Needed the high-arity support for zipOrAccumulate in my project, and since those overloads were already implemented locally I decided to open this PR. Added a single test covering the new overloads as a starting point; if the approach and style are acceptable, I can extend the coverage to the remaining variants.

Redo of #3778, I didn't pay attention and work on the main branch with that and it wasn't possible to change the source branch for the PR.


tibtof:
@serras thanks for the review and feedback. And sorry for giving you more work with the pipeline fixes than what I did for this PR 😔

serras:
@tibtof no problem at all, thanks for your contribution! In theory all the API files update and styling should be applied automatically, but it doesn't work quite well on forks :/

PR #3793, opened 2025-11-28
Unify handling of addSuppressed

kyay10:
Only behaviour change is that `guaranteeCase` now matches the behaviour of `resourceScope` in that `guaranteeCase({ throw e }) { throw CancellationException() }` Now throws `e` with the cancellation suppressed. 

The logic is spelled-out on purpose, to make editing easier in the future. In particular, we might want to deprioritize `CancellationException`.

### DIFF
--- a/arrow-libs/resilience/arrow-resilience/api/android/arrow-resilience.api
+++ b/arrow-libs/resilience/arrow-resilience/api/android/arrow-resilience.api
@@ -89,7 +89,6 @@ public final class arrow/resilience/CircuitBreaker$State$Open : arrow/resilience
 	public fun getOpeningStrategy ()Larrow/resilience/CircuitBreaker$OpeningStrategy;
 	public final fun getResetTimeout-UwyO8pc ()J
 	public final fun getStartedAt ()Lkotlin/time/TimeMark;
-	public fun hashCode ()I
 	public fun toString ()Ljava/lang/String;
 }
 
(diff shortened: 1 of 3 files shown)


arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/CircuitBreaker.kt
```diff
@@ -168,91 +183,73 @@ private constructor(
...
-    }
+    return Either.catchOrThrow<ExecutionRejected, _> { protectOrThrow(fa) }
   }
 
   /**
    * Returns a new task that upon execution will execute the given task, but with the protection of this circuit breaker.
    * If an exception in [fa] occurs it will be rethrown
    */
-  public tailrec suspend fun <A> protectOrThrow(fa: suspend () -> A): A {
+  public suspend fun <A> protectOrThrow(fa: suspend () -> A): A {
```
serras: I think losing the `tailrec` here is not good. With `tailrec` we ensure that running `protectOrThrow` would not give a stack overflow -- something that is quite important if this needs to run for a lot of iterations.

kyay10: It's no longer recursive. It uses `update` now instead of `compareAndSet` and recursing

serras:
I don't want to sound harsh, but I would prefer not to have this kind of "doing lots of things together" PRs. You may touch a lot of files -- for #3773 does that -- but there should one clear goal in each PR. Here I see at least:
- rewriting some functions using `?` in a clearer format,
- using raise combinators when possible,
- making `SagaScope` use `ResourceScope` under the hood,
-  changing the implementation of `Process` in SuspendApp (what is the impact in JS? for example).

These may require approval from different people, but GitHub has an all-or-nothing behavior for approvals.

kyay10:
Not harsh at all! I did admittedly go overboard with some refactorings here. I'll split this up into separate PRs!

kyay10:
Not harsh at all! I did admittedly go overboard with some refactorings here. I'll split this up into separate PRs!

arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/CircuitBreaker.kt
nomisRev: IIRC I used `compareAndSet` specifically so I could implement it as `tailrec` 😁

kyay10:
The PR now only refactors the suppression logic in `ResourceScope`, `AutoCloseScope`, and `guaranteeCase`, all of which are intimately related.

serras (approved):

PR #3794, opened 2025-11-29
kyay10/saga-as-resource

kyay10:
The error suppression logic now matches `resourceScope`. Even though `SagaBuilder` is `@PublishedApi`, it was never used in any `inline` function, so it's technically `private`. For simplicity, I just deprecated it, but theoretically it could be removed


arrow-libs/resilience/arrow-resilience/build.gradle.kts
```diff
@@ -6,8 +6,7 @@ kotlin {
   sourceSets {
     commonMain {
       dependencies {
-        api(projects.arrowCore)
-        implementation(libs.coroutines.core)
+        api(projects.arrowFxCoroutines)
```
serras: My only concern with this PR is that we start depending on `fx-coroutines` in the `resilience` package. @nomisRev do we have any rules in place about this? I remember we wanted to keep these concerns somehow separate.

kyay10: Mu justification for this was that resilience is part of Arrow Fx, and the core of Arrow Fx seems to be fx-coroutines.

arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/Saga.kt
```diff
@@ -129,19 +133,11 @@ internal class SagaBuilder(
   override suspend fun <A> saga(
     action: suspend SagaActionStep.() -> A,
     compensation: suspend (A) -> Unit
-  ): A =
-    guaranteeCase({ action(SagaActionStep) }) { res ->
```
kyay10, line 133: The logic of this change is:

- the `finalizer` lambda doesn't call any coroutines code, so `NonCancellable` is pointless
- when `res` is null, `finalizer` does nothing, so `guaranteeCase` simply catches and rethrows, which is pointless
- when `res` is not null, `guaranteeCase` simply runs the `action`, and then the finalizer here (and again, `NonCancellable` is irrelevant)

Hence, guaranteeCase's code can be simplified to the following (assumptions: `finalizer` returns immediately when `res` is null, and `finalizer` doesn't suspend):
```kotlin
private suspend fun <A> guaranteeCase(
  fa: suspend () -> A,
  finalizer: suspend (value: A?) -> Unit
): A {
  val res = fa()
  finalizer(res)
  return res
}
```
which is just `also`!
There is a nasty bug in that implementation as well, which is that compensation doesn't get scheduled for `null` values! This fixes that as well.
Obviously, this is code left for binary compatibility, so I can roll this back, but it felt nicer just so that I can remove `guaranteeCase` and `runReleaseAndRethrow`

PR #3795, opened 2025-11-29
kyay10/deprecate-saga

kyay10:
`sagaScope` is an interesting name... maybe we should call it `sagaTransact` or even just `transact`. 
An alternative to this PR would be to have `ResourceScope: SagaScope` and not deprecate `SagaScope`.

### DIFF
--- a/arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/Saga.kt
+++ b/arrow-libs/resilience/arrow-resilience/src/commonMain/kotlin/arrow/resilience/Saga.kt
@@ -1,3 +1,5 @@
+@file:Suppress("DEPRECATION")
+
 package arrow.resilience
 
 import arrow.atomic.Atomic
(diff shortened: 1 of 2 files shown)


serras:
As mentioned in Slack, I'm against this change to our API. Even though we could implement `Saga` in terms of `ResourceScope` (as done in other PR), I think the concerns are different enough to have different types. I think it would be quite suprising for users to start a `saga` block and suddenly get access to `autoCloseable`, for example.

kyay10:
Absolutely fair. How would you feel instead about `resourceScope` allowing you to `saga`? `SagaScope` then remains, and `ResourceScope : SagaScope` with default implementations (this is akin to how `ResourceScope : AutoCloseScope`). The only annoying part is we'd have to move the `SagaScope` interface into `arrow-fx-coroutines`. 
I'm happy already with just the implementation moved, but I think it'd be nice to have "the one API" that lets people both `install` and `saga` (i.e. `onRelease` and `compensate` respectively). Keeping `Saga` types is absolutely fine by me.

kyay10:
Closed in favour of #3806

PR #3796, opened 2025-11-29
Use guaranteeCase in CircuitBreaker instead of addSuppressed

kyay10:
Part of the effort to unify `addSuppressed` usages, this delegates to `guaranteeCase`; as a nice bonus, the code is cleaner!

### DIFF
--- a/arrow-libs/resilience/arrow-resilience/build.gradle.kts
+++ b/arrow-libs/resilience/arrow-resilience/build.gradle.kts
@@ -6,9 +6,7 @@ kotlin {
   sourceSets {
     commonMain {
       dependencies {
-        api(projects.arrowCore)
-        implementation(libs.coroutines.core)
-        implementation(projects.arrowExceptionUtils)
+        implementation(projects.arrowFxCoroutines)
       }
     }
     commonTest {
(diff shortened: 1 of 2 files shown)


arrow-libs/resilience/arrow-resilience/build.gradle.kts
```diff
@@ -6,8 +6,7 @@ kotlin {
   sourceSets {
     commonMain {
       dependencies {
-        api(projects.arrowCore)
-        implementation(libs.coroutines.core)
+        api(projects.arrowFxCoroutines)
```
serras: Same concern as with #3794, are we OK depending on `fx-coroutines` in the `resilience` module? @nomisRev

serras: @kyay10 after we merge #3773, we can move `guaranteeCase` (and `Bracket`) to the `arrow-exceptions-utils` module, and then we can depend on that instead.

kyay10: Sounds good to me! I'll open a new (draft) PR based on #3773 that does that move, and we can undraft it and merge it after #3773 is merged

serras: Alas, I've realized that moving `bracket` to `arrow-exception-utils` would require adding a dependency on `kotlinx.coroutines` for that module, and transitively to `arrow-core` and `arrow-autoclose`. And this is something we don't want to do, at least at this point.

serras:
@kyay10 sorry for all the back and forth. I thought splitting `arrow-exception-utils` would solve the dependency on `arrow-fx-coroutines` problem, but it doesn't. I think adding that dependency to `arrow-resilience` is not that bad, and that allows us to unify behavior everywhere. So I would say, let's go with this approach!

serras:
@kyay10 I've rebased the branch on the latest `main`. Feel free to merge if the rebased version is OK for you.

PR #3803, opened 2025-11-30
Add holdsIn contracts

kyay10:
Kept as Draft for now since it poisons the binary


kyay10:
I have a feeling this might not be binary-poisoning anymore. @serras can you confirm if that's true?

PR #3804, opened 2025-11-30
Make ResourceScope implement SagaScope

kyay10:
Alternative to #3795
This unfortunately necssitates moving Saga into arrow-fx-coroutines.


kyay10:
Closed in favour of #3806 
This wasn't easily workable anyway. `Saga.bind()` and `Resource.bind()` conflicted on JVM; maybe there's some crazy suppressions we could use to make it work, but that's already a bad sign

PR #3806, opened 2025-12-01
Introduce resourceSaga

kyay10:
An alternative to #3804 and #3795, this introduces a new handler `resourceSaga` that gives access to both `ResourceScope` and `SagaScope`, in such a way that the compensators and finalizers run in the expected order (which is different from just doing `resourceScope { saga {...}.transact()}`). I went for contexts immediately because making an intersection interface results in conflicting JVM signatures because both `Resource.bind` and `Saga.bind` have the same signature


kyay10:
Converted to a draft temporarily until we add `context` bridges for `Resource` and `Saga`

kyay10:
Turning this into a discussion (#3836) because this code is stale, and it's ultimately adding a single funciton, `resourceSaga`, whose implementation isn't very interesting

PR #3812, opened 2025-12-10
Prevent error accumulation in `withNel` when RCE caught

kyay10:
This is a precedent set by `ior`, which was carried over into `iorAccumulate` and by extension `accumulate` itself. This further extends this rule to `withNel`. The previous behaviour was surprising since a user might expect `ra.withNel { raise(foo.nel()) }` to be equivalent to `ra.raise(foo)` (where `ra: RaiseAccumulate`), but that wasn't true in the presence of `try-catch`.

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/raise/Builders.kt
@@ -172,14 +172,20 @@ public inline fun <Error, A> iorAccumulate(noinline combineError: (Error, Error)
 internal fun <Error> Raise<Error>.IorRaiseAccumulate(
   state: Atomic<Any?>,
   combineError: (Error, Error) -> Error,
-): RaiseAccumulate<Error> = RaiseAccumulate(IorAccumulate(state, combineError, this)) { e -> raise(EmptyValue.combine(state.get(), e, combineError)) }
+): RaiseAccumulate<Error> {
+  val accumulate = IorAccumulate(state, combineError, this)
+  return RaiseAccumulate(accumulate, accumulate, accumulate::raiseSingle)
+}
 
 @ExperimentalRaiseAccumulateApi
 private class IorAccumulate<Error>(
   private val state: Atomic<Any?>,
   private val combineError: (Error, Error) -> Error,
   private val raise: Raise<Error>,
-) : Accumulate<Error> {
+) : Accumulate<Error>, Raise<NonEmptyList<Error>> {
+  fun raiseSingle(e: Error): Nothing = raise.raise(EmptyValue.combine(state.get(), e, combineError))
+  override fun raise(r: NonEmptyList<Error>) = raiseSingle(r.reduce(combineError))
+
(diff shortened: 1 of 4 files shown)


serras:
@kyay10 Could you maybe give an example of where you'll see the change of behavior? I'm having a hard time understanding the whole implications of this change.

kyay10:
The tests prefixed with `tryCatchRecoverRaiseWithNel` are the ones where behaviour changes.
I'll focus on one of them:
```kotlin
  @Test fun tryCatchRecoverRaiseWithNel() = runTest {
    accumulate(::merge) {
      try {
        withNel {
          accumulate(1)
          raise(nonEmptyListOf(2, 4))
        }
      } catch (_: Throwable) { }
      raise(3)
    } shouldBe nonEmptyListOf(1, 3)
  }
```
The previous behaviour is that `withNel` gave you a `RaiseNel<Error>: Raise<NonEmptyList<Error>`, whose raise function is defined as `accumulateAll(r).value`
This means that the call to `raise(nonEmptyListOf(2, 4))` would immediately accumulate `2` and `4`, adding them to the accumulated errors permanently.
Now, `withNel` gives you access to a custom `Raise<NonEmptyList<Error>>`. The behaviour of it is that it will *never* accumulate errors directly, instead it will `raise` somehow. For this particular case, it will ultimatey translate to `raise(nonEmptyListOf(1, 2, 4))` targeted at the `Raise` provided by `merge`.
It will then be stopped by the `catch`, so those errors will (intentionally) disappear forever. Hence, the final list of accumulated errors will just be `1, 3`, since the info about `2, 4` was caught and intentionally discarded
We already have this behaviour in tests for `iorAccumulate` because we had it for `ior`.
In particular, this was already true (although we didn't have a test for it, which I added here):
```kotlin
  @Test fun tryCatchRecoverRaise() = runTest {
     accumulate(::merge) {
      try {
        accumulate(1)
        raise(2)
      } catch (_: Throwable) { }
      raise(3)
    } shouldBe nonEmptyListOf(1, 3)
  }
```
With the same logic applying: `raise(2)` ends up desugaring to a `raise(nonEmptyListOf(1, 2))` targeted at `merge`, but it gets stopped and discarded by the `catch`

This PR effectively makes `RaiseAccumulate<Error>` an intersection of `Accumulate<Error>`, `Raise<Error>`, and `Raise<NonEmptyList<Error>>`, although it hides away the latter and only exposes it through `withNel`.

This reasoning here is why we don't have `Accumulate<Error>: Raise<Error>`, even though we easily could. Not only are they different concerns, `accumulate(x).value` is inheritly different from `raise(x)`, since the former guarantees that `x` is accumulated, while the latter allows for the `RaiseCancellationException` to be caught. I personally like that `raise` only has one (side) effect, namely throwing an exception. `accumulate` also has one (side) effect, which is adding the error to the accumulator. `Value<T>.value` has only one (side) effect, which is throwing if the value doesn't exist.

At the same time, you can argue that this is pointless. We have a precedent in `ior` and `iorAccumulate`, yes, but perhaps we're over-constraining our implementations. I'd love to hear more if you think we shouldn't extend that precedent to this API

PR #3813, opened 2025-12-10
Tolerate access to `RaiseAccumulate.Error` values in `accumulating`

kyay10:
The goal is to allow usage patterns where `accumulating` values can depend on other `accumulating` values. 
I'm not super happy with this yet. It doesn't play well with property delegation (because you can't have context parameters on `getValue` yet, and even then, the design of that feature is unclear).

### DIFF
--- a/arrow-libs/core/arrow-core/api/android/arrow-core.api
+++ b/arrow-libs/core/arrow-core/api/android/arrow-core.api
@@ -983,7 +983,7 @@ public final class arrow/core/raise/Raise$DefaultImpls {
 	public static fun invoke (Larrow/core/raise/Raise;Lkotlin/jvm/functions/Function2;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
 }
 
-public class arrow/core/raise/RaiseAccumulate : arrow/core/raise/Accumulate, arrow/core/raise/Raise {
+public class arrow/core/raise/RaiseAccumulate : arrow/core/raise/Accumulate, arrow/core/raise/Raise, arrow/core/raise/Tolerance {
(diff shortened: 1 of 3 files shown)


arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/raise/RaiseAccumulateSpec.kt
```diff
@@ -37,61 +37,80 @@ class RaiseAccumulateSpec {
...
-        1
-      }
-      val y by accumulating {
-        raise("bye")
-        2
-      }
-      x + y
+      val x = accumulating<Int> { raise("hello") }
+      val y = accumulating<Int> { raise("bye") }
+      x.value + y.value
```
serras, line 42: Why are these tests changed? We need to ensure that using `by` works as expected here

kyay10: `by` can't be supported in this PR, annoyingly. This is the biggest reason why I've kept it a draft

kyay10:
Closed due to nicheness concerns. Maybe we can revisit when contextual `getValue` is supported.
For the keen reader, you can recreate the functionality of this PR yourself. The main part is some `val RaiseAccumulate.Value<*>.isError`, which can be implemented as:
```kotlin
val RaiseAccumulate.Value<*>.isError get() = try { value; false } catch(_: Throwable) { true }
```
You can copy `Tolerance`, and provide your own `accumulate { }` and `accumulating { }` blocks that have `Tolerance` contexts.

PR #3819, opened 2025-12-13
Dynamically tolerate `value` calls in `accumulating`

kyay10:
Alternative to #3813. This preserves the property delegation syntax (yay!) and has a very reasonable story around context parameters (while #3813 would be annoying to use). The only downside is that it is *dynamic*, meaning that you can't guess its behaviour locally from just looking at the code. I think that might be a good fit here, though. 
`accumulating` now does 2 jobs: it recovers from `raise` by accumulating, and it also delays any `Error.value` calls.


kyay10:
Closed due to nicheness concerns.

PR #3832, opened 2025-12-18
ktor typesafe routing

mrf7:
Adds support for typesafe routing support to the ktor server raise handlers.
~~Temporarily named all the functions `*OrError` instead of `*OrRaise` because the conflicting overloads on params and type arguments were funky. Maybe there should be a module for the type safe routing variants and a module for the regular variants? Would be a little restrictive to only use one or the other without weird override conflicts~~ Nevermind i tried the rename and it seems fine

### DIFF
--- a/arrow-libs/integrations/arrow-raise-ktor-server-resources/api/arrow-raise-ktor-server-resources.klib.api
+++ b/arrow-libs/integrations/arrow-raise-ktor-server-resources/api/arrow-raise-ktor-server-resources.klib.api
@@ -1,5 +1,5 @@
 // Klib ABI Dump
-// Targets: [iosArm64, iosSimulatorArm64, iosX64, js, linuxArm64, linuxX64, macosArm64, macosX64, mingwX64, tvosArm64, tvosSimulatorArm64, tvosX64, wasmJs, watchosArm32, watchosArm64, watchosSimulatorArm64, watchosX64]
+// Targets: [androidNativeArm64, androidNativeX64, iosArm64, iosSimulatorArm64, iosX64, js, linuxArm64, linuxX64, macosArm64, macosX64, mingwX64, tvosArm64, tvosSimulatorArm64, tvosX64, wasmJs, watchosArm32, watchosArm64, watchosSimulatorArm64, watchosX64]
 // Rendering settings:
 // - Signature version: 2
 // - Show manifest properties: true


arrow-libs/integrations/arrow-raise-ktor-server/src/commonMain/kotlin/arrow/raise/ktor/server/routing/typesafe/routingTypesafe.kt
```diff
@@ -0,0 +1,79 @@
...
+import io.ktor.http.HttpStatusCode
+import io.ktor.server.resources.delete
+import io.ktor.server.resources.get
+import io.ktor.server.resources.head
+import io.ktor.server.resources.options
+import io.ktor.server.routing.Route
+import io.ktor.server.routing.RoutingContext
+import io.ktor.server.routing.patch
+import io.ktor.server.routing.post
+import io.ktor.server.routing.put
```
tKe: ⚠️ These are pulling in the wrong handlers and will result in trying to `call.receive<TRoute>()` instead of using the resources plugin.

arrow-libs/integrations/arrow-raise-ktor-server/src/commonMain/kotlin/arrow/raise/ktor/server/routing/typesafe/routingTypesafe.kt
```diff
@@ -0,0 +1,79 @@
+@file:Suppress("API_NOT_AVAILABLE")
+
+package arrow.raise.ktor.server.routing.typesafe
```
tKe: Can we use the `.resources` package naming to be consistent with the ktor package?

I'm not sure if it's best to fully align with 
```kotlin
package arrow.raise.ktor.server.resources
```

or keep to a subpackage of 
```kotlin
package arrow.raise.ktor.server.routing.resources
```

arrow-libs/integrations/arrow-raise-ktor-server/src/commonMain/kotlin/arrow/raise/ktor/server/routing/typesafe/routingTypesafe.kt
```diff
@@ -0,0 +1,79 @@
...
+import io.ktor.server.resources.head
+import io.ktor.server.resources.options
+import io.ktor.server.routing.Route
+import io.ktor.server.routing.RoutingContext
+import io.ktor.server.routing.patch
+import io.ktor.server.routing.post
+import io.ktor.server.routing.put
+import io.ktor.utils.io.KtorDsl
+
+public typealias TypeSafeRespondingRaiseRoutingHandler<TRoute, TResponse> = suspend context(Raise<Response>) RoutingContext.(TRoute) -> TResponse
```
tKe: Can we be consistent with this naming? We seem to have `Resourced`, `TypeSafe`, and `Typed` (for the tests). I suggest sticking with `Resources` to align with the `Resources` plugin.

arrow-libs/integrations/arrow-raise-ktor-server/src/commonMain/kotlin/arrow/raise/ktor/server/routing/typesafe/routingReceivingTypesafe.kt
```diff
@@ -0,0 +1,45 @@
...
+import io.ktor.server.resources.patch
+import io.ktor.server.resources.post
+import io.ktor.server.resources.put
+import io.ktor.server.routing.RoutingContext
+import io.ktor.utils.io.KtorDsl
+
+public fun interface ResourcedReceivingRespondingRaiseRoutingHandler<TRoute, TBody, TResponse> {
+  context(_: Raise<Response>)
+  public suspend fun RoutingContext.handle(route: TRoute, body: TBody): TResponse
+}
```
tKe: This shouldn't need to be a `fun interface` and could just be a `typealias` for the appropriate lambda (making the usage below simpler).

The reason for the `fun interface` before was to rely on SAM usage to deprioritise it against the zero-arity equivalents without received parameters (as otherwise the compiler finds it ambiguous if between a no-arg lambda or a one-arg lambda with implicit `it` parameter). 

In this case, both candidates should be distinct by needing either one (explicit or implicit) or two explicit parameters on the lambda.

arrow-libs/integrations/arrow-raise-ktor-server/build.gradle.kts
```diff
@@ -10,6 +10,7 @@ kotlin {
     commonMain {
       dependencies {
         api(libs.ktor.server.core)
+        implementation(libs.ktor.server.resources)
```
tKe: I'd rather we avoid bringing in a dependency on the `ktor-server-resources` plugin for all consumers of this library as it's unnecessary unless explicitly wanting to use this feature.

That said, I'm not sure if gradle [feature variants](https://docs.gradle.org/current/userguide/how_to_create_feature_variants_of_a_library.html#header) are easily achievable with kotlin multiplatform libraries.

An alternative might be to add another arrow module with the resources support included, that depends on this one. Any thoughts @nomisRev / @serras?

arrow-libs/integrations/arrow-raise-ktor-server/src/jvmTest/kotlin/arrow/raise/ktor/server/response/RespondOrRaiseTypedTest.kt
```diff
@@ -0,0 +1,131 @@
+package arrow.raise.ktor.server.response
```
tKe: Can we relocate the tests to the same package?

arrow-libs/integrations/arrow-raise-ktor-server/src/commonMain/kotlin/arrow/raise/ktor/server/routing/typesafe/routingTypesafe.kt
mrf7: Moved everything to a separate module and decided to use `arrow.raise.ktor.server.resources` to align with ktors packaging as well as the package depth of the other raise routing functions

mrf7:
Moved everything using resources into its own module to avoid forcing the resources dependency on all users. This had the side effect of needing to make the typed `respondOrRaise` functions public. Also added some extra test cases to make make sure to check all the new functions

arrow-libs/integrations/arrow-raise-ktor-server/build.gradle.kts
mrf7: Agreed, went with the new arrow module approach instead of feature variants since that seems to be the default approach for the rest of the modules

arrow-libs/integrations/arrow-raise-ktor-server/src/commonMain/kotlin/arrow/raise/ktor/server/routing/typesafe/routingReceivingTypesafe.kt
mrf7: Ah I see, makes sense to me changed it to a typealias

arrow-libs/integrations/arrow-raise-ktor-server/src/commonMain/kotlin/arrow/raise/ktor/server/routing/typesafe/routingTypesafe.kt
mrf7: Ah good catch, I caught the same issue somewhere else and wrote a test for it but i guess i missed these ones. fixed these and added a test case that covers all exposed functions to make sure they work as expected.

settings.gradle.kts
```diff
@@ -147,6 +147,9 @@ project(":arrow-resilience-ktor-client").projectDir = file("arrow-libs/integrati
 include("arrow-raise-ktor-server")
 project(":arrow-raise-ktor-server").projectDir = file("arrow-libs/integrations/arrow-raise-ktor-server")
 
+include("arrow-raise-ktor-server-resources")
```
serras, line 150: Could you please add a reference to the project in other places where needed?

- [Kover](https://github.com/arrow-kt/arrow/blob/main/build.gradle.kts#L76)
- [Dokka](https://github.com/arrow-kt/arrow/blob/main/build.gradle.kts#L105)

tKe:
@mrf7 I've pushed a commit to avoid the internal api change as I'm wary of exposing it publicly and confusing the API surface. Looking back, I'd probably want to remove those internal functions as they only help mitigate the need to use `with(body) { handle(...) }` for the receiving calls due to the 0-artiy/1-arity lambda fun-interface workaround.

For your callers though, it's simple enough to invoke the body directly and indeed composes nicely with the non-receiving calls. Feel free to adjust the implementation as desired though.

I've also added the kover/dokka project references and generated the ABI (`:arrow-raise-ktor-server-resources:updateLegacyAbi`).

mrf7:
> simple

Makes sense to me, I'm good with those changes as is. Thanks!

PR #3837, opened 2025-12-29
NonEmptyCollection builders

kyay10:
The builders work by giving the caller a `MonotoneMutableList/Set` and forcing the caller to return a `NonEmpty & List/Set`. Operations on `MonotoneMutableList/Set` that are sure to add items, like `add` and `addAll(NonEmptyCollection)`, have a contract that gives their receiver the apt intersection type.

This PR also removes most usages of `NonEmptyList/Set()`, as in the unsafe constructors, and most usages of `PotentiallyUnsafeNonEmptyOperation`. This also meant that a lot of warning suppressions for `WRONG_INVOCATION_KIND` and `LEAKED_IN_PLACE_LAMBDA` were finally removed!

The PR doesn't compromise on inline bytecode size, either. In particular, it would be easy to call inline lambdas multiple times to get `AT_LEAST_ONCE` to comply. Instead, I've extensively used `do-while`, which resulted in cleaner code, anyway.

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/MonotoneCollectionBuilder.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/MonotoneCollectionBuilder.kt
@@ -14,13 +14,10 @@ import kotlin.contracts.contract
 @SubclassOptInRequired(PotentiallyUnsafeNonEmptyOperation::class)
 public interface MonotoneCollectionBuilder<in E> {
   // underscored so that addAll(Nec<E>) extension overload takes precedence
-  // Turning this into an abstract class doesn't work well because
-  // contracts on methods don't work well for some reason?
-  @IgnorableReturnValue
-  public fun _addAll(elements: Collection<E>): Boolean
+  // Turning this into an abstract class doesn't work well because of KT-83602
+  public fun _addAll(elements: Collection<E>)
 
-  @IgnorableReturnValue
-  public fun _add(element: E): Boolean
+  public fun _add(element: E)
 
   /**
    * Marker interface for builders that guarantee at least one element has been added.
@@ -30,34 +27,31 @@ public interface MonotoneCollectionBuilder<in E> {
   public interface NonEmpty
 }
 
-@IgnorableReturnValue
-public fun <E> MonotoneCollectionBuilder<E>.add(element: E): Boolean {
+public fun <E> MonotoneCollectionBuilder<E>.add(element: E) {
   contract { returns() implies (this@add is NonEmpty) }
   return _add(element)
 }
 
-@IgnorableReturnValue
-public fun <E> MonotoneCollectionBuilder<E>.addAll(elements: Iterable<E>): Boolean = when (elements) {
+public fun <E> MonotoneCollectionBuilder<E>.addAll(elements: Iterable<E>): Unit = when (elements) {
   is Collection -> addAll(elements)
-  else -> {
-    var result = false
-    for (item in elements) if (add(item)) result = true
-    result
-  }
+  else -> for (item in elements) add(item)
 }
 
-@IgnorableReturnValue
-public fun <E> MonotoneCollectionBuilder<E>.addAll(elements: NonEmptyCollection<E>): Boolean {
+public fun <E> MonotoneCollectionBuilder<E>.addAll(elements: NonEmptyCollection<E>) {
   contract { returns() implies (this@addAll is NonEmpty) }
-  return _addAll(elements)
+  _addAll(elements)
 }
 
-@IgnorableReturnValue
-public fun <E> MonotoneCollectionBuilder<E>.addAll(elements: Collection<E>): Boolean = _addAll(elements)
+public fun <E> MonotoneCollectionBuilder<E>.addAll(elements: Collection<E>): Unit = _addAll(elements)
 
-public fun <C> C.isNonEmpty(): Boolean where C : Collection<*>, C : MonotoneCollectionBuilder<*> {
-  contract { returns(true) implies (this@isNonEmpty is NonEmpty) }
-  return isNotEmpty()
+@SubclassOptInRequired(PotentiallyUnsafeNonEmptyOperation::class)
+private abstract class MonotoneMutableCollectionImpl<E>(private val underlying: MutableCollection<E>) : MonotoneCollectionBuilder<E>, NonEmpty {
+  final override fun _addAll(elements: Collection<E>) { underlying.addAll(elements) }
+  final override fun _add(element: E) { underlying.add(element) }
+
+  final override fun equals(other: Any?) = underlying == other
+  final override fun hashCode() = underlying.hashCode()
+  final override fun toString() = underlying.toString()
 }
 
 /**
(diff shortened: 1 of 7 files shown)


kyay10:
Note to self: There seems to be a [strange compilation mishap](https://github.com/arrow-kt/arrow/actions/runs/20586650442/job/59124177327) with `NonEmptySet` and default methods. I'll work around it for now

arrow-libs/core/arrow-core-high-arity/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
```diff
@@ -23,7 +22,22 @@ public inline fun <A, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10, R> NonEmptyList<A
   map: (A, T1, T2, T3, T4, T5, T6, T7, T8, T9, T10) -> R,
 ): NonEmptyList<R> {
   contract { callsInPlace(map, InvocationKind.AT_LEAST_ONCE) }
-  return all.zip(t1.all, t2.all, t3.all, t4.all, t5.all, t6.all, t7.all, t8.all, t9.all, t10.all, map).wrapAsNonEmptyListOrThrow()
```
serras, line 26: Please try not to change `inline` code as much as possible. One should be especially careful with such a big piece of code, for two reasons:
- It makes more difficult to us to change anything in the API. With the current approach we could improve `zip` and `wrapAsNonEmptyListOrThrow`, and everybody would benefit; whereas this is a very low-level implementation.
- The bytecode size of the user of the library is very inflated, and this is a concern in Android environments.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/MonotoneCollectionBuilder.kt
```diff
@@ -0,0 +1,199 @@
...
+
+import arrow.core.MonotoneCollectionBuilder.NonEmpty
+import kotlin.contracts.ExperimentalContracts
+import kotlin.contracts.InvocationKind
+import kotlin.contracts.contract
+
+/**
+ * A builder for collections that can only grow by adding elements.
+ * Removing elements is not supported to preserve monotonicity.
+ */
```
serras, line 13: I don't think this is required at all. A `buildNonEmptyList` function should have just a single contract: by the end of the block you should have at least one element (regardless of using `add`, `addAll`, `remove`, or whatever you want). Here you're introducing a completely different notion.

arrow-libs/core/arrow-core-high-arity/src/commonMain/kotlin/arrow/core/NonEmptyList.kt
kyay10: In this particular case, `zip` is already inline, so the bytecode size concern is out the window.
I don't quite see how `wrapAsNonEmptyListOrThrow` can ever be "improved", but regardless, the final code doesn't do anything remotely like it, since it instead statically guarantees that the list is non-empty.
Is your concern one of code duplication then? As in, that it's a maintenance burden for Arrow to have 2 copies of `zip`, one for `Iterable`, and another for `Nel`?
If so, I can extract a general "Iterator zip" algorithm out of this that assumes the iterators are non-empty. Then, the `Nel` implementation trivially then just summons `iterator()`s, and the `Iterable` implementation summons `iterator()`s, and early-returns with an empty list if any iterator happens to be empty (i.e. `hasNext()` returns false immediately). 
I didn't do this immediately because it'd end up being `@PublishedApi`. If duplication is a concern, though, then it's reasonable to expose it as such.
It's also completely reasonable if you'd instead prefer the original implementation. I simply wanted to reduce `wrapAsNonEmptyListOrThrow` calls, as well as any `WRONG_INVOCATION_KIND` warnings.

arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/MonotoneCollectionBuilder.kt
kyay10: Practically, I don't see how it can be done any other way while offering a *static* guarantee that the list is non-empty. Sure, you can `throw` if the resulting list is non-empty, but that's far from a static guarantee. One can always create intermediate `List`s and `remove` and `add` to them, as long as they always have a call that guarantees the non-emptiness of the builder. 

I think it's interesting to see that the Arrow functions affected got by just fine with `add` and `addAll`, never needing `remove`.

Just to make sure I'm fully clear, this API *guarantees* that your list is non-empty. The compiler *will* complain if you don't have an `add` or `addAll(NonEmptyCollection)`, or your own custom method that has the apt contract (where, of course, you'd be expected to ultimately call `add` or `addAll`). The only way to get around this guarantee is through a bad user-defined `contract` (to which there are compiler checkers coming eventually), or a cast to `MonotoneCollectionBuilder.NonEmpty`. This is similar to how you can already break `NonEmptyList` if you cast its `all` to `MutableList`, at least on the JVM.

If it helps, `MonotoneCollectionBuilder` is a `suspend`-less `SequenceScope`/`FlowCollector`. `add`/`addAll` correspond to `yield`/`yieldAll` and `emit`/`emitAll`.

`MonotoneCollectionBuilder` is designed as an *effect*. It's basically a write-only accumulator (maybe we should name it as such). It doesn't know, or care, how the data is stored, just that it's written somewhere. In fact, it's exactly the effect version of the `Writer` monad. This is incredibly useful because it doesn't force the implementation to use a `List`, any monoid will do. That's why the `Set` implementation is so easy, since that's another monoid.
What the `buildNonEmptyX` functions do is that they let you use any semigroup because they guarantee at least one `add` or non-empty `addAll` call. `NonEmptyList` and `NonEmptySet` are semigroups.
One can even imagine extending `MonotoneCollectionBuilder` (which ought to be called `Writer` or `Accumulator`) further by adding effect combinators like what we have for `Raise`. A `mapWriter` is easily implementable, and serves as a more efficient way to `map` a List.

<details><summary>Some caffienated handwaving about connections to monoids</summary>
There's a deeper connection here in the form of `MonotoneCollectionBuilder<T>.() -> Unit` actually being kind of a free monoid over `T`, and hence isomorphic to `List<T>`. (Technically, it's actually isomorphic to `() -> List<T>` because of exceptions and non-termination and such, and you can add `suspend` modifiers to both, too, with the isomorphism staying).
Handwaving a bit, sequencing of `MonotoneCollectionBuilder<T>.() -> Unit` functions corresponds exactly to the monoid's `combine`, with `val id: MonotoneCollectionBuilder<T>.() -> Unit = { }` serving as the identity for the monoid.
If you've ever seen how a monoid is modelled in CT, it's through a bunch of morphisms. In some way, `MonotoneCollectionBuilder<T>` lets you map any `T` to a morphism, with the morphism composition being the sequencing.
`MonotoneCollectionBuilder<T>.() -> Unit` is the "mother of all monoids", while `List<T>` is the free monoid, analogously to (multi-shot) `Continuation` being the "mother of all monads", while `Free` is the free monad.
There's something about "initial" vs "final" encoding here that I don't know enough to comment about. What I do konw is that `MonotoneCollectionBuilder<T>.() -> Unit` is more performant than a `List<T>` because it doesn't involve intermediate lists, in a similar way to how continuations are more performant than `Free` because it doesn't involve intermediate values.
</details>

kyay10: If it helps, here's a very simple version of the API to play around with ([playground](https://play.kotlinlang.org/#eyJ2ZXJzaW9uIjoiMi4zLjAiLCJwbGF0Zm9ybSI6ImphdmEiLCJhcmdzIjoiIiwianNDb2RlIjoiIiwibm9uZU1hcmtlcnMiOnRydWUsInRoZW1lIjoiaWRlYSIsImNvZGUiOiJpbXBvcnQga290bGluLmNvbnRyYWN0cy4qXG5mdW4gbWFpbigpIHtcbiAgICB2YWwgbGlzdDogTGlzdDxJbnQ+ID0gYnVpbGROZWwge1xuICAgICAgICAvL3RlbGwoNDIpIC8vIHVuY29tbWVudCB0byBnZXQgY29kZSBjb21waWxpbmdcbiAgICAgICAgdGhpcyAvLyBpZiB0aGVyZSdzIG5vIGB0ZWxsYCBpbiB0aGUgY29kZSBhYm92ZSwgY29tcGlsZXIgZXJyb3JzIHdpdGggXCJSZXR1cm4gdHlwZSBtaXNtYXRjaFwiXG4gICAgfVxuICAgIHByaW50bG4obGlzdClcbn1cblxuaW50ZXJmYWNlIFdyaXRlcjxpbiBUPiB7XG4gICAgZnVuIF90ZWxsKHZhbHVlOiBUKVxuICAgIGludGVyZmFjZSBOb25FbXB0eVxufVxuXG5mdW4gPFQ+IFdyaXRlcjxUPi50ZWxsKHZhbHVlOiBUKSB7XG4gICAgY29udHJhY3QgeyByZXR1cm5zKCkgaW1wbGllcyAodGhpc0B0ZWxsIGlzIFdyaXRlci5Ob25FbXB0eSkgfVxuICAgIF90ZWxsKHZhbHVlKVxufVxuXG5hYnN0cmFjdCBjbGFzcyBMaXN0V3JpdGVyPFQ+IGludGVybmFsIGNvbnN0cnVjdG9yKHByaXZhdGUgdmFsIHVuZGVybHlpbmc6IE11dGFibGVMaXN0PFQ+KTogV3JpdGVyPFQ+LCBMaXN0PFQ+IGJ5IHVuZGVybHlpbmcge1xuICAgIG92ZXJyaWRlIGZ1biBfdGVsbCh2YWx1ZTogVCkgeyB1bmRlcmx5aW5nLmFkZCh2YWx1ZSkgfVxuICAgIG92ZXJyaWRlIGZ1biB0b1N0cmluZygpID0gdW5kZXJseWluZy50b1N0cmluZygpXG4gICAgY29tcGFuaW9uIG9iamVjdCB7IG9wZXJhdG9yIGZ1biA8VD4gaW52b2tlKCk6IExpc3RXcml0ZXI8VD4gPSBJbXBsKCkgfVxuICAgIHByaXZhdGUgY2xhc3MgSW1wbDxUPjogTGlzdFdyaXRlcjxUPihtdXRhYmxlTGlzdE9mKCkpLCBXcml0ZXIuTm9uRW1wdHlcbn1cblxuaW5saW5lIGZ1biA8VCwgVz4gYnVpbGROZWwoYmxvY2s6IExpc3RXcml0ZXI8VD4uKCkgLT4gVyk6IExpc3Q8VD4gd2hlcmUgVzogV3JpdGVyLk5vbkVtcHR5LCBXOiBMaXN0V3JpdGVyPFQ+ID0gXG4gICAgYmxvY2soTGlzdFdyaXRlcigpKS5pZkVtcHR5IHsgZXJyb3IoXCJZb3UndmUgYnJva2VuIG15IEFQSVwiKSB9In0=)). Let me know if you find any way to break it!

serras: The question here for me is whether providing this API is useful enough to include it in `arrow-core`, given that at this point we have no way to enforce that at least one call to `add` or `addAll` is done. We really made a big effort to reduce Arrow's API to its minimum, and this PR goes into the opposite direction.

kyay10: > given that at this point we have no way to enforce that at least one call to add or addAll is done

Do you mean that `buildNonEmptyList` doesn't enforce that? Because it does! See the playground link above. It's all compile-time checked, relying on contracts and a lambda with an intersection return type.

I can have a go at reducing the API surface here if you'd like. I was aiming to have an API similar to `MutableList/Set` (without the `remove` methods), but I think only a fraction of the API is really necessary, and we can always extend it later

kyay10:
FWIW, I can move the builders to a new module, and add an `implementation` dependency on it in core. This way, the API isn't exposed to outside users automatically, but we can still rely on it internally.

PR #3839, opened 2025-12-30
Cleanup and optimize Nec serialization

kyay10:


kyay10:
It looks like some of the deserializers were public. Are we concerned about binary compatibility here? If so, I can add back the (now unused) deserializers

serras:
In principle we should try to keep binary compatibility, because one never knows when reflection may kick in.

However, I'm not entirely sure here about what we are gaining by splitting the `NonEmptyList` and `NonEmptySet` (de)serializers. It's true we can use a slightly more optimized structure in `NonEmptySet`, but is this worth the change?

kyay10:
The (admittedly micro) optimization here is using `wrapAsNonEmptyXOrNull` instead of `toNonEmptyXOrNull`. The splitting happens as a consequence because you can't use `wrapAsNonEmptySetOrNull` on a `List`!
I could likely make a more convoluted version with one (de)serializer that also uses `HashSet` vs `ArrayList` based on a parameter, but it seemed like an unnecessary hassle to avoid duplicating a few lines of code.

Note: using `wrapAsNonEmptyXOrNull` technically assumes a good faith `List` and `Set` deserializer implementation, as in, one that doesn't store the deserialized `List` or `Set` and mutate it later, removing elements. I think that's a rather reasonable expectation.

PR #3840, opened 2025-12-30
Cleanup Nec tests

kyay10:


arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/raise/EffectSpec.kt
```diff
@@ -603,10 +603,8 @@ class EffectSpec {
   }
 
   @Test fun accumulateReturnsEveryError() = runTest {
-    checkAll(Arb.list(Arb.int(), range = 2..20)) { errors ->
-      either<NonEmptyList<Int>, List<String>> {
-        mapOrAccumulate(errors) { raise(it) }
-      } shouldBe errors.toNonEmptyListOrNull()!!.left()
+    checkAll(Arb.nonEmptyList(Arb.int(), range = 2..20)) { errors ->
+      merge { mapOrAccumulate(errors) { raise(it) }.head } shouldBe errors
```
serras: Why this change? Changing the body of tests is tricky, because we may not be testing the same behavior as before.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/raise/SingletonSpec.kt
```diff
@@ -129,179 +124,74 @@ class SingletonSpec {
...
-      } else {
-        a.mapValuesNotNull { it.value }
-      }
-
-      singleton(
-        { recovered },
-        { a.bindAll() },
-      ) shouldBe expected
+      val recovered = emptyMap<Int, Int>()
+      singleton({ recovered }) { a.bindAll() } shouldBe run { a.mapValues { (_, v) -> v ?: return@run recovered } }
```
serras: I'm not super-sure about this change either. The fact that the early return moves to `run` is quite convoluted, to be honest, and it's hard for me to know whether the test has changed or not.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
```diff
@@ -505,29 +473,16 @@ class IterableTest {
...
-              "$accA${
-                either.value.fold("") {accB, entry -> "$accB$entry"}
-              }"
-            }
-            .left()
-        } else {
-          nelslist.filterIsInstance<Either.Right<Int>>().map { it.value }.right()
-        }
+      val expected = list.mapNotNull { it.leftOrNull() }.toNonEmptyListOrNull()?.flatten()?.reduce(String::plus)?.left()
+        ?: list.mapNotNull { it.getOrNull() }.right()
```
serras: On a general note: these tests do not try to be short, but to be as _clear_ as possible. I don't think we should replace the longer test with a two-liner, especially if it uses some non-trivial control flow with `?:`.

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/NonEmptyListTest.kt
```diff
@@ -473,23 +455,19 @@ class NonEmptyListTest {
   @Test
   fun plusIterable() = runTest {
     checkAll(
-      Arb.list(Arb.int(), range = 0..10),
+      Arb.nonEmptyList(Arb.int(), range = 1..10),
```
serras, line 458: Using `nonEmptyList` directly here seems quite useful to make the tests much more concise :)

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/raise/EffectSpec.kt
kyay10: The `.head` is actually unreachable here, which I didn't see because we have suppressed it.
I'll undo this change for now. 
I eventually want to do a run through the code and remove any unreachable code (because it's a stronger guarantee that the compiler says some code is unreachable vs our tests checking at runtime that it's unreachable. In fact, removing such code means we're guaranteeing that users get those guarantees statically).
<details><summary>Here's an explanation of why the change made sense, if you're curious</summary>
If we can agree that this change is valid:
```kotlin
either<NonEmptyList<Nothing>, NonEmptyList<String>> {
  mapOrAccumulate(errors) { raise(it) }
}.merge() shouldBe errors
```
Notice that the only 3 changes are that `toNonEmptyListOrNull!!` is no longer necessary, `Nel<Int>` is now `Nel<Nothing>`, and we're `merge`ing the `Either` instead of doing `.left()`.
Why is the `.merge()` valid here? Well, `Nel<Nothing>` is actually equivalent to `Nothing`, so we never return normally (and in fact, the compiler knows this, and warns if we add any code to the end of it)
Then, you can trivially simplify the above to:
```kotlin
merge { mapOrAccumulate(errors) { raise(it) } } shouldBe errors
```
The point being that there could never be a `Right` in the code once we're using `Arb.nonEmptyList`.
The `.head` was my poor attempt at showcasing that, since it exhibits that `Nel<Nothing>` really is isomorphic to `Nothing`.
</details>

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/raise/SingletonSpec.kt
kyay10: The change makes sense if you keep in mind that `raise` is supposed to operate exactly like non-local returns, hence the usage of `return@run`.
Personally, the previous test was less clear since it had `mapValuesNotNull`, which raises eyebrows since that map might not have the same size then. Obviously, that's not the case, but the test is still thus confusing (and there's more tests in that same style that are incredibly more confusing IMO, so I wanted to have a consistent new style)

This one, though, can be simplified to just `if (a.containsValue(null)) recovered else a`, so I'll make that change

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/IterableTest.kt
kyay10: the previous code was difficult to follow for me, suffering from that same issue of "oh no, what if the `filter` decreases the size of the list" issue. I'll have a go at rewriting this so it's clearer!

kyay10:
Thanks for the nudges in the right direction! The changes are even clearer as a result

arrow-libs/core/arrow-core/src/commonTest/kotlin/arrow/core/EitherTest.kt
```diff
@@ -241,18 +240,8 @@ class EitherTest {
...
-        all.filterIsInstance<Left<String>>().fold("") { acc, t -> "$acc${t.value}" }.left()
-      } else {
-        all.filterIsInstance<Right<Any?>>().map { it.value }.let {
-          Tuple9(it[0], it[1], it[2], it[3], it[4], it[5], it[6], it[7], it[8]).right()
-        }
-      }
-
-      res shouldBe expected
+      Either.zipOrAccumulate(String::plus, a, b, c, d, e, f, g, h, i) { _, _, _, _, _, _, _, _, _ -> }.leftOrNull() shouldBe
+        listOf(a, b, c, d, e, f, g, h, i).mapNotNull { it.leftOrNull() }.toNonEmptyListOrNull()?.reduce(String::plus)
```
serras: As before, this test may be in theory equal, but it's testing it in a very different way.

serras: Just to be clear, I'd be very happy if we had _both_ tests (we can just have a couple of `shouldBe` in the same test), but I don't think we should _remove_ any test that's already there.

PR #3864, opened 2026-02-08
Deprecate `unweave`

### LINKED ISSUE #3675: Counter-intuitive behavior of `unweave` in Iterable / Sequence
Say we have 3 initial collections (after transformations)
```
[16, 17, 18]
[3, 4]
[5, 6]
```
and then we want to unweave.. but what does it mean?
Probably we wanted to interleave all 3 collections? If so, then the output should be `[16, 3, 5, 17, 4, 6, 18]`

But current `unweave` does something quite strange - it starts from the last two collections and then moves backwards interleaving them.
So first we have `[3, 4] interleave [5, 6]` -> `[3, 5, 4, 6]`
then we get to the first element and `[16, 17, 18] interleave [3, 5, 4, 6]` -> `[16, 3, 17, 5, 18, 4, 6]`
Is this the original intent?

Also the recursion approach used in implementation might be slow.. ?

serras:
Nobody seems to understand what it does, and the usage is non-existent.

Fixes #3675

### DIFF
--- a/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
+++ b/arrow-libs/core/arrow-core/src/commonMain/kotlin/arrow/core/Iterable.kt
@@ -817,6 +817,7 @@ public fun <A> Iterable<A>.interleave(other: Iterable<A>): List<A> =
  * ```
  * <!--- KNIT example-iterable-15.kt -->
  */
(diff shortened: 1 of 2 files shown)


kyay10 (approved):
Looks incredibly niche, so yeah no reason to keep it around

PR #3879, opened 2026-03-03
Use common atomics on non-JVM platforms

serras:
Since we officially support from Kotlin 2.1 on, I think we should make ourselves a favor and use the common atomics. The reason why it's not done for all platforms is due to binary compatibility: in JVM (and Android) `arrow-atomic` simply wraps the JVM primitives, so changing that would be breaking.

@nomisRev @kyay10 what do you think?


serras:
Note: I'm postponing this until 2.4, since at that point only standard library from 2.1 on is supported, and there common atomics appeared.

PR #3928, opened 2026-06-25
feat: add QUERY verb to arrow-raise-ktor-server and -resources

tKe:
ktor-server-resources plugin doesn't currently expose a `query` extension yet, so I've invoked `resource/method/handle` manually.


nomisRev (approved):
Looks good! Thank you @tKe. I'll create a Ktor issue/pr for the missing query method in resources.

PR #3930, opened 2026-06-26
fix(suspendapp): avoid deadlock and align JVM behaviour on system.exit

### LINKED ISSUE #3846: [Bug] SuspendApp freezes app on System.exit

tKe:
fixes #3846

This fix essentially sidesteps the deadlock by not attempting to cancel/graceful-shutdown when the hook is invoked for any reason other than SIGINT/SIGTERM.

The rationale is that `System.exit` equivalents on other platforms doesn't trigger cancellation of the SuspendApp, and the other hook invocation cause (last non-daemon thread terminated) is not expected through the normal usage of SuspendApp (that is, as the entry point on the main thread for a service/application).

Instead, this fixes the Signal handling, captures the signal, and limits triggering cancellation to only when signalled.

### DIFF
--- a/arrow-libs/suspendapp/suspendapp-test-runner/src/test/kotlin/SuspendAppTest.kt
+++ b/arrow-libs/suspendapp/suspendapp-test-runner/src/test/kotlin/SuspendAppTest.kt
@@ -1,5 +1,10 @@
+import io.kotest.inspectors.shouldForAtLeastOne
+import io.kotest.inspectors.shouldForNone
 import io.kotest.inspectors.shouldForOne
+import io.kotest.matchers.collections.shouldBeIn
 import io.kotest.matchers.shouldBe
+import io.kotest.matchers.shouldNotBe
+import io.kotest.matchers.string.shouldContain
 import io.kotest.matchers.string.shouldEndWith
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.test.runTest
@@ -19,7 +24,17 @@ abstract class SuspendAppTest : ProcessProvider {
     val (process, output) = execute("fail")
     process.exitValue() shouldBe 255
     output.shouldForOne { it.line shouldBe "resource clean complete" }
-      .shouldForOne { it.line shouldEndWith "IllegalStateException: BOOM!" }
+      .shouldForAtLeastOne { it.line shouldEndWith "IllegalStateException: BOOM!" }
+  }
+
+  @Test
+  fun exit() = runTest {
+    val (process, output) = execute("exit")
+    process.exitValue() shouldBe 42
+    output
+      .filter { it.source == "stdout" }
+      .shouldForNone { it.line shouldBe "resource clean complete" }
+      .last().line shouldBe "Running ExitProcess"
   }
 
   @Test
@@ -27,7 +42,7 @@ abstract class SuspendAppTest : ProcessProvider {
     val (process, output) = execute("childfail")
     process.exitValue() shouldBe 255
     output.shouldForOne { it.line shouldBe "resource clean complete" }
-      .shouldForOne { it.line shouldEndWith "IllegalStateException: boom." }
+      .shouldForAtLeastOne { it.line shouldEndWith "IllegalStateException: boom." }
   }
 
   @Test
@@ -57,9 +72,21 @@ abstract class SuspendAppTest : ProcessProvider {
   @Test
   fun waitAndSignalSigint() = waitAndSignal(Signal.SIGINT)
 
-  private fun waitAndSignal(signal: Signal) = runTest {
-    val (process, output) = execute("wait") {
-      delay(1.seconds)
+  @Test
+  fun waitAndTimeout() = runTest {
+    val (process, output) = execute("timeout") {
+      delay(0.5.seconds)
+      sendSignal(Signal.SIGTERM)
+    }
+    // TODO: inconsistent exit codes across platforms, for now just check it's not a success
+    process.exitValue() shouldNotBe 0
+    output.shouldForAtLeastOne { it.line shouldContain "Timed out waiting for 5000 ms" }
+      .shouldForNone { it.line shouldContain "resource clean complete" }
+  }
+
+  private fun waitAndSignal(signal: Signal, mode: String = "wait") = runTest {
+    val (process, output) = execute(mode) {
+      delay(0.5.seconds)
       sendSignal(signal)
     }
     process.exitValue() shouldBe signal.code + 128
--- a/arrow-libs/suspendapp/suspendapp/src/jvmMain/kotlin/arrow/continuations/Enviroment.jvm.kt
+++ b/arrow-libs/suspendapp/suspendapp/src/jvmMain/kotlin/arrow/continuations/Enviroment.jvm.kt
@@ -63,6 +63,8 @@ private object JvmProcess : Process {
               e.printStackTrace()
             }
           }
+        } else {
+          System.err.println("WARNING: SuspendApp Shutdown Hook invoked without being signalled. No SuspendApp cancellation will occur.")
         }
       }
     }
@@ -96,8 +98,9 @@ private object JvmProcess : Process {
             handle?.handle(sig)
           }
         }
-    } catch (_: Throwable) {
-      /* Ignore */
+    } catch (ex: Throwable) {
+      System.err.println("WARNING: SuspendApp was unable to install a signal handler for SIG$signal. Signalled cancellation may be impacted.")
+      ex.printStackTrace()
     }
 
   override fun runScope(context: CoroutineContext, block: suspend CoroutineScope.() -> Unit): Unit =
--- a/arrow-libs/suspendapp/suspendapp/src/jvmMain/kotlin/arrow/continuations/Enviroment.jvm.kt
+++ b/arrow-libs/suspendapp/suspendapp/src/jvmMain/kotlin/arrow/continuations/Enviroment.jvm.kt
@@ -2,6 +2,7 @@ package arrow.continuations
 
 import arrow.AutoCloseScope
 import java.util.concurrent.atomic.AtomicBoolean
+import java.util.concurrent.atomic.AtomicInteger
 import kotlin.coroutines.CoroutineContext
 import kotlin.system.exitProcess
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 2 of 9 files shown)


serras:
@tKe could you rebase after the merging of #3917? In fact, is there any relation between both PRs that need to be taken into account?

arrow-libs/suspendapp/suspendapp/src/jvmMain/kotlin/arrow/continuations/Enviroment.jvm.kt
```diff
@@ -14,42 +15,85 @@ internal actual fun AutoCloseScope.process(): Process = JvmProcess
...
+        } catch (_: IllegalStateException) {
+          // Shutdown hook already running, ignore
+        }
       }
     }
   }
 
   override fun onSigTerm(block: suspend (code: Int) -> Unit): Unit =
-    addSignalHandler("SIGTERM") { block(15) }
+    addSignalHandler("TERM", block)
```
serras, line 86: Why is the change from `SIGTERM` to `TERM` needed?

tKe: `sun.misc.Signal` takes the signal name sans SIG prefix.
The example from the JavaDoc is:


> Signal objects are created based on their names. For example:
>  
> ```
> new Signal("INT");
> ```

Previously these were throwing `IllegalArgumentException` on construction that were being caught and ignored.

tKe:
@serras rebasing this, which includes job name fixes for the js tests, highlighted that the exitApp change has introduced a regression in then wasm/js signal handling as it now reports an exitCode of 255 not 128+sig. I'll include a fix in this rebase.

arrow-libs/suspendapp/suspendapp/src/jvmMain/kotlin/arrow/continuations/Enviroment.jvm.kt
nomisRev: Should we perhaps just `e.printStackTrace()` here? I guess this is what was swallowing the `IllegalArgumentException` that was thrown for using `SIGINT` vs just `INT`.

Alternatively we do not catch it at all and just let it throw?

arrow-libs/suspendapp/suspendapp/src/jvmMain/kotlin/arrow/continuations/Enviroment.jvm.kt
nomisRev: Good catch @tKe

arrow-libs/suspendapp/suspendapp/src/jvmMain/kotlin/arrow/continuations/Enviroment.jvm.kt
```diff
@@ -14,42 +15,85 @@ internal actual fun AutoCloseScope.process(): Process = JvmProcess
...
     }
 
     val hook = Thread(::shutdown, "Arrow-kt SuspendApp JVM ShutdownHook")
     Runtime.getRuntime().addShutdownHook(hook)
     return {
       if (!isShutdown.get()) {
-        Runtime.getRuntime().removeShutdownHook(hook)
+        try {
+          Runtime.getRuntime().removeShutdownHook(hook)
+        } catch (_: IllegalStateException) {
```
nomisRev, line 78: Does this occur? It should be gated by `isShutdown`, right?

arrow-libs/suspendapp/suspendapp/src/jvmMain/kotlin/arrow/continuations/Enviroment.jvm.kt
```diff
@@ -14,42 +15,85 @@ internal actual fun AutoCloseScope.process(): Process = JvmProcess
...
+        - Since that wait cannot resolve, the shutdown hook will deadlock.
+
+        Other platform behaviour:
+        - Other targets do not have a JVM-style shutdown hook with equivalent process-exit semantics,
+          and currently do not attempt graceful shutdown either, except when signalled.
+
+        So, graceful shutdown is gated solely on SIGINT/SIGTERM capture, both to align behaviour with
+        other platforms and to avoid the potential deadlock of an explicit `System.exit`.
+        */
+        if (lastSignal.get() != -1) {
```
nomisRev, line 57: We probably want to add an `else` case here with some logging?

arrow-libs/suspendapp/suspendapp/src/jvmMain/kotlin/arrow/continuations/Enviroment.jvm.kt
```diff
@@ -14,42 +15,85 @@ internal actual fun AutoCloseScope.process(): Process = JvmProcess
...
-  private fun addSignalHandler(signal: String, action: suspend () -> Unit): Unit =
+  private fun addSignalHandler(signal: String, action: suspend (code: Int) -> Unit): Unit =
     try {
       var handle: SignalHandler? = null
-      @Suppress("ASSIGNED_VALUE_IS_NEVER_READ")
       handle =
-        Signal.handle(Signal(signal)) { prev ->
-          runBlocking { action() }
+        Signal.handle(Signal(signal)) { sig ->
+          runBlocking { action(sig.number) }
```
nomisRev, line 96: This violates the "exit code formula" `Exit Code=128+Signal Number`.

The exit code formula `Exit Code=128+Signal Number` is not being respected here since now we set `2` for `SIGINT`, and `15` for `SIGTERM`. This seems consistently implemented in the other platforms.

arrow-libs/suspendapp/suspendapp-test-runner/src/test/kotlin/SuspendAppTest.kt
```diff
@@ -57,9 +72,21 @@ abstract class SuspendAppTest : ProcessProvider {
...
-  private fun waitAndSignal(signal: Signal) = runTest {
-    val (process, output) = execute("wait") {
-      delay(1.seconds)
+  @Test
+  fun waitAndTimeout() = runTest {
+    val (process, output) = execute("timeout") {
+      delay(0.5.seconds)
+      sendSignal(Signal.SIGTERM)
+    }
+    // TODO: inconsistent exit codes across platforms, for now just check it's not a success
```
nomisRev, line 81: Seems like an inconsistency introduced in this PR?  See other comment https://github.com/arrow-kt/arrow/pull/3930/changes#r3519183429.

nomisRev:
@tKe couple of small questions and remarks. Feels like a relatively large change, but it mostly tries to prevent hanging indefinitely which I guess is okay. What is a bit strange to me is that now _silently_ in those cases the shutdown will not happen.

Tagging @alexandru since he recently also worked here and is familiar with semantics of shutdown and would like to hear his thoughts if he has time to chime in.

arrow-libs/suspendapp/suspendapp/src/jvmMain/kotlin/arrow/continuations/Enviroment.jvm.kt
tKe: Unlikely though it is, there's the possibility the runtime shutdown process has been started but our shutdown hook thread has not yet started or set the `isShutdown` flag. Attempting to remove hooks once the shutdown process has started will fail.

arrow-libs/suspendapp/suspendapp/src/jvmMain/kotlin/arrow/continuations/Enviroment.jvm.kt
tKe: This is consistent with the other platforms where the onSigInt/onSigTerm blocks receive the signal number (i.e. `2` / `15`). We could name the parameter of the lambda for clarity and avoid using unqualified `code` in most places but I didn't want to make too many tweaks outside of the main area of impact.

The other platforms handlers tend to do the `sig+128` maths to convert an exit code in the block passed to `onSigInt`/`onSigTerm`.
i.e. `{ code -> exitAfter(128 + code) { block() } }`

For the JVM, we let the JVM handle the exit code calculation based on the default signal handler so we needn't do the calculation here (although we could theoretically call System.halt we don't know if other shutdown hooks have finished or not, nor have control of the ordering)

arrow-libs/suspendapp/suspendapp/src/jvmMain/kotlin/arrow/continuations/Enviroment.jvm.kt
tKe: It's the invocation of `Signal(signal)` that throws when the signal is unknown.
The `Signal.handle` invocation may also throw if the signal is reserved by the VM/OS. 
It might be safer to `e.printStackTrace` or add a custom log warning that signal handling for SuspendApp may be impacted. Otherwise the cases where we don't cancel on signal as we haven't installed the handler would be hard to diagnose.

arrow-libs/suspendapp/suspendapp-test-runner/src/test/kotlin/SuspendAppTest.kt
tKe: Currently, I believe the inconsistency is on JVM still exiting with the correct signalled exit code (128+sig) but the other platforms are exiting with exit code 255 due to the timeout exception.

```
JvmSpec:waitAndTimeout exitCode == 143
JsSpec:waitAndTimeout exitCode == 255
WasmJsSpec:waitAndTimeout exitCode == 255
NativeSpec:waitAndTimeout exitCode == 255
```

tKe:
Thanks for the review @nomisRev. I've addressed or replied to your feedback.

It is a large change, but ultimately the shutdown hook not cancelling the app on `exitProcess` is aligned with the other platforms doing the same, and the new `exitApp` from @alexandru provides an alternative nicely.

It also aligns the platform handling of `SIGHUP` which will currently terminate without cancellation/shutdown on JS/Wasm/Native and JVM now (whereas before it would cancel on JVM).

Only `SIGQUIT` differs amongst the other platforms, with the JVM dumping stacktrace but otherwise continuing (expected JVM functionality), and the other platforms exiting without cancellation with exit code 131.

Otherwise the change only impacts when SuspendApp isn't on a non-daemon thread (which would be very unexpected usage) as it would otherwise have unregistered the shutdown hook before it would be invoked due to last non-daemon thread terminating.

For the expected usage patterns that aren't calling `exitProcess` (which previously deadlocked and never shutdown gracefully anyway), the impact is that we no longer cancel on SIGHUP.

serras:
@nomisRev are you OK with the reviewed code? I’d like to merge this one, since it seems it fixes an important problem.

tKe:
(sorry for the bump)

@nomisRev was there any outstanding concerns with this change?

PR #3936, opened 2026-06-29
Mark suspendapp-ktor ResourceScope.server as ignorable

tKe:


arrow-libs/suspendapp/suspendapp-ktor/src/commonMain/kotlin/arrow/continuations/ktor/KtorServer.kt
```diff
@@ -103,7 +105,7 @@ private suspend fun EmbeddedServer<*, *>.release(
     environment.log.info(
       "prewait delay of ${preWait.inWholeMilliseconds}ms, turn it off using io.ktor.development=true"
     )
-    delay(preWait.inWholeMilliseconds)
+    delay(preWait)
```
serras, line 108: Is this change intented? I’d prefer to use the Duration overload of delay instead of the integer one, since it’s more explicit about the units.

tKe: This is switching to use the delay(Duration) overload instead of delay(Long). IntelliJ was complaining about the deprecated Long overload usage.

serras: Sorry, I somehow reversed the + and -

PR #3954, opened 2026-07-21
Make Arrow Optics compiler plug-in work in the presence of `@Serializable`

serras:


kyay10:
Is there no way to generate methods within the kotlinx-serialization-generated companion object? It might have you accessing FIR directly, but I think you can grab the outer class of a companion object, so you can check its optics annotation. There's even a predicate for that (I think it's called `under` or something like that, for a class whose outer class is annotated)

serras:
It turns out that FIR plug-ins don't see each others' declarations. There's an [open issue](https://youtrack.jetbrains.com/issue/KT-62194/K2-Two-compiler-plugins-interference-in-generated-companion-object).

